PR #576, opened 2015-01-30
Fixes #575

### LINKED ISSUE #575: latexmlc: Create subdirectory structure for --log if it does not exist
When I do: 

```
latexmlc file.tex --log=/non/existing/folder/file.ltxlog
```

latexmlc simply dies with: 

```
Fatal:I/O:forbidden Couldn't open log file /non/existing/folder/file.ltxlog : No such file or directory
```

It would be good if latexmlc could create subdirectory structures for the log file if it does not exist. 
/cc @dginev KWARC/localmh#194

dginev:
Same treatment for ```--log``` as we do for ```--destination``` in terms of auto-making their directory.

(Aside: It is a bit awkward to do dev work while libxml is raising validity errors everywhere, but I am pretty certain this works as intended)

### DIFF
--- a/lib/LaTeXML/Common/Config.pm
+++ b/lib/LaTeXML/Common/Config.pm
@@ -191,11 +191,12 @@ sub read {
       -sections => 'OPTIONS/SYNOPSIS', output => \*STDOUT);
   }
 
-  # Check that destination is valid before wasting any time...
-  if ($$opts{destination}) {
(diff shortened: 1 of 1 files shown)


dginev:
And I am realizing these are "system write options" rather than "I/O" options, but it seems a waste to make an extra comment just to fix a slightly off comment. I can do that if you want me to.

PR #585, opened 2015-02-27
Patch ePub image packing for files in subdirectories

dginev:
Hi,

I stumbled on this problem while trying to create an ePub for my new blog posts (soon to be online) and realized that the Manifest couldn't pick up any resources in the subdirectories under OPS.

In my concrete example I needed to generate a manifest ```OPS/content.opf``` file that contained an item for a nested image file:

```xml
<item id="app-assets-images-pointers.png" href="app/assets/images/pointers.png" 
      media-type="image/png"/>
```

I used the core File::Find module for traversing the subdirectories, and sanitized the produced item IDs to be NCNames. 

I also removed some code duplication in the Pack post-processor and made sure LaTeXML.cache is never added to the archive. 

If you are happy with using File::Find, this is ready to merge. I haven't added a test for it yet though, sorry.

### DIFF
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -12,6 +12,7 @@
 package LaTeXML::Post::Manifest::Epub;
 use strict;
 use warnings;
+use File::Find qw(find);
 
 our $uuid_tiny_installed;
 
@@ -40,8 +41,6 @@ sub new {
   my ($class, %options) = @_;
   my $self = $class->SUPER::new(%options);
   return $self; }
-
-use Data::Dumper;
 
 sub initialize {
   my ($self, $doc) = @_;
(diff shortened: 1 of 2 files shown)


dginev:
Just occured to me, the recursive traversal of subdirectories, as well as the ```pathname_to_ncname``` function can be refactored as individual subroutines in ```LaTeXML::Util::Pathname```. If you would like to keep them there, I can refactor with an extra commit here.

dginev:
Since I saw some movement, I will do a gentle ping on this PR. No urgency to merge and feedback is welcome!

dginev:
@brucemiller wondering if your silence indicates you haven't seen the PR yet? Not sure whether to ping you or not now :)

brucemiller:
Oh; forgotten, I guess.  I think I wanted to comment about whether we needed to add a new dependency (Find::File), or whether, possibly the existing find_file works for that (or could)?

BTW: your code reminded me of the new (to me!) substitution keyletter; you can write:
```
my $foo = $bar = s/this/that/r;
```
w/o changing $bar! *is* that new? (wonder if it's good style)

dginev:
Please don't use that :) These idioms are very dangerous, as they are easy to get wrong and hard to debug/maintain. I had read recently that having more than one assignment operator per line is just asking for trouble and should be avoided at all costs. Which doesn't mean you should agree with that... 

I'll take a look at find_file and see if it's a sensible refactoring.

brucemiller:
Actually the second = should have been ```=~```, so it isn't quite an assignment. An alternative in the pod which does have 2 assignments is
```
($foo = $bar) =~ s/this/that/;
```
which does confuse me :>

dginev:
```perl
($foo = $bar) =~ s/this/that/;
```

I find myself compelled to make a Perl joke. This syntax isn't too far away visually from the table flipping operator:
```
 (╯°□°）╯︵ ┻━┻) 
```
Reference: http://knowyourmeme.com/memes/flipping-tables-%E2%95%AF%E2%96%A1%EF%BC%89%E2%95%AF%EF%B8%B5-%E2%94%BB%E2%94%81%E2%94%BB

dginev:
```
I think I wanted to comment about whether we needed 
to add a new dependency (Find::File),
or whether, possibly the existing find_file works for that (or could)?
```

The way I am reading the LaTeXML::Util::Pathname API, there is no way to a) recursively search a directory and b) match all files of a certain extension.

I am searching for all CSS and PNG files under the ePub resource path, in order to include them in the manifest. Not sure whether we want to reimplement a core Perl module under LaTeXML, File::Find is shipping as a Perl core module from 5.8.8 all the way to 5.20.1. But if you would prefer having a wrapper for it in Util::Pathname, I can add that.

dginev:
This PR has been laying around for a while, just rebased it on master, still works well.

Should I rework it to not use File::Find and ping the thread again?

brucemiller:
Is File::Find part of Core? Seems to be in Fedora 22.

PR #593, opened 2015-04-14
Fixes #592

### LINKED ISSUE #592: Preserve information for broken citations in CrossRef
For practical (backwards-compatibility) reasons, I am currently developing a setup where LaTeXML is run on fragments without a bibliography, but with the actual ```\cite{}``` commands left in the source.

I find myself needing the actual citation keys in the final HTML output, so that we can post-process them externally to LaTeXML. The keys are already present in the LaTeXML XML but seem to get stripped out during Post::CrossRef. @brucemiller : would you be receptive to keeping the broken bibrefs into the final HTML, maybe with a special CSS class, so that they can be hidden, or made red like the rest of the errors, or turned into he TeX-like question mark, based on the user's preference? I can do the necessary work once we develop a strategy.

Example source:
```tex
\documentclass{article}
\begin{document}
Request: Preserve content of \cite{missing:citations} in post-processing.
\end{document}
```

LaTeXML XML:
```xml
  <para xml:id="p1">
    <p>Request: Preserve content of <cite>[<bibref bibrefs="missing:citations" separator="," show="Refnum" yyseparator=","/>]</cite> in post-processing.</p>
  </para>
</document>
```

XML after CrossRef post-processing (no stylesheet applied):
```xml
<document xmlns="http://dlmf.nist.gov/LaTeXML" xml:id="Document">
  <para xml:id="p1" fragid="p1">
    <p>Request: Preserve content of <cite>[]</cite> in post-processing.</p>
  </para>
</document>
```

dginev:
1. There was an else conditional for a missing citation that wasn't being reported as a warning, which I now handled.

2. Instead of returning nothing, I now pass the original citation key to the post-processed XML, with a special ```ltx_missing_citation``` class. 

This is all I need in #592, and I am open to reworking it to a reasonable alternative. I think this PR catches the gist of the request, feel free to merge if you like it.

The XML snippet produced for #592 is now:
```xml
<document xmlns="http://dlmf.nist.gov/LaTeXML" xml:id="Document">
  <para xml:id="p1" fragid="p1">
    <p>Request: Preserve content of <cite>[<ref class="ltx_missing_citation" idref="missing:citations" title="missing:citations">missing:citations</ref>]</cite> in post-processing.</p>
  </para>
</document>
```

I am unsure which attributes are worth copying over, since I am not familiar with the uses of each of them. I set all of them to the key value for now, until I understand more.

### DIFF
--- a/lib/LaTeXML/Post/CrossRef.pm
+++ b/lib/LaTeXML/Post/CrossRef.pm
@@ -429,9 +429,13 @@ sub make_bibcite {
               title   => [$doc->trimChildNodes($title || $keytag)],
               attr    => { idref => $id,
                 href => orNull($self->generateURL($doc, $id)),
-                ($title ? (title => orNull($title->textContent)) : ()) } }); } } }
+                ($title ? (title => orNull($title->textContent)) : ()) } }); } }
+      else {
+        $self->note_missing('warn', 'Entry for citation', $key);
+        push(@data, { refnum => [$key], title => [$key], attr => { idref => $key, title => $key, class => "ltx_missing_citation" } }); } }
     else {
-      $self->note_missing('warn', 'Entry for citation', $key); } }
+      $self->note_missing('warn', 'Entry for citation', $key);
+      push(@data, { refnum => [$key], title => [$key], attr => { idref => $key, title => $key, class => "ltx_missing_citation" } }); } }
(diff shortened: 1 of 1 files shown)


brucemiller:
redid the patch slightly differently, but should have the same effect (see #592)

PR #596, opened 2015-04-15
[labels] option for RDFa auto-preservation of TeX labels

dginev:
Adds a [labels] option to lxRDFa.sty, which auto-generates RDFa for LaTeX labels.

Example use:
```
latexmlc --preload=article.cls --preload=[labels]lxRDFa.sty 
  --whatsin=fragment --whatsout=fragment --format=html
  'literal:\section{Foo}\label{bar}test' 
```

Output:
```html
<article class="ltx_document" prefix="dcterms: http://purl.org/dc/terms/">
  <section id="S1" class="ltx_section" property="dcterms:alternative" content="bar">
    <h2 class="ltx_title ltx_title_section"><span class="ltx_tag ltx_tag_section">1 </span>Foo</h2>

    <div id="S1.p1" class="ltx_para">
      <p class="ltx_p">test</p>
    </div>
  </section>
</article>
```

### DIFF
--- a/lib/LaTeXML/Package/lxRDFa.sty.ltxml
+++ b/lib/LaTeXML/Package/lxRDFa.sty.ltxml
@@ -16,6 +16,14 @@ use warnings;
 use LaTeXML::Package;
 
 #======================================================================
+# Package Options
+DeclareOption('labels', sub {
+  Let(T_CS('\lxRDF@original@label'),T_CS('\label'));
+  DefMacro('\label Semiverbatim','\lxRDF@original@label{#1}\lxRDFa{property=dcterms:alternative,content=#1}');
+});
+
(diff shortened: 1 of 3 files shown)


dginev:
Also, I think I am using a very appropriate DC term, description states:
```
An alternative name for the resource.
```

dginev:
And just one more related minor enhancement - I am copying over all RDFa attributes to the document root when providing embeddable "fragment" output.

dginev:
Any chance to see this merged / given feedback on soon? It's the last blocking item for my workflow and I would love to have it in master. I promise not to be annoying for a while after :>

dginev:
Thanks a bunch!!

PR #598, opened 2015-04-16
Consistent Model Interpretation

dginev:
In sTeX, we are currently seeing inconsistent XML produced on different runs of LaTeXML, as shown here:
https://github.com/KWARC/sTeX/issues/68#issuecomment-93775585

Interestingly, we were getting inconsistent output only when LaTeXML was starting up from scratch. During daemonized runs the output was consistently the same. This lead me to believe there are possible non-deterministic interpretations of the sTeX schema.

We are using an already pre-compiled model, as in:
```
(Loading compiled schema /home/dreamweaver/localmh/ext/sTeX/schema/omdoc+ltxml.model... 0.03 sec)...... 0.14 sec)
```

Which reduced the possible causes.

As it turns out, we need to be extra careful with sorting hash keys in ```LaTeXML::Common::Model``` and ```LaTeXML::Common::Model::RelaxNG```, as the different orders were creating a different interpretation of the sTeX schema. I have tested this PR and can confirm it solves the sTeX inconsistencies.

### DIFF
--- a/lib/LaTeXML/Common/Model.pm
+++ b/lib/LaTeXML/Common/Model.pm
@@ -297,12 +297,12 @@ sub getXPath {
 
 sub getTags {
   my ($self) = @_;
-  return keys %{ $$self{tagprop} }; }
+  return sort keys %{ $$self{tagprop} }; }
 
 sub getTagContents {
   my ($self, $tag) = @_;
   my $h = $$self{tagprop}{$tag}{model};
-  return $h ? keys %$h : (); }
+  return $h ? sort keys %$h : (); }
 
 sub addTagContent {
   my ($self, $tag, @elements) = @_;
@@ -313,7 +313,7 @@ sub addTagContent {
 sub getTagAttributes {
   my ($self, $tag) = @_;
   my $h = $$self{tagprop}{$tag}{attributes};
-  return $h ? keys %$h : (); }
+  return $h ? sort keys %$h : (); }
 
 sub addTagAttribute {
   my ($self, $tag, @attributes) = @_;
--- a/lib/LaTeXML/Common/Model/RelaxNG.pm
+++ b/lib/LaTeXML/Common/Model/RelaxNG.pm
@@ -69,9 +69,9 @@ sub loadSchema {
 
   # The resulting @schema should contain the "start" of the grammar.
   my ($startcontent) = $self->extractContent('#Document', @schema);
-  $$self{model}->addTagContent('#Document', keys %$startcontent);
+  $$self{model}->addTagContent('#Document', sort keys %$startcontent);
   if ($LaTeXML::Common::Model::RelaxNG::DEBUG) {
-    print STDERR "========================\nStart\n" . join(', ', keys %$startcontent) . "\n"; }
+    print STDERR "========================\nStart\n" . join(', ', sort keys %$startcontent) . "\n"; }
 
   # NOTE: Do something automatic about this too!?!
   # We'll need to generate namespace prefixes for all namespaces found in the doc!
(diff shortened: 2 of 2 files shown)


dginev:
It may well be the case that not all of these sorts are needed for fixing the issue, but I would much rather be on the safe side for future extensions, instead of having to fear encountering this problem again, as the methods and their use evolves.

brucemiller:
Almost forgot about this.  Although it's hard to see why the order should affect your processing... (maybe this will just hide some other error?)

cmaeder:
More needs to be done, see https://github.com/KWARC/LaTeXML/issues/6

brucemiller:
@cmaeder  Presumably more needs to be done somewhere, but I don't think this has anything to do with the base LaTeXML.  It's not the order of elements or attributes that is the issue, but the way you've computed the value of the attribute.  Probably the same sort of thing, however, needing make sure the order of items you're concatenating is reproducible.

cmaeder:
I only call latexmlc as given by https://github.com/KWARC/LaTeXML repeatedly. (I do not compute any value or concatenate items.)

brucemiller:
think about what you're saying...

dginev:
@cmaeder it is important to realize we have a number of people working on a number of projects, all of which calling some of the executables in the LaTeXML, master.
 * Bruce is the creator, developer and maintainer of the core LaTeXML (this repo) and is usually very responsive with fixing core problems. He is also usually kind enough to educate us at KWARC whenever we hit some very intricate situation in the sTeX bindings. Some sTeX errors are caused by master, which is why we have pull requests like this one.
 * On the other hand, at KWARC we have sTeX which includes:
  - an authoring syntax,
  - a document model based on the OMDoc ontology,
  - a set of LaTeX packages realizing the syntax and
  - a set of bindings for LaTeXML, OMDoc post-processing stylesheets and experimental native extensions to LaTeXML.

  sTeX is a KWARC project, created/developed/maintained by Michael, with some help from me for the LaTeXML bindings. Any issues you encounter while converting sTeX documents to OMDoc are first and foremost sTeX issues and you should be opening issues at https://github.com/KWARC/sTeX/issues
* The KWARC/LaTeXML fork has historically served sTeX and sTeX only, so it is a bad place to open new issues - no one will feel responsible for them. I have just disabled the issue tracker to avoid future confusion. 

Again, it is great that you are using LaTeXML for your work, and that you are reporting the problems you run into, but please bear with us in reporting in a way that makes maintaining and prioritizing easier.

dginev:
Merged in KWARC/LaTeXML and I think I added a good fix for the "for" values being jumbled up. @cmaeder if you update both LaTeXML and sTeX things should look better. Feel free to open new issues if you run into other problems.

cmaeder:
if you really want problem reports from users, you should not simply close down the issue tracker.

From my user POV the issue tracker for https://github.com/KWARC/LaTeXML seemed ideal. How should I know if to report here or to  https://github.com/KWARC/sTeX/issues?

If you know what the proper issue tracker is, why don't you simply create an issue there an reference the original problem?

Should I make a new issue for the "make test" failure? (I forget what the other issue was.)

dginev:
Well, I first asked nicely, then I closed it, I think I escalated things appropriately. Also, as I mentioned above, the KWARC/LaTeXML repo is a spin-off for the sTeX project, so any bug related to it can be considered an sTeX issue. We have had the long term desire to fully obsolete that fork, but we're not quite there yet.

If you are using LaTeXML in a project based on sTeX, always open the issue on the sTeX tracker, unless you are fully certain you are seeing a core LaTeXML error. I will happily take the duty of escalating relevant issues to the core LaTeXML repo, just as I did this time.

brucemiller:
And if you *do* think that you've uncovered a core LaTeXML issue, it would likely be best to open a ticket under it's tracker brucemiller/LaTeXML, rather than the KWARC/LaTeXML tracker.  Of course, in such cases, it would be most useful if you can work out an example of the bug that involves just latex and other packages, without sTeX, since I'm not equipped to debug the latter.

PR #600, opened 2015-04-21
Downcase citation keys only for lookup

dginev:
This is a minor change with an important impact for my server-side use of LaTeXML. Key points of understanding:
 * BibTeX keys are case insensitive. LaTeXML uses a lower-cased canonical form to accommodate that, which is great.
 * While lookup should work in a case-insensitive manner, the actual key value shouldn't be changed, in terms of reporting the key back to the log or output.
 * In my use, I need the missing citations to be recorded on output with the exact key written in the source, in order to have some sanity and not worry about special normalizations LaTeXML has a preference towards.

This PR preserves the functional behaviour, while leaving the key intact for output. Feedback welcome!

### DIFF
--- a/lib/LaTeXML/Post/CrossRef.pm
+++ b/lib/LaTeXML/Post/CrossRef.pm
@@ -386,8 +386,7 @@ sub fill_in_bibrefs {
 sub make_bibcite {
   my ($self, $doc, $bibref) = @_;
 
-  # NOTE: bibkeys are downcased when we look them up!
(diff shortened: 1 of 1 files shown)


brucemiller:
Seems like a safe change, as far as it goes. But I think I've encountered cases (arXiv?) where inconsistent case was used in citations, so you may still need to downcase to look them up externally.

dginev:
Thanks!

PR #601, opened 2015-04-30
grffile.sty.ltxml

dginev:
It extends the pathname capabilities of graphicx, which (I am almost certain) is already supported by native LaTeXML.

We use this LaTeX style at Authorea, so I am sending over a stub binding to the LaTeXML master.


brucemiller:
This seems innocuous and helpful; thanks!

brucemiller:
Oh, but you've got to add the file to the MANIFEST before it works;  I went ahead and did that.

dginev:
Well, "it works" without being added to the manifest, I just didn't think ahead to distro concerns. The manifest is used for generating the various tarballs and packages, which I know, but didn't think of. Thanks for the reminder!

PR #602, opened 2015-04-30
fancyref.sty.ltxml

dginev:
This is a barebones implementation, the fancyref package can do more (and has a dependency on varioref). However, I thought I would request the basic support merged in as a start, and see if advanced use cases show up - users can then complain.


dginev:
This PR ended up out of sync, and I think @brucemiller didn't like the idea of only having a partial implementation, so closing here for now.

PR #603, opened 2015-05-04
Keep minipage inline-blocks adjacently inline, when specified

dginev:
Example solution to keeping adjacent {minipage} content in the same inline context, in order to allow for horizontal stacking.


brucemiller:
Your basic idea was correct, but can avoid needing the extra parameter inline.  Even so, it took this example for me to fully realize that latex's tabular (and latexml's ltx:tabular) should have display:inline-table generally, in order to simulate latex!  Thanks; I haven't merged this push, but the equivalent effect, or better, should be in place.

PR #604, opened 2015-05-06
Daemon-friendly listings and literal strings

dginev:
It took me a while to realize why the listing language styles were being reloaded on every daemonized conversion job, and then I saw this wonderfully helpful commented out code.

I would like to make a very warm request for including it, as the speedup it grands in daemon mode is in the order of 5+ seconds, while classic mode wouldn't experience much of a hit.

I also included a very helpful 10 character snippet of the mouth content when reporting that a Literal String is being processed, so that the logs are a lot more informative. For example this helped me figure out I had to be looking at \lstset:
```
(Processing content Literal String \subsectio...
(Processing content Literal String \lstset{ %...
(Digesting TeX Anonymous String...
(Processing content Literal String \end{docum...
(Processing content Literal String \section{{...
(Processing content Literal String \lstset{ %...
(Digesting TeX Anonymous String...
(Processing content Literal String \end{docum...
(Processing content Literal String \section{L...
(Processing content Literal String \lstset{ %... 0.24 sec) 0.26 sec) 0.25 sec) 0.49 sec) 0.49 sec) 0.49 sec)
```

Would be super happy to see this merged!

### DIFF
--- a/lib/LaTeXML/Core.pm
+++ b/lib/LaTeXML/Core.pm
@@ -239,6 +239,7 @@ sub initializeState {
   my $gullet  = $stomach->getGullet;
   $stomach->initialize;
   my $paths = $state->lookupValue('SEARCHPATHS');
+  $state->assignValue('InitialPreloads' => 1, 'global');
   foreach my $preload (@files) {
     my ($options, $type);
     $options = $1 if $preload =~ s/^\[([^\]]*)\]//;
@@ -258,6 +259,7 @@ sub initializeState {
     LaTeXML::Package::InputDefinitions($preload, type => $type,
       handleoptions => $handleoptions, options => $options);
   }
+  $state->assignValue('InitialPreloads' => undef, 'global');
   return; }
 
 sub writeDOM {
--- a/lib/LaTeXML/Core/Mouth.pm
+++ b/lib/LaTeXML/Core/Mouth.pm
@@ -34,7 +34,7 @@ sub create {
     $options{shortsource} = "$name.$ext";
     return $class->new($options{content}, %options); }
   elsif ($source =~ s/^literal://) {    # we've supplied literal data
-    $options{source} = 'Literal String' unless defined $options{source};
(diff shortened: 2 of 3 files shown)


brucemiller:
It's a performance gain for preloading in the daemon, but a loss for command line use. The reason is that it normally loads the language files only until it finds the needed language. This means that in the common case, using common languages, only the first one gets loaded.

So, your suggestion is good but should be conditional upon (1) running in daemon mode and (2) in preload phase.  Are there any flags (like your REMOTE_REQUEST) that we could leverage? Or we could add?

dginev:
I think loading everything if we are doing ```--preload=listings.sty``` is the correct intuition here. Preload has the implied semantics of "load everything in advance", daemon or no daemon.

So all we need is the following addition to ```LaTeXML::Core::initializeState```:
```perl
$state->assignValue('InitialPreloads' => true, 'global');
  foreach my $preload (@files) {
  # ...
  }
$state->assignValue('InitialPreloads' => undef, 'global');
```
Which we can then consult in listings.sty.

Do you agree? If so I will add a commit to this PR and request merging again. (Or you can beat me to it)

brucemiller:
Yeah, well... it's close enough. I don't read *quite* that semantics into "preload" --- sometimes it's pre-loading something that wouldn't have been loaded otherwise (eg latexml.sty switches).
But it's close enough, so please go ahead...

dginev:
Done. And I tested with:

Loads everything:
```
latexmlc --preload=listings.sty 'literal:test'
```

Loads only the core:
```
latexmlc  'literal:\documentclass{article}\usepackage{listings}\begin{document}test\end{document}'
```

Ready for merge!

brucemiller:
looks good; Thanks!

kohlhase:
To me this begs the question: can we leverage this for preloading ```tikz``` as well? That would make a huge difference in our sTeX conversions.

PR #605, opened 2015-05-07
Safe I/O option for latexml.sty

dginev:
This is an important security enhancement for processing third-party TeX code in a web server context.

The PR disables all commands that perform a direct call to ```Package::Input``` or directly create a file Mouth. It may not be complete, but it is an important first step.


dginev:
That is yet another setup, where even my intuition of "string data" vs "file data" breaks, and you need to "taint" everything in a given directory.

But ok, I will close here and implement a local workaround.

brucemiller:
So that this eventually gets looked at a bit more carefully, and not forgotten, I've opened #607.

PR #608, opened 2015-05-11
Preserve cite macro information as class attribute

dginev:
The goal of this pull request is to preserve more of the information we originally had in the TeX source to the XML output of LaTeXML.

In particular, I have added a hook for preserving the macro name of (natbib and core) citation commands (such as ```\citep``` or ```\citet```) as an optional argument to the internal ```\@@cite``` macro. I am happy to refactor if you find flaws in the approach.

My personal goal here is to apply different CSS styling to the different macros, when dealing with an HTML version of the input.


brucemiller:
kinda odd, but OK... :>

PR #609, opened 2015-05-12
Simple crossref fill-in for missing citations

dginev:
For cases such as natbib, the fill-in "show" pattern becomes quite complex (e.g. ```AuthorPhrase1, AuthorPhrase2``` or similar). 

In the case for missing citations, I think we agreed in one of the previous pull requests, that we just want to propagate the citation key back into the XML, as a form of user-facing error report. This behaviour is currently broken under natbib, where you could get ```<cite>``` contents such as ```[]``` or ``` , ```.

So, with this pull request I am doing a simple circumvention of the ```show``` parser in the case of missing citations, which ensures we get the citation key into the final XML output.


dginev:
A quick merge here would be really great, let me know if there is anything to discuss.

PR #610, opened 2015-05-13
Preserve RDFa in MathML

dginev:
This is a stop gap measure to preserve all RDFa attributes in top-level math elements, akin to the way LaTeXML has been preserving image-related attributes.

Maybe it helps move the discussion forward, maybe it gets merged. I have tested and it does exactly what I would expect it to - the MathML <math> element can be seen carrying RDFa attributes e.g. in LaTeXML's HTML output.

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -78,9 +78,12 @@ sub outerWrapper {
       'altimg-width'  => $math->getAttribute('imagewidth'),
       'altimg-height' => $math->getAttribute('imageheight'),
       'altimg-valign' => ($depth ? -$depth : undef)); }        # Note the sign!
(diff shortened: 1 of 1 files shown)


dginev:
Also, you can assume I have time available right now to work on the "RDFa preservation in MathML", and it is somewhat of a priority for me, so I am ready to react on any feedback on how to implement this better.

Again, these are LaTeXML-native attributes, so I am unsure we want to group them with the same treatment we were discussing about foreign attributes.

brucemiller:
Ah, so they're native! "It's a TRAP!!!"

Actually, your patch seems pretty innocuous, and probably essentially correct, as far as it goes. I would think that similar code would want to to be in stylizeContent and/or pmml_mi, pmml_mo, cmml_ci?  Should I wait for you to look into that, or just merge what you have for now?

dginev:
I would say let's merge and test this now, and evolve it to further elements later, when we have the need and certainty it's an acceptable approach.

I am still wondering on how to force certain types of RDFa to float up to certain elements, so it bears some extra thought. But merging here would be super!

PR #615, opened 2015-05-24
Search for profiles in LaTeXML's --paths

dginev:
Until now, LaTeXML would only look for profile ```.opt``` files in its own resource directory, ignoring user-supplied paths.

However, this hinders using custom profiles, and is also problematic for testing plugin modules. As ```make test``` executes the tests using the built ```blib``` directory, prior to installing LaTeXML, it becomes a necessity to be able to specify the built directories as search paths for LaTeXML, as in this example:
https://github.com/KWARC/LaTeXML-Plugin-MathWebSearch/commit/2b8838b3e5dc6a6105572c27d0dc52075b646c7b#diff-d0d26286eac2af991850ff3ab61960c8R16

My other request is that after merging here we also increment the minor version number of LaTeXML and push a new release to CPAN, so that I can have a successfully building plugin release of ```LaTeXML::Plugin::MathWebSearch```. This is part of our testing process for establishing an effective and convenient plug-in ecosystem.

### DIFF
--- a/lib/LaTeXML/Common/Config.pm
+++ b/lib/LaTeXML/Common/Config.pm
@@ -308,7 +308,7 @@ sub _obey_profile {
   if ($profile ne 'custom') {
     if (defined $$PROFILES_DB{$profile}) {
       %$profile_opts = %{ $$PROFILES_DB{$profile} }
-    } elsif (my $file = pathname_find($profile . '.opt', paths => [],
+    } elsif (my $file = pathname_find($profile . '.opt', paths => $$opts{paths},
         types => [], installation_subdir => 'resources/Profiles')) {
       my $conf_tmp = LaTeXML::Common::Config->new;
       $conf_tmp->read(_read_options_file($file));


kohlhase:
I support this change, it would make profiles much more useful.

dginev:
My LaTeXML plugin keeps generating CPAN test failures that I get daily email digests of, so pinging this PR again. It's a nice general capability I think, but feedback is welcome as always.

dginev:
Should I ping here again? @brucemiller let me know what you would like to see done differently here.

brucemiller:
Well, you have to ping me to force me out of avoiding saying:  This is basically a very good idea, but is also part of the Grand Configuration scheme.  If we make profiles actually useful, then people will use them, and then we'll have a bigger compatibility issue when we finally want to redo it.  So, I don't really want to do it. As it is now, if I make profiles obsolete, I've only got one person to deal with (not that it will be easy to convince that person! :> )

dginev:
Oh, it's too useful to be merged. Got it :>

However, I can still remove the test case and have my plugin sitting on CPAN and nastily writing into the LaTeXML namespace. We're talking about Perl here. You can't stop me :> And it's probably a very low risk that anyone else will write LaTeXML plugins in the immediate future, plus this change (1 line) is incredibly easy to undo.

In any case, I would appreciate if you close such PRs without merging and with the explanation, so that I make progress using other "methods". If it goes stale no one benefits. Thanks!

brucemiller:
Yup, too useful. It isn't desirable to reject basically good ideas out-of-hand, nor is it always easy to figure out how to morph good ideas into great ideas.  But maybe there's an evolutionary solution here.

So, where I eventually want to go is more useful "configuration" files, or we can even call them "profiles" to avoid a change; has a slightly different connotation to me, but I could adapt. Ultimately, the language for options should become richer, but that can just be absorbed in the current profiles. Likely recursive inclusion of profiles is easily accommodated as well. If we end up wanting to change the option language incompatibly, that'll be a disruption in either case, so that doesn't matter either.

So, I guess the potential blocker for me is naming. I can live with "profile", but I'm kinda concerned with the generic "opt" extension; particularly if they're going to be searched for over a wider portion of the filesystem, rather than one locked directory.  It appears that "opt" is currently assumed, or you can explicitly specify "foo.opt". Hmm... What would be a better, more purposeful extension?  I'd have liked "latexml", but some shortsited person seems to have used that for something else.  And how disruptive would it be to change the extension now, and before the wider user community starts using them?

dginev:
We can change all .opt files on a short notice. I think it is still safe to say that I have write access to any repository that has its own .opt file (I would be surprised to hear otherwise).

If that will make this PR go through, sure, let's decide on an extension and do the refactoring - I can take care of that.

How about something like ```.lxopt``` or ```.ltxml.opt```?

kohlhase:
Actually removing options would be a two-person deal :-). 

But I agree with Bruce that the ```.opt``` extension is not nice. I never liked that, since there is no way to remember the relation between "profile" and ```.opt``` (even though I understand where the ```.opt``` extension comes from). So how about ```.ltxprofile``` or ```.ltxml.profile```? As there are few profiles having long and expressive extensions may be a good thing.

dginev:
Just to make clear, I wasn't suggesting removing anything from KWARC repositories, I only said it is possible for me to mass-refactor all file extensions to a different string. Renames shouldn't be controversial, and if LaTeXML imposes a new name, the rest of the repos have to conform anyway.

kohlhase:
@dginev That was exactly what I understood.

brucemiller:
As this is public record, I probably should point out that the short-sited person who used the extension "latexml" is *me*! ( I should have tried for something like mydoc.documentcustomization.ltxm :> )

```.ltxprofile``` seems nicely unambiguous.  Renaming the files and changing ```.opt``` to ```.ltxprofile``` is trivial enough; synchronizing with various plugins slightly less so.  But is profile advertsed as ```--profile=fragment``` or ```--profile=fragment.opt```?  I think the former, in which case the change would be transparent; but I notice the latter also works.

@dginev Assuming you're ok with this, should I just go ahead & merge this and wait for a 2nd pull, or wait for it all at once?

dginev:
I think you should merge here first, for convenience's sake.

Also, breaking changes usually require a version bump, so that we can have dependency sanity, so i will also ask you (again) to release on CPAN after the rename. Even if we just move to "0.8.1.1", I can add that to the Makefile.PL of all LaTeXML-depending modules, so that the sync works well.

dginev:
What I am trying to avoid is having a plugin move to ```.ltxprofile``` and then fetch a LaTeXML from CPAN that expects ```.opt```

brucemiller:
right. but of course, the change will precede formalizing a release. And any modules that depend on cpan will currently get 0.8.1 anyway. Unless they're doing cpanm or something, in which case they're asking for the bleeding edge.

dginev:
I suggest we do the file extension renaming, as well as other breaking changes, immediately before we make a CPAN release. In other words, do the renames last. Exactly to avoid such confusions about the current 0.8.1 and the bleeding edge.

dginev:
Similarly, the value of this PR that you just merged (thanks!) will be offered to my CPAN-breaking plugin, only after LaTeXML updates itself on CPAN and not before.

PR #619, opened 2015-05-25
Legible error messages for timeouts and interrupts

### LINKED ISSUE #618: Legible timeout error message
@fred-wang couldn't make heads-or-tails from the timeout Fatal error I had introduced in LaTeXML.pm some time back, and he is a power user.

That is certainly a sign that particular Fatal message needs to be improved in terms of user-facing clarity. And it's not hard to do so, just recording here for bookkeeping. I'll take care of this.

dginev:
Fixes #618 .

Not exactly sure if undefing the ```_DIE_``` signal is the optimal way of getting rid of the die Fatal error, but it works exactly as I expect it. With this PR the interrupt and timeout Fatals are cleanly displayed without any extra-wrappers.

An example with ```latexmlc --timeout=2 'literal:\def\foo{\foo\foo}\foo'```:
```
(Digesting TeX Anonymous String...
(Processing content Literal String \def\foo{\...
Fatal:conversion:timeout Conversion timed out after 2 seconds!
	at Literal String \def\foo{\; line 1 col 23
	In \foo from at String; line 1 col 18
	 <= Core::Gullet[@0x3750258] <= Core::Stomach[@0x3750150] <= ...
1 fatal error at /home/dreamweaver/git/my-LaTeXML/blib/script/../lib/LaTeXML/Common/Error.pm line 69.


Conversion complete: 1 fatal error.
Status:conversion:3
```

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -209,7 +209,7 @@ sub convert {
   # 2 Beginning Core conversion - digest the source:
   my ($digested, $dom, $serialized) = (undef, undef, undef);
   my $convert_eval_return = eval {
-    local $SIG{'ALRM'} = sub { die "Fatal:conversion:timeout Conversion timed out after " . $$opts{timeout} . " seconds!\n"; };
+    $$latexml{timeout} = $$opts{timeout}; # For reporting purposes
     alarm($$opts{timeout});
     my $mode = ($$opts{type} eq 'auto') ? 'TeX' : $$opts{type};
     $digested = $latexml->digestFile($source, preamble => $current_preamble,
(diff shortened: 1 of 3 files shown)


dginev:
I also moved the alarm handler in ```withState``` so that I have access to the LaTeXML error-reporting API. That is not available from LaTeXML.pm.

It can probably get even better in the long-run, but it's solving the immediate problem already.

dginev:
Also note that this improves the core processing error message, the post-processing timeouts are a completely separate block of code (only available in LaTeXML.pm)

dginev:
I am guessing this PR is obsolete now, closing.

PR #621, opened 2015-05-28
nameref.sty.ltxml

dginev:
* Added support for nameref.sty.ltxml
 * Fixed an argument counting bug for \autoref, cleaned up double definition
 * Trimmed "tag"s from shown titles in CrossRef
 * Added classes for ref macro names 
  * this bit is openly Authorea-motivated, as we are doing the labels and referencing outside of LaTeXML and would like to know the macro information, so that we created the right label
 * Added tests

### DIFF
--- a/lib/LaTeXML/Package/hyperref.sty.ltxml
+++ b/lib/LaTeXML/Package/hyperref.sty.ltxml
@@ -16,6 +16,7 @@ use warnings;
 use LaTeXML::Package;
 
 RequirePackage('url');
+RequirePackage('nameref');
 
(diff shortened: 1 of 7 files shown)


brucemiller:
For testing purposes, I went ahead & merged.  But I'd still be happy to take an updated nameref.sty! :>

dginev:
Cool, thanks for the merge! It's probably wise you grabbed the current state, as I am a bit distracted at the moment with preparing the Authorea infrastructure for \nameref and \autoref (as we're doing them client-side, for scalability reasons).

I'll get back to them some time next week though.

dginev:
Another note - I am about to start relying on the ```ltx_refmacro_macroname``` classes to figure out which ref macros need to be completed (externally to LaTeXML), so I would be really grateful if you let me know before renaming them to something different and/or decide on removing them. Ideally we just keep them there, it's just helpful extra information.

PR #638, opened 2015-07-10
Reversion for label

dginev:
The actual change here is a single line, which sets the label reversion to:
```
   reversion   => '\label{#1}', 
```

The rest of the pull request is updating all tests files. @brucemiller do you see any issues with making label revertable? It makes preserving the literal math tex even better (esp. for eqnarrays)

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -2649,7 +2649,7 @@ DefConstructor('\label Semiverbatim', sub {
       my %labels = map { ($_ => 1) } $label, split(/\s+/, $node->getAttribute('labels') || '');
       $document->setAttribute($node, labels => join(' ', sort keys %labels));
       $document->setNode($savenode); } },
-  reversion   => '',
+  reversion   => '\label{#1}',
   properties  => { alignmentSkippable => 1 },
   afterDigest => sub {
     my $label = CleanLabel(ToString($_[1]->getArg(1)));
--- a/t/alignment/eqnarray.xml
+++ b/t/alignment/eqnarray.xml
@@ -8,7 +8,7 @@
     <equationgroup class="ltx_eqn_eqnarray" xml:id="S0.EGx1">
       <equation labels="LABEL:e1" xml:id="S0.E1" refnum="1" frefnum="(1)">
         <MathFork>
-          <Math xml:id="S0.E1.m4" tex="\displaystyle x=17y" text="x = 17 * y">
+          <Math xml:id="S0.E1.m4" tex="\displaystyle x=17y\label{e1}" text="x = 17 * y">
             <XMath>
               <XMApp>
                 <XMTok meaning="equals" role="RELOP">=</XMTok>
@@ -38,7 +38,7 @@
                 </Math>
               </td>
               <td align="left">
-                <Math mode="inline" xml:id="S0.E1.m3" tex="\displaystyle 17y" text="17 * y">
+                <Math mode="inline" xml:id="S0.E1.m3" tex="\displaystyle 17y\label{e1}" text="17 * y">
                   <XMath>
                     <XMApp>
                       <XMTok meaning="times" role="MULOP">⁢</XMTok>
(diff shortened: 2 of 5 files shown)


brucemiller:
Yeah, actually, I do.  From my point of view, it really isn't "math" (of course, neither is \nonumber), but is part of the containing document that just happens to need to be stored in the math.  The TeX for equations (or pictures) is for 2 purposes, generating images and user's cut-n-paste.  In those cases, the label is potentially error causing, though probably not, and also annoying cruft for the user.

So, I really am not too fond of this idea.  And in fact, I should have objected to \nonumber, on the same grounds, but you caught me in a weak moment.

Before you object: things like eqnarray are peculiar monstrosities.  The aim, is to have the TeX strings for the cells (which we have), and also TeX strings for the complete equations (which we haven't yet).  The former allow creating a table (at document level) of the images of math fragments; the latter allow to provide cut-n-paste of the logical (if not necessarily semantic) equations.  As you can see, in neither case are \label an \nonumber relevant or desirable.

And also, remember the "literal" TeX is not necessarily what the author wrote.

dginev:
The practical need for the label is so that math processing engines that deal with equation numbers, can properly recover the label for later referencing. MathJaX has the capability of both generating equation numbers and later ```\ref```ing to them, which is something currently used at Authorea.

The principle reason is that even though it is called "math mode" clearly not all TeX code in that mode modifies the mathematical content. There can be equation numbers, vertical and horizontal spacing, labels, metadata, definitions of new macros, and other non-mathematical instructions.

I think recording all present TeX primitives is a fair expectation, even if it may be contrary to the intention of recording the math-only instructions.

As we have discussed before, maybe we need two separate reversion modes, one "pragmatic" and one "strict".

brucemiller:
How does MathJax deal with \ref?  Where do you PUT the \ref? (that's normally within the document text, not math).

And most of that stuff doesn't modify the mathematical content; it modifies the connection between the mathematical content and the document.  Except for defining new macros which might modify the content, but are expanded away and you just get the modified content.

At any rate, it's got nothing to do with "pragmatic", it's just what you want; call it "Authorea mode".

dginev:
The ref is just a formula with just a ref in it: ```\[\ref{foo}\]```. 

I was actually thinking that "strict" mode would be what I want, where all TeX commands are preserved as closely as possible to the original, while the "pragmatic" mode is what you need for LaTeXML to do its math handling well.

I am happy to combine recording primitives with our older discussion of trying to record unexpanded macros, so that we get the most user-friendly high level TeX for copy-paste, assuming the user has the same definitions as the paper.

And yes, my practical need is related to Authorea, but the general discussion is something we've also discussed in the context of DRMF and Planetary.

brucemiller:
```\[\ref{foo}\]``` ?!?!? Surely you're joking.

Of course, I'm not privy to all Planetary discussions (and I'm mostly thinking of PlanetMath), but PlanetMath wouldn't need to preserve such nonsense in the TeX string of the math because it is dealing with documents and MathML in a consistent fashion --- or so it seems to me.

I don't quite understand what the 3rd paragraph refers to.

It sounds like what you & DRMF want is to redefine {equation} in terms of {verbatim}. Problem solved.

brucemiller:
Really, if you want to reassemble some sort of eqnarray, with labels & nonumbers, you should be checking the attributes on the ltx:equation's that comprise the structure you're trying to rebuild. (ie. perserving \nonumber was a mistake)

brucemiller:
Regrettably, I can't really accept this pull, and after thinking about it reverted the previous one for \nonumber.

I think if this is need and worth doing, it probably should involve some sort of reconstruction of the TeX at the ltx:equationgroup level; perhaps equationgroup could have a tex attribute. Then you're in a position to reconstruct the labels, nonumbers, tags, intertext and whatever else you're wanting, as suggested in the previous comment, and also know whether it's an eqnarray, split, align, etc environment that you should be using rather than assuming eqnarray.

Otherwise, it really sounds like you don't want LaTeXML to do anything at all with the equation, and probably should redefine them as verbatim.

brucemiller:
Oh, and a clarification of my 1st comment: the synthesized equations from an eqnarray, along with their TeX strings, *is* present in the LaTeXML XML, but it doesn't (yet) migrate to the html, as I wasn't clear how/where it should be placed; it essentially should be available but hidden from casual view.

dginev:
sigh

PR #642, opened 2015-07-13
whatsout math returns image content when imagesrc attr is set

dginev:
First CICM pull request! Moritz wanted an easy access to the SVG content of the math conversion, see discussion here:
https://github.com/dginev/LaTeXML-Plugin-ltxpsgi/issues/3

I am quite unsure if File::Slurp is the best way to read the content, but this actually works, so that you get only SVG when you do:
```
latexmlc --whatsin=math 'literal:abc' --whatsout=math --mathsvg 
```

But I would really appreciate advice and a code review by @brucemiller , since I'm not sure I fully understand how imagesrc attributes interact with formulas, in particular when there is parallel markup.


physikerwelt:
@hcohl: You can track the status of our SVG issue here.

PR #646, opened 2015-08-06
Upgrades to LaTex.pool.ltxml

lkohlhase:
I added the definition for \loop and for all the \newcounts, \newdims, \newifs etc. I found in latex.ltx. 

I also added some very short macros that seemed like aliases suchs \dischyph=\- .


brucemiller:
Thanks! looks good.

Issue #652, opened 2015-08-16
RFE: Provide a LaTeXML-maybeAltimg.js script


fred-wang:
In LaTeXML/lib/LaTeXML/resources/javascript/ we have a script to optionally run MathJax.

@gpetty mentioned that the SVG images generated by the --mathsvg options are linked as altimg image. I guess there are options to generate PNG fallback too.

So I propose to have a LaTeXML-maybeAltimg.js script to optionally convert all the math tags into img tags with the math@altimg as the img@src attribute and math@altext value used for img@alt.

fred-wang:
(I'll try & work on that)

brucemiller:
Cool! Actually, I'd been talking to @dginev about setting up something similar, but going a bit further. My idea was to generate & serve MathML with both an svg & png annotations (presumably the latter would be set as the altimg?).  Then use feature detection to select (1) MathML (essentially only real Gecko's), (2) svg image, if supported, and finally (3) png image.  Perhaps it could automatically adjust to what annotations are present (eg. svg)?

I'm thinking it might be better to have 1 "smart" script that can handle most cases rather than a plethora of mixtures, but I'm not sure.

Bonus points if there're hooks to choose/force preferences with cookies! :> (then could even have the means to select MathJax, if user wishes)

fred-wang:
I think we had discussions in the past about the "right" way to propose the image fallback and the MathML + altimg / semantics annotation were proposed but I don't remember what LaTeXML generates at the end (and this may also be dependent of command line options...). Since I was not able to use the --mathsvg command, I just assumed it was in altimg as mentioned on the other issue.

Note that the CSS vertical-align property is important to correctly align the image of an inline formula with the surrounding text but I'm not sure whether LaTeXML generates it (or whether it's actually possible/easy to calculate it). The width/height values also help to avoid reflow when the image is loaded.

The cookie stuff should be easy, although there might be a privacy concern here (one cookie will be created for each domain)... but probably people already using MathJax don't care about that.

fred-wang:
Regarding merging all in one "smart" script: I guess this would also mean adding an option for the CSS-based rendering of https://github.com/brucemiller/LaTeXML/issues/655

FYI, the following script provides the CSS/MathJax fallback and cookie features (however it checks for mpadded support instead of sniffing the browser user agent string):
http://fred-wang.github.io/mathml-warning.js/
https://github.com/fred-wang/mathml-warning.js

dginev:
[...]

PR #654, opened 2015-08-16
Add a maybeAltimg script to lib/LaTeXML/resources/javascript. #652

fred-wang:
Here is a proposal for #652.

Only tested with Chromium on Linux.

(I was not able to use the --mathsvg option though)

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -290,6 +290,7 @@ lib/LaTeXML/resources/CSS/ltx-apj.css
 lib/LaTeXML/resources/CSS/ltx-listings.css
 lib/LaTeXML/resources/CSS/ltx-svjour.css
 lib/LaTeXML/resources/CSS/ltx-ulem.css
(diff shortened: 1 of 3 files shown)


fred-wang:
No activity for more than one year, so I'm closing this pull request and deleting my fork.

dginev:
Up to you @fred-wang . As far as I can see in #652 there was no deadline on this issue and we have pushed it back to a further-along milestone. I'm sure if you need this in master sooner / now Bruce could take a fresh look and see if this can be reevaluated.

Similar to the delay of this PR, I have a number of minor bits that are unlikely to get looked into soon, such as the [secureio](https://github.com/dginev/LaTeXML-Plugin-secureio) plugin, and I have relegated them to plugins until the time comes where their main issues can be tackled.

In any case, sending a "ping" that shows continuing interest in the issue could have given this extra priority, it's always hard to pick the next item on the list. I personally haven't looked into this PR in particular, since I assumed it is for Bruce to evaluate it next.

PR #661, opened 2015-09-15
Redetermine colorize logging any time we rebind STDERR

dginev:
I was bitten by the colors while parsing STDERR from latexmls, which binds STDERR **after** the Error module is loaded.

Using a package-level var for ```COLORIZED_LOGGING``` is nice for performance, so I kept that as-is, but also added the checks to ```bind_log``` and ```flush_log``` which are responsible for changing STDERR.

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -650,6 +650,7 @@ sub bind_log {
     *STDERR_SAVED = *STDERR;
     *STDERR       = *$log_handle;
     binmode(STDERR, ':encoding(UTF-8)');
+    $LaTeXML::Common::Error::COLORIZED_LOGGING = -t STDERR;
     $$self{log_handle} = $log_handle;
   }
   return; }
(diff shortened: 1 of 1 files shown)


dginev:
Thanks for merging this Bruce! For the brief time this was a problem, Authorea was left without LaTeXML error messages, as latexmls was **always** receiving colored messages. Great to have our reports back!

brucemiller:
On 09/16/2015 01:19 PM, Deyan Ginev wrote:
> Thanks for merging this Bruce! For the brief time this was a problem,
> Authorea was left without LaTeXML error messages, as latexmls was
> *always* receiving colored messages. Great to have our reports back!

Oh! Sorry for the delay in merging (was bogged in something else).

dginev:
Well, it wasn't your problem and we can always use our own fork if push comes to shove. So far using the official master repo has been great though, I hope we can keep tracking that and avoid any risks of diverging!

PR #662, opened 2015-09-16
Change xsl inclusions.

angerhang:
This is suggested for two reasons:

(1) changing `import` to 'include' doesn't really have any impact on `LaTeXML-all-xhtml.xsl` itself since this file only centralizes the inclusions but doesn't really contain any template that will be influenced by the precedence of stylesheet.

(2) for the use cases as I reported in https://github.com/KWARC/sTeX/issues/133, the developers will tend to include the LaTeXML stylesheet directly, so it is much better if the LaTeXML's stylesheets use `include` as well, to avoid cases like the null-function of the deletion function, since very likely the users will have a copy template in the root stylesheet to include all the elements, which overwrites the deletion functions.
```
<xsl:template match="*">
  <xsl:copy><xsl:apply-templates select="*|@*|text()"/></xsl:copy>
</xsl:template>
<xsl:template match="@*"><xsl:copy-of select="."/></xsl:template>
``` 

P.S. for someone who is not really familiar with the stylesheet, like me, messing with `XSLT` is really a danger zone :>

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-all-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-all-xhtml.xsl
@@ -20,19 +20,19 @@
 
   <!-- Include all LaTeXML to xhtml modules -->
   <!-- Note that you can include these in your own stylesheet using urns like:
-       <xsl:import href="urn:x-LaTeXML:XSLT:LaTeXML-common.xsl"/>
+       <xsl:include href="urn:x-LaTeXML:XSLT:LaTeXML-common.xsl"/>
   -->
 
-  <xsl:import href="LaTeXML-common.xsl"/>
-  <xsl:import href="LaTeXML-inline-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-block-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-misc-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-meta-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-para-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-math-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-tabular-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-picture-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-structure-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-bib-xhtml.xsl"/>
-  <xsl:import href="LaTeXML-webpage-xhtml.xsl"/>
+  <xsl:include href="LaTeXML-common.xsl"/>
+  <xsl:include href="LaTeXML-inline-xhtml.xsl"/>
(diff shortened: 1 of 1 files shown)


brucemiller:
I don't think I want to do this; import was intentionally used so that you *can* override the built-in templates.  I would have thought that was a feature that sTeX would particularly want (my DLMF sheets use it).

I don't quite follow the situation in the referenced issue. It sounds like either a name clash for templates (in which case both of the templates would be the wrong one half the time), or a template that is defined in your enclosing stylesheet that you *want* to be ignored (using the LaTeXML one instead).  If the latter, why define that template at all?

angerhang:
In fact let me explain the situation that I countered in more detail:
There is a stylesheet `htmlpost.xsl` that uses 
```xsl
<!-- the fallback (mostly for LaTeXML-generated XHTML: 
     copy the whole thing recursively, until there is something to do -->
<xsl:template match="*">
  <xsl:copy><xsl:apply-templates select="*|@*|text()"/></xsl:copy>
</xsl:template>
<xsl:template match="@*"><xsl:copy-of select="."/></xsl:template>
<!-- some general LaTeXML attributes we want to lose -->
<xsl:template match="@fragid"/>
```
to copy everything over, but it crashes with
```xsl
  <!-- By default, these disappear -->
  <xsl:template match="ltx:resource"/>
```
which is the delete template for resource. In this situation, if we we use import, the resource tag will appear in the places that we don't want it to be. That's why `include` would have been a better choice.

There are a few sTeX specific templates in `htmlpost.xsl` such that merely using LaTeXML's stylesheet wouldn't be the best idea. I think this kind of situation is not uncommon, as people who write their own bindings very likely also have some control sequences that they want to add into the stylesheets. 

I probably might have overlooked other aspects, but as always feel free to point them out : )

brucemiller:
I still don't quite get it. Your copying template in htmlpost.xsl is for copying LaTeXML's markup into the html?  If so (for whatever reason), shouldn't that be more explicitly controlled, such as with a named template, or a mode, or ... ?

dginev:
@angerhang can you explain further? I am afraid I'm also a bit confused by the discussion so far, still unsure about the concrete trade-offs we are looking at.

dginev:
Upon rereading I think there are other workarounds here (named modes, rearranging rules, etc.) and since DLMF uses the import approach already this is very likely a "wontfix" for now.

If it turns out there is a fundamental flaw with "import" we can always backtrack.

angerhang:
Thanks for closing.

At the moment sTeX's html branch is waiting for the LaTeXML's 0.8.3 release for the `format=html5` feature, and after the adoption I think this change is no longer needed.

dginev:
The next release is 0.8.2, but ```--format=html5``` is already available in LaTeXML. I am guessing you are thinking of the HTML5 output for sTeX?

Which features are you waiting for exactly? Or you just want a tagged version for stability?

angerhang:
As detailed in https://github.com/KWARC/sTeX/issues/135, I think we are waiting for ''with the upcoming improvements in 0.8.3 (ETA early 2016) you could have an entire format-based profile, allowing to select or deselect various post-processors.''

PR #665, opened 2015-09-25
Unset whatsout was raising a warning, silence it

dginev:
This is a cosmetic PR, following up on my last encoding patch, aiming to avoid an annoying warning when ```--whatsout``` is not specified to latexmlc.

Here is the issue requesting it:
https://github.com/KWARC/sTeX/issues/136


brucemiller:
sounds good.

PR #666, opened 2015-10-02
Defensive upper-bounds for ImageMagick conversion jobs

dginev:
This was a bit tricky to get right, but I think I got the best possible solution here.

I started with a pure-Perl solution adding:
```perl
push(@args, 'memory-limit', '512MiB',
                      'map-limit', '1GiB',
                      'disk-limit', '2GiB',
                      'time-limit', 300);
```
before the first 'Set' operation. However, with testing I found out that these options **override** the environmental variables, which means that users would be unable to modify these LaTeXML settings.

The solution in the PR allows users to specify ImageMagick environmental variables **from outside** to LaTeXML, and thus adjust the settings as needed. I only set a default (arXMLiv-motivated) upper bound for memory, map, disk and time, if they were not already set prior to the latexml process starting. I tested this extensively and it works exactly as expected.

@brucemiller had mentioned he would enjoy also having a time upper bound - and I bring good news - ImageMagick indeed has such a bound, and it seems to work rather nicely for the examples I have. I set it to 5 minutes (300 seconds) for this PR, as it seemed a reasonable time to wait for an image conversion job. It indeed works well - I tested it with 1 second as the bound and it performed exactly as you would expect.

The disk limit is set surprisingly (to me) high, because I had worst-case examples (selected arXiv ```.eps``` images) that took 1.7 GB of scratch disk space before it generated the final 330KB PNG image (!!). 

The really good news for me: when an ImageMagick job terminates because of surpassing the upper bounds, the Perl wrapper manages to receive a clean return code and - and this is crucial for arXMLiv - **correctly clean up the temporary files from the disk**. Which avoids memory overflows due to accumulated failed jobs.

I have very high hopes that with this PR merged, CorTeX will be able to convert all of arXiv without further ImageMagick issues.

### DIFF
--- a/lib/LaTeXML/Util/Image.pm
+++ b/lib/LaTeXML/Util/Image.pm
@@ -38,6 +38,12 @@ our $DPI            = 90;               # [CONSTANT]
 our $DOTS_PER_POINT = ($DPI / 72.0);    # [CONSTANT] Dots per point.
 our $BACKGROUND     = "#FFFFFF";        # [CONSTANT]
 
+# We also provide defaults for certain defensive environmental variables, if not yet set
+$ENV{MAGICK_DISK_LIMIT} = "2GiB" unless defined $ENV{MAGICK_DISK_LIMIT};
+$ENV{MAGICK_MEMORY_LIMIT} = "512MiB" unless defined $ENV{MAGICK_MEMORY_LIMIT};
+$ENV{MAGICK_MAP_LIMIT} = "1GiB" unless defined $ENV{MAGICK_MAP_LIMIT};
+$ENV{MAGICK_TIME_LIMIT} = "300" unless defined $ENV{MAGICK_TIME_LIMIT};
+
 # Note that Image::Size my, itself, use Image::Magick, if available,
 # as a fallback for getting image size & type!!!
 sub image_type {


dginev:
Although, as indicated by the solution, I can adjust the arXMLiv conversion jobs without LaTeXML ever being aware of these limits, as the environmental variables are external to the process.

I think this is actually a customization solution we have overlooked in LaTeXML so far - having our own environmental variables. They are a great solution for certain runtime bounds (as exemplified here) that could be passed independently of where exactly the LaTeXML library is embedded. And we could remove the ```--timeout```, ```--expire``` and related options from the command-line, simplifying the user experience.

dginev:
As mentioned above, the PR is not urgent for my arXMLiv conversion, as I could add these limits externally to LaTeXML, which I have done today. I have a CorTeX "tex_to_html" worker implemented as a LaTeXML plugin here:
https://github.com/dginev/LaTeXML-Plugin-Cortex

Added the environmental settings here:
https://github.com/dginev/LaTeXML-Plugin-Cortex/blob/master/bin/latexml_worker#L50

Currently rerunning, will see if any arXiv article still has significant image conversion problems.

brucemiller:
Looks good.  As soon as I merge this back into my local repo, I'll push some further cleanup of the id management in postprocessing and more informative error messages during post processing.  I also took the liberty (cough) of removing the redundant "Fatal:blah..." from LaTeXML.pm; the ```$@``` should have sufficient info w/o upping the error counts.

As far as environment variables simplifying the user experience: I'm more inclined to think the opposite, actually.

brucemiller:
OBTW: It'd be nice if you updated before rerunning, since then we (hopefully) can make sense of the Fatals.  It only takes, like 5 minutes to run 5000 doc, right :>

dginev:
Yep, but I will rerun all fatals anyway. I am currently completely focused on having a robust and reliable conversion build system, since that is the foundation for both arXMLiv and LLaMaPUn.

We'll certainly make LaTeXML better after that, and I am quite open to updating and rerunning often - it will be a lot of fun to do so, once the framework is 100% reliable. I just caught another weird error that needs addressing (related to lifetime of temporary files and the way the archives get written to disk), so still some work to go. And found that HULK is currently having a hard limit at 256 CPUs for me...

dginev:
FYI the current rerun uses the latest LaTeXML, I updated everything after adding the two patches mentioned in my last comment. Let's see how it goes.

brucemiller:
Oh, before I forget to mention: I just pushed a little patch to my previous commit such that we prepend ```Fatal::conversion:unknown``` to an ```$@``` if we die with an error message, but it doesn't start with Fatal.  This will help bookkeeping, but doesn't do anything about errors leaking out.

dginev:
I'll add that now, since I am restarting the entire conversion harness anyway - the Rust folks helped me localized a file descriptor leak, it seems Rust's automagic file closing isn't automagic enough in certain scenarios (still unsure if I fixed this properly).

Issue #611, opened 2015-05-14
Support for nameref.sty (needed by hyperref.sty)


dginev:
As Authorea has now entered beta testing of LaTeXML, we have started receiving helpful user reports of various issues and missing features. I will be opening LaTeXML issues for those, and will introduce a new "Authorea" label for them, so that I can keep track myself (and give them priority).

This one is quite straightforward: hyperref supports a range of extended referencing macros, ```\nameref``` being one of the most notable ones. Support for those is missing in LaTeXML (I think we only have the "core" of hyperref implemented?) so we should think of adding the missing bits.

Assigning this issue to myself for now, but please feel free to steal it from me if it seems fun.

brucemiller:
That's actually in nameref.sty but I can't tell if that's always loaded or only with certain options.

dginev:
This works out of the box in pdflatex:
```latex
\documentclass{article}
\usepackage{hyperref}
\begin{document}

\section{First}\label{sec:first}
  \nameref{sec:second}

\section{Second}\label{sec:second}
  \nameref{sec:third}

\section{Third}\label{sec:third}
  \nameref{sec:first}

\end{document}
```

Hyperref auto-loads nameref.sty, and a few others, as can be seen from the log:
```
/usr/local/texlive/2013/texmf-dist/tex/latex/hyperref/hyperref.sty
(/usr/local/texlive/2013/texmf-dist/tex/generic/oberdiek/hobsub-hyperref.sty
(/usr/local/texlive/2013/texmf-dist/tex/generic/oberdiek/hobsub-generic.sty))
(/usr/local/texlive/2013/texmf-dist/tex/latex/graphics/keyval.sty)
(/usr/local/texlive/2013/texmf-dist/tex/generic/ifxetex/ifxetex.sty)
(/usr/local/texlive/2013/texmf-dist/tex/latex/oberdiek/auxhook.sty)
(/usr/local/texlive/2013/texmf-dist/tex/latex/oberdiek/kvoptions.sty)
(/usr/local/texlive/2013/texmf-dist/tex/latex/hyperref/pd1enc.def)
(/usr/local/texlive/2013/texmf-dist/tex/latex/latexconfig/hyperref.cfg)
(/usr/local/texlive/2013/texmf-dist/tex/latex/url/url.sty))

Package hyperref Message: Driver (autodetected): hpdftex.

(/usr/local/texlive/2013/texmf-dist/tex/latex/hyperref/hpdftex.def
[...]

PR #669, opened 2015-10-13
Basic support for the family of nameref macros

dginev:
I wanted to see if we can get #611 closed with a simple binding extension. Some of the details around the page-enhanced referencing simply don't apply to HTML, and while they can be reimagined, I don't have the incentive to get the perfectly right now.

The docs for the package are [here](http://mirror.ctan.org/macros/latex/contrib/hyperref/nameref.pdf)


brucemiller:
sounds reasonable....

PR #705, opened 2016-01-10
Unified error-reporting in core+post processing

dginev:
The core+post workflow now has unified error-reporting, with more robust reports for post-proc and archive whatsouts.

As it turns out, if there was a Fatal error (via a Perl die) in post-processing, when creating an archive as output, the entire log was not being written out, as the archive was never created in the first place.

The reason for that was that the ```eval``` guard was wrapping the entire post-processing call, wrapping also the archive creation logic.

In this PR I have improved on that end, and finally unified the error-reporting codes (albeit in a hacky way) by:
 * Making the ```eval``` guard only wrap ```$latexmlpost->ProcessChain```, hence always writing an archive with a log file (in the CorTeX setup).
 * Ensuring that the final ```status_code``` is the max of the combined core and post-processing status codes, so that a no problem core run (code 0) and a fatal post run (code 3) return an overall fatal code, and vice versa.
 * I also added a minor update to Config.pm, where any destination ending with a ```.zip``` extension forces the ```archive``` whatsout setting. I had a few test runs where latexml was writing out HTML5 in a .zip file, which felt very unintuitive.

All of these problems were originally created by me, so I take the full blame for them lurking around for this long. The solution is still slightly patchy, and can be streamlined once we have a flow where both core and post-processing use the same State object.

I would appreciate it if we could merge this PR (or a slight improvement of it) soon, so that the CorTeX reruns can move forward, even if it's not the ultimate state of the joint error-reporting.

Feedback welcome!

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -19,6 +19,7 @@ use File::Temp;
 File::Temp->safe_level(File::Temp::HIGH);
 use File::Path qw(rmtree);
 use File::Spec;
+use List::Util qw(max);
 use LaTeXML::Common::Config;
 use LaTeXML::Core;
 use LaTeXML::Util::Pack;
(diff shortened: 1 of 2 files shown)


dginev:
There are currently near 30,000 fatal error job runs in CorTeX which I would like to rerun after we merge (a version of) this PR -- I am hoping they will reveal some interesting low-hanging fruits for improving the robustness in post-processing, likely many of the issues are calling methods on undef and the like.

brucemiller:
Not sure I follow all the details, but it looks right :>

dginev:
Thanks! Will rerun now :-)

PR #713, opened 2016-02-11
Sanitize is essentially cruft by now

dginev:
As said, removing some unneeded code from the high-level driver, as it arguably has no effect at all by now.

Also removed some trailing whitespaces, my editor does that automatically now.


lib/LaTeXML.pm
```diff
@@ -682,23 +680,11 @@ sub flush_log {
   $$self{log} = q{};
   return $log; }
 
-sub sanitize {
-  my ($self, $log) = @_;
-  if ($log =~ m/^Fatal:internal/m) {
```
dginev: this check almost never (truly never?) was true, so the entire call is pointless.

brucemiller:
seems good; thanks!

PR #714, opened 2016-02-11
Adds defined method for Config objects

dginev:
This is a convenience method that allows me to check if certain options have been "defined" (without actually getting their content) from peripheral services, such as the latexmls socket server.


dginev:
Oh and again, removing some trailing whitespace.

PR #719, opened 2016-02-19
Fallback to files without extensions, when no .tex file is present

dginev:
1.52% of arXMLiv's conversion currently exhibits "no message" fatals, a lot of which seem to be caused by a failure to detect the main TeX file in the source archives.

This PR checks if there is no .tex file present in the archive, and if so checks all files without extensions as possible candidates for a main source file.

I have tried it on a couple of arXiv papers and it converted both to no problems, which seemed promising!


dginev:
Oh and I just remembered to change the category for missing sources to "invalid"

brucemiller:
Awesome!

PR #720, opened 2016-02-20
Check single-file archives, as some may have %auto-ignore from arXiv

dginev:
The size of the diff is misleading here, it's mostly a re-indent of a block. The PR does a single change - removes the early exit if there is a single .tex file in a ZIP archive.

Instead, it still performs the heuristic check, which can determine the .tex file is invalid and throw a "Fatal:invalid:Archive" rather than a rather bizarre alternative error. It also has the nice side-effect of filtering out withdrawn papers from the corpus.

If we merge here, I would need to rerun both Fatals and no-problems (as many withdrawn papers may have ended as "no problem" snippets), and will add a special "invalid" category for the reports, based on the "Fatal:invalid" messages.


dginev:
Oh, here is a link explaining auto-ignore:
https://arxiv.org/help/faq/mistakes#auto_ignore

PR #721, opened 2016-02-21
Write invalid fatal info to log

dginev:
My last PR was not explicit enough with writing the invalid fatal to the log file (it's in LaTeXML.pm, so it is hard-printed, rather than using the Error API)

So here is one more commit that adds the info to the log, and makes it parseable by systems such as CorTeX.


dginev:
I also fine-tuned the Mouth file-read fatals to fit the categorization scheme a bit nicer.

PR #731, opened 2016-04-09
Fix the empty items when CommaList parameterType actually takes a list

angerhang:
In https://github.com/KWARC/LaTeXML-Plugin-sTeX/issues/20 we discovered that parameter of type `CommaList` becomes empty when we provide it with a list. The primary cause is that the original `CommaList` when taking a list of arguments always pushes an empty `@tokens` to the `@items` and we construct an Array based on the item list.

Because the if else snippet does not capture the situation when the `$token` is a regular token and thus the tokens we want were never added to the `@tokens`

```perl
 if ($token->equals(T_END)) {
          push(@items, Tokens(@tokens));
          last; }
        elsif ($token->equals($comma)) {
          push(@items, Tokens(@tokens)); @tokens = (); }
        elsif ($token->equals(T_BEGIN)) {
          push(@tokens, $token, $gullet->readBalanced->unlist, T_END); }
```

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -469,7 +469,8 @@ DefParameterType('CommaList', sub {
         elsif ($token->equals($comma)) {
           push(@items, Tokens(@tokens)); @tokens = (); }
         elsif ($token->equals(T_BEGIN)) {
-          push(@tokens, $token, $gullet->readBalanced->unlist, T_END); } }
+          push(@tokens, $token, $gullet->readBalanced->unlist, T_END); }
+        else {push(@tokens, $token); }}
       if ($typedef) {
         @items = map { [$typedef->reparseArgument($gullet, $_)]->[0] } @items; } }
     else {


dginev:
@brucemiller a quick read through the CommaList parameter type is sufficiently confusing for me to wonder if this is indeed a real problem.

In particular the [inner while loop](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/Package/TeX.pool.ltxml#L465) doesn't seem to be pushing any of the argument tokens into the ```@tokens``` array.

brucemiller:
Very odd, and sorta look(ed) unfinished and/or untested; it's not used anywhere else in LaTeXML yet. I wonder if it was intended to be restricted to something like ```{{stuff},{stuff},...}``` and give an error otherwise.  Your correction interprets CommaList as ```{balanced-stuff,...}```, which is sensible.

I merged the request, after reformatting (you might look at using the pre-commit hook (in .git/hooks).
Thanks!

angerhang:
Great thanks for merge :)

PR #738, opened 2016-04-29
Auto-load LaTeX when braced form is seen in TeXFileName parameters

dginev:
The real-world example I am looking at had a ```\input{preamble.tex}``` at the top of its main file, which worked just fine when typeset with ```pdflatex```.

To emulate that, I suggest that LaTeXML keeps an eye for braced ```{filename}``` wrappers in the bare ```TeXFileName``` paramater type, and load the LaTeX.pool definitions when spotted.

I probably have a blind spot for unexpected side-effects, especially since I am not familiar with all cases in which ```TeXFileName``` is used. So I am open to refactor if you can predict side-effects.

Thanks for reviewing!

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -340,6 +340,16 @@ DefParameterType('TeXFileName', sub {
       && (($cc = $token->getCatcode) != CC_SPACE) && ($cc != CC_EOL) && ($cc != CC_COMMENT) && ($cc != CC_CS)) {
       push(@tokens, $token); }
     $gullet->unread($token) unless ($cc == CC_SPACE) || ($cc == CC_EOL) || ($cc == CC_COMMENT);
+    my $lead_cc = @tokens && $tokens[0]->getCatcode();
+    if ($lead_cc == CC_BEGIN) {
+      my $trail_cc = @tokens && $tokens[-1]->getCatcode();
+      if ($trail_cc == CC_END) {
+        # A begin-end wrapper indicates latex style {filename} use,
+        # so first unwrap,
+        @tokens = @tokens[1..$#tokens-1];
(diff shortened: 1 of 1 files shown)


brucemiller:
Have you checked what the behaviour of ```\input{foo}``` in *plain* tex is?  Does it strip the braces or does it look for "{foo}.tex" ?

dginev:
Yes, I have checked, the plain tex behaviour is exactly the latexml behaviour:

```bash
$ tex main.tex 
This is TeX, Version 3.14159265 (TeX Live 2015/Debian) (preloaded format=tex)
(./main.tex
! I can't find file `{preamble}'.
l.2 \input{preamble}
```

But the latex behavior just works:
```bash
$ latex main.tex 
This is pdfTeX, Version 3.14159265-2.6-1.40.16 (TeX Live 2015/Debian) (preloaded format=latex)
 restricted \write18 enabled.
entering extended mode
(./main.tex
LaTeX2e <2016/02/01>
Babel <3.9q> and hyphenation patterns for 81 language(s) loaded.
(./preamble.tex (/usr/share/texlive/texmf-dist/tex/latex/base/article.cls
Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
(/usr/share/texlive/texmf-dist/tex/latex/base/size10.clo))
No file main.aux.
) [1] (./main.aux) )
Output written on main.dvi (1 page, 232 bytes).
Transcript written on main.log.
```

So my suggestion is heuristic here - due to the fact that braces are
  1. unsupported for file names on some file systems 
  2. traditionally considered dirty / bad style in filenames

We remove the current latexml support for "brace-wrapped filenames a-la plain tex" and replace it with "heuristic enabling of latex". The reason being that the real-world examples I have seen have so far always conformed with the latex-oriented behaviour.

brucemiller:
OK, let's try it.

Issue #736, opened 2016-04-28
Machine-readable (xpointer) locators on demand


angerhang:
We are trying to retire the KWARC fork of LaTeXML and the first primary difference we need to synchronize is the locator and I will just copy the comment that Deyan had on this: 

> We need to add a capability to latexml for switching to the stex-style
> (xpointer-style really) locators. Ideally it should be a trigger in the
> latexml.sty.ltxml binding.
> 
> For example, you can already do:
> 
> --preload=[ids]latexml.sty
> 
> to request adding ids to all XML nodes created by latexml. So I would
> imagine an option that allows for the xpointer-style locators as
> 
> --preload=[ids,xpointer_locators]latexml.sty
> 
> We can implement that internally by flipping a switch, that switches
> between latexml's default locators and our minimally modified stex ones
> (they're in Mouth.pm).

dginev:
The very simple summary of this issue is:

We need machine-readable locators, when we write them down in attributes.

This is an sTeX use case so far, but it goes beyond that. Any web editor powered by latexml would benefit from machine-readable locator attributes, as it could then have fine-grained automatic connection between the source and rendered content. See #101 for the full discussion in that direction.

This issue can be seen as a prerequisite for #101, as making the current locators machine-readable is a low difficulty task on the way of making them as precise as possible.

brucemiller:
Hi all; with #1013, we have basic support for structured locators (as opposed to the former strings), which also cover ranges; a `toAttribute` method on these objects turns them into XPath.  I'll ping @tkw1536 and/or @dginev to offer a short example of how to use them before I close.  Note also there's another open issue #101 relating to the accuracy of the locators which may need further work.

tkw1536:
The newly merged PR does not fix #101 -- the locators are about as accurate as they were before. However, they are now accessible in a machine-readable format. 

One useful example would be to have the following code in a binding somewhere:

```perl
Tag('ltx:quote', afterClose => sub {
    my ($document, $node, $box) = @_;
    return unless $box;
    $node->setAttribute('srcref', $box->getLocator->toAttribute); });
```

Then any tex code like

```latex
\begin{quote}
some text here
\end{quote}
```

would end up with a srcref attribute somewhat like this: 

```xml
    <quote srcref="example.tex#textrange(from=9;0,to=11;11)">
      <p>some text here</p>
    </quote>
```

[...]

PR #739, opened 2016-04-30
[feature] optional xpointer syntax for locators

angerhang:
As what we have talked about in #736 , we are making this locator optional by doing the switch 
via  `--preload=[ids,xpointer_locators]latexml.sty`. `make test` passed without a problem.


dginev:
Can you refactor this into two functions - ```get_locator_narrative``` and ```get_locator_xpointer``` and call the respective one from the top-level ```getLocator``` ? It's a bit messy as it is in the PR.

dginev:
Also rather than have a boolean option here, it may make sense to have several ones, and just extend as appropriate.

So ```locator_xpointer``` and ```locator_narrative``` both make sense, and you can set a global value ```LOCATOR_STYLE``` that defaults to ```narrative``` and can be set to ```xpointer```.

 In fact it starts feeling even closer to a [key=val] pair, but those are not common for packages, so it's fine with two explicit options. Sorry I changed my mind since my first email, but the more I think about it, the more the idea fleshes itself out. It's also interesting to see what Bruce's feedback will be here.

angerhang:
This pull request has been updated. The separation does make things more tidy : )

lib/LaTeXML/Core/Mouth.pm
```diff
@@ -159,24 +159,51 @@ sub stringify {
...
-      my $cm = ($c < 1       ? 0       : $c - 1);
-      my $cn = ($n - $c > 50 ? $c + 40 : $n - 1);
-      my $p1 = ($c0 <= $cm ? join('', @$chars[$c0 .. $cm]) : ''); chomp($p1);
-      my $p2 = ($c <= $cn  ? join('', @$chars[$c .. $cn])  : ''); chomp($p2);
-      $msg .= "\n  " . $p1 . "\n  " . (' ' x ($c - $c0)) . '^' . ' ' . $p2; }
-    return $msg; }
-  else {
-    return "at $$self{source}; line $l col $c"; } }
+  # Make Xpointer style locator optional
+  $STATE->assignValue('LOCATOR_STYLE' => 'narrative') unless $STATE->lookupValue('LOCATOR_STYLE');
```
dginev: This check should happen once in ```latexml.sty.ltxml``` rather than every time when getLocator() is called. The locator function is called extremely often, so we should make sure it does as little as possible.

Especially with the way you've handled the conditional, you can assume an unset value leads to the default.

So something like:
```perl
 my $locator_style = $STATE->lookupValue('LOCATOR_STYLE')
 if ($locator_style && ($locator_style eq 'xpointer')) {
   return get_locator_xpointer(@_);
 } else {
   return get_locator_narrative(@_); } }
```

suffices here.

lib/LaTeXML/Package/latexml.sty.ltxml
```diff
@@ -44,6 +44,9 @@ DeclareOption('noprofiling', sub { AssignValue('PROFILING' => 0, 'global'); });
 DeclareOption('mathparserspeculate',   sub { AssignValue('MATHPARSER_SPECULATE' => 1, 'global'); });
 DeclareOption('nomathparserspeculate', sub { AssignValue('MATHPARSER_SPECULATE' => 0, 'global'); });
 
+DeclareOption('xpointer_locators', sub { AssignValue('LOCATOR_STYLE' => 'xpointer', 'global'); });
+DeclareOption('locator_narrative', sub { AssignValue('LOCATOR_STYLE' => 'narrative', 'global'); });
```
dginev: And one last bit here is making the names consistent, e.g. ```narrative_locators``` instead of ```locator_narrative```. Though Bruce may have better suggestions than me.

lib/LaTeXML/Core/Mouth.pm
angerhang: This is indeed much more efficient. Thanks for the tips :)

dginev:
Just to note, this is a blocker PR for sTeX's transition to a LaTeXML plugin to move forward, since the xpointer locators are heavily relied on in the KWARC stack.

brucemiller:
This just deals with the format of the locator (a separate discussion #736) discusses having the locator accurate. I'm not so sure that I want to deal with this in this way, however; it introduces another unnecessary switch, and puts complication in surprising places.  And kinda weird behaviour: with the swtich on, you'd get xpath info in your error messages.

In the evolution of LaTeXML's notion of turning an object into a string, the number of variations has grown (ToString, Stringify, UnTeX,...) well, there's an ```$object->toAttribute``` method (but not yet a ToAttribute function). Currently Document will use the latter method when placing an object into an attribute; the Constructor Compiler should (but doesn't yet) do that, as well. 

So, the more elegant solution, IMHO, would be to make locator an object and let the machinery give you the appropriate form at the appropriate time.  You'd get a "human readable" locator in messages and xpath locator in XML attributes.

brucemiller:
AND it would be Simpler!

dginev:
I like the locator object idea design-wise. The only question is how much would that slow the sTeX transition to the master LaTeXML, since I think the xpointer locators are quite heavily used in the KWARC stack.

But maybe this is something @angerhang can look into? Unsure.

Up to you whether you want this PR close right away, but it sounds like we should create an alternative one that adds the new infrastructure.

dginev:
As a point on syntax, today I got to look at [commonmarker](https://github.com/gjtorikian/commonmarker) which I understand is the currently used markdown parser by github (that renders this very issue comment).

In HTML, locators are not standardized, so you're pretty free to invent any scheme desirable for annotating (also XPointer is not HTML-native). What they have decided on is a very minimalist:
```
data-sourcepos="1:1-59:3"
```

with the standard `row:column` index pairs. Something to consider, given that this PR only addresses the primary XML output.

brucemiller:
This patch should be superceded by #1013, so we'll go ahead and close.

Issue #741, opened 2016-05-05
Timeout ignored in pathological math example


dginev:
In [this gist](https://gist.github.com/dginev/e83af1f8703adf9ca851ce9ca9cebdc1) I have attached a pathological TeX snippet that fails to terminate in some 70 seconds, even after it hits a 20 second timeout alarm signal.

It's an anonymized source from Authorea, where an author made a typo and accidentally turned the rest of a block into a huge formula. Note that this times out without math parsing even enabled, as in Authorea we are using the equivalent of:

```bash
latexmlc pathological_math_error.tex  
  --noparse --timeout=20 --format=html5 
  --whatsin=fragment --whatsout=fragment
```

There are two separate issues I would like to solve here:
 1. Make the 20 second timeout succeed and truly terminate processing (it seems to interrupt the absorbtion of math boxes)
 2. Try to optimize latexml's processing for this example, so that it takes a magnitude faster, if possible

Ideas welcome, I'll take a look.

dginev:
Reading up on Perl's [safe signals](http://www.theregister.co.uk/2009/05/21/perl_safe_signal_handling/), shows a very reliable workaround of invoking latexml prefixed with ```PERL_SIGNALS=unsafe``` which avoids the delayed + safe signal handling in modern Perl.

That immediately kills the process but risks corrupting Perl's internal datastructures, which isn't an option for a long-living server for example.

I am also wondering why the timeout is reported as a warning here:
```
Warning:perl:warn Perl warning
	at /tmp/scratch.tex; line 23 col 39
		(in cleanup) 
	Fatal:timeout:timedout Conversion timed out
		at /tmp/scratch.tex; line 23 col 39
```

As that message is printed exactly at the second the alarm is triggered, so it seems the signal is handled (but probably wrongly?) by LaTeXML's Error module.

dginev:
I analyzed some of the stack traces and indeed, whenever the alarm lands in the middle of libxml processing, it fails to interrupt the C call. It is actually handled at the right time in Perl - the log messages are printed just in time. Yet the processing isn't terminated, and it seems the full job continues until it is fully complete, even returning the final result + log correctly.

So now I'm looking for a reasonable strategy for interrupting libxml (although the article I linked to seems to suggest there isn't a truly reasonable one)

dginev:
Maybe for this issue the best tactic is to understand why latexml takes over a minute here (mostly in ligature application it seems) and optimize away the slowness.

[...]

PR #742, opened 2016-05-05
Disable math ligatures for --noparse option

dginev:
Basic fix for #741 , which disables math ligatures entirely, when the noparse option is specified.

A more advanced fix would attack the problem at its root and e.g. only apply ligatures to the last N children of the current parent, using efficient libxml methods such as a combination of ```->lastChild``` and ```->previousSibling```.

### DIFF
--- a/lib/LaTeXML/Core.pm
+++ b/lib/LaTeXML/Core.pm
@@ -53,6 +53,7 @@ sub new {
         @{ $options{graphicspaths} || [] }], 'global');
   $state->assignValue(INCLUDE_STYLES => $options{includeStyles} || 0, 'global');
   $state->assignValue(PERL_INPUT_ENCODING => $options{inputencoding}) if $options{inputencoding};
+  $state->assignValue(NOMATHPARSE => $options{nomathparse} || 0, 'global');
   return bless { state => $state,
     nomathparse => $options{nomathparse} || 0,
     preload => $options{preload},
@@ -280,7 +281,7 @@ sub writeDOM {
 
 __END__
 
-=pod 
+=pod
 
 =head1 NAME
 
@@ -302,11 +303,11 @@ But also see the convenient command line script L<latexml> which suffices for mo
 
 =item C<< my $latexml = LaTeXML::Core->new(%options); >>
 
-Creates a new LaTeXML object for transforming TeX files into XML. 
+Creates a new LaTeXML object for transforming TeX files into XML.
 
  verbosity  : Controls verbosity; higher is more verbose,
               smaller is quieter. 0 is the default.
- strict     : If true, undefined control sequences and 
+ strict     : If true, undefined control sequences and
               invalid document constructs give fatal
               errors, instead of warnings.
  includeComments : If false, comments will be excluded
@@ -332,7 +333,7 @@ B<OBSOLETE> Use C<$latexml->convertFile("literal:$string");> instead.
 
 =item C<< $latexml->writeDOM($doc,$name); >>
 
-Writes the XML document to $name.xml. 
+Writes the XML document to $name.xml.
 
 =item C<< $box = $latexml->digestFile($file); >>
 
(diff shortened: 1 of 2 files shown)


brucemiller:
I don't see why doing ligatures ought to be connected with math parsing (or frankly, why you'd not want to parse: you get better MatML).

dginev:
1. Why we don't want to parse math at Authorea - since latexml is one of many renderers we have on Authorea, and it is easier to accept TeX math embedded in HTML from all of them and have a single frontend setup. It's an unfortunate byproduct of the math-on-the-web state in browsers.

2. Why I think not parsing math should also opt out from math ligatures - as demonstrated in #741 there is a huge performance flaw in the way ligature definitions pass arguments, especially when math parsing is disabled. And since we want to pass the TeX directly to an external parser, the ligature information will be lost for us anyway, so we'd rather not spend time on it.

I do intend to push for a switch to server-side MathML generation at Authorea in the long run, but that day will only come after a few other pieces fall into place. Until then, I think it is consistent with the intent behind "--noparse" to disable ligatures as well.

dginev:
A different way to phrase this, is that the ```--noparse``` setup (combined with a few other non-traditional options) make LaTeXML an equivalent drop-in replacement for Pandoc's latex-to-html conversion, which is what I am trying to deprecate at Authorea.

The first step in that process is to demonstrate a very robust, and as performant as possible, LaTeXML server setup, which produces output comparable to Pandoc's, but has much wider coverage over TeX/LaTeX. Once that is stable enough it becomes easier to push towards making LaTeXML the reference conversion engine at Authorea, and start a move to server-side MathML, which brings with it a significant amount of refactoring and testing. A bit of a roundabout journey, but the technical circumstances are such.

brucemiller:
So, another way of putting it, you want an equivalent to something that doesn't do as much, and so parsing *and* ligatures are wasted effort --- all the moreso, given a coincidental bug in pathological cases.  Otherwise, I don't really see the two as necessarily that connected; that you'd always want both or neither.  In fact, it would seem that in the original usecase for noparse, you'd still want the ligatures formed.

The eventual config setup intends to be more explicit about enabling/disabling the various processing phases.  I'd prefer a separate option (cmdline, or just internal var) to control ligatures, than coupling them.

dginev:
Yes, that's entirely accurate. I could live with a new option, I am just wondering of the use case where ligatures are enabled but parsing isn't. What is that purpose of that?

dginev:
And I should clarify I always mean "math ligatures" when I say "ligatures" in this particular issue.

brucemiller:
You'd be the one to know, really. Some LLamapun usecase, where you want to parse yourself, or later.  But you may still want the "ligatures" formed.  Math ligatures are more related to lexing, than parsing (though the border may be a bit fuzzy here); there's all sorts of combining of operator tokens, roman characters, numbers, etc.

dginev:
Ah! Interesting. I wonder if I can find a counter-example that shows lexing ligatures is also ambiguous and should be left to the parsing stage? Probably there are some...

But when it comes to the reason behind llamapun-like tools wanting to use ```--noparse```, it is to avoid as much of the inference as possible, so that they can deal with the entirety of it. In fact, if asked today I would prefer to not even work with the output of latexml itself (the XMath XML) but instead plug in directly in the processing pipeline and receive the exact input that the latexml grammar receives.

In other words, I am less interested in ```--noparse``` now, and a lot more interested in writing grammar plugins for latexml, that do indeed parse, as part of the core processing. And those can benefit from the ligatures.

Additionally, the number of math ligatures is quite manageable (less than 10?) and should be very easy to migrate over to any math parsing experiment that builds on latexml, even if it decides to go the --noparse route.

There is no active code in LLaMaPUn today that uses ```--noparse```, the only semi-active experiment is based on enabled parsing with a special math grammar plugin. But thanks for thinking of all of my hats!

brucemiller:
To the first comment, uh yeah, I could point out several ambiguities, but that's your job here!
To the second comment, I think we're gradually migrating to the same position; who knows who's moved the farthest!  But why spoil a good argument with unnecessary agreement.

dginev:
Well, I think I have reassured myself that ```--noparse``` should take away all math-related goodness, people should be using the math parsing after all. Let them suffer for refusing it :> 

That always keeps my Authorea enhancements moving in a straightforward way, without extra options and complexity.

brucemiller:
well, you were *supposed* to be convincing *me*!!  But I guess I'll go along with (paraphrasing) F'em if they can't take a joke.

dginev:
Sweet, thanks!!

I haven noticed that it is much easier to convince myself than you, guess I'm lucky that ended up convincing you anyway :>

PR #748, opened 2016-05-12
Use lean generation of Fatal messages for timeouts

dginev:
As discussed in #741

### DIFF
--- a/lib/LaTeXML/Common/Error.pm
+++ b/lib/LaTeXML/Common/Error.pm
@@ -75,9 +75,10 @@ sub Fatal {
       $LaTeXML::BAILOUT = 1;
       push(@details, "Recursive Error!"); }
     $state->noteStatus('fatal') if $state && !$ineval;
+    my $detail_level = ($category eq 'timeout') ? 0 : 2;
     $message
       = generateMessage(colorizeString("Fatal:" . $category . ":" . ToString($object), 'fatal'),
-      $where, $message, 2, @details);
+      $where, $message, $detail_level, @details);
     # If we're about to (really) DIE, we'll bypass the usual status message, so add it here.
     # This really should be handled by the top-level program,
     # after doing all processing within an eval
--- a/lib/LaTeXML/Common/Error.pm
+++ b/lib/LaTeXML/Common/Error.pm
@@ -75,7 +75,7 @@ sub Fatal {
       $LaTeXML::BAILOUT = 1;
       push(@details, "Recursive Error!"); }
     $state->noteStatus('fatal') if $state && !$ineval;
-    my $detail_level = ($category eq 'timeout') ? 0 : 2;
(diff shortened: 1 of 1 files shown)


dginev:
I'm actually both grateful and shocked to find the infrastructure already in place to request messages without the stack trace and context.

Such an easy pull request!

dginev:
@brucemiller I came up with some follow-up enhancements on this line, which will benefit from having this PR merged first. 

Anything I can do to aid that process? The other small PRs also offer important stability improvements to long-lived latexmls job chains, with timeouts.

brucemiller:
I'd been meaning to check whether there was an appropriate override; presumably if verbosity is 1 or 2, it ought to produce the longer message even if it is due to a timeout.  The thing is, sometimes that potentially deep stack trace is the only clue you've got to where things went wrong.    So I did check and it looks like it just sets the shorter message for any timeout; maybe could fix that?

dginev:
I think I understand.

However, my counter-argument here is that debugging is not the primary use case for the timeout parameter, and should not be a main concern in the message generation. Timeouts tend to be invaluable in ensuring production reliability of latexml runs (both for cortex/arxiv and Authorea), in order to have consistent performance. Having error information on top of that is great, but it is perfectly OK to only have a minor trace/hint -- in fact it's OK to have no message at all.

If the job is of interest it can be studied with a larger timeout, or without any, in isolation. The harm of the timeout not being reliable is much more dangerous (and trickier to work around), as it causes daemon processes to be come unresponsive, in an essentially unpredictable manner to the outside, and that makes sysadmin life very difficult.

My next PR in this line is actually to add one more category where I expect the "silently stop as soon as possible" behaviour, which is the "too_many_errors" fatal. In that case we have 100 prior error messages to look at, which should provide more than enough context.

Does this sound reasonable?

dginev:
But if you're asking to switch back to detail level 2 for verbosity larger than the default one, that sounds reasonable as well.

dginev:
Ok, I have committed the suggestion, feel free to review again.

If you decide to merge, there is a new trick GitHub allows to reduce the commit clutter. After you click "Merge pull request", there is a small green arrow that lets you choose between: 

  * "Confirm merge" - merge as-is, with the entire commit history
  * "Squash and merge" - merge as a single commit, where each individual commit message from the original history will be recorded as an item in the commit's description.

I encourage you to try the squash option, it's quite nice, especially if there are spurious commits in the PR.

brucemiller:
exactly, for the last question.  While debugging is not the primary goal of timeout, it seems like having the option of getting a trace is a reasonable tool for figuring out why it's hanging at a particular point.  Perhaps you could run it interactive, wait the timeout period and then ^C. 

Oh, see new comment. will add this comment, then look at the new diff...

dginev:
Thanks!

PR #749, opened 2016-05-14
Add travis support

physikerwelt:
[ @dginev edited with description]

Thanks to @physikerwelt  for opening the pull request and initiating the Travis discussion via email. I assume the travis file is borrowed from KWARC/LaTeXML.

Let me write a brief justification for @brucemiller since I suspect he hasn't been sucked in the Travis madness.

The idea is that while CPAN Testers would provide us a test harness for the releases, Travis can already provide us with quick regression testing on every push to git. It is usual to combine the build with a badge, you can see the "build |passing" button in the README of [KWARC/LaTeXML](https://github.com/KWARC/LaTeXML) for an example.

After merging the ```.travis.yml``` file, you also need to enable builds from the travis site, after which each push to git will be build and recorded, so that there is a clear visual indication if the latest repository head has a successfully building test suite. Here is the [travis page](https://travis-ci.org/KWARC/LaTeXML) for the KWARC fork as an example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: perl
+perl:
+  - "5.20"
+  - "5.18"
+  - "5.14"
+  - "5.10"
+
+sudo: false
+addons:
+  apt:
+    packages:
+     - libdb-dev
+     - libxml2-dev
+     - libxslt1-dev
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -350,6 +350,7 @@ our $kpse_cache = undef;
 
 sub pathname_kpsewhich {
   my (@candidates) = @_;
+  return unless $kpsewhich;
   build_kpse_cache() unless $kpse_cache;
   foreach my $file (@candidates) {
     if (my $result = $$kpse_cache{$file}) {
@@ -364,6 +365,7 @@ sub pathname_kpsewhich {
 
 sub build_kpse_cache {
   $kpse_cache = {};    # At least we've tried.
+  return unless $kpsewhich;
        # This finds ALL the directories looked for for any purposes, including docs, fonts, etc
   my $texmf = `"$kpsewhich" --expand-var \'\\\$TEXMF\'`; chomp($texmf);
   # These are directories which contain the tex related files we're interested in.


dginev:
Updated description + title, thanks @physikerwelt for adding the PR!

dginev:
Also here are the official Travis docs for setting up a Perl project, for reference:
https://docs.travis-ci.com/user/languages/perl

My experience with building Perl projects with travis has been great, and I consider their build harness extremely quick and convenient.

physikerwelt:
@dginev thank you for adding the full story.
Note, that there is currenlty the problem with the permission problem
https://travis-ci.org/physikerwelt/LaTeXML/jobs/130250797#L855
That should probably be investigated first.
It occours independent of the format of the travis file (cf https://github.com/physikerwelt/LaTeXML/tree/travis)

dginev:
I think there is need to commit the executable bit of latexmlc, as shown here in a similar discussion:
http://stackoverflow.com/a/15572639

brucemiller:
I'm fine with committing the travis build file, if it's ready.  I'm not clear on where this permission problem is.  It's happening in 2 (somewhere in t/expansion, t/digestion) of the normal tests, and also 2 of the daemon tests.  The latter cause a fail since the daemon is picking up the unexpected message, but that message shouldn't be happening(?) at all.  I don't think it's the x bit, since all other tests run fine.

dginev:
Travis actually found a rather awesome bug we have in the ```kpsewhich``` logic. Well, I found it, but Travis showed it.

First [here is a working Travis build](https://travis-ci.org/dginev/LaTeXML/builds/132175131) obtained by two commits that simply returned before the shell-out backticks on ```$kpsewhich```. [Commits here](https://github.com/dginev/LaTeXML/commits/7c48d6ffa10c91c8b7a8cf45edb6810ec48864cd)

Explanation: On Travis, by default we have no texlive, so the [kpsewhich fishing](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/Util/Pathname.pm#L348) fails and sets the Util::Pathname class variable to undef.

Hence, what then happens in the various backticks on that variable, and I think [this one](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/Util/Pathname.pm#L368) in particular, is to call the empty command in shell, which returns a permission denied error.

I think a simply fix is to do a quick return if there is no kpsewhich available, will add a very quick commit here to get that done. But since the PR is on Mortiz's fork, he'll have to cherry-pick it himself. (Or I can close here and make a second PR from my fork)

dginev:
Yep, I think we can make progress here. Here is the latest [good travis build](https://travis-ci.org/dginev/LaTeXML/builds/132176086), for which I need @physikerwelt to cherry-pick commit 1fc5638de0e7d1af71a1ef8cc849fe8f1abfb1b6

And of course I need a nod and a wink from @brucemiller to get it merged thereafter. Happy we squashed a sneaky bug in the process!

dginev:
And while you're at it, I would also suggest adding the ```5.24``` Perl version, so that we have 5 to test against, including a relatively new one.

physikerwelt:
@dginev thank you for looking into that. I'll cherry pick your commit and rerun the tests later today.

brucemiller:
Y'all are confusing me with all your branches, forks and pr's!  The kpsewhich patch is pretty trivial --- once you've realized what the problem is! Thanks @dginev!  Should I wait for a PR on that, or just fix it?

dginev:
Moritz agreed to cherry-pick it, so please wait for it to land in this PR, trivial or not :>

That's what you get for having open source contributors!!

dginev:
P.S. Actually, if you want it in master as soon as possible @brucemiller no one can stop you from cherry-picking that commit yourself, merging the PR will still work. Just don't add a new separate commit, which will hit us with a git conflict.

physikerwelt:
I'm working on that right now. I still need to add the 5.24 version

physikerwelt:
It runs most of the tests, however some are skipped due to missing libraries.

physikerwelt:
Do you have an idea which of the library listed in https://github.com/physikerwelt/mathpipe/blob/master/.travis.yml#L12-L29 could be the missing bit? I could add them all, but that would not be nice.

dginev:
Yep, the skipped tests are due to not having texlive on the machines, and I think it's fine to skip them in Travis (it gets 2x slower with the texlive installation, which is huge).

Also, not having texlive is the reason we caught this sneaky bug anyway, all LaTeXML developers tend to have texlive on their local machines and it's a blind spot to be missing a TeX installation.

Good thing I suggested 5.24, it seems we caught another change to Perl's core packages:
```
Can't locate Parse/RecDescent.pm in @INC (@INC contains: /etc/perl /usr/local/lib/perl/5.14.2 /usr/local/share/perl/5.14.2 /usr/lib/perl5 /usr/share/perl5 /usr/lib/perl/5.14 /usr/share/perl/5.14 /usr/local/lib/site_perl .).
```

I suggest that Bruce merges this PR and adds Parse::RecDescent to the dependencies explicitly, also so that he gets to test his own Travis setup and get used to the new bells and whistles :>

dginev:
To be perfectly clear, all Travis needs is having Parse::RecDescent in Makefile.PL, since it executes:
```
cpanm --quiet --installdeps --notest .
```

to build the dependencies.

physikerwelt:
@dginev I thought https://travis-ci.org/physikerwelt/LaTeXML/jobs/132295208#L170 means that travis does not know 5.24?

dginev:
Oh, I completely missed that, thanks @physikerwelt . Should get a coffee. Not to mention RecDescent is already in Makefile.PL, so I'll just delete my last 2 comments.

physikerwelt:
I removed the latest commit from the PR. And consider my job as done;-)

dginev:
Yep, I think we'll wait until 5.24 becomes available in Travis, until then let's just merge here and move on with life. Thanks @physikerwelt !

brucemiller:
Thanks!

dginev:
@brucemiller people love the status badges too, if you want to add one: https://docs.travis-ci.com/user/status-images/

PR #750, opened 2016-05-22
Introducing UNSAFE_FATAL, trimming insertion context traces

dginev:
A number of stability improvements for daemonized processing here:
 1. The daemon initialization code was never wrapped in a timeout, I have added one
 2. Marked some Fatal errors as explicitly "unsafe", so that daemon processes are forced to reinitialize from scratch (motivation below)
 3. The reporting of "insertion context" for absorbtion errors prints enormous XML dumps on screen, which are almost always unnecessary/overwhelming, even to a developer. I have found the context of 5 levels deep to be both brief and sufficiently informative for the examples I have. Maybe this should also depend on verbosity?
 4. Minor: improved ALRM message text a little

**Motivation:**

Some Fatals have proven to corrupt the State. In particular certain deep recursions (inf. loops?) in the absorbtion stage leave follow-up jobs on the same daemon process to Fatal with ```internal:<recursion>``` errors. I have experienced this on a production server where hundreds of jobs ended up broken due to a malformed State, which was quite unfortunate. It seemed safest to mark any terminations caused by signals, as well as "too_many_error" fatals, as requiring a new init phase.

I have coupled this with a couple of robustness patches to the latexmls socket server as well, which can be [found here](https://github.com/dginev/LaTeXML-Plugin-latexmls/commits/master).

I have a couple of examples that exhibit the described behaviour, but they are very pathologically broken and are probably not too interesting (definitely not valid TeX). Again, mostly ensuring production stability of long-lived daemon processes here.

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -101,7 +101,12 @@ sub initialize_session {
   my $latexml;
   my $init_eval_return = eval {
     # Prepare LaTeXML object
+    local $SIG{'ALRM'} = sub { die "Fatal:conversion:init Failed to initialize LaTeXML state\n" };
+    alarm($$self{opts}{timeout});
+
     $latexml = new_latexml($$self{opts});
+
+    alarm(0);
     1;
   };
   ## NOTE: This will give double errors, if latexml has already handled it!
@@ -252,6 +257,11 @@ sub convert {
       $state->popDaemonFrame;
       $$state{status} = {};
   });
+  if ($LaTeXML::UNSAFE_FATAL) {
+    # If the conversion hit an unsafe fatal, we need to reinitialize
+    $LaTeXML::UNSAFE_FATAL = 0;
+    $$self{ready} = 0;
+  }
   if ($eval_report || ($$runtime{status_code} == 3)) {
     # Terminate immediately on Fatal errors
     $$runtime{status_code} = 3;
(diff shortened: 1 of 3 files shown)


dginev:
Still related to #741 and extensions thereof

brucemiller:
You've combined to distinct sets of changes. The first one (latexml, Error) I can't object with.

The second one, you may not have noticed, but you added the argument limiting the depth of context on *every* call to getInsertionContext.  I'm thinking that perhaps a better approach would be for that method to default to a "nominal" depth (say 5 :> ), rather than the full context.  Better yet, localizing it there, would give the option of defaulting to different depths, depending on verbosity.

dginev:
Completely agree, I did a quick change there since I saw you've already thought of having the parameter and wanted to have the code reviewed before I do something more complex.

I'll refactor to move the 5 in the definition of ```getInsertionContext``` for say verbosity of 1 or less, and keep the current "deep" behaviour for verbosity of 2 or larger. Thanks!

dginev:
I made an experiment with editing this on GitHub, but now I both don't know if the new changeset passes the test suite, and I left poor commit messages. I guess I won't try that again. Will let you know when I've ran the tests.

dginev:
Confirmed tests pass, should be ready to merge.

Issue #710, opened 2016-01-26
Bibliography: Problems with A&A adsurl


goska:
I am experiencing problems with conversion of documents and bibliographies containing <code>adsurl</code>s of Astronomy and Astrophysics articles, which contain strings <code>%26</code> (url encoded & in A&A). LaTeXML generates malformed ``bibentry`` elements for such bib records: closing ``bib entry`` tag and and any ``bib-data`` following ``bib-data`` with ``adsurl`` role are missing. Using LaTeXML 0.8.1 running under MacOS X 10 with TeXLive, aa style bibliography and standard LaTeX report document class.
For example:
```latex
@ARTICLE{2009A&A...505..385A,
   author = {{Andrei}, A.~H. and {Souchay}, J. and {Zacharias}, N. and {Smart}, R.~L. and 
	{Vieira Martins}, R. and {da Silva Neto}, D.~N. and {Camargo}, J.~I.~B. and 
	{Assafin}, M. and {Barache}, C. and {Bouquillon}, S. and {Penna}, J.~L. and 
	{Taris}, F.},
    title = "{The large quasar reference frame (LQRF). An optical representation of the ICRS}",
  journal = {\aap},
archivePrefix = "arXiv",
   eprint = {0907.2403},
 keywords = {catalogs, reference systems, quasars: general, methods: data analysis, astrometry},
     year = 2009,
    month = oct,
   volume = 505,
    pages = {385-404},
      doi = {10.1051/0004-6361/200912041},
   adsurl = {http://adsabs.harvard.edu/abs/2009A%26A...505..385A},
  adsnote = {Provided by the SAO/NASA Astrophysics Data System}
}
@ARTICLE{2008AcA....58...89U,
```
yields
```xml
        <bibentry key="2009A&amp;A...505..385A" type="article" xml:id="bib.bib12">
          <bib-name role="author">
            <surname>Andrei</surname>
            <givenname>A. H.</givenname>
          </bib-name>
          <bib-name role="author">
            <surname>Souchay</surname>
            <givenname>J.</givenname>
          </bib-name>
          <bib-name role="author">
            <surname>Zacharias</surname>
            <givenname>N.</givenname>
          </bib-name>
          <bib-name role="author">
            <surname>Smart</surname>
            <givenname>R. L.</givenname>
          </bib-name>
          <bib-name role="author">
            <surname>Vieira Martins</surname>
            <givenname>R.</givenname>
          </bib-name>
          <bib-name role="author">
[...]

PR #753, opened 2016-05-30
Add support for unescaped % characters in adsurl

goska:
I have added support for unescaped % characters in adsurl (bug report #710), by adding the code suggested by you (using Semiverbatim) to aa_support.sty.ltxml. It's my first attempt at creating a pull request.


brucemiller:
Cool! I assume this is standard enough amongst AA folks to include.  But I'm puzzled: Did you test this with the git version or the released one?  I don't think that ```Semiverbatim``` will actually accept ```%```, but you'd need ```Verbatim```.

goska:
I have probably done something irregular. I have made changes to the latest aa_support.sty.ltxml pulled from GitHub, and have tested it with the aa.cls.ltxml (also the latest version pulled from GitHub) and  modified aa_support.sty.ltxml in the same directory as the bib file I was converting, but running an older 0.8.1 installation of LaTeXML from Mac Ports.When I tried using Verbatim, I got a fatal Perl error (I haven't checked if the syntax I used was correct - sorry).

brucemiller:
Exactly: Verbatim wasn't yet defined in 0.8.1, so that gives an error.  OTOH, Semiverbatim in 0.8.1 treats % a normal char, which causes errors *other* places it was used!  So, the current version redefines Semiverbatim to leave % alone, but adds a new Verbatim that neutralizes %.  In summary, we'd like a patch against the current git version, but it won't work with 0.8.1.

dginev:
> but running an older 0.8.1 installation of LaTeXML from Mac Ports

To put it briefly, please always test against the repository head when filing pull requests, to avoid tricky inconsistencies.

And if you're already excited about using latexml enough to be issuing pull requests, switching to a git-based local installation is a great idea anyway.

goska:
OK - thanks. I'll change to Verbatim, but will have to continue to use the other version myself. Switching to the git-based installation locally is not that straightforward for me, as I have to use the 0.8.1 release for my work to be compatible with the production version. But I will install the git-based version on another machine.

goska:
I have replaced Semiverbatim with Verbatim and tested with LaTeXML pulled from GitHub.

brucemiller:
Thanks folks! (I'll reformat & recommit)

PR #758, opened 2016-06-08
Increasing test timeouts, raspberry pi is too slow

dginev:
So, I was installing LaTeXML on my Raspberry Pi 2, because why not, when I noticed the daemon tests failing in mysterious ways. I even think I have seen this class of failures before on CPAN Testers. Exciting!

Next, I investigated, and realized the Pi was just way too slow. I was using the default setup, which happens to be the measly 700 Mhz speed (random [details](https://learn.adafruit.com/introducing-the-raspberry-pi-2-model-b/performance-improvements) for the curious), and it was timing out on the heavier tests for the daemon - basically anything that had math parsing enabled.

I decided to do the obvious and double the test timeouts, from 5 to 10 seconds - and voila! All tests pass and I have a working (and terribly slow) LaTeXML installation on the Raspberry Pi 2!

I hope this change will also make CPAN Testers build more frequently, since it is possible they are also using rather slow hardware, at least for some of the OSes


brucemiller:
Yeah, I bet CPAN testers is just a bunch of Raspberry Pi's!!!

PR #763, opened 2016-06-10
Fixes #612 ; Full update to Common::Config doc

### LINKED ISSUE #612: Document all allowed options in LaTeXML::Common::Config
To my shock, several options are still undocumented, such as ```--dbfile``` and ```--graphicimages``` in the ```Config.pm``` POD. 

I think the reason is that I removed the documentation for any options I had not yet added support for in the joined latexmlc executable. But as I improved on the added support, I forgot to re-add the documentation.

So I should get around to cleaning up the docs soon.

dginev:
* Minor cleanup of some option defaults
 * Restructured the POD to follow the new structure of the latexmlpost POD
 * Copied over all option docs I could find from latexmlpost and latexml
 * Proofread

Probably not perfect, but hopefully exhaustive.

### DIFF
--- a/lib/LaTeXML/Common/Config.pm
+++ b/lib/LaTeXML/Common/Config.pm
@@ -369,8 +369,9 @@ sub _prepare_options {
       else {
         ${ 'LaTeXML::' . $ltx_class . '::DEBUG' } = 1; } } }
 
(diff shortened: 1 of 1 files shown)


dginev:
Mergeable?

brucemiller:
looks so...thanks

PR #777, opened 2016-07-03
Don't load external DTD in testing, better failure messages

dginev:
Thanks to Bruce for catching this! Should be a nice boost in robustness + speed of the test suite.


brucemiller:
Thanks!!!

PR #793, opened 2016-07-29
Fixes #792 (regression with latexmlc documentation)

### LINKED ISSUE #792: latexmlc does not open the documentation
Up to a few weeks ago, the commands `latexml* --help` used to print the help directly in the terminal.

Then, the behavior changed into printing just the version information and opening the man page (which is fine for me). On the latest git version & Ubuntu 14.04, the commands `latexml --help` and `latexmlpost --help` are equivalent to `man latexml` and `man latexmlpost` (at least, on my system).

However, `latexmlc` does not follow this behavior. The command `latexmlc --help` just prints the version number on the terminal and nothing else. If I try to access the man page via `man latexmlc`, I get a quite scant manpage with the following synopsis:
```
SYNOPSYS
       See the OPTIONS section in LaTeXML::Common::Config for usage
       information.
         Also consult latexmlc --help
```
To get a clue of what the options for `latexmlc` are, I have to type `man LaTeXML::Common::Config` and go through a couple of screens before getting into the section "OPTION SYNOPSIS" that indeed describes the options for `latexmls/latexmlc`.

I think this is not really "user friendly", since it makes quite hard for a user to find the information he is looking for. It would be nice, imho, to

- at least redirect `latexmlc --help` to `man latexmlc`, as it already happens with `latexml` and `latexmlpost`;
- create a separate manpage for `latexmlc`, as it used to be a few weeks ago (afair).

I hope this is not just a problem with my configuration, so let me kindly know if you can reproduce the bug.

dginev:
I had rearranged the docs, and had forgotten that I am using an advanced sections selector of pod2usage, which made ```latexmlc --help``` return an empty value, rather than the synopsis string.

This pull request fixes this.

### DIFF
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -232,13 +232,13 @@ __END__
 
 =head1 NAME
 
-C<latexmlc> - A omni-executable for LaTeXML, capable of
-  stand-alone socket-server and (soon) web service conversion.
+C<latexmlc> - An omni-executable for LaTeXML, capable of
+  stand-alone, socket-server and web service conversion.
+  Supports both core processing and post-processing.
 
 =head1 SYNOPSYS
 
 See the OPTIONS section in L<LaTeXML::Common::Config> for usage information.
-  Also consult latexmlc --help
 
 =head1 DESCRIPTION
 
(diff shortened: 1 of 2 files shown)


brucemiller:
thanks!

PR #796, opened 2016-09-01
fixes typo in LaTeXML::Core::Token documentation

teepeemm:
This fixes the typo at the bottom of http://dlmf.nist.gov/LaTeXML/manual/coremodules/LaTeXML_Core_Token.html
in the description of getCSName

### DIFF
--- a/lib/LaTeXML/Core/Token.pm
+++ b/lib/LaTeXML/Core/Token.pm
@@ -373,7 +373,7 @@ Return a string representing C<$object>.
 =item C<< $string = $token->getCSName; >>
 
 Return the string or character part of the C<$token>; for the special category
-codes, returns the standard string (eg. C<T_BEGIN->getCSName> returns "{").
+codes, returns the standard string (eg. C<< T_BEGIN->getCSName >> returns "{").
 
 =item C<< $string = $token->getString; >>
 


teepeemm:
I have `ln -s ../../tools/pre-commit .git/hooks` setup.  Does that automatically happen before a commit?  Or do I need to invoke it manually?  Does it matter if I'm committing from NetBeans?

Is it possible to do whatever Travis-Ci does before I commit?  Or is that basically a quick `make` to ensure I've not done anything boneheaded?  Or is Travis-Ci the same as pre-commit?

teepeemm:
Although looking more closely, it appears getCSName returns the PRIMITIVE_NAME by default, so that T_BEGIN->getCSName is actually "Begin".  So I may have fixed a typo in the coding, but not the documentation.

brucemiller:
Yes, you're right: it returns "Begin" 'cause that's what TeX does, so my POD markup and the doc itself was messed up.

Aside: the git hook just *checks* the formatting (and complains if it doesn't like it), but doesn't actually reformat. ```tools/latexmllint``` will reformat it for you, if you want.  (But none of this will likely come into play in this case)

brucemiller:
I guess I forgot to pull this.... Thanks!

PR #808, opened 2016-11-02
Fix typo in command line call

ianhbell:
Also clarify that the flags should be passed to the latexmlpost command (I hope I got the markup right)


brucemiller:
Thanks for the clarification!

Issue #580, opened 2015-02-13
Binding request: siunitx & friends, enumitem, type1cm, listingsutf8, braket


matteosecli:
I've just encountered some unsupported packages while converting one of my documents, that prevent my document to be converted. They are:
- siunitx.sty,
- enumitem.sty,
- type1cm.sty,
- listingsutf8.sty, 
- braket.sty.

I think that the most important is "enumitem", because it allows lists in the form a) b) c) d) etc..., that are a daily matter in writing LaTeX documents. Then, the second most important is "braket", because it's widely used if you write quantum mechanical stuff. I don't really have the necessary knowledge to help you (otherwise I would), but I don't think that translating "braket" macros into HTML should be so difficult, since the commands can be rewritten in plain LaTeX.
Thank you for your efforts and for your program; it's AMAZING! So far, it's the best in what it does, and I've seen huge improvements during these years.
Keep on this way! :D

brucemiller:
Thanks for the kind words; it's appreciated!

* I think that type1cm.sty is pretty much ignorable, so I've created an empty binding for that.
* I'm hoping that the encoding issues that make listingsutf8 necessary aren't relevant to latexml, so I've defined listingsutf8 to simply load listings; let me know if that works out.
* The two packages siunitx and braket have some interesting semantics and so really should be done.
* and enumitem is also an interesting and useful package, but will take some time to integrate into latexml.

So, sofar, I've only fixed the least useful parts! Hopefully won't take till 0.9 to get done, but I'll leave that Milestone for now

matteosecli:
Thank you for the quick response!
I've just compiled and tried the new version, and now my document compiles (if you wanna use it as a test for the other packages I've reported, it's in one of my repos: https://github.com/matteosecli/QMC/blob/master/LaTeX/main.tex). It's -- obviously -- full of errors about the "braket" package, but apart from that the result is still nice. The listingsutf8 package works correctly, now; and I agree with you that "type1cm" doesn't matter for a HTML output, it was just for a better PDF.
I've read a little bit the documentation and the code that you used to implement some packages; I've also read about sTeX and the importance of semantic. I hope I will learn enough to eventually give some help, but it will take me quite a long time because I've a lot of stuff to program in other projects and I don't know Perl at all.
In the meantime, I will keep reporting bugs! ;)

brucemiller:
[...]

PR #817, opened 2016-11-25
basic experimental bindings for sinutis, hepunits and friends

mazenafif:


dginev:
I think it is too early to merge these into the latexml master, since they are just placeholders. There is an open issue at #580 for them, and I know that @urabenstein has been looking into writing out the full bindings.

A PR with fleshed out bindings is certainly welcome, but until then I will close here. Thanks!

dginev:
And another note is about where the PR was coming from - the ```Authorea:master``` tracks latexml and includes Authorea-specific extensions, so it shouldn't be a source for a PR heading back into upstream. Instead, it should come from a fork of the brucemiller/latexml repo in a special feature branch.

PR #824, opened 2017-01-01
Consistently use the XML::XML_NS namespace uri

dginev:
A little cosmetic improvement I spotted while studying the code.

The trailing whitespaces removed are a side-effect of my code editor (sublime text), which right trims lines automatically. Also cosmetically appealing.


brucemiller:
looks reasonable; Thanks!

PR #839, opened 2017-05-22
add onlyPreamble checks for \documentclass and \documentstyle

dginev:
Add `onlyPreamble` guards for class-selection LaTeX macros. The absence of these errors became apparent during an Authorea debugging session, so I'm contributing them back upstream.

The goal is that given input such as:
```
\documentclass[11pt,letter]{article}


\begin{document}

\documentclass[letterpaper,10pt]{article}
...
\end{document}
```

We would like LaTeXML to report a preamble violation error as in:
```
Error:unexpected:\documentclass The current command '\documentclass' can only appear in the preamble
```

This is all setup already for usepackage, so just leveraging that infrastructure in this PR.

Bonus: My sublime editing environment removes trailing whitespace (that is actually a tiny speedup for RawTeX I believe)


brucemiller:
Thanks!

PR #842, opened 2017-06-15
Add archlinux setup instructions

tkw1536:
This commit adds instructions to the web page on how to install latexml
on archlinux and in particular how to setup dependencies.


dginev:
I would always recommend installing from source via `cpanm`, also for archlinux (did that myself, worked like a charm). Good to collect the list of prerequisite packages of course.

PR #845, opened 2017-06-23
More updates to archlinux setup instructions

tkw1536:
This PR makes more updates to the archlinux setup instructions -- in particular incorporating the suggestion from @dginev in the last commit to install via CPAN instead.


dginev:
For the record, my suggestion was to install via `cpanm`, which should also be platform-independent. 

Can be as simple as:
```
cpanm git://github.com/brucemiller/LaTeXML.git
```

or if you are working from the clone itself:
```
cpanm .
```

Adding a `--notest` in there makes it quite fast for doing quick builds while developing and still using the simple `latexmlc` (and friends) commands without specifying any concrete paths. This is definitely the easiest way to work if you don't have `sudo` access to a machine, everything can live in the user's home directory.

Readme is quite good too:
https://github.com/miyagawa/cpanminus/

tkw1536:
Ok, I misunderstood then. I will update the documentation further.

brucemiller:
On 06/23/2017 12:57 PM, Tom Wiesing wrote:
> Ok, I misunderstood then. I will update the documentation further.

I'm thinking that this would be a good addition to the "CPAN installation"
subsection?

brucemiller:
A little more on the topic:
I think that it's important to point out that you
can install the released LaTeXML (& all its dependencies)
from CPAN; and also important to point out that you
can install the development version (& dependencies)
using cpanm, conveniently and w/o sudo.

But it's also important to point out how to
install as much as possible using your systems
software repository, as that allows you to keep
all that software up-to-date w/o remembering all
the different places you need to run updates.
Unfortunately, even the linux systems tend to
need special instructions for each since the
package naming conventions are so different.

dginev:
Sounds great! Btw knowing Tom, he may want to also create an example LaTeXML docker image and upload it to https://hub.docker.com/ , given that he's had plentiful Docker experience at KWARC.

In fact funnily enough there seems to already be one:
https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=latexml&starCount=0

The usual question is what a docker image is actually useful for - and the main answer is quickly spinning up microservices of a certain kind. A common example nowadays, with cloud latex startups popping up left and right, is starting an array of worker machines with texlive capability, which becomes very easy once you have:
https://hub.docker.com/r/harshjv/texlive-2015/

Whether there is any usefulness of preparing this also for the main latexml versions, and whether Tom should be the one doing that is another question, so this is just a random thought and not a real short-term suggestion. We're yet to dockerize our LaTeXML setup at Authorea, as we're yet to farm out that work, but once we do we would probably use docker hub for that.

tkw1536:
@dginev Looking at the docker image you linked that seems to be using the version that is in the debian repository - which may not be the best idea. 
If I do make one (we should probably have a separate issue on that), I would very much prefer basing it on a different method so that we do not have to rely on the distributions to upgrade their repositories.

tkw1536:
I have written and added a paragraph on how to install via cpanm.

PR #847, opened 2017-07-08
Change shebang to /usr/bin/env/perl

tkw1536:
On certain machines, when run directly, all scripts would default to use
the perl interpreter located in /usr/bin. This can cause problems on
machines where perl is located in a different path, or where multiple
perl instances are present on a system (such as on Mac when using
homebrew).

This PR updates the shebang, the line which starting with #!
indicating the interpreter to use when running a file, from
/usr/bin/perl to /usr/bin/env perl. This makes sure not to hard-code the
perl path to /usr/bin, but to instead search $PATH for the appropriate
file.

### DIFF
--- a/bin/latexml
+++ b/bin/latexml
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl -w
+#!/usr/bin/env perl
 # /=====================================================================\ #
 # |  latexml                                                            | #
 # | main conversion program                                             | #
--- a/bin/latexml
+++ b/bin/latexml
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl -w
 # /=====================================================================\ #
 # |  latexml                                                            | #
 # | main conversion program                                             | #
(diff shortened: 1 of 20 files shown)


brucemiller:
hmm... any security implications? i'll have to look into that.

dginev:
Tom's change looks to be the best practice:
http://perldoc.perl.org/perlrun.html

But why did travis fail?

dginev:
Travis failed because the env form does not allow passing arguments, such as `-w` to the interpreter. And there are trade-offs it seems:
https://unix.stackexchange.com/a/29620

dginev:
As can be seen in the wisdom in the unix stackexchange comment i shared, the `-w` can be upgraded to an explicit `use warnings;` in each of the executables, which solves the travis failure. Personally I am happy either way, I have used both forms in the past and my setups are simple ("standard"?) enough to work with either form.

tkw1536:
I can update the PR accordingly, which should fix the Travis Tests.

brucemiller:
For the record, there is a slight downside to this: If you run ```valgrind``` on latexml, it no longer "sees" perl, but (apparently) checks the memory usage of env(?).  The easy workaround is
```
valgrind perl path/to/latexml ...
```
Just so you know...

PR #851, opened 2017-07-13
Add new DefHook and AddHook functions

tkw1536:
This PR adds two new functions to Package.pm called `DefHook` and
`AddHook`.

These allow to define hooks for existing macros. A hook can work like a
Primitive, a Macro or a Constructor. However, instead of only
performing the specified actions, a hook also runs the existing implementation.
The primary use-case is to introduce side-effects to an existing macro.

The `DefHook` function implements creating and registering a new control
sequence that perfoms a user-given side-effect, whereas the `AddHook`
function only registers an existing control-sequence as a hook for a
second already-existing control sequence.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -55,6 +55,7 @@ our @EXPORT = (qw(&DefAutoload &DefExpandable
     &DefPrimitive  &DefPrimitiveI
     &DefRegister &DefRegisterI &LookupRegister &AssignRegister &LookupDimension
     &DefConstructor &DefConstructorI
+    &DefHook &AddHook
     &dualize_arglist &createXMRefs
     &DefMath &DefMathI &DefEnvironment &DefEnvironmentI
     &convertLaTeXArgs),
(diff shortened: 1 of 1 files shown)


tkw1536:
Me and @brucemiller had a talk about this yesterday, and would like to hear some more comments in particular from @dginev. 

We are both essentially in agreement that the `DefHook` functionality is useful, but also see some problems with it. 

It is essentially a nice shortcut for
```latex
\let\originalmacro\macro
\def\macro{\hookmacro\originalmacro}
\def\hookmacro{% Hooked code
}
```

This can ([and is](https://github.com/brucemiller/LaTeXML/blob/5a242423d7b4b9e1840579fbb50c18572ecbfa03/lib/LaTeXML/Package/algorithmic.sty.ltxml#L21-L32)) easily written with the help of existing macro, however if done multiple times and taking into account macro arguments, this can become tedious. 

Furthermore, the implementation in this PR runs the hooked in code first, and then proceeds to run the existing macro. In practice both scenarios are useful. 
Furthermore, sometimes not all arguments are needed in the hook macro, and only some of them should be passed along. 
Together, this can lead to a lot of variety in use-cases that may be difficult to implement with a single (or two) `sub`s; a TeX based implementation may lead to more flexibility.

brucemiller:
To clarify my POV, hopefully without biasing the discussion too much: I certainly appreciate the sentiment of handy concise hooks, daemons & wrappers, I've got reservations whether it is worth it.  This kind of hook happens somewhat frequently (not as much as I thought, though), but there are always variations like which args are needed by the hook, which order the hook & original should run, whether the hook is a macro, primitive or constructor.  So while having DefHook makes it concise, it obscures the basic (& usually simple) TeX that's going on behind it.  And it;s another LaTeXML-specific bit of code that has to be remembered and understood, not to mention perhaps needing to be reimplemented in some other  (cough) language.

dginev:
> And it;s another LaTeXML-specific bit of code that has to be remembered and understood, not to mention perhaps needing to be reimplemented in some other (cough) language.

(heavy cough) I would have to agree 100% with Bruce here (cough continues)

dginev:
Also I've been talking to functional people recently so my gut feeling when I read
> introduce side-effects

was DON'T! Haha. Hooking into existing macros is a very opaque technique, and makes programming in tex a bigger nightmare than it has to be. I would almost always recommend adding a new API on top rather than silently overhauling the existing macro.

Sadly the technique _is_ common, but can't we let TeX package writers deal with this entirely in TeX? Do we need it in writing latexml bindings? I recall doing the Let() jumps and hoops, but it actually made me happy to be explicit about the low-level bits. At the same time I have an obvious bias to please not add more to Package.pm until I stop (cough)-ing.

brucemiller:
On 07/13/2017 11:18 AM, Deyan Ginev wrote:
> Also I've been talking to functional people recently so my gut feeling
> when I read
>
>     introduce side-effects
>
> was DON'T!

Well, a healthy respect for principles of functional programming
is a good thing, so long as you recognize that a program with
no side effects is actually useless.

But actually the notion of "side effect" is a bit of red-herring, here.

> Haha. Hooking into existing macros is a very opaque
> technique, and makes programming in tex a bigger nightmare than it has
> to be. I would almost always recommend adding a new API on top rather
> than silently overhauling the existing macro.
>
> Sadly the technique /is/ common, but can't we let TeX package writers
> deal with this entirely in TeX? Do we need it in writing latexml
> bindings?

Ironically, I think that's exactly why we do need it!
The New Paradigm (wow!) is to try loading the raw .sty files,
whenever possible. Ideally we only have to hook into a few places
to get the XML/semantic benefits, depending on the programming
style of the package --- eg. pgf has a nice set of driver hooks designed in.
Sometimes it's best to redefine a macro, others to piggy-back a hook
onto it.

>At the same time I
> have an obvious bias to please not add more to Package.pm until I stop
> (cough)-ing.

Yeah, it was probably unfair to Tom's argument for me to bring that aspect
into it, knowing how you'd respond :>

tkw1536:
Judging from the discussion on this, I do not really see the need for this any more. Closing.

PR #852, opened 2017-07-13
Add TexLive to Travis Tests

tkw1536:
This PR updates the Travis tests to run both with and without TexLive installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,30 @@
 language: perl
-perl:
-  - "5.20"
-  - "5.18"
-  - "5.14"
-  - "5.10"
+sudo: required
+dist: trusty
 
-sudo: false
 addons:
   apt:
     packages:
      - libdb-dev
      - libxml2-dev
      - libxslt1-dev
+
+perl:
+  - "5.20"
+  - "5.18"
+  - "5.14"
+  - "5.10"
+
+env:
+  - TEX=none
+  - TEX=texlive
+
+before_install:
+  - if [[ "$TEX" != "none" ]]; then sudo apt-get update -q; fi
+  - if [[ "$TEX" == "texlive" ]]; then sudo apt-get install -y texlive-full -q; fi
+
+install:
+  - cpanm -v --installdeps --notest .
+
+script:
+  - perl Makefile.PL && make test


dginev:
Don't you need a matrix or some such in the yml file? It would be awesome if we can travis-test both the with and without setups, that's really great! I just don't want us to use the no-texlive successes.

Really cool idea!

tkw1536:
> Don't you need a matrix or some such in the yml file? It would be awesome if we can travis-test both the with and without setups, that's really great!

This already makes a matrix. From [the travis documentation](https://docs.travis-ci.com/user/languages/perl/):
> For Perl projects, `env` and `perl` can be given as arrays to construct a build matrix.

You can have a look at the [Travis Test](https://travis-ci.org/brucemiller/LaTeXML/builds/253347396) for this PR, there are 2*4=8 test cases, for each perl version there is one with `TEX=none` and one with `TEX=texlive`

> I just don't want us to use the no-texlive successes.

In order for the build to count as passing, all subtests have to pass. That means both with and without texlive, for each given `perl` version the tests have to succeed.

dginev:
Super cool! I'm two thumbs up for this then, great upgrade

dginev:
Btw I would also be for dropping the `5.18` version and adding the latest perl version supported by Travis, so that we stay compliant with the latest Perl. My ubuntu has `5.24` wondering if Travis has that too?

tkw1536:
Travis should have that one too. According to [their documentation](https://docs.travis-ci.com/user/languages/perl/) they use `perlbrew` which goes up `5.26` I believe.

tkw1536:
I would update the PR, but I am not sure why the exact versions that are currently tested have been choosen, so I will leave it to you to change `travis.yml` (allow edits from maintainers is on).

brucemiller:
Too bad they don't have a way to specify all perls back to 5.10 (say).   I'd certainly like to test against the newest perls! I don't recall who wrote the original travis.yml, but I'd forgotten about its existence until I needed to disable 5.10 for some XS testing.  But since this patch does (at least) double duty, I wonder if we should prune some of the middle versions? 5.14, 5.20, 5.24 or something?  I'd hate to overstay my welcome with travis :>

PR #854, opened 2017-07-19
Sort entries in Manifest File

tkw1536:
Over time, several file names inside the `MANIFEST` file have gotten out of order. This PR accomplishes two things (1) it sorts the lines to be in order again and (2) adds a script to (semi-) automatically regenerate the MANIFEST file in a sorted fashion the future.


brucemiller:
This seems like it could occasionally be useful, at least for testing branches; I'd be wary of using it on master, since you want to add files to the MANIFEST only very intentionally, and making sure that they're also under version control.

PR #858, opened 2017-08-02
Set @filelist macro before reading file (fixes #857)

### LINKED ISSUE #857: Weird @filelist behaviour when loading cls file
The ``@filelist`` macro seems to be  empty when reading in a class file. Consider the following situation:

``filelist.tex`` contains:
```latex
\documentclass{filelistclass}
\begin{document}
  The file list inside of the cls file was `\filelistclass'. 
\end{document}
```

``filelistclass.cls`` contains:
```latex
\let\filelistclass\@filelist
\LoadClass{article}
```

``filelistclass.cls.ltxml`` contains:
```perl
InputDefinitions('filelistclass', type => 'cls', noltxml => 1, withoptions => 1);
1;
```

When running this with `pdflatex filelist.tex` we get
The file list inside of the cls file was ‘filelistclass.cls’.
indicating that `\@filelist` is `filelistclass.cls`. 

When running this with `latexml filelist.tex` we instead get
```html
<p>The file list inside of the cls file was ‘’.</p>
```
indicating that `\@filelist` is empty.  

I am working on a fix for this, will make a pull request in a bit.

tkw1536:
When loading a file through `InputDefinitions`, the `@filelist` macro was
only set after the file has finished loading (see #857). This results in the macro
being empty during the processing of a file.

This commit fixes the issue by setting the `@filelist` macro before
loading of the file. Furthermore, it adds a testcase to prevent this bug
from reappearing in the future.


brucemiller:
Looks good, but you forgot to add your test case files to the MANIFEST

brucemiller:
DOH!!! It's the first file in the commit !?!?!

PR #860, opened 2017-08-04
Keyval Improvements and xkeyval support

tkw1536:
Tl;DR: This PR adds major improvements to the handling of KeyValue arguments, mostly based on the `xkeyval` package. 

I have been working on this for the past few weeks and I am going to try to summarize the changes that I have made. I probably forgot something, so please have a look at the diffs as well. 

The KeyVals logic, previously contained in `KeyVals.pm` and partially in `TeX.pool.ltxml` and `Package.pm`, has been completely refactored and split into two files `KeyVals.pm` and `KeyVal.pm`. 
  * `KeyVal.pm` provides a new `KeyVal` class. This gives access to *a single* key, allowing it to be read and set. The file also contains a refactored `DefKeyVal` implementation (previously found in `Package.pm`. It is fully backwards compatible, so no rewriting of existing bindings is necessary. With the new class, `DefKeyVal` behaves much like its native TeX counterpart would. Furthermore, this refactoring adds extended support for the `xkeyval` package, such as `command`, `choice` and `boolean` keys. 
  * `KeyVals.pm` provides the `KeyVals` class. Each instance of it represents a set of Key-Value pairs. This class has also been refactored, making the code cleaner and adding support for advanced `xkeyval` features such as key prefixes, setting keys in more than one family, and more. 
Unfortunately, the constructor of the `KeyVals` class is not completely backward compatible, so amongst some others, the `listings` package binding had to be slightly updated. The new constructor is fully documented with POD and making the change for other bindings (if any) should be straightforward. 
  * In `TeX.pool.ltxml`, the `RequiredKeyVals` and `OptionalKeyVals` argument types have been given additional arguments and variants with '*' and '+' have been added. This is again based on the `xkeyval` package, and has been documented using in-line comments. The existing argument types are fully backward compatible. 
* `keyval` package bindings have updated to match the new `KeyVals` class. 
* `xkeyval` and `xkvview` package bindings have been added. 
  * support for defining and setting KeyVals using this package
  * support for `\DeclareOptionX` and friends, which is an extended version of the LaTeXs `\DeclareOption`, both for `sty` and `cls` files. 
* lots and lots of test cases. 

This does not support the pointer system of `xkeyval`. It is also not possible to preset keys. I plan to make a second pull request for these in the near future. 

/cc @brucemiller


dginev:
This sounds like truly great work!

> lots and lots of test cases.
And I can't celebrate this enough! :tada:

tkw1536:
Pull requested rebased and updated after offline suggestions from @brucemiller:

* made sure that the first in `DefPrimitive` is called `$stomach` even though it is unused. 
* renamed test cases in `t/keyval_dox` to `t/keyval_options` for clarity

brucemiller:
Looks good, except that you didn't change to keyval_options in the MANIFEST(?)

tkw1536:
Oops, I have fixed that now.

brucemiller:
Looks good... and impressive! Thanks!

PR #862, opened 2017-08-04
Also record labels as metadata for longtables, when requested in lxRDFa

dginev:
This PR is a follow-up to #596 and I would appreciate a quick review & merge, if possible.

It turns out that my generic recording of `\label` as metadata did not extend to packages redefining it, such as longtable.

The problem there is that longtable redefines (via `Let`) the `\label` macro on each environment start, which makes it impossible to maintain the RDFa hook I introduced globally in the preamble. However, I _can_ hook into the lower level `\@longtable@label` macro, and it all works great again!

Example test being:
```tex
\documentclass{article}
\usepackage{longtable}
\usepackage[labels]{lxRDFa}
\begin{document}

\begin{longtable}[]{@{}lccc@{}}
1 & 2 & 3 & 4 \\

\caption{\label{table:long} caption}
\end{longtable}

I should be able to refer to Table \ref{table:long}.

\end{document}
```

with the expected snippet of metadata produced:
```html
<figcaption class="ltx_caption" property="dcterms:alternative" content="table:long">
```

Tested locally and all works well.

### DIFF
--- a/lib/LaTeXML/Package/lxRDFa.sty.ltxml
+++ b/lib/LaTeXML/Package/lxRDFa.sty.ltxml
@@ -19,7 +19,10 @@ use LaTeXML::Package;
 # Package Options
 DeclareOption('labels', sub {
   Let(T_CS('\lxRDF@original@label'),T_CS('\label'));
+  Let(T_CS('\lxRDF@original@longtable@label'),T_CS('\@longtable@label'));
   DefMacro('\label Semiverbatim','\lxRDF@original@label{#1}\lxRDFa{property=dcterms:alternative,content=#1}');
+  DefMacro('\@longtable@label Semiverbatim',
+           '\lxRDF@original@longtable@label{#1}\lxRDFa{property=dcterms:alternative,content=#1}');
 });
 
 ProcessOptions();


brucemiller:
Hmm... I guess the original option "labels" was a patch from you, and I accepted it?  But...

It kinda seems the wrong way to go; since we'd conceivably need to mung every package that plays around with ```\label```.  Shouldn't it be more correct to use Tag (or something similar) to check for any labels attributes, and add RDF when appropriate?

dginev:
But then I would need to add that rule to all heading-like tags? Or attach it to `*` which would hurt performance badly. 

I double-checked the bindings, i think this is the only case one needs to make a special exception. All other table packages, such as tabularx or tabulary keep using the standard label.

Happy to backtrack when it becomes unmanageable (which I hope will never happen)?

brucemiller:
Have you actually tried whether it really hurts performance? I'd bet it's not as bas as you think (but might be).

dginev:
I mean I have a perfectly safe 3-line PR that does everything one needs with no issue...

dginev:
Maybe a compromise - if we merge here as-is, I promise to rework this using `Tag` the next time I find a missing label macro that isn't supported.

brucemiller:
On 08/24/2017 10:32 AM, Deyan Ginev wrote:
> Maybe a compromise - if we merge here as-is, I promise to rework this 
> using |Tag| the next time I find a missing label macro that isn't supported.

Better trade: I merge, you swallow a few "i told you so"'s next time
I start whining about XS ?

dginev:
XS? I love XS! I'm sure it will work out in the end :) Thanks for merging :+1:

brucemiller:
OK, I'm convinced!

PR #867, opened 2017-08-29
Fix missing runtime_class warning

tkw1536:
When generating `LaTeXML::MathGrammar`, recent versions of `Parse::RecDescent` were generating a warning about the missing `runtime_class` option. This warning indiciated that the grammar may stop working with future versions of the RecDescent package.

This PR adds the missing option to the call inside of Makefile.PL, ensuring that the default runtime `Parse::RecDescent` will continue to be used.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -174,7 +174,7 @@ sub compile_MathGrammar {
 pure_all :: $(INST_LIBDIR)/LaTeXML/MathGrammar.pm
 
 $(INST_LIBDIR)/LaTeXML/MathGrammar.pm: lib/LaTeXML/MathGrammar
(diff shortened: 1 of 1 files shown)


brucemiller:
Cool! Thanks!!

brucemiller:
whoops... this doesn't actually work; at least not with Parse::RecDescent 1.965.  if you do ```make distclean``` to start fresh, then perl Makefile.PL and make, you get
```
Usage: perl -MLocalTest - <grammarfile> <classname>
```
it seemingly doesn't accept the 3rd arg on the line.

tkw1536:
Oops, didn't consider that older dependency versions do not have the third argument. I will need to add a fallback case.

PR #868, opened 2017-08-30
Bugfix for older Parse::RecDescent versions

tkw1536:
Older versions of Parse::RecDescent did not support the third argument
to parser pre-compilation introduced in
b5aa3d04474538ea0d601d99e3e45d7a4df89aaa. This caused installation
problems for LaTeXML on older systems.

This PR fixes the problem by introducing a fallback that does not
use the third argument. This only gets triggered when the MathGrammar is
not created.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -175,6 +175,9 @@ pure_all :: $(INST_LIBDIR)/LaTeXML/MathGrammar.pm
 
 $(INST_LIBDIR)/LaTeXML/MathGrammar.pm: lib/LaTeXML/MathGrammar
 	$(PERLRUN) -MParse::RecDescent - lib/LaTeXML/MathGrammar LaTeXML::MathGrammar Parse::RecDescent
+	@$(PERLRUN) -e 'exit(1) unless -e "MathGrammar.pm";' || \
+		(echo "Parse::RecDescent failed to created parser, trying to use old implementation. " && \
+		$(PERLRUN) -MParse::RecDescent - lib/LaTeXML/MathGrammar LaTeXML::MathGrammar)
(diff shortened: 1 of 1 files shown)


tkw1536:
@brucemiller can you check if this works on your problematic systems before merging?

Makefile.PL
```diff
@@ -175,6 +175,9 @@ pure_all :: $(INST_LIBDIR)/LaTeXML/MathGrammar.pm
 
 $(INST_LIBDIR)/LaTeXML/MathGrammar.pm: lib/LaTeXML/MathGrammar
 	$(PERLRUN) -MParse::RecDescent - lib/LaTeXML/MathGrammar LaTeXML::MathGrammar Parse::RecDescent
+	@$(PERLRUN) -e 'exit -e ("$(INST_LIBDIR)/LaTeXML/MathGrammar.pm") ? 0 : 1;' || \
```
dginev: `exit -e` ? Don't you mean `perl -e`?

tkw1536:
No, I actually mean -e inside the perl code to check if the given file exists.

Makefile.PL
dginev: the leading `@` on this line confused me enough to lose sight of the perlrun for my older question. What does leading `@` do?

tkw1536: Prevents the command from being echoed out. I added that to keep users from being confused by a 'fallback' message in the command, even though it is not relevant when not called.

dginev: I see, thanks for educating me!

dginev: I remember having known this years ago and then forgetting it, as bash syntax often goes.

tkw1536:
OK, I managed to test this on an older machine, and need to make one more change for this to work.

PR #869, opened 2017-08-31
Refactor LaTeXML::Core::Box blessed objects

tkw1536:
Previously, objects representing a `LaTeXML::Core::Box` were an array, whereas objects representing `LaTeXML::Core::Whatsit` (which inherit from Box) were a hash.  This led to duplication of several methods in the Whatsit class.

This PR refactors `LaTeXML::Core::Box` (and `LaTeXML::Core::List`, `LaTeXML::Core::Comment`, `LaTeXML::Core::Whatsit` that inherit from it) to be a blessed hash reference instead. This makes code within between the four classes consistent with each other, and also makes code more readable due to named access of properties.

### DIFF
--- a/lib/LaTeXML/Core/Box.pm
+++ b/lib/LaTeXML/Core/Box.pm
@@ -41,7 +41,10 @@ sub Box {
 
 sub new {
   my ($class, $string, $font, $locator, $tokens, %properties) = @_;
-  return bless [$string, $font, $locator, $tokens, {%properties}], $class; }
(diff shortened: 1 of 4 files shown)


tkw1536:
oops, looks like one of the tests is suddenly failing on this (I remember them going through). I will have to have another look.

tkw1536:
Fixed, was caused by overlooking a special case in `LaTeXML::Core::Box->getProperty`.

PR #872, opened 2017-09-28
Add error when readBalanced fails to balance at end of Mouth

dginev:
This was a bit of a shocker in Authorea editing. There were obvious pieces of wrong LaTeX content that never returned an error from latexml, such as:

```
\textbf{some text here
```

Importantly, this is being converted in fragment mode, and has no `\end{document}` in _the same mouth_ to throw a grouping error. This revealed a particular case of unreported errors in LaTeXML:

**When a Gullet->readBalanced call does not have a closing brace, the read silently succeeds to slurp all content until the end of the Mouth**

This can cause some serious issues if the error is committed early in a large chunk of latex, naturally. So I suggest throwing an informative error, and then proceeding as usual. It is good enough for users to correct their mistake in online editing. Example interface from Authorea:
selection_999 032

Feedback welcome on how exactly to structure the error.

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -361,6 +361,11 @@ sub readBalanced {
       push(@tokens, $token); }
     elsif ($cc == CC_MARKER) {
       LaTeXML::Core::Definition::stopProfiling($token, 'expand'); } }
+  if ($level > 0) {
+    # TODO: The current implementation has a limitation where if the balancing end is in a different mouth,
+    #       it will not be recognized.
+    Error('expected', "}", $self, "Gullet->readBalanced ran out of input in an unbalanced state.");
+  }
   return Tokens(@tokens); }
 
 sub ifNext {
@@ -819,7 +824,7 @@ sub readInternalMuGlue {
 
 __END__
 
-=pod 
+=pod
 
 =head1 NAME
 
@@ -963,7 +968,7 @@ and return the value.  Returns undef if the next token isn't such a register.
 =item C<< $number = $gullet->readNumber; >>
 
 Read a L<LaTeXML::Common::Number> according to TeX's rules of the various things that
-can be used as a numerical value. 
+can be used as a numerical value.
 
 =item C<< $dimension = $gullet->readDimension; >>
 


dginev:
To reproduce the error in the master latexml from the command-line, here is a one-liner:

```
latexmlc --whatsin=fragment --preload=article.cls 'literal:\textbf{example...'
```

I would clearly expect this to throw an error, such as the on from this PR

brucemiller:
Thanks;  Actually, I have a nagging doubt that there was a reason to leave w/o error, but can't remember why; perhaps the usually-expected error that follows is a clearer one?  Anyway, merge and see what happens :>

dginev:
I was investigating the potential reasons, and the most pertinent bit of logic was the fact some Mouths `autoclose` and some do not. There is some subtle intention there that I half-understood, but I tested around several conditions and none seemed to print a decent error for this specific case. Maybe it is even subtler!

We'll be informed upon the next arXiv rerun :>

dginev:
I may have found out why it was there. I am now getting a rather unpleasant error when loading elsarticle.cls, on a vanilla empty-ish input:
```
gullet->readbalanced ran out of input in an unbalanced state. /usr/local/share/perl/5.22.1/LaTeXML/Package/elsart_support.sty.ltxml line 236 Input is empty In Core::Gullet[@0x30a84f18] /usr/local/share/perl/5.22.1/LaTeXML/Package/elsart_support.sty.ltxml line 236 <= Core::Definition::Primitive[\newenvir... <= Core::Stomach[@0x30ace348] <= Core::Gullet[@0x30a84f18] <= ..
```

Hm, need a way to keep the error for the good cases, but kick it out of here.

dginev:
Basically it's the technique used internally by RawTeX, I'm amazed we don't have any tests that use that and would fail? 

```perl
$stomach->getGullet->readingFromMouth(LaTeXML::Core::Mouth->new($text), sub {
```

dginev:
Well, while I have connected the error to the piece of code using a new Mouth, I haven't yet connected to the exact reason `readBalanced` would fail on a perfectly balanced input.

dginev:
Exact RawTeX line with this error is:
```perl
RawTeX('\newenvironment{pf}{\begin{@proof}[\proofname]}{\end{@proof}');
```

dginev:
One-liner to reproduce:
```
latexmlc --preamble='literal:\documentclass{elsarticle}' literal:test
```

dginev:
Oh my.

After staring here for an unreasonable amount of time... Is there a typo in the RawTeX string? I printed a whole lot of Dumper messages and it now looks as if the ending:
````
{\end{@proof}
```

should have an additional ending brace:
```
{\end{@proof}}
```

Does this make sense? I can commit that right away.

dginev:
I have verified this is the issue, will push directly to master (since we're talking 1 character)

dginev:
Done: a09061140dbb5be4231ae97453f4ada8badeb01e

Hope you don't mind my first commit to master @brucemiller ! I think this PR may catch some interesting RawTeX oversights now, more convinced it was correct to add this error message.

brucemiller:
bravo & welcome ! :>

PR #881, opened 2017-10-23
update Mailing list

kohlhase:
it has moved,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ doc/site/examples/tabular/tabular.pdf
 doc/site/examples/tabular/tabular.xml
 doc/site/index.xml
 doc/site/releases.tex
+auto
--- a/README.pod
+++ b/README.pod
@@ -17,7 +17,7 @@ where you can retrieve and browse the current source,
 along with an L<Issue tracker|https://github.com/brucemiller/LaTeXML/issues>
 and L<Wiki|https://github.com/brucemiller/LaTeXML/wiki>.
 
-For general discussion feel free to join the L<mailing list|http://lists.kwarc.info/mailman/listinfo/project-latexml>.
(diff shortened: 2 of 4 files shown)


dginev:
Wow, just saw this PR and realized I sent my last list email to the wrong location... explains why it bounced!

brucemiller:
2 questions:  Is "auto" an ignorable file I should know about, or some special .gitignore directive?

And I had the impression that "kwarc"  (and the domain) was moving to erlangen, and services moved servers but not url's... Not complaining, just curious why the change.

kohlhase:
On 15.11.17 15:41, bruce miller wrote:
>
> 2 questions: Is "auto" an ignorable file I should know about, or some
> special .gitignore directive?
>
when I run LaTeX from emacs, it makes a directory "auto" with *.el files
in it, and it bothers me seeing those, feel free to revert this change.
>
> And I had the impression that "kwarc" (and the domain) was moving to
> erlangen, and services moved servers but not url's...
>
yes, that was the plan why we set up mailman on kwarc. But FAU will only
allow a mail server on a domain, if you completely turn it over to them
(i.e. turn over kwarc.info, lists.kwarc.info is not sufficient), and I
was not prepared to do that. Thus the suboptimal transition.
>
> Not complaining, just curious why the change.
>
I am complaining as you see :-).

Michael

tkw1536:
> when I run LaTeX from emacs, it makes a directory "auto" with *.el files in it, and it bothers me seeing those, feel free to revert this change.

Since this is editor specific, and not every LaTeXML contributor uses emacs, it sounds like something that belongs in a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

brucemiller:
LaTeXML contributors that don't use emacs should generally keep quiet about it.

dginev:
> LaTeXML contributors that don't use emacs should generally keep quiet about it.

:no_mouth:

brucemiller:
Thanks!

PR #885, opened 2017-11-15
Preserve image graphic in fragment when not found

dginev:
So, one more interesting use case for me today, with an interesting issue.

I needed to run through some latex through latexml that was "detached" from its auxiliary files (crucially: images), but I wanted to eventually reconnect the HTML result to the original images. I am happy to do some post-processing after latexml, in essence all I needed was to have the includegraphics file argument preserved in the final `src` attribute of the HTML `img` tag.

That's really easy to do! And in fact should be better practice than the current master behavior, which is to leave an empty `src=""` attribute, which is technically invalid / discouraged HTML.

I also caught a tiny bug with one conditional (for activate latexmlc post-processing) and disabled a somewhat useless guard that stops you from using `--graphicimages` in certain modes. I think there are legitimate use cases now to allow that.

Maybe not the ideal fix - open to feedback. The true core of this PR is the line:
```
+    $node->setAttribute('imagesrc', $node->getAttribute('graphic'));
```

which is what I truly need.

P.S. I ended up laughing to myself after finally seeing (or yet again remembering?) the double entendre of the `--graphicimages` and `--nographicimages` options. To the careless observer this may look like we are running our own content safety filters!

### DIFF
--- a/lib/LaTeXML/Common/Config.pm
+++ b/lib/LaTeXML/Common/Config.pm
@@ -431,12 +431,13 @@ sub _prepare_options {
   # Any post switch implies post (TODO: whew, lots of those, add them all!):
   $$opts{math_formats} = [] unless defined $$opts{math_formats};
   $$opts{post} = 1 if ((!defined $$opts{post}) &&
-    (scalar(@{ $$opts{math_formats} }))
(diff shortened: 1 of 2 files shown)


brucemiller:
Looks reasonable; Thanks!

PR #886, opened 2017-11-16
Ensure mathsvg produces svg in whatsout=math when html format

dginev:
Me again. Turns out my PR yesterday was operation on `--whatsout=math` but only for the `xml` format target. I added some extra robustness so that the html family out whatsouts is also recognized.

I don't need this in a rush, since I can happily use the `xml` format myself, but may be useful in general.


brucemiller:
Thanks!

PR #889, opened 2017-11-20
Robust preloads for daemon reruns, unlink cache for mathsvg

dginev:
This PR is a bit more critical as it fixes a long-standing (and shocking) bug with daemonized runs.

Namely, the preload `insertPI` logic was wrongly modifying the preloads directly in the latexml object, so reruns would see all preloads as extension-free and mistreat classes as packages.

if we `--preload=article.cls`, the first run would run correct, but overwrite the value to `article` (no extension) and the second daemonized run would wrongly try to load article as a package. Pretty shocking! It is painfully obvious with `--mathsvg` as the resulting `\usepackage{article}` makes the SVG conversion fail entirely, which lead me to debugging & fixing.

Additionally, I made sure to remove the cache file, which is something I borrowed from latexmlmath.


brucemiller:
Sounds reasonable; Thanks!

PR #906, opened 2017-12-17
Use library to generate XML in maketests instead of system()

bfirsh:
Oddly, there were two `gen_xml` methods. The one being used was shelling out to `latexml`. I am running a development environment where I haven't run `make install` and everything is in the working directory, so this doesn't work.

The other `gen_xml` method uses the library, which works with my local development environment. A nice side-effect is this is the same approach used by `make test`, which seems more resilient somehow.

Also, `convertAndWriteFile()` was broken, so I've fixed that.


brucemiller:
Yeah, hmm... I added that alternative in the middle of some debugging session; I don't think it was really intended to be permanent. Thanks!

PR #907, opened 2017-12-17
Add bindings for cvpr.sty

bfirsh:
A simple one to get me started. :)

These are the bindings for the CVPR conference submission style, [details of which are here](http://cvpr2017.thecvf.com/submission/main_conference/author_guidelines). It's not on CTAN, but many Arxiv papers use it, such as [this paper](https://arxiv.org/abs/1512.03385v1) ([Latexml rendering, with error](https://www.arxiv-vanity.com/papers/1512.03385/)).

For lack of a better place to put the test, I've added a set of tests called `lib`, which are intended to map 1-1 with files in `lib/` for simple binding tests.


bfirsh:
If you're interested, here is the error when trying to use `InputDefinitions` on cvpr.sty:

```
latexmlc (LaTeXML version 0.8.2; revision 79039244)
processing started Sun Dec 17 10:41:08 2017

(Digesting TeX egpaper_final...
(Processing content /app/local/cvpr2017/latex/egpaper_final.tex...
(Loading /app/lib/LaTeXML/Package/LaTeX.pool.ltxml... 0.18 sec)
(Loading /app/lib/LaTeXML/Package/article.cls.ltxml... 0.02 sec)
(Loading /app/lib/LaTeXML/Package/cvpr.sty.ltxml...
(Processing definitions /app/local/cvpr2017/latex/cvpr.sty...
(Processing definitions /app/local/cvpr2017/latex/eso-pic.sty...
(Processing definitions /app/local/cvpr2017/latex/cvpr_eso.sty... 0.01 sec)
(Loading /app/lib/LaTeXML/Package/keyval.sty.ltxml...
(Processing definitions /usr/share/texlive/texmf-dist/tex/latex/graphics/keyval.sty... 0.02 sec) 0.07 sec)File /app/lib/LaTeXML/Package/cvpr.sty.ltxml had an error:

Fatal:unexpected:<endgroup> Attempt to pop last locked stack frame
    at /app/local/cvpr2017/latex/eso-pic.sty; line 167 col 26
    In Core::Stomach[@0x55f7f6e089c8] at /app/local/cvpr2017/latex/eso-pic.sty; line 167 col 26
     <= Core::Definition::Primitive[End] <= Core::Stomach[@0x55f7f6e089c8] <= Core::Gullet[@0x55f7f6e08920] <= Core::Definition::Constructor[\usepac... <= Core::Stomach[@0x55f7f6e089c8] <= Core::Gullet[@0x55f7f6e08920] <= Core::Definition::Constructor[\usepac... <= Core::Stomach[@0x55f7f6e089c8]
1 fatal error
```

brucemiller:
That's a quite nice patch! Particularly for a, um, not-so-much Fan of Perl! :>

But when I looked at your error log, it looked like the underlying problem was some avoidable glitch in processing eso-pic.sty (seemed to otherwise deal with cvpr OK).  So I fetched the article to play with.  And now that I have a working `--includestyles` (thank you very much), I discovered that there were just a couple of bugs/oversights in LaTeXML that kept it from processing error free.  And in fact, fixing a couple more bugs allow it to produce an almost perfect output!

Alas, a couple of those issues have some subtleties (eg. the notion of Preamble isn't really handled correctly), so I'm still doing some testing before I commit.  But wanted to give you some feedback.

bfirsh:
Nice! That sounds great.

One thing I quite liked about just picking out the relevant stuff from the style is it also stopped it doing all of the unnecessary stuff that conference styles do (redefine maketitle, redefine abstract, change heading styles in this case) which we don't want to come through in Engrafo.

I suppose for LaTeXML, you want to respect these styles don't you? For Engrafo, I suppose we could add some stylesheets/postprocessing to remove them again. Differently sized/emboldened headings are a particularly annoying thing that some papers do (and cvpr.sty, actually...).

brucemiller:
I've just pushed a bunch of commits that allow LaTeXML to process eso-pic, cvpr and other styles without errors.  It also fixes a few formatting glitches that the article turned up.

Thanks for the report and offered patch!

brucemiller:
Oh, I wanted to also comment on your last remark, @bfirsh.  Yeah, interesting design topic. In the early stages of the program, I was just happy to get the structure right, and felt like all the styling was irrelevant and could be added back with CSS.  But over time, folks (including me) kept wanting to preserve as much styling as possible, as well.

The tricky part is trying to record & implement all the styling that the original LaTeX would have had, but do it in a way that makes it easy to remove or override in the end product.  Unfortunately, currently most of the ad-hoc latex styling gets recorded in css style attributes which are rather hard to override.  Any alternative scheme I've thought of would need some layer of indirection, possibly creating a per-document css stylesheet.  This is where I do *not* consider CSS to be my friend :>

bfirsh:
Hmm, yes, that is tricky. Perhaps it could be a runtime option to leave out custom styling.

The approach we've taken for Engrafo is a post processor to just get rid of all the inline styles, and that's working fine so far.

PR #921, opened 2018-01-07
auto-detect UTF-8 encoding in Mouth::file, when no encoding is given

### LINKED ISSUE #919: Wrong encoding in HTML output of references
I use the following test documents,
test.tex:
```
\documentclass[a4paper,12pt,twoside,openright]{book}
\usepackage[english]{babel}		
\usepackage[utf8]{inputenc} 
\usepackage{natbib}
\begin{document}
\tableofcontents
\chapter{Präfatßio}
Löräm ipßum dolor ßit amet, conßäctetur adipisiki älit, sed äüsmod tempor incidunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exärcitation ullamco laboris nisi ut aliquid ex ea commodi conßequat. Quis aüte iüre reprähenderit in voluptate velit eße cillüm dolore äu fugiat nulla pariatur. Exceptäür sint öbcäcat cupiditat nön proident, ßünt in kulpa kwi offitßia deserunt mollit anim id äßt laborum \cite{Wuersst1887}.
\nocite{*}
\bibliographystyle{apalike}
\bibliography{test}
\end{document}
```
and test.bib:
```
@BOOK{Wuersst1887,
  title = {{The german umlauts - A pain in the ähß}},
  publisher = {Springer, Berlin},
  year = {1887},
  author = {Hantß Würßt},
  note = {Deutsche Ausgabe: Würßt, Die deutschen Umlaute ÄÖÜß, Springer, Berlin 1871},
  owner = {asmaier},
}
```
On Mac OS X with basictex and latexml 0.8.2 I do the following
```
$ latexml test.tex > test.xml
$ iconv -f iso-8859-1 -t utf-8 test.xml > testutf8.xml   # Necessary, because of problem #917
$ latexmlpost  --dest=test.html testutf8.xml
```
When I open test.html in Chrome I see the following:
latexmlumlauts
Note the encoding issues in the reference to the book and also in the author and title of the bibliography entry. 
The bibliography file test.bib is stored as utf-8:
```
$ file test.bib 
test.bib: BibTeX text file, UTF-8 Unicode text
```
The corresponding PDF output of `latex` and `bibtex` doesn't show these issues. So I guess, this could be an encoding issue in the bibtex processing of `latexmlpost`.

dginev:
Fixes #919 (experimental)

I tried my hand at adding a dependency and try to proactively guess the encoding of a given file, looking for  UTF-8 lines.

I expect the biggest drawback here would be (large?) performance slowdown, as the naive first stab would test every line as a file is being read-in, for every file latexml ingests.

However the examples at #919 all work correctly, which makes it a decent proof-of-concept. Maybe gets a discussion started?

### DIFF
--- a/lib/LaTeXML/Core/Mouth/file.pm
+++ b/lib/LaTeXML/Core/Mouth/file.pm
@@ -16,12 +16,14 @@ use LaTeXML::Global;
 use LaTeXML::Common::Error;
 use LaTeXML::Util::Pathname;
 use Encode;
+use Encode::Detect::Detector;
+
 use base qw(LaTeXML::Core::Mouth);
 
 sub new {
   my ($class, $pathname, %options) = @_;
   my ($dir,   $name,     $ext)     = pathname_split($pathname);
-  my $self = bless { source => $pathname, shortsource => "$name.$ext" }, $class;
+  my $self = bless { source => $pathname, shortsource => "$name.$ext", encoding => $STATE->lookupValue('PERL_INPUT_ENCODING') }, $class;
   $$self{fordefinitions} = 1 if $options{fordefinitions};
   $$self{notes}          = 1 if $options{notes};
   $self->openFile($pathname);
@@ -63,11 +65,17 @@ sub getNextLine {
       close($fh); $$self{IN} = undef;
       return; }
     else {
+      if (!$$self{encoding}) {
+        my $detected = detect($line);
+        if ($detected && $detected eq 'UTF-8') {
+          $$self{encoding} = 'UTF-8';
+        }
+      }
       push(@{ $$self{buffer} }, LaTeXML::Core::Mouth::splitLines($line)); } }
 
   my $line = shift(@{ $$self{buffer} });
   if (defined $line) {
-    if (my $encoding = $STATE->lookupValue('PERL_INPUT_ENCODING')) {
+    if (my $encoding = $$self{encoding}) {
      # Note that if chars in the input cannot be decoded, they are replaced by \x{FFFD}
      # I _think_ that for TeX's behaviour we actually should turn such un-decodeable chars in to space(?).
       $line = decode($encoding, $line, Encode::FB_DEFAULT);
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -72,6 +72,7 @@ WriteMakefile(NAME => 'LaTeXML',
     'Time::HiRes'       => 0,
     'URI'               => 0,
     'version'           => 0,
+    'Encode::Detect::Detector' => 0,
     # If we have an "old" version of XML::LibXML,
     # we also need XPathContext.
     # But we can't determine that additional dependence


lib/LaTeXML/Core/Mouth/file.pm
```diff
@@ -63,11 +65,17 @@ sub getNextLine {
...
+        if ($detected && $detected eq 'UTF-8') {
+          $$self{encoding} = 'UTF-8';
+        }
+      }
       push(@{ $$self{buffer} }, LaTeXML::Core::Mouth::splitLines($line)); } }
 
   my $line = shift(@{ $$self{buffer} });
   if (defined $line) {
-    if (my $encoding = $STATE->lookupValue('PERL_INPUT_ENCODING')) {
+    if (my $encoding = $$self{encoding}) {
```
dginev: I also snuck in an optimization here - we were calling `lookupValue` on every line, but we can really just call it once when the Mouth object gets created.

dginev:
At least as far as `make test` is concerned, I am not seeing any slowdown, in fact I am seeing identical performance. Then again I haven't double-checked that the tests aren't using an `inputencoding` option that is explicitly set, in which case the new code paths won't be triggered.

lib/LaTeXML/Core/Mouth/file.pm
```diff
@@ -63,11 +65,17 @@ sub getNextLine {
       close($fh); $$self{IN} = undef;
       return; }
     else {
+      if (!$$self{encoding}) {
+        my $detected = detect($line);
+        if ($detected && $detected eq 'UTF-8') {
+          $$self{encoding} = 'UTF-8';
```
dginev: I also ensured that this check will be performed only until the first UTF-8 line is spotted for a given file.

Sadly if a file is plain ascii (as all .sty files are for example), this `detect()` call will be invoked on every single line. So it better be cheap.

dginev:
Interestingly, the magic happens in C: 
https://metacpan.org/source/JGMYERS/Encode-Detect-1.01/Detector.xs

and this is the code probing for UTF-8:
https://metacpan.org/source/JGMYERS/Encode-Detect-1.01/src/nsUTF8Prober.cpp

So in theory, we could adapt it without the dependency... especially if we were using XS already ourselves...

brucemiller:
Hmm, I don't follow the double encoding reasoning here; the fact that it "works" for a unicode TeX file but not for a bibliography suggests that there's something broken while munging the bib source into pseudo-TeX.

dginev:
It is probably worth to spend more time for me and read into the exact differences. 

But you say it works with regular latex and I wonder - does it? A quick and dirty test is:

```
latexml 'literal:проба' --dest=test.xml
```

which creates double-encoded unicode in the final XML. I think most examples where it works are simply correct latex and include a unicode-friendly package definition, such as `\usepackage[utf8]{inputenc}`, and hence work.

brucemiller:
Ugh, I think you're right; I'd overlooked the inputenc. And our bibtex fails because the real bibtex, again, doesn't actually do any TeX processing and thus doesn't much care about encoding.

dginev:
Yeah, this line of upgrades for unicode-friendly bibtex definitely goes to the "above & beyond" category. Not really something you can casually do in classic latex, tends to blow up in your face rather fast.

dginev:
But it is 2018, and being unicode-friendly isn't exactly a radical request either...

dginev:
A less ambitious fix without all the experimental magic here could just make sure to pass the value of `$STATE->lookupValue('PERL_INPUT_ENCODING')` at the end of the core processing to the bibtex processor in MakeBibliography.

All it really takes to claim we're as good as latex, since inputenc will get propagated onto the bibliography. Direct `latexml foo.bib` calls will still have the issue of course, you'd need `--inputencoding=utf8` to make them work.

dginev:
Yes, I feel convinced that's a good PR make. Will close here after testing that.

brucemiller:
Yeah, that sounds better and avoids (for the time being) a non-core dependency. I'm going to have to dig deeper into the hypothetical xs, though. Something's broke there.

A "Reasonable" approach might be to assume utf8 (of which ASCII is a proper subset), unless a command line switch is used, or `inputenc` is loaded.  The catch, particularly for the xs full-file slurp, is to make sure we haven't decoded wrongly before we've seen a load of inputenc.

brucemiller:
Still a catch: with latexmlpost, you may not know what the original encoding of the tex document was? (or is it hidden away somewhere?)

dginev:
Ah, `latexmlpost`, of course. You're completely right, I was in latexmlc dreamland - for latexmlpost we would need to add an `inputencoding` processing instruction to the intermediate XML ... Or some such. A bit of a hassle but it seems to be the "classic" solution.

brucemiller:
Yeah,  but the eventual latexml(c) has to be able to act like either or both of latexml and latexmlpost; that is it still will need to start from an xml file as well as from a tex file.

Recording the original encoding on a PI seems a solution, hacky tho' it may be.

brucemiller:
Actually, particularly in light of #931, I think this approach is wrong (or rather, compounds a mistake?).  As I understand it, TeX & friends are at *best* 8bit safe; the inputenc work by defining the upper half as macros that expand into the appropriate glyphs. Even utf8 inputenc only does some kind of simulation of multibyte chars? (not sure whether XeTeX just makes that more efficient, or is fully unicode).
So, to stick to that approach, really Mouth, and probably Token's, should deal exclusively with bytes.

I suspect that's what we do, but it needs to be more explicit rather than accidental; If `PERL_INPUT_ENCODING` comes into play, it probably should only be for XeTeX simulation(?).

As far as bibtex is concerned; it's also merely 8bit safe; it passes through anything it doesn't understand. In the case of processing bibliiography during postprocessing, probably we can extract the input encoding from the base document's PI's.  For standalone, it will have to be declared on the commandline (or sniffed out, if you insist ;> )

Issue #200, opened 2011-03-17
LaTeXML support for CVs


dginev:
[Originally Ticket 1524]

I have been asked whether LaTeXML could be used to create CVs for websites. I think that is a good feature that would gear more normal users to benefit from LaTeXML. Of course, this is definitely not a first priority.

From my understanding, some of the complex parts are creating bindings for CV packages such as [moderncv](http://www.ctan.org/tex-archive/macros/latex/contrib/moderncv/) and to support bibliography magic (multi-section bibliographies) as for example does [bibtopic](http://www.ctan.org/tex-archive/macros/latex/contrib/bibtopic/).

I am filing the ticket here for a lack of a better place and giving it low priority.

dginev:
Just to show this is possible (and I do apologize I am linking to my own bio), this is a moderncv-based resume, created via LaTeXML: http://prodg.org/cv .

I would like to make arrangements for the responsible code to swim back into the master repo, but it needs some polish first. At the moment, it contains:
 * A moderncv.sty.ltxml binding
 * A LaTeXML-CV.xsl post-processing stylesheet (because I used bootstrap)
 * A ltx-cv.css stylesheet for making it look pretty

The issue is that I made them too specific for the context of my personal website, so they need some disentangling back into mint status. For instance, the only reason I needed a custom stylesheet (that just extended on top of the HTML5 one) is that I was doing some Bootstrap-based styling.

brucemiller:
Interesting, since I just updated my cv using moderncv...

Most of moderncv looks "easy", if tedious; and the styling takes some creativity; so the 1st & 3rd items shouldn't be hard to integrate.

Trickier is how to deal with the 2nd item: to easily integrate extra stylesheets, or to avoid the need to do so.  The catch with bootstrap, is that, perhaps inevitably, you have to completely buy-in at the outset and use official bootstrap classes on many critical elements.

You *could*, in principle, arrange for the appropriate elements to get appropriate classes added during conversion. It's not quite in the spirit of LaTeXML to do such complete restyling at that stage, and probably isn't trivial to insert such behaviour at all the right places.

Perhaps more in the right spirit would be a way of customizing the way classes are added to element upon conversion?

dginev:
I agree that moderncv is easy but tedious, it is 100% styling work.

I think we can do with a simpler solution that is already compliant with the way LaTeXML deals with CSS. Namely:
[...]

PR #924, opened 2018-01-09
moderncv.sty binding

dginev:
Finally making progress on my promise in #200 and starting to generalize my personal CV binding sketch towards supporting all of moderncv.sty.

Would like some feedback on what may be the sane approach to deal with the special CV styling however - a bit more concretely now that I have a first test CV and attached PDF and HTML files to compare. I "chickenized" my own CV as a first test, it is not an exhaustive one by any means and I should add at least one of the official templates to ensure the coverage makes some sense.

There's also the open question of allowing to use the various `\moderncvtheme` definitions, maybe a native read-in of the TeX could get more of that auto-ingested?

In any case, this PR contains the working setup from my private website (responsible for [my cv](https://prodg.org/cv) ), and should be a good starting point to adding support for CVs. Feedback more than welcome!


dginev:
I'm also wondering if the `RequireResource` command can be / could be enhanced to be used with remote resources, since the only way to get the well-looking HTML output right now is to convert with this extensive invocation:

```
latexmlc t/moderncv/cs_cv.tex --dest=t/moderncv/cs_cv.html \
   --stylesheet=LaTeXML-CV.xsl \
   --css="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css" \
   --css="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap-theme.min.css" \
   --javascript="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/js/bootstrap.min.js"
```

brucemiller:
I Like it! Well, the results, anyway. I'm not so convinced about the need for the XSLT; that it's even necessarily CV specific, or maybe about what it exactly does. It seems that it could be handled more generically in the standard structure stylesheet, possibly with appropriate css (maybe hiding some of the output by default?).

dginev:
I double-checked with a friend who is a CTO of a job-application company, in which "CV imports" are a fundamental acquisition channel, and indeed they are working entirely with PDF and Word files. To the best of her knowledge there is no HTML standard for CVs in wide use today, and no one is really using HTML as a CV import channel for them.

Or in other words, we are "free to innovate" with the HTML target for this PR.

I was actually looking for an easy out, so that I just adopt some "CV standard" for HTML, but it seems we'd have to invent something like one. Will be back to this issue soon.

dginev:
This ties into your XSLT comment - the stylesheet applies the Bootstrap `class` markup, as an add-on that is entirely HTML-specific and does not belong in the binding files.

The big problem remains that Bootstrap is not a latexml CSS requirement, and it really shouldn't be. I should up my CSS game and find a solution that is native, instead of requiring a new framework. Then the XSLT can go away and only use the CSS file.

dginev:
I've heard (only) great things about flexbox, modern Bootstrap is based on it, and it looks like flexbox grids are simple to code in CSS ([random tutorial here](https://css-tricks.com/dont-overthink-flexbox-grids/))

So I will take a stab at reworking what I have to pure flexbox

brucemiller:
Yeah, flex has crept into the stock LaTeXML css, too :>

And, yeah, we don't want to build in assumptions about css or js frameworks, they come & go so often, and folks have their own preferences.  OTOH, waving hands, it would be nice if we could more easily enable the use of random frameworks.

It seems the common way of applying these frameworks is to add specific classes from the framework onto your html markup, which then gets whatever properties & behaviour of the kit. So (waving hands more franticly) that it wouldn't be hard to develop a simple tool to that used either css selectors or xpath (either on the latexml xml or html) and added the appropriate css class to the matching items.
(and in fact sorta seems like something CSS should have done in the first place; ie. a certain selector inherits all the properties of some other class).

dginev:
Hm, sounds handy, but maybe that is premature abstraction at this point? Since we haven't had the need for external frameworks yet, and it is very hard to predict the future...

Should I postpone my flexbox rewrite here, or you think that sounds good for now?

brucemiller:
On 04/05/2018 03:13 PM, Deyan Ginev wrote:
> Should I postpone my flexbox rewrite here, or you think that sounds good 
> for now?

Probably a good thing if it can easily be covered by flexbox

brucemiller:
Since I just merged TagsNTitles, you'll probably have to remake your test case (don't think anything else would be affected).

dginev:
Making it work was easier than I thought, but making the CV look good with flexbox has a bit of a learning curve... makes you appreciate bootstrap's magic when dealing with an entire doc :+1:

dginev:
Alright, ready for review here again @brucemiller . I have:
 * Redone the HTML styling using flexbox
 * Simplified the class names, styling scheme, and even the internal binding setup
 * Rebased to the latest master and updated my unnumbered section redefinition (tricky change!)
 * Integrated the XSLT rules with the main stylesheets, as a separate `cv` component
 * Updated all tests

Feel free to take a look at the latest `cs_cv.html` file in the PR, though you'd need to carry over the `.css` files for it to look good.

I'm still short of covering all moderncv features, but basic CVs can look reasonable!

dginev:
Oh, and I forgot - improved user experience ergonomics. You don't need _any_ special options now, a simple invocation will get you an HTML CV:
```
latexmlc cs_cv.tex --dest=cs_cv.html
```

Done.

dginev:
Tiny fix: had piggybacked on the wrong XSLT rule which turned the mobile number into an email link. Much better now.

dginev:
Another fun bit - you can actually generate the CV as an epub file that Calibre understands and shows nicely, which makes it distributable akin to a PDF, if such a future ever becomes mainstream. Cool side effect!

brucemiller:
The tests, css, (other stuff?) need to go into the MANIFEST, too.  Otherwise, let me know when you're done tweaking...it's looking good.

dginev:
I am done tweaking here, but the MANIFEST is confusing me. Can I just run a command and have it auto-update instead of adding each file by hand? I think there was a `make manifest`, but I haven't used it for the longest time.

dginev:
I tried and remembered that `make manifest` regenerates the entire file and ruins the manually curated structure, so I went and added the remaining files by hand.

Should be ready to merge, I'm done working on this PR :+1:

brucemiller:
Cool! Thanks!!

Issue #926, opened 2018-01-12
Math parsing problem of `\boldsymbol{0}` ?


asmaier:
When converting the following formula to MathML (Mac OS X, v0.8.2 of LaTeXML), it seems the number 0 somehow get's lost:

```
$ latexmlmath "\displaystyle\boldsymbol{0}" --preload=amsmath
<?xml version="1.0" encoding="UTF-8"?>
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="\displaystyle\boldsymbol{0}" display="block">
  <mn/>
</math>

```
One can see in the MathML output, that the number 0 is missing. It doesn't happen with other numbers like `\boldsymbol{1}`:
```
$ latexmlmath "\displaystyle\boldsymbol{1}" --preload=amsmath
<?xml version="1.0" encoding="UTF-8"?>
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="\displaystyle\boldsymbol{1}" display="block">
  <mn>𝟏</mn>
</math>
```
It also doesn't happen, if I change the formula and write the 0 without `\boldsymbol`:
```
$ latexmlmath "\displaystyle 0" --preload=amsmath
<?xml version="1.0" encoding="UTF-8"?>
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="\displaystyle 0" display="block">
  <mn>0</mn>
</math>
```
What is going on?

dginev:
This should be an easy fix for a classic Perl error where you have an `if` conditional testing if something is defined, but the literal string `"0"` ends up false in Perl. Sometimes they slip in, it's a great catch

brucemiller:
Yeah; I have no doubt that this happened, and probably exactly how @dginev says. But it seems that it's already been fixed.  If that turns out not to be true, please reopen.  Thanks for the report.

asmaier:
When I try the same example with the LaTeXML version from HEAD, I do see the following:

```
$ latexmlmath "\displaystyle\boldsymbol{0}" --preload=amsmath
<?xml version="1.0" encoding="UTF-8"?>
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="\displaystyle\boldsymbol{0}" display="block">
  <mn>ð</mn>
</math>
```

and also

```
$ latexmlmath "\displaystyle\boldsymbol{1}" --preload=amsmath
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #938, opened 2018-02-03
binmode not needed if we always output bytes

dginev:
Simple enough (now that we have a proper understanding), all result prints expect bytes now, i.e. the result to be printed has already been utf-8 encoded.

### DIFF
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -160,13 +160,9 @@ if ($result) {
     if (!open($output_handle, ">", $opts->get('destination'))) {
       print STDERR "Fatal:I/O:forbidden Couldn't open output file " . $opts->get('destination') . ": $!";
       exit 1; }
-    if (!$is_archive) { # If we're not outputing binary data, encode to UTF-8
-      binmode($output_handle, ":encoding(UTF-8)"); }
     print $output_handle $result;
     close $output_handle;
   } else {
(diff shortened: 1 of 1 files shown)


dginev:
Fixes #926 for latexmlc

dginev:
well, almost. A certain path through latexmlc still has wide characters as a serialization never gets utf-8 encoded. Which is why I was wondering if we shouldn't put the encode call already in `serialize_aux` in Document, since I think this is the only serializer that does not explicitly do an encoding to bytes.

The current `master` branch shifts that responsibility to the executables, notably `latexml`, which means I need to be extra careful in `latexmlc` to not skip a case. Luckily some of the tests for the daemon runs test the logs!

dginev:
Added a commit that should handle things with the current approach.

brucemiller:
The thing that got us in trouble was the inconsistency: one rare case being already encoded when most other stuff was characters.  I'd prefer to keep things as consistent and localized as possible.  So, I'm less inclined to change Core::Document's approach.  But you're right that latexmlc has to deal with a variety of objects and we need a way to know (or use polymorphism) whether each object is encoded or not.

dginev:
Wouldn't a convention that any `toString` method returns an encoded string be more consistent?

dginev:
The other type of object that latexmlc can return are archives, and that is just binary data, so it follows the `result is just bytes` convention

PR #947, opened 2018-02-06
Expose math lexemes in final output

dginev:
This is a llamapun-motivated pull request. I have been creating some token models over arXiv, and would like to take them to the next level by incorporating an ascii, linguistics-friendly, representation of all formulas.

To that extent, I would like to dispense with any remnants of TeX, and instead use the closest possible form to the one used by latexml's grammar input.

This was simultaneously easier than I expected - in terms of introduced the feature in MathParser, as `$lexemes` already existed in a very near ideal form. But also harder than expected from the usability side as I just couldn't find a way around introducing a new option called `--mathlex`, as well as a dedicated `llamapun.sty.ltxml` binding to avoid introducing a new option in the core latexml processing as well.

If it wasn't for the MathParser changes, one could imagine the rest of the toggling UX as a separate dedicated plugin. But making new options from a plugin isn't that direct either... as it is inevitable to have to patch Config.pm.

But back to the fun part. With the current PR one can run for instance:

```
latexmlc
   --whatsin=math --whatsout=math --pmml
   --preload=llamapun.sty  --mathlex
   'literal:P(E) = {n \choose k} p^k (1-p)^{ n-k}'
```

and get back:
```xml
<Math xmlns="http://dlmf.nist.gov/LaTeXML" xmlns:m="http://www.w3.org/1998/Math/MathML" mode="inline" xml:id="p1.m1" tex="P(E)={n\choose k}p^{k}(1-p)^{n-k}" text="P * E = (n binomial k) * p ^ k * (1 - p) ^ (n - k)" lexemes="UNKNOWN:P OPEN:( UNKNOWN:E CLOSE:) RELOP:equals OPEN:( FRACOP:binomial ARG:start UNKNOWN:n ARG:end ARG:start UNKNOWN:k ARG:end  CLOSE:)  UNKNOWN:p POSTSUPERSCRIPT:start UNKNOWN:k POSTSUPERSCRIPT:end OPEN:( NUMBER:1 ADDOP:minus UNKNOWN:p CLOSE:) POSTSUPERSCRIPT:start UNKNOWN:n ADDOP:minus UNKNOWN:k POSTSUPERSCRIPT:end" fragid="p1.m1">
  <m:math alttext="P(E)={n\choose k}p^{k}(1-p)^{n-k}" display="inline">
    <m:semantics>
      <m:mrow>
        <m:mrow>
          <m:mi>P</m:mi>
          <m:mo>⁢</m:mo>
          <m:mrow>
            <m:mo stretchy="false">(</m:mo>
            <m:mi>E</m:mi>
            <m:mo stretchy="false">)</m:mo>
          </m:mrow>
        </m:mrow>
        <m:mo>=</m:mo>
        <m:mrow>
          <m:mrow>
            <m:mo>(</m:mo>
            <m:mfrac linethickness="0pt">
              <m:mi>n</m:mi>
              <m:mi>k</m:mi>
            </m:mfrac>
            <m:mo>)</m:mo>
          </m:mrow>
          <m:mo>⁢</m:mo>
          <m:msup>
            <m:mi>p</m:mi>
            <m:mi>k</m:mi>
          </m:msup>
          <m:mo>⁢</m:mo>
          <m:msup>
            <m:mrow>
              <m:mo stretchy="false">(</m:mo>
              <m:mrow>
                <m:mn>1</m:mn>
                <m:mo>-</m:mo>
                <m:mi>p</m:mi>
              </m:mrow>
              <m:mo stretchy="false">)</m:mo>
            </m:mrow>
            <m:mrow>
              <m:mi>n</m:mi>
              <m:mo>-</m:mo>
              <m:mi>k</m:mi>
            </m:mrow>
          </m:msup>
        </m:mrow>
      </m:mrow>
      <m:annotation encoding="application/x-llamapun">UNKNOWN:P OPEN:( UNKNOWN:E CLOSE:) RELOP:equals OPEN:( FRACOP:binomial ARG:start UNKNOWN:n ARG:end ARG:start UNKNOWN:k ARG:end  CLOSE:)  UNKNOWN:p POSTSUPERSCRIPT:start UNKNOWN:k POSTSUPERSCRIPT:end OPEN:( NUMBER:1 ADDOP:minus UNKNOWN:p CLOSE:) POSTSUPERSCRIPT:start UNKNOWN:n ADDOP:minus UNKNOWN:k POSTSUPERSCRIPT:end</m:annotation>
    </m:semantics>
  </m:math>
</Math>
```

I think the lexeme string itself is also interesting to look at for complex examples (already here with `\choose` actually), so I will leave more comments.


dginev:
Also I used `llamapun` for the new names, since I didn't know what to name the binding and the annotation mime-type, so I used the application that motivated things.

As to the specific lexeme string, here it is broken down next to its tex form:
```tex
$P(E) = {n \choose k} p^k (1-p)^{ n-k}$
```

lexemes:
```
OPEN:( ATOM:nk CLOSE:) UNKNOWN:n ADDOP:minus UNKNOWN:k 
UNKNOWN:P OPEN:( UNKNOWN:E CLOSE:) RELOP:equals
ATOM:(nk) UNKNOWN:p POSTSUPERSCRIPT:k 
OPEN:( NUMBER:1 ADDOP:minus UNKNOWN:p CLOSE:) POSTSUPERSCRIPT:-nk
```

It is clearly buggy! Guess not as reliably sequential as I hoped in the lexemes generation, will try to improve now.

dginev:
Ok, pushed a commit that gets the order right, but i couldn't find an easy way to incorporate the children lexemes, so punting on those until i get a great idea (from me or from feedback :>)

Now my example gets an annotation of:
```
UNKNOWN:P OPEN:( UNKNOWN:E CLOSE:) RELOP:equals
ATOM:(nk) UNKNOWN:p POSTSUPERSCRIPT:k 
OPEN:( NUMBER:1 ADDOP:minus UNKNOWN:p CLOSE:) POSTSUPERSCRIPT:-nk
```

I am a bit tempted to downcase the roles since the caps make it hard to read, but that's quite minor. 

My next task is to get these subtree atoms/postsuperscripts etc. into a properly expanded multi-token form

dginev:
progress! indents are from me, commit just added.

```
UNKNOWN:P OPEN:( UNKNOWN:E CLOSE:) RELOP:equals 

ARG:start 

ARG:start
   FRACOP:binomial ARG:start UNKNOWN:n ARG:end ARG:start UNKNOWN:k ARG:end
ARG:end

ARG:start
   OPEN:( 
       ARG:start 
                FRACOP:binomial
                ARG:start UNKNOWN:n ARG:end 
               ARG:start UNKNOWN:k ARG:end 
      ARG:end 
   CLOSE:)
ARG:end 

ARG:end 

UNKNOWN:p POSTSUPERSCRIPT:k OPEN:( NUMBER:1 ADDOP:minus UNKNOWN:p CLOSE:) 

POSTSUPERSCRIPT:n-k 

```

dginev:
The last remaining item from this example may be to figure out what happens when things like `postsuperscript` have complex structrures in them. I am fond of the arg:start and arg:end delimiters to preserve argument structure, which will also be quite handy for grammars.

dginev:
Ok, I think I have this example nailed down. Deleted some of my intermediate comments to reduce clutter.

Again, for the tex formula:
```tex
$ P(E) = {n \choose k} p^k (1-p)^{ n-k} $
```

The resulting lexemes are now:
```
UNKNOWN:P OPEN:( UNKNOWN:E CLOSE:) RELOP:equals

OPEN:( FRACOP:binomial ARG:start UNKNOWN:n ARG:end ARG:start UNKNOWN:k ARG:end  CLOSE:)

UNKNOWN:p POSTSUPERSCRIPT:start UNKNOWN:k POSTSUPERSCRIPT:end 

OPEN:( NUMBER:1 ADDOP:minus UNKNOWN:p CLOSE:) 

POSTSUPERSCRIPT:start UNKNOWN:n ADDOP:minus UNKNOWN:k POSTSUPERSCRIPT:end
```

which feels ideal. Hence requesting feedback from @brucemiller : 
  * Do you think lowercasing the roles for this export makes sense? (minor...)
  * More major, I decided to use the second child of the XMDual, so that I preserve the presentational queues from latex. Do you think that makes sense, given that the idea is to expose the forms to linguistic tools (e.g. external grammars, token models)
 * My start/end "fence tokens" are named a bit ad-hoc, abusing a bit the `role:meaning` convention, as I am using start/end as the meaning entry. Any intuitions what would be better naming? I kind of like this approach as it feels somewhat self-explanatory.
 * Any high level feedback about merging the PR? Would you like tests added?

I may want to try some more complex formulas, e.g. all the formulas in my equations example at latexml.mathweb.org/editor, possibly turn them into tests... but will wait for feedback before I jump on that.

dginev:
The Cauchy-Schwarz Inequality also looks decently lexematized with the current state:
csineq


```
OPEN:( 
  SUMOP:sum POSTSUBSCRIPT:start UNKNOWN:k RELOP:equals NUMBER:1 POSTSUBSCRIPT:end POSTSUPERSCRIPT:start UNKNOWN:n POSTSUPERSCRIPT:end
  UNKNOWN:a POSTSUBSCRIPT:start UNKNOWN:k POSTSUBSCRIPT:end UNKNOWN:b POSTSUBSCRIPT:start UNKNOWN:k POSTSUBSCRIPT:end 
CLOSE:) POSTSUPERSCRIPT:start NUMBER:2 POSTSUPERSCRIPT:end 

RELOP:less-than-or-equals 

OPEN:( 
  SUMOP:sum POSTSUBSCRIPT:start UNKNOWN:k RELOP:equals NUMBER:1 POSTSUBSCRIPT:end POSTSUPERSCRIPT:start UNKNOWN:n POSTSUPERSCRIPT:end
  UNKNOWN:a POSTSUBSCRIPT:start UNKNOWN:k POSTSUBSCRIPT:end POSTSUPERSCRIPT:start NUMBER:2 POSTSUPERSCRIPT:end 
CLOSE:) 
OPEN:( 
  SUMOP:sum POSTSUBSCRIPT:start UNKNOWN:k RELOP:equals NUMBER:1 POSTSUBSCRIPT:end POSTSUPERSCRIPT:start UNKNOWN:n POSTSUPERSCRIPT:end
  UNKNOWN:b POSTSUBSCRIPT:start UNKNOWN:k POSTSUBSCRIPT:end POSTSUPERSCRIPT:start NUMBER:2 POSTSUPERSCRIPT:end
CLOSE:)
```

dginev:
After a coffee it seems a bit more obvious to rename things like

* `POSTSUBSCRIPT:start` / `POSTSUBSCRIPT:end`

into the more consistent with the internal notation

* `OPEN:postsubscript` / `CLOSE:postsubscript`

and similarly for `OPEN:arg` / `CLOSE:arg`.

Does that sound better?

dginev:
Got another fix in by inspecting the Ramanujan identity in the examples i mentioned. Here is the result (still using the start/end lexemes, until we discuss):

ramanujan


```
FRACOP:divide 
ARG:start NUMBER:1 ARG:end 
ARG:start
    OPEN:( 
        UNKNOWN:square-root ARG:start
            UNKNOWN:phi UNKNOWN:square-root ARG:start NUMBER:5 ARG:end 
        ARG:end 
        ADDOP:minus UNKNOWN:phi 
    CLOSE:)
    UNKNOWN:e 
    POSTSUPERSCRIPT:start 
        FRACOP:divide ARG:start NUMBER:2 ARG:end ARG:start NUMBER:5 ARG:end 
        UNKNOWN:pi 
    POSTSUPERSCRIPT:end 
ARG:end

RELOP:equals

NUMBER:1 ADDOP:plus 
FRACOP:divide 
ARG:start
    UNKNOWN:e POSTSUPERSCRIPT:start ADDOP:minus NUMBER:2 UNKNOWN:pi POSTSUPERSCRIPT:end 
ARG:end
ARG:start
    NUMBER:1 ADDOP:plus 
    FRACOP:divide 
    ARG:start
        UNKNOWN:e POSTSUPERSCRIPT:start ADDOP:minus NUMBER:4 UNKNOWN:pi POSTSUPERSCRIPT:end 
    ARG:end 
    ARG:start
        NUMBER:1 ADDOP:plus 
        FRACOP:divide 
        ARG:start 
            UNKNOWN:e POSTSUPERSCRIPT:start ADDOP:minus NUMBER:6 UNKNOWN:pi POSTSUPERSCRIPT:end 
        ARG:end 
        ARG:start 
            NUMBER:1 ADDOP:plus 
            FRACOP:divide 
            ARG:start
                UNKNOWN:e POSTSUPERSCRIPT:start ADDOP:minus NUMBER:8 UNKNOWN:pi POSTSUPERSCRIPT:end 
            ARG:end 
            ARG:start NUMBER:1 ADDOP:plus ID:ldots ARG:end
        ARG:end
    ARG:end
ARG:end
```

dginev:
More special cases handled, example:
```tex
$ x,\quad\quad \text{for} |q|<1. $
```

lexemes in output:
```
UNKNOWN:x PUNCT:,
ATOM:for VERTBAR:| UNKNOWN:q VERTBAR:| RELOP:less-than NUMBER:1 PERIOD:.
```

brucemiller:
Sounds cool! Could really drive and enhance some upcoming development... or clash with it? Perhaps some offlist discussion would be good.

dginev:
Definitely a more exploratory piece of work here, happy to discuss at length in a more casual setting than infinitely long github comments :joy: I have high hopes it is in full synergy with other upcoming upgrades, or adaptable to such a point.

brucemiller:
So, I'm inclined to go ahead and accept this PR on the grounds that it "can't hurt", even though it's clearly not finished.  It'd be good to be able to play with it!  Object soon if you don't agree :>

dginev:
I am in favor! Will rebase it quickly.

dginev:
Rebase completed, it was a single space indent...

dginev:
Yes, XMDual is tricky. This PR takes the approach that the lexeme is what "the author wrote", with the intent of _preserving_ any semantic hints, but not incorporating any guesses, be they about lexical semantics, or parse trees.

Would need to look into specific examples of duals, but preserving the token sequence coming from the gullet would be the ideal solution from my current perspective.

dginev:
I remembered my favourite dual example - piecewise functions. So here is the code for a simple [indicator function](https://en.wikipedia.org/wiki/Indicator_function) for example:

```tex
\mathbf {1} _{A}(x) :=
  \begin{cases}
   1 & \text{if } x \in A,\\
   0 & \text{if } x \notin A.
  \end{cases}
```

image

The XML is nothing to sniff at, so attaching it as a gist [here](https://gist.github.com/dginev/5331372cf5853b9bb9d05bbb1627203a) or the comment will get huge.

When generating the lexemes with the current PR, we get:
```
NUMBER:1 
POSTSUBSCRIPT:start UNKNOWN:A POSTSUBSCRIPT:end
OPEN:( UNKNOWN:x CLOSE:) 
RELOP:assign 
OPEN:{ 
  NUMBER:1 ATOM:if UNKNOWN:x RELOP:element-of UNKNOWN:A PUNCT:,
  NUMBER:0 ATOM:if UNKNOWN:x RELOP:not-element-of UNKNOWN:A 
PERIOD:.
```

As usual, line-breaks and indentation for the lexemes is manually done by me, for easier reading - it's just a space-separated line in the output.

dginev:
While I am unsure if this representation of the dual is the best we can do, it certainly isn't terrible - all visual elements are preserved.

One question is whether the `{cases}` delimiters need to be wrapped more explicitly (with `ARG:start` and `ARG:end` as done for e.g. fraction arguments), especially given that the line-break isn't preserved as a token and you need to rely on the comma punctuation.

Another minor bit is whether I need to escape all spaces inside the text content of lexemes, so that I get `ATOM:if\ ` and preserve the space given in the `\text` macro. 

I think when I wrote the XMDual handler, I explicitly made the choice to use the second, presentation, branch in order to stay true to the authored input.

brucemiller:
Ugh! My local copy of dlmf stuff is already updated to master... does this need some sort of rebase?

brucemiller:
Curious: how do you do that? I did a `git rebase` which rebased it to *your* master, I assume. How do I rebase it to the original master (something about "upstream", I assume)?

dginev:
Steps I take from the checkout of my fork (which is `origin` and your master is my `upstream` remote).

1. First I update my fork's master:
```
git fetch upstream master
git checkout master
git rebase upstream/master
git push
```

2. Then I rebase the branch in question:
```
git checkout math-lexemes
git rebase master
git push origin math-lexemes --force
```

It tends to require force, as the rebase isn't reconcilable with the github-hosted branch when pushed.

dginev:
I tend to be painfully explicit with the rebase parameters, to be certain I know what git will do _before_ I run the command, since wrong rebases can be a pain to rewind back (bunch of resets...)

brucemiller:
Ah, I see... OK, now I've got a whole mess of lexemes embedded in an xml file. Hmmm....

dginev:
Sounds about right - I am indenting them by hand to proof them, as they're in this linearized stream form. But as it happens the space separated stream is perfect for transferring the lexemes over to linguistic applications, grammars, etc.

brucemiller:
for immediate purposes, I may just hack MathParser.pm to print the lexeme streams for each numbered eqn to a file...

dginev:
> for immediate purposes, I may just hack MathParser.pm to print the lexeme streams for each numbered eqn to a file...

I take it you're in a hurry :> It may be easier to debug that way, but you can also write a tiny Perl script using XML::LibXML that snipes the attributes from the result XML file.

brucemiller:
Yeah you're right; first I thought I could get xmllint to do it, but the output isn't quite what we need.  Wanted to write a script, but the boilerplate needed before you're actually doing xpath is annoying... until I remembered latexmlfind!  Added a few more options :>

brucemiller:
Maybe I'll want to learn how to do a PR on a PR, but I'm not there yet... :>

I could see wanting to factor this out of MathParser, perhaps Util or Common. The idea being to have an encoding of not just before and after parsing, but after some "analysis" (whether by external or internal tools).

dginev:
Well I would really like to wrestle on the topic - I see any "analysis" (scare quotes or not) to be external to the core latexml suite, and either a follow-up step (as does the llamapun suite) or as a plugin (as does my Marpa grammar plugin).

Whether the lexeme generation code lives in MathParser or a different module I don't have strong opinions on per se, but it seemed consistent with the existing machinery that provides the `text` attribute - which is very similar conceptually, just for a different type of reuse.

dginev:
Also, to make a "counter-offer PR", you can branch off from this branch:
```
git checkout math-lexemes
git checkout new-lexemes-proposal
# ... add some commits ... 
git push --set-upstream origin new-lexemes-proposal
````

And you can open a new PR from the github interface when visiting the site :+1: 
The one tricky bit is that you may need to specify the remote, if you have both my fork and your master in the same repo, don't remember that offhand.

**Comment by dginev** [comment]
edited my last comment, had a crucial typo

**Comment by brucemiller** [comment]
Well, I did say "internal or external".  But maybe our different attitudes are due to my seeing "analysis" as fundamentally a middle step, between core & post processing, rather than after LaTeXML is done, which makes XMath the medium, rather than some abused form of MathML or OpenMath.

Given how unmarketable XMath is likely to be, presumably this is a controversial POV.
I don't know how long-winded the rationalization needs to be, but I guess there are two main points:
* LaTeXML/XMath schema is the bottleneck representation through which a variety of forms of TeX pass from Core into a variety of target formats.  Particularly, a rich & complete intermediate representation allows for much enriching of the user-interfaces of each target format.  So, that intermediate seems the efficient place for the effort, rather than on the tails of all those other formats.
* As well suited as MathML or OpenMath are for representing the finished product (for varying purposes), they seem to me extremely ill suited to representing the intermediate/semi-determined form of the math.

**Comment by brucemiller** [comment]
'Course maybe I missed your point, and/or it may be I have a bit of unintentional NIH.

**Comment by dginev** [comment]
Yeah, I'm coming from a perspective where we tried to adopt XMath for developing (equation) analysis services with ~10 grad students back in the day, and it turned out:
 1. the learning curve was significant,
 2. reaching XMath already has non-trivial choices made if math parsing is enabled (think invisible times) 
 3. if math parsing is disabled, we really would like to see the lexemes, rather than a tree, to build our own tree. Writing a grammar, feeding the input in ML models, etc all work easier on a token stream.
 4. if the math is parsed, e.g. for math search, it's already post-analysis and we'd much rather work on canonical mathml, which is standard and there is hope tools can be reused beyond latexml.

To address XMath directly, I think it is vital and convenient within LaTeXML, and any internal modifications should leverage it extensively. But any external applications tend to build on standard components that work on plaintext or HTML (most frequently), and the XMath ends up just something to "unwrap" and "rewrap" on input/output.

This lexeme PR was always intended for _external_ use (as mentioned, my focus was llamapun, and obtaining a token model over arXiv), hence my desire to add a lexeme output serialization.

**Comment by dginev** [comment]
> seeing "analysis" as fundamentally a middle step, between core & post processing, rather than after LaTeXML is done

Indeed, and as I mentioned, I've experimented with some plugins that fit this model, in particular for inserting a math grammar. But I would definitely claim this is a specific type of use case, that is closer to a "latexml developer" flow, where the focus is on improving the latexml results, as contrasted with say "corpus linguistic analysis".

I consider linguistic analysis over a latexml-generated HTML corpus of documents an equally viable use case, and in pragmatic terms a flow that makes using standard representation and NLP tools more viable.

**Comment by brucemiller** [comment]
Some apples 'n oranges here, I think.

If XMath is non-standard and impractical, did you have any better success using a standard representation like MathML or OpenMath for the analysis?  The learning curve's no shallower, but at least it's reusable. But by the time the math is in MathML or OpenMath format, such crimes as the implicit times are infinitely worse! (BTW: I'd call the infinite times issue not an inherent weakness of XMath, but a symptom of the fact that neither the grammar, nor the parse engine, nor the analysis are there yet).

A token stream is certainly simpler, but not standard. It's certainly more appropriate for parsing. though --- in fact, that's exactly what LaTeXML does. But the key is that it then applies that result back into the XMath to proceed.  I'd think that model applies equally well to something external like llamapun: If tokens are what the tool wants to work with, fine!  But at least in a LaTeXML context, a goal should be to then reinsert the knowledge gained back into the XMath tree (or create the tree).  Otherwise, although you might be able to construct cmml, I'd think pmml that resembles the author's intent would be way out of reach.

I think your search comment misses the point in that cmml is exactly good for representing the "understood" math; and good for indexing in search.  The problem is how to get there.

**Comment by brucemiller** [comment]
So, what is it that we're disagreeing on?

**Comment by dginev** [comment]
Well we were mostly wrestling due to me requesting it :> 

But I guess my point leans towards 1) "this PR can be simple", and 2) a MathParser refactor may be needed to make the serialization a utility, and to make a nicely pluggable "analysis" module that executes roughly around MathParser.

I was going to bring up the MathParser refactors I have been thinking of when we got to discussing grammars again at some later point.

**Comment by brucemiller** [comment]
Sure, OK, the refactoring can wait till there's something to go between.  But I'd still like to make it slightly more complicated: Namely separate presentation & content traversals to build lexeme sequences after parsing, and the ability to store both of those as well for later annotation.

**Comment by dginev** [comment]
Cool, sure. I'm surprised at the interest, given how little love my "semantic tex attribute" issue got back in the day (see #432 )

**Comment by brucemiller** [comment]
? Had nothing to do with lack of love for the attributes, but rather lack of love for the limitations of MathML annotations.

**Comment by dginev** [comment]
Sure, it just never took off - that's all I meant. If we start adding multiple lexeme serializations, we can consider adding multiple tex serializations too while we're at it.

**Comment by dginev** [comment]
---

Unwinding back to the top of the discussion, just bookkeeping here that the PR is considered blocked until we also transfer any relevant font+style information to the lexeme serialization. So that, as a basic example, `bold x` is serialized as a distinct lexeme from a regular mathematical `x`. 

The key observation being that while font+style differences in regular language mainly implies emphasis or structural role (e.g. headings) of the same word, while in mathematical expressions the underlying symbol is different - both emphasis and structural information are not employed in formulas.

**Comment by dginev** [comment]
Let me add another interesting example I hadn't thought of when preparing the PR - over accents, which are not scripts. In particular, the conical function syntax in the left-hand side of equation 14.20.2 of the DLMF:

image

Here is the current lexeme string produced with the PR, with my manual indent as usual:
```
OVERACCENT:widehat ARG:start UNKNOWN:Q ARG:end
POSTSUPERSCRIPT:start ADDOP:minus UNKNOWN:mu POSTSUPERSCRIPT:end 
POSTSUBSCRIPT:start
  ADDOP:minus FRACOP:divide
    ARG:start NUMBER:1 ARG:end 
    ARG:start NUMBER:2 ARG:end  
  ADDOP:plus UNKNOWN:i UNKNOWN:tau 
POSTSUBSCRIPT:end
```

This seems reasonable I think? Just jotting it down as an extra example (and archival)

**Comment by dginev** [comment]
Greetings! I have just added a piece of code that I would like to claim could move this PR from "demo" to "experimental", and allow it to enter a release, even in this imperfect form. I am of course also happy to reimplement if I get actionable feedback!

The additions from today:
 * I added a crude approximation to "obtaining the Font information" for the leaf node lexeme serialization. I am sure Bruce knows a better way, but this was a quick stab that seems to work decently well for basic examples.
 * I dropped the `UNKNOWN` role from the serialization, seeking some compromise ground after I got feedback that the grammatical role feels unnecessary. I still consider it useful to have the *full* assumed grammatical information that latexml's MathParser has, in order to enable reasonable comparisons between math grammars. Dropping unknown doesn't look anything in this regard, as one can assume any lexeme without a role is unknown.
 * Similarly, I concatenated the font information via a dash, to contrast between it and the colon separator for the role.

Perfect? No. But good enough for meaningful experiments? I think so, and can name two separate classes that I would be personally interested in.

Revealing example:
```
latexmlc --preload=bm.sty --preload=llamapun.sty --pmml --mathlex 'literal:$x+\boldsymbol{y}=0'
```

produces the lexemes:
```
italic-x ADDOP:plus bold-italic-y RELOP:equals NUMBER:0
```

File: lib/LaTeXML/MathParser.pm [hunk 1]
```diff
@@ -680,19 +693,85 @@ sub parse_single {
...
   my $i       = 0;
   my $lexemes = '';
+
   foreach my $node (@nodes) {
-    my $role = $self->getGrammaticalRole($node);
-    my $text = getTokenMeaning($node);
-    $text = 'Unknown' unless defined $text;
-    my $lexeme = $role . ":" . $text . ":" . ++$i;
-    $lexeme =~ s/\s//g;
+    my $lexeme = $self->node_to_lexeme($node) . ":" . ++$i;
```
dginev: ah, I forgot I was reusing the method in the internal parse, I need to be a little more careful here - my font changes broke a test I'm afraid.

dginev:
Removed a small hiccup, now fully separating the experimental lexeme syntax serialization from the internal parser lexemes. Could be cleaned further, but I prefer waiting for feedback.

brucemiller:
Yeah, the font size stuff is silly! :>  I'd thought there was the beginnings of a notion of math-meaningful font attributes, but can't quite find it.  I do find the inverse: there's a couple of methods in Common::Font regarding "pursestyle", which give the parts of a font which are (presumably) *not* meaningful.  Perhaps that's a good place to start with a new method to get the meaningful parts, which would be only family, series, shape, I guess.

dginev:
Quick update here, as discussed, now explicitly only using the family, series and shape of the font, as relative to the default text font. The example from my previous comments with the direct sum now produces:

```
italic-x SUMOP:direct-sum bold-italic-y RELOP:equals NUMBER:0
```

which seems quite reasonable.

Issue #948, opened 2018-02-07
Bold zeros in math lost during transformation to HTML


goska:
Applies to LaTeXML 0.8.2.
In math fragments, zeros tagged in bold font are lost during transformation to HTML with MathML, that is: XML fragment
` <XMTok font="bold" meaning="0" role="NUMBER">0</XMTok>` gets transformed to
`<mn></mn>`
Other digits appear unaffected. I tested with 1:
`<XMTok font="bold" meaning="1" role="NUMBER">1</XMTok>`
was transformed into
`<mn>𝟏</mn>`.
The same input LaTeX code (from which quoted XML was derived) worked OK in LaTeXML 0.8.1. Is there any way I could fix this problem?

dginev:
I think Bruce just fixed this in #926 -- but you would need the latest GitHub version to take advantage of the fix.

Or of course 0.8.3, when we release it.

goska:
On brief reading, #926 mentions specifically output to STDOUT, but I was  writing to an HTML file.

dginev:
You can give me the exact input that reproduces the problem and I can test if it is still present in the latest master branch.

I can verify that:
```
 latexmlc --preload=amsmath.sty 'literal:$\boldsymbol{0}$' --dest=test.html
```

produces the zero correctly. So it should be fixed.

goska:
OK, your test indicates that the problem is indeed fixed in the latest version. Thanks.

dginev:
Sure thing! I actually spotted more encoding issues while testing, so your issue was helpful in improving our encoding robustness. PR is at #949 

I will close here, since things seem OK.

PR #949, opened 2018-02-07
Two more wrongly encoded cases in LaTeXML.pm

dginev:
While investigating #948 I spotted a broken bold 0 encoding when printing to stdout from latexmlc.

Turns out there were two cases that needed new patches
 * One was the fatal error case for core processing, which was missing the encode call
 * The other was the XML format case in post-processing, which was leveraging the duck-typing of all document objects (i.e. all of them having `toString`) to return differently encoded values -- since exactly one of our `toString` calls isn't returning bytes. So I added an explicit check there.

Keeps feeling awkward... I still don't care enough to raise a larger stink about refactoring :> If I hit this once again I might.

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -277,6 +277,8 @@ sub convert {
     my $log = $self->flush_log;
     $serialized = $dom if ($$opts{format} eq 'dom');
     $serialized = $dom->toString if ($dom && (!defined $serialized));
+    # Using the Core::Document::serialize_aux, so need an explicit encode into bytes
+    $serialized = Encode::encode('UTF-8', $serialized) if $serialized; 
 
     return { result => $serialized, log => $log, status => $$runtime{status}, status_code => $$runtime{status_code} }; }
   else {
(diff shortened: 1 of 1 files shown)


dginev:
One way of seeing this may be that **every** use of the `serialize_aux`-based toString requires/expects encoded unicode. Which may be an indication that the encode call could be pushed more tightly into the actual method?

brucemiller:
Actually that's not true. I should have caught that in your comments, but `serialize_aux` is internal to implementing `LaTeXML::Core::Document->toString` so (virtually) "nobody" uses it directly, but *everybody* uses `toString` and they expect characters. (should correct the comments).  As I said elsewhere, only `XML::LibXML::Document->toString` (and `toStringHMTL`) return bytes and those really should be locked away to prevent confusion

dginev:
Ah, but I phrased my claim extremely carefully this time.

All `toString` calls for the _final_ output of the created Document by latexml, be it Core, Post, or some fragment in the `--whatsout=fragment` route, is currently expected to be bytes. All classic executables, as well as LaTeXML.pm, do a `print` assuming the payload is already in bytes.

This is the entire realm I was quantifying over when I said **every**, I was being quite careful.

The internal `toString` methods are not something i have studied closely recently, but I do recall the main workhorse internally tend to be the `ToString` and `Stringify` global functions instead.

Anyway, I think we are saying the same thing from 2 perspectives. Maybe a renaming all Document methods away from `toString` into a `serialize` or `serializeFinal` call, that is always bytes, and mandating all `toString` calls are chars is the right refactor? And we refactor away any uses of the `toString` calls in the executables and Writer, to be sure.

brucemiller:
Oh! I forgot to merge... is this one still needed?

dginev:
It is still needed, yes.

I thought you left it open so that I rework it so that there is a specially named method that always encodes to Unicode, but if you have decided that is a long-term goal, merging here is still a good choice. Fixes a problem.

brucemiller:
Ah, not really; a nice safe general approach will be good, but I'm not waiting.

Issue #951, opened 2018-02-25
Infinite loop while Building


dginev:
I have extracted a minimal example from the infinite loop reported at https://github.com/arxiv-vanity/engrafo/issues/280

which converted [1404.2371](https://arxiv.org/abs/1404.2371) from arXiv. The minimal example is:

```tex
\documentclass{article}
\usepackage{amsmath,amssymb,amsthm}

\def\qed{\framebox(6,6){}}

\begin{document}

\begin{proof}
proof here
\end{proof}

\end{document}
```

which produces the log:
```
$ latexml nthroot.tex

latexml (LaTeXML version 0.8.2; revision a23e6ac2)
processing started Sun Feb 25 11:54:07 2018

(Digesting TeX nthroot...
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/TeX.pool.ltxml...
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/eTeX.pool.ltxml... 0.01 sec)
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/pdfTeX.pool.ltxml... 0.01 sec) 0.14 sec)
(Processing content /home/dreamweaver/testbed/infinite_loop_latexml_post/nthroot.tex...
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/LaTeX.pool.ltxml... 0.16 sec)
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/article.cls.ltxml... 0.02 sec)
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/amsmath.sty.ltxml...
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/amsbsy.sty.ltxml... 0.00 sec)
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/amstext.sty.ltxml... 0.00 sec)
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/amsopn.sty.ltxml... 0.00 sec) 0.06 sec)
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/amssymb.sty.ltxml...
(Loading /home/dreamweaver/perl5/lib/perl5/LaTeXML/Package/amsfonts.sty.ltxml... 0.00 sec) 0.03 sec)
[...]

PR #952, opened 2018-02-25
Remove class of infinite loop errors during insertion point lookup

dginev:
This PR adds a looping guard to `find_insertion_point` addressing the generic part of #951 

The idea is to recognize cases where we auto-open intermediate nodes that have already been auto-closed during the same recursive chain, and terminate immediately with a malformed error.

The PR reduces the example in #951 to a single second runtime, with the error:
```
Error:malformed:ltx:rect <ltx:rect> loops infinitely when opening an intermediate <ltx:text>
	at /home/dreamweaver/testbed/infinite_loop_latexml_post/nthroot.tex; line 9 col 1
	Currently in XMLDocument[94225025257712]<ltx:document _box="LaTeXML::Core::Whatsit=HASH(0x..." _font="Font[serif,medium,upright,10,b..." _ID_counter_p_="1" xmlns="http://dlmf.nist.gov/LaTeXML">...<ltx:proof _box="LaTeXML::Core::Whatsit=HASH(0x..." _font="Font[serif,medium,upright,10,b...">...<ltx:para _box="LaTeXML::Core::Box=HASH(0x55b2..." _font="Font[serif,medium,upright,10,b..." xml:id="p1">...<ltx:p _box="LaTeXML::Core::Box=HASH(0x55b2..." _font="Font[serif,medium,upright,10,b...">...
	In Core::Document[@0x55b2796bc9b0] at /home/dreamweaver/testbed/infinite_loop_latexml_post/nthroot.tex; line 9 col 1
	 <= Core::Definition::Constructor[\frameb... <= Whatsit[\pic@framebox,Pair[Float[6],F... <= Core::Document[@0x55b2796bc9b0] <= ...
```

and adds the malformed `ltx:rect` directly into its the `ltx:p`.

The main goal here is to have a generic guard that reduces the number of busy loops when converting arXiv, irrespective of the specific schema issue.

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -883,7 +883,7 @@ sub getInsertionContext {
 # This will move up the tree (closing auto-closable elements),
 # or down (inserting auto-openable elements), as needed.
 sub find_insertion_point {
-  my ($self, $qname) = @_;
+  my ($self, $qname, $closedto) = @_;
   $self->closeText_internal;    # Close any current text node.
   my $cur_qname = $$self{model}->getNodeQName($$self{node});
   my $inter;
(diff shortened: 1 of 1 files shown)


brucemiller:
Part of me wants to not do this 'cause obviously after we fix #951, this kind of thing will *never* happen again! but...

Actually, can' t this be even simpler: You may want to autoclose, then autoopen, but once you've started autoopen, you don't want to autoclose at all; you don't even need to compare tags.

dginev:
Is that simpler or just the inverse of what I did?

dginev:
Ah, you mean have a boolean flag instead of the string comparison, got it. Let me add a commit ...

brucemiller:
Yep, exactly. Actually, I already have it in my `Core::Document` -- it also helps to debug the underlying issue.  But thanks for the inspiration!

dginev:
Hey! Not so fast :> 

At least take a look at my commit, which changes the error to:
```
Error:malformed:ltx:rect <ltx:rect> failed auto-open through <ltx:text> at inadmissible <ltx:text>
```

dginev:
And my commit didn't show because you closed the PR, tsk tsk, now you get a new one :>

PR #956, opened 2018-03-01
Soften heuristic, was breaking mathbf macro

### LINKED ISSUE #931: Math fonts needs surrounding {-}
I'm still working with `latexmlc` and it works fine. Here is my input:
```
latexmlc \
  --includestyles\
  --format=xhtml\
  --whatsin=math\
  --whatsout=math\
  --pmml\
  --cmml\
  --nodefaultresources\
  --linelength=90\
  --preload=LaTeX.pool\
  --preload=article.cls\
  --preload=amsmath.sty\
  --preload=amsthm.sty\
  --preload=amstext.sty\
  --preload=amssymb.sty\
  --preload=eucal.sty\
  --preload=[dvipsnames]xcolor.sty\
  --preload=url.sty\
  --preload=hyperref.sty\
  --preload=[ids]latexml.sty\
  --preload=texvc\
  literal:"\frac{a}{b}"
```
However, when I enter `\mathbf{x}` there is a warning about "mathbf" should only appear in math mode and the output is a html fragment. Same for `\mathcal{x}`. When I enter `{\mathbf{x}}` everything works again...

dginev:
Fixes #931 . 

Simple case of over-eager heuristics, any macro that started with `\math` was treated as a signal that math mode was already activated during `--whatsin=math` processing. Clearly not the case.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -1485,7 +1485,7 @@ DefMacro('\@ensuremath{}', sub {
     else { (T_MATH, $stuff->unlist, T_MATH); } });
 
 # Magic check that math-mode trigger follows
-our $MATHENVS = 'math|displaymath|equation*?|eqnarray*?'
+our $MATHENVS = 'displaymath|equation*?|eqnarray*?'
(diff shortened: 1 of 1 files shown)


brucemiller:
Screwy, but reasonable :>

Issue #806, opened 2016-10-22
Latexmlc epub generation fails on Windows


charlyms:
Hello,
I tried to convert some arxiv.org-stuff from source to epub but failed with Windows 10:

```
C:\Users\Klaus\Eigene Dokumente\Anne\HERUS A CO Atlas from SPIRE Spectroscopy of local ULIRGs~>latexmlc herusfts_ajs_forastroph.tex --destination=book.epub
latexmlc (LaTeXML version 0.8.2) : these path directories do not exist: C:/Users/Klaus/Eigene\ Dokumente/Anne/HERUS\ A\ CO\ Atlas\ from\ SPIRE\ Spectroscopy\ of\ local\ ULIRGs~
(Loading C:\Strawberry\perl\site\lib\LaTeXML\Package\TeX.pool.ltxml...
(Loading C:\Strawberry\perl\site\lib\LaTeXML\Package\eTeX.pool.ltxml... 0.00 sec)
(Loading C:\Strawberry\perl\site\lib\LaTeXML\Package\pdfTeX.pool.ltxml... 0.01 sec) 0.20 sec)
latexmlc (LaTeXML version 0.8.2)
processing started Sat Oct 22 20:24:47 2016...

Calibre cannot open the file:
Traceback (most recent call last):
  File "site-packages\calibre\utils\ipc\simple_worker.py", line 286, in main
  File "site-packages\calibre\ebooks\oeb\iterator\book.py", line 64, in extract_book
  File "site-packages\calibre\customize\conversion.py", line 245, in __call__
  File "site-packages\calibre\ebooks\conversion\plugins\epub_input.py", line 238, in convert
  File "site-packages\calibre\utils\localunzip.py", line 231, in extractall
  File "site-packages\calibre\utils\localunzip.py", line 187, in _extractall
AttributeError: 'NoneType' object has no attribute 'replace'
```

Trying with Ubuntu 16.04 I succeed.

```
charlyms@ubuntu:~/Dokumente/HERUS A CO Atlas from SPIRE Spectroscopy of local ULIRGs~$ latexmlc herusfts_ajs_forastroph.tex --destination=book.epub
latexmlc (LaTeXML version 0.8.1) : these path directories do not exist: /home/charlyms/Dokumente/HERUS\ A\ CO\ Atlas\ from\ SPIRE\ Spectroscopy\ of\ local\ ULIRGs~
[...]

PR #960, opened 2018-03-19
Ensure raw binmode when writing archive files in latexmlc

### LINKED ISSUE #827: Latexmlc (LaTeXML 0.8.2) epub generation fails on Windows 10
There are multiple issues with epub generation.

Command line latexmlc --dest "%1.epub" "%1.tex" produces an epub with only mimetype file and META-INF folder containing container.xml.

The same script when run through TeXWorks throws an error

Can't locate loadable object for module XML::LibXML in @INC (@INC contains: C:\Strawberry\perl\site\lib D:/texlive/2016/tlpkg/tlperl/lib d:/texlive/2016/tlpkg/tlperl/site/lib d:/texlive/2016/tlpkg/tlperl/lib .) at C:\Strawberry\perl\site\lib/XML/LibXML.pm line 156.

and does not generate any epub.

On removing the path definition d:\texlive\2016\tlpkg\tlperl\bin from TeXWorks preferences, produces a new error

Cwd.c: loadable library and perl binaries are mismatched (got handshake key 0AA80080, needed 0AF00080)

dginev:
Fixes #827 and #806 

This is a platform-specific fix, and in fact I need to double-check this keeps the epub generation as-is on Linux. 

It turns out that without an explicit `:raw` binmode, printing out binary data on Windows 10 leads to corrupted archives, for reasons I do not fully understand at the moment. 

While debugging, I first verified the archive integrity, all EPUB-needed files were added to the Archive::Zip object and then serialized via IO::String. However, the same payload lead to different `.epub` file contents when e.g. the compression level was changed, leading to the OPS subdirectory completely disappearing when the archive is viewed in 7zip. 

In the end, it turned out the produced archives were partially readable by 7zip (my Win10 archive reader of choice), but were silently corrupted at some hard to predict point in the file-writing process.

Adding an explicit `:raw` binmode to the file handle in latexmlc resolved this fully, and created a fully functioning `.epub` file in Windows 10.

### DIFF
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -162,6 +162,9 @@ if ($result) {
     if (!open($output_handle, ">", $opts->get('destination'))) {
       print STDERR "Fatal:I/O:forbidden Couldn't open output file " . $opts->get('destination') . ": $!";
       exit 1; }
+    if ($whatsout eq 'archive') {
+      binmode($output_handle, ':raw');
+    }
     print $output_handle $result;
     close $output_handle;
   } else {
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -546,7 +546,7 @@ sub convert_post {
   # If our format requires a manifest, create one
   if (($$opts{whatsout} =~ /^archive/) && ($format !~ /^x?html|xml/)) {
     require LaTeXML::Post::Manifest;
-    my $manifest_maker = LaTeXML::Post::Manifest->new(db => $DB, format => $format, %PostOPS);
+    my $manifest_maker = LaTeXML::Post::Manifest->new(db => $DB, format => $format, log=>$$opts{log}, %PostOPS);
     $manifest_maker->process(@postdocs); }
   # Archives: when a relative --log is requested, write to sandbox prior packing
   if ($$opts{log} && ($$opts{whatsout} =~ /^archive/) && (!pathname_is_absolute($$opts{log}))) {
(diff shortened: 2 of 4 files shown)


dginev:
I am also about to add a test case here, which will keep us away from regressions.

brucemiller:
On 03/19/2018 12:16 PM, Deyan Ginev wrote:
> It turns out that without an explicit |:raw| binmode, printing out 
> binary data on Windows 10 leads to corrupted archives, for reasons I do 
> not fully understand at the moment.

For unixy systems, files are always just bytes;
the binmode only determines a layer which may
encode, say Unicode, into bytes.

But as I understand it, windows actually
distinguishes binary from text files, so
binmode for a zip apparently matters.

dginev:
@brucemiller indeed, that seems to be the case.

I have now added an end-to-end test for epub generation via latexmlc, which runs the executable via the shell, and then checks the resulting file has the expected data integrity and epub contents.

I stumbled on a couple of validator issues while debugging via:
```
latexmlc literal:test --dest test.epub --log test.log
```

so added some minor upgrades to keep the epub valid.

dginev:
With the upgrades in this PR, the basic epub test is almost valid, with the exception of some extremely picky CSS rules. But we can merge as-is, as it resolves the Windows-specific problem.

css_epub_validation

dginev:
Courtesy of Travis, I can now also claim the PR keeps the epub generation operational in linux 👍

brucemiller:
Cool! Thanks!!

brucemiller:
Didn't notice exactly where it comes from, but when I make test, now, I get an extra "No obvious problems\n Wrote 931_testdPmT.epub" in the output.... Doesn't cause the tests to fail, but distracts from the quick visual confirmation.

dginev:
Yes, it's from this specific PR, it's an extra print from running latexmlc externally. I'll try to contain it and submit another PR.

PR #961, opened 2018-03-22
Teasing away Perl-level dies for 100+ error document

dginev:
Just took a quick look at the Perl-level fatal errors of `1802.08337` in arXiv, and teased them away so that the Fatal comes from having 100+ errors. The document itself has some serious problems with tikz interpretation, so I won't attempt to resolve it to a non-error state now.


brucemiller:
Cool! Thanks!!

PR #962, opened 2018-03-22
Teasing away Perl-level die for error arxiv doc 1802.05385

dginev:
This PR upgrades the mentioned arXiv doc from a Fatal status to an Error status with a rather decent looking HTML output.


brucemiller:
And thanks!...

PR #963, opened 2018-03-24
make DirectoryList tolerant of spaces

dginev:
The specific use in arXiv 1802.01430 was causing a Fatal error:
```
\graphicspath{ {./Figs/} } 
```

as the argument was being misinterpreted. Please check if this fix is the most accurate upgrade, it certainly makes the document work with a decent HTML output.


brucemiller:
May not be the only way, but looks like it should work.

PR #964, opened 2018-03-25
when core returns an empty document, ensure post still succeeds

dginev:
This causes cortex-specific troubles, as the return archive is empty, and hence no log information can be gathered. 

This PR ensures if the core processing returns a completely empty document, without any elements set, it will still produce an HTML document and add the log file to the archive, when the target is `.zip`.

The specific arXiv document that hit this was 1802.05735


brucemiller:
Hm.... ok.

PR #966, opened 2018-03-26
Initialize branch TagsNTitles; replaces attributes @refnum,@frefnum,@…

brucemiller:
…rrefnum with more flexible ltx:tags containing multiple ltx:tag, each with @role to indicate what each tag is used for; this allows formatting within the reference numbers, extensibility for new types of tags (eg. hyperref's autoref), possibilities for consolodating duplicated code and making it easier to implement endnotes, glossaries, etc. These changes will be disruptive of some bindings; More to come!


lib/LaTeXML/Post/Scan.pm
```diff
@@ -422,14 +399,14 @@ sub bibitem_handler {
...
       fragid      => orNull($self->inPageID($doc, $id)),
-      authors     => orNull($doc->findnode('ltx:bibtag[@role="authors"]', $node)),
-      fullauthors => orNull($doc->findnode('ltx:bibtag[@role="fullauthors"]', $node)),
-      year        => orNull($doc->findnode('ltx:bibtag[@role="year"]', $node)),
-      number      => orNull($doc->findnode('ltx:bibtag[@role="number"]', $node)),
-      refnum      => orNull($doc->findnode('ltx:bibtag[@role="refnum"]', $node)),
-      title       => orNull($doc->findnode('ltx:bibtag[@role="title"]', $node)),
-      keytag      => orNull($doc->findnode('ltx:bibtag[@role="key"]', $node)),
-      typetag     => orNull($doc->findnode('ltx:bibtag[@role="bibtype"]', $node))); }
+      authors     => orNull($doc->findnode('ltx:tags/ltx:tag[@role="authors"]', $node)),
```
dginev: Since `ltx:tags` is used for more than bibliographies, isn't the new selector potentially more general than simply over bibliography tag elements? And is that intended?

brucemiller: Indeed; I replaced sequences of `ltx:bibtag` by an `ltx:tags` containing any number of `ltx:tag`. But `ltx:tags` is also used to collect the single `ltx:tag` that were previously allowed at the beginnings of various sectional, floats, theorems, etc (there can now also be multiple `ltx:tag`). And they're also allowed in equations and such.  So, yeah, that was the idea that a collection of tags would become fully general.

t/alignment/algx.xml
```diff
@@ -8,128 +8,168 @@
...
+</listingline>
+      <listingline xml:id="alg1.l3"><tags>
+          <tag><text fontsize="80%">3:</text></tag>
+          <tag role="refnum">3</tag>
+        </tags>     <text font="bold">while</text> <Math mode="inline" tex="r\not=0" text="r not-equals 0" xml:id="alg1.l3.m1">
           <XMath>
             <XMApp>
-              <XMTok fontsize="80%" meaning="not-equals" name="not-=" role="RELOP">≠</XMTok>
-              <XMTok font="italic" fontsize="80%" role="UNKNOWN">r</XMTok>
-              <XMTok fontsize="80%" meaning="0" role="NUMBER">0</XMTok>
```
dginev: unsure if this is minor, but the fontsize attribute disappeared from the individual math tokens? Seems unrelated as a change.

t/alignment/array.xml
```diff
@@ -5,7 +5,12 @@
 <document xmlns="http://dlmf.nist.gov/LaTeXML">
   <resource src="LaTeXML.css" type="text/css"/>
   <resource src="ltx-article.css" type="text/css"/>
-  <section refnum="1" xml:id="S1">
+  <section xml:id="S1">
+    <tags>
+      <tag>1</tag>
+      <tag role="refnum">1</tag>
+      <tag role="typerefnum">§1</tag>
```
dginev: why not `trefnum` for a role, given there is `frefnum`? I am unsure if it's better to argue for a single letter convention, or longer attribute names such as `refnum_formatted` and `refnum_type` or some such.

Also curious to observe that they really have the same refnum *role*, but hold distinct `variant`s for that role. Though I suspect adding another attribute is overhead that's not its time right now.

t/alignment/algx.xml
brucemiller: I think this was due to a bugfix that came about as a side-effect of treating tags more uniformly: sometimes the formatter for a tag was something like `\bf\thesomething`; Hopefully the current code is more consistent about wrapping a group around so that the sideeffects don't leak out, like the small font did here.

t/alignment/longtable.xml
```diff
@@ -7,7 +7,12 @@
 <document xmlns="http://dlmf.nist.gov/LaTeXML">
   <resource src="LaTeXML.css" type="text/css"/>
   <resource src="ltx-article.css" type="text/css"/>
-  <table frefnum="Table 1" refnum="1" xml:id="S0.T1">
+  <table xml:id="S0.T1">
+    <tags>
+      <tag>Table 1</tag>
```
dginev: just as a readability remark, I'm unsure what the tag with no attribute signifies. Looking at the diff I assume it is the `frefnum` that was originally on the table element? If so, maybe it warrants a role attribute specifying that?

t/daemon/formats/tei.xml
```diff
@@ -825,6 +882,9 @@ Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor
         </div>
       </div>
       <div>
+        <!-- The element tags
+            is currently not supported for the main body.
+        -->
```
dginev: Extremely minor, but seeing this same comment on every div feels a bit verbose.

dginev:
For what is worth, I skimmed the code and left minor comments. The changeset is very impressive! So I can not claim any reasonable quality control, but at least I understand the diffs better.

Depositing so many refnum-like tags does feel redundant while skimming, but it's probably worth the trouble given how much machinery requires the different variants. I can definitely believe this makes writing the XSLTs much easier. So all 👍 from me, let me know if you'd like me to take a closer look at some specific bit.

lib/LaTeXML/Post/Scan.pm
dginev: Cool!

t/alignment/array.xml
brucemiller: I spent a lot of time, and back-and-forth, with the names, trying to improve over "frefnum" :>  There are 3 base cases roughly equivalent to the former `frefnum`, `refnum` and `rrefnum`, but hopefully with more explanatory names, but perhaps those names can be further clarified.

* No role is a "formatted refnum"; it is used for the display in or alongside the object being tagged (eg. items, equations) Exception is that objects with a title or caption, the tag is essentially duplicated within the title/caption so the one in the `ltx:tags` is likely never used.
* `@role=refnum` is the plain refnum, essentially `\thecounter` which LaTeX would use to fill in `\ref`.  Note that the author typically gives some context like `see section \ref{foo}`, and that not all objects get a refnum, in which case latex uses the refnum of the parent.
* `@role=typerefnum` is a "typed refnum" for use when generating references to an object, typically automated, such as back-references and such. There's no context there, so the type name or symbol is helpful.

Hope that helps; if you can think of better names...

t/alignment/longtable.xml
brucemiller: see reply to previous comment about what the lack of `@role` means. Either can be default or explicit, if there's a meaningful name for it.

t/alignment/array.xml
dginev: Thanks! Got it.

Well, `refnum_formatted` and `refnum_typed` do sound better to me, but they are also long, so probably not ideal. I still slightly lean towards recommending them...

dginev: The other, more informal-sounding, bunch of names that come to mind are things like `fmtnum` and `typnum`, but I don't really find the appeal in such dashing shorthands nowadays.

brucemiller:
OK, let's cross our fingers and do it!

dginev:
oh wow, it shipped! :rocket: 

Time to do some testing...

PR #969, opened 2018-04-13
Integrate logs from separate bibliography converts in post-processing

### LINKED ISSUE #916: Postprocessing doesn't show errors during processing of bibliography
When processing the bibliography `latexmlpost` only shows the number of errors encountered, but doesn't give any hint, what went wrong:

```
latexmlpost --verbose --verbose --verbose --dest=equations.html equations.xml 

(MakeBibliography equations.html processing... [Converting bibliography equations/test.bib ...


Conversion complete: 1 error.
Status:conversion:2 
... converted!] [using bibliographies test.bib] [154 bibentries, 9 cited] [DBStatus: 8195/0 objects] 13.45 sec)
```
With a big `*.bib` file, it can be quite painful to track down these errors. It would be nice, if `latexmlpost` would at least give the line number, where it encountered problems parsing or processing the `*.bib` file.

dginev:
Fixes #916 

A bit tricky, since managing the STDERR redirects is getting wild, but it works nicely. Unsure about the final log to print, attaching screenshots with this version. I did not want to indent the messages, so that automated regexes (e.g. from cortex) can still parse the log as usual.

Here are the logs converting the the issue example, with `master`:

master

And the logs after this PR is installed:

new

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -210,6 +210,7 @@ sub convert {
   my $latexml = $$self{latexml};
   $latexml->withState(sub {
       my ($state) = @_;    # Sandbox state
+      $$state{status} = {};
       $state->pushDaemonFrame;
       $state->assignValue('_authlist', $$opts{authlist}, 'global');
       $state->assignValue('REMOTE_REQUEST', (!$$opts{local}), 'global');
@@ -255,7 +256,6 @@ sub convert {
       my ($state) = @_;    # Remove current state frame
       $$opts{searchpaths} = $state->lookupValue('SEARCHPATHS'); # save the searchpaths for post-processing
       $state->popDaemonFrame;
-      $$state{status} = {};
   });
   if ($LaTeXML::UNSAFE_FATAL) {
     # If the conversion hit an unsafe fatal, we need to reinitialize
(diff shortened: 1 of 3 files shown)


lib/LaTeXML.pm
```diff
@@ -255,7 +256,6 @@ sub convert {
       my ($state) = @_;    # Remove current state frame
       $$opts{searchpaths} = $state->lookupValue('SEARCHPATHS'); # save the searchpaths for post-processing
       $state->popDaemonFrame;
-      $$state{status} = {};
```
dginev: this was the only place I had to touch the actual daemon core logic.

In essence, instead of immediately reset-ing the status hash after a conversion run is completed, I now reset it before a fresh run is started. This way I can still reuse the hash for passing on the log to a parent converter object.

lib/LaTeXML/Common/Error.pm
```diff
@@ -357,11 +357,9 @@ sub MergeStatus {
     if ($type eq 'undefined' or $type eq 'missing') {
       my $table = $$external_status{$type};
       foreach my $subtype(keys %$table) {
-        $$status{$type}{$subtype} ||= 0;
         $$status{$type}{$subtype} += $$table{$subtype};
       }
     } else {
-      $$status{$type} ||= 0;
```
dginev: I have gotten used to being extra cautious from my Ruby days, let me know if you want these guards back @brucemiller

dginev: works just fine without them naturally

brucemiller:
Looks good; let me know when you're done :>

dginev:
Oh, i am done here.

brucemiller:
Cool! Thanks!!

PR #970, opened 2018-04-17
do not Fatal when dimension flow goes awry, recover with an Error

dginev:
Recovers 20 out of ~650 Fatals in the single month (feb 18) arXiv rerun I did yesterday.

Basic gain is avoiding a hard Fatal, and producing a reasonable Error-severity HTML output.

### DIFF
--- a/lib/LaTeXML/Common/Dimension.pm
+++ b/lib/LaTeXML/Common/Dimension.pm
@@ -14,6 +14,7 @@ use strict;
 use warnings;
 use LaTeXML::Global;
 use LaTeXML::Common::Object;
+use LaTeXML::Common::Error; 
 use base qw(LaTeXML::Common::Number);
 use base qw(Exporter);
 our @EXPORT = (qw(&Dimension));
@@ -64,6 +65,28 @@ sub pointformat {
 sub attributeformat {
   my ($sp) = @_;
   return sprintf('%.1fpt', LaTeXML::Common::Number::roundto($sp / 65536, 1)); }
+
+sub getDepth { # ??? 
+  #  sometimes Dimension(0) is used as a default value
+  #  and errors in the flow invoke getDepth on that dimension
+  #
+  #  probably safe to assume asking for the depth of a dimension 
+  #  was caused by a bug in execution, and recover by returning the dimension itself?
+  my ($self) = @_;
+  Error('expected','Box', undef, 'getDepth invoked on a Dimension, returning self');
+  return $self;
+}
+# Might as well add the defaults for width and height while we're at it.
+sub getWidth {
+  my ($self) = @_;
+  Error('expected','Box', undef, 'getWidth invoked on a Dimension, returning self');
+  return $self;
+}
+sub getHeight {
+  my ($self) = @_;
+  Error('expected','Box', undef, 'getHeight invoked on a Dimension, returning self');
+  return $self;
+}
 
 #======================================================================
 1;
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -1429,11 +1429,11 @@ DefRegister('\wd Number', Dimension(0),
 DefRegister('\dp Number', Dimension(0),
   getter => sub {
     my ($n) = @_;
-    my $stuff = LookupValue('box' . $_[0]->valueOf);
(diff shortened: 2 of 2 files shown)


brucemiller:
The fix to TeX seems reasonable, but the one to Dimension seems perverse. Can you point to a case where that occurs? What kind of error does it produce?

dginev:
Sure, the arXiv job I am testing on is:
https://arxiv.org/abs/1802.01810

Running `latexml zariski_semigroup.tex` on the sources, the Fatal I see after fixing only TeX.pool.ltxml is:

```
Fatal:perl:die Perl died
	at /home/dreamweaver/testbed/100/67608/zariski_semigroup.tex; line 854 col 114
	Can't locate object method "getDepth" via package "LaTeXML::Common::Dimension" at (eval 85) line 2577, <$IN> line 854.
	In Core::Definition::Primitive[\pgfmaths... from pgfmath.code.tex.ltxml line 316
	 <= Core::Stomach[@0x5627594e9de0] <= Core::Definition::Primitive[Begin] <= Core::Stomach[@0x5627594e9de0] <= Core::Definition::Constructor[\hbox B... <= Core::Stomach[@0x5627594e9de0] <= Core::Definition::Primitive[\setbox N... <= Core::Stomach[@0x5627594e9de0] <= Core::Definition::Primitive[Begin] <= Core::Stomach[@0x5627594e9de0] <= Core::Definition::Constructor[\hbox B... <= Core::Stomach[@0x5627594e9de0] <= Core::Definition::Primitive[\setbox N... <= Core::Stomach[@0x5627594e9de0] <= Core::Definition::Constructor[\begin{... <= Core::Stomach[@0x5627594e9de0]
```

There are a lot (31) errors prior to this Fatal, and patching the pgf interpretation for this document may require a lot more work. I did not trace the full macro chain.

dginev:
Oh, the cortex ZIP may be easier to unpack, you can find it at (assuming captcha was previously accepted):
https://corpora.mathweb.org/entry/import/67608

dginev:
Actually it is still mysterious (i don't want to say "buggy" just yet), that before any of the fixes the error I am seeing on my local laptop is:
```
Fatal:misdefined:LaTeXML::Package::Pool::__ANON__ Can't call method 'valueOf' on an undefined value
```

So still weird to have it in the 100+ errors. Then again, I just updated my texlive locally, and some of the worker machines have a 2015 version even. So this may be a pgf-related discrepancy.

brucemiller:
Yeah, but also, shouldn't that one be fixed by the TeX patch, rather than by the Dimension patch?

dginev:
I just checked the log recorded on the server machine, and indeed perfectly regular stream of 100 errors. So my fix isn't even fixing the reported cortex problem, it just improves my local installation.

... I am starting to get tempted into thinking of using a Docker-ized worker setup with a fixed texlive version, for sanity's sake.

dginev:
re: which patch fixes what. 

1. the TeX.pool patch fixes the `valueOf on undef` Fatal.
2. after which, rerunning exposes the `getDepth` on a Dimension Fatal, which also needs a patch.

After both patches are added, the document converts to an Error status with reasonable HTML, on my recent texlive installation.

dginev:
Btw, to defend that my Dimension patch is perverse - those 3 subroutines I added will only be called in perverse circumstances, and we'll know about them from the logged error messages. So it's reasonable ointment for a perverse problem, in my mind.

brucemiller:
I just wanted to look at a case that was "fixed" by the Dimension patch to see if a slightly more correct fix would work

dginev:
Got it. Let me know if I can offer any further data.

brucemiller:
Well, actually I can't find any fatals due to calling getDepth on something inappropriate.

dginev:
I'd rather we merge here before the next rerun to reduce some Fatal count. But if you think I shouldn't touch Dimension.pm I will undo that changeset. You can also just add the part you like to master and close the PR.

I'll use that as a sign to start a rerun :+1:

brucemiller:
Ah, I didn't catch that this was a new error that you got after the 1st patch.  So, I applied the patch to TeX (and also for `\wd` and `\ht'), tried things out, and am convinced that my implementation of wd,ht,dp in the pgfmath parser was to blame. I'm not sure I've got it completely right now, but it's at least plausible, and avoids a Fatal :>

dginev:
Cool, initiating rerun, thanks!

PR #973, opened 2018-04-22
Fix misc fatal:misdefined:ANON problems

dginev:
I took a pass through one example document, but also went through all line numbers in the 92 Fatal messages reported at:
```
   fatal >  misdefined >  LaTeXML::Package::Pool::__ANON__
```

https://corpora.mathweb.org/corpus/%2Fdata%2Farxiv_1802%2F/tex_to_html/fatal/misdefined

and patched them. Just some minor code hardening.


brucemiller:
Cool; thanks!

PR #974, opened 2018-04-23
use double quotes for XPath attribute values

dginev:
Fixes a Perl fatal in post-processing caused by the amazing:
```
\newtheorem*{thm1.2'}{Theorem 1.2$'$}
```

of 1802.04737 in arXiv.

Apparently XPath has no way of escaping apostrophes, so the attribute value selectors need to always be in double quotes.

### DIFF
--- a/lib/LaTeXML/Post.pm
+++ b/lib/LaTeXML/Post.pm
@@ -491,7 +491,7 @@ sub associateNode {
     # But ONLY if that XMDual is the "direct" parent, or is parent of XRef that points to $current
     if (my $container = $document->findnode('parent::ltx:XMDual[1]', $sourcenode)
       || ($sid &&
-        $document->findnode("ancestor-or-self::ltx:XMDual[ltx:XMRef[\@idref='$sid']][1]",
+        $document->findnode('ancestor-or-self::ltx:XMDual[ltx:XMRef[@idref="'.$sid.'"]][1]',
           $sourcenode))) {
       $sourcenode = $container; } }
   # If the current node is appropriately visible, use it.


brucemiller:
Not so sure that quoting is enough, since somebody will eventually sneak a double quote in there.  I'd made an attempt to sanitize a similar case recently; basically strip out such chars. It's not right either, but...  In any case, obviously that patch didn't migrate to all the necessary places... where was it?

dginev:
Possible, all I claim is that this PR upgrades 1802.04737 from a fatal to a warning status :+1:

dginev:
This one-liner can still be useful btw.

brucemiller:
But the idref is just a copy of an xml:id; that xml:id should never have gotten quotes in it. So the patch just hides an error... need to find out how it got that way.

dginev:
Oh, that's easy, in LaTeX.pool there is a `defineNewTheorem` subroutine responsible for building out the infrastructure for `\newtheorem` definitions.

And that subroutine creates `idprefix`-es that are not sanitized for apostrophes and such. And I think they remain unsanitized? At least it looks that way from the final markup.

brucemiller:
Yeah, I think that dealing with it in `NewCounter`  fixes it at the source, and hopefully handles any other cases that might sneak through. Probably can close this one.

dginev:
Merging this one-liner is still better hygiene though, and doesn't hurt :+1: up to you, great that you found a fix at the source!

brucemiller:
I see where you're coming from, but it's probably better to let the bug show itself, than to hide it here.  I'm cursed by trying to paper over register reading errors with that insidious "missing" that keeps showing up. (one of these days....).  So, sorry, I think I'll close w/o the patch. Thanks though.

PR #977, opened 2018-04-27
Adding textcase.sty test

dginev:
I was using this test to develop, but now that @brucemiller beat me to the punchline, I'm offering it as a PR to ward off regressions


brucemiller:
Cool! But as long as you're testing, wouldn't it be good to embed some of the special junk (math, `\ref`,..) that cause it to split into separate `\uppercase`s?

dginev:
Good call with adding the complex cases. I updated with the documentation examples from CTAN at:
http://mirrors.nxthost.com/ctan/macros/latex/contrib/textcase/textcase.pdf

And regenerated the PDF and XML for the new test. I noticed two problems when inspecting by hand:
 - Unicode macro seem iffy - `\ae` was capitalized in the PDF, but not in the XML
 - Citations are left unfilled during core processing so the trick with renewing the `\citeform` macro is a no-op for latexml, and the citation text remains unfilled.

But at least no errors and meaningful output! And tables, environments, seem to work ok.

dginev:
I think I'm ready to get back to PR comments.

I pushed a commit that fixes the `\ae` capitalization issue, but at the cost of the custom binding for the latex-native `\MakeUppercase` and `\MakeLowercase` macros regressing. 

Interestingly if I replace the binding with the raw code from `latex.ltx` they work again!

So feel free to examine my last commit, which should entirely fix the macro capitalization issue.

As to the citation-specific parts - it is a limitation we should probably ignore for now, until we want to model it explicitly, as you mentioned via email. Could be good to merge!

PR #983, opened 2018-05-02
\@protected@testopt definition from latex.ltx, allows loading ucs.sty

### LINKED ISSUE #982: ucs.sty raw support
We received a report that `--includestyles` interpretation over ucs.sty leads to an infite loop. It looks like a small and focused target to improve the raw interpretation of latexml, so adding a new issue.

A minimal latex document reproduces under `latexml --includestyles test.tex`:
```tex
\documentclass{article}
\usepackage{ucs} 
\begin{document}
test
\end{document}
```

dginev:
Fixes #982 

There was a rather nasty infinite loop when loading `ucs.sty` with the `--includestyles` option, which kept allocating new memory.

Turns out using the native definition of `\@protected@testopt` from latex.ltx allows for the package to load completely smoothly and error-free. So I'm submitting this PR to address the original report on the mailing list.

That said, looking at the documentation of ucs:
https://mirrors.sorengard.com/ctan/macros/latex/contrib/ucs/ucs.pdf

I am tempted to add a trivial binding for latexml, that sets the input encoding to utf8. Will wait for feedback from @brucemiller

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -5080,7 +5080,14 @@ DefMacro('\@testopt{}{}', sub {
     my ($gullet, $cmd, $option) = @_;
     ($gullet->ifNext(T_OTHER('[')) ? $cmd->unlist
       : ($cmd->unlist, T_OTHER('['), $option->unlist, T_OTHER(']'))); });
-Let('\@protected@testopt', '\@testopt');    # ?
+RawTeX(<<'EoTeX');
+\def\@protected@testopt#1{%%
+  \ifx\protect\@typeset@protect
+    \expandafter\@testopt
+  \else
+    \@x@protect#1%
+  \fi}
+EoTeX
 
 Let('\l@ngrel@x', '\relax');                # Never actually used anywhere, but...
 DefMacro('\@star@or@long{}', '\@ifstar{\let\l@ngrel@x\relax#1}{\let\l@ngrel@x\long#1}');


brucemiller:
Patch looks reasonable.

But, taking a very quick glance at ucs's docs, I'm not sure that would do it. They suggest `\usepackage[utf8x}{inputenc}`, so shouldn't the encoding already be utf8? (or do we need a utf8x.def binding?)

dginev:
Right, the patch makes the raw interpretation _possible_, but it does not make it correct.

Adding a ucs binding that is a simple alias for utf8 inputenc with empty stubs for all ucs-specific macros (since they may be used by the authors) could be a decent solution.

I can't say much more without actual tests, but will ask the original reporter in the email thread.

dginev:
Ah, I just realized I shouldn't have asked the reporter a second too late. The log suggests ucs is auto-loaded when using inputenc with `--includestyles` on, as seen at:
```

(Loading /opt/local/lib/perl5/vendor_perl/5.26/LaTeXML/Package/inputenc.sty.ltxml...
(Processing definitions /usr/local/texlive/2016/texmf-dist/tex/latex/ucs/utf8x.def... 0.15 sec) 0.54 sec)
(Processing definitions /usr/local/texlive/2016/texmf-dist/tex/latex/ucs/ucs.sty...

```

So I'm now agreeing with your remark about ut8x.def needing a binding that prevents further inputs.

brucemiller:
hmm... how's utf8x.def compare to utf8.def ?

dginev:
significantly? I don't understand either in sufficient depth.

The extended Unicode is something I have only loaded to get full arabic support, as the usual utf8 has some gaps. However, it is discouraged, as you run into subtle conflicts with modern packages, see e.g. 
https://tex.stackexchange.com/questions/13067/utf8x-vs-utf8-inputenc

So it's probably not worth spending any significant time on it beyond parity with the regular utf8, since latexml should already offer a larger level of support as far as unicode is concerned.

dginev:
So I would suggest as the easiest patch to adjust `SetInputEncoding` in inputenc.sty.ltxml to map `utf8x` to `utf8`, and proceed as usual.

brucemiller:
On 05/03/2018 12:52 AM, Deyan Ginev wrote:
> So I would suggest as the easiest patch to adjust |SetInputEncoding| in 
> inputenc.sty.ltxml to map |utf8x| to |utf8|, and proceed as usual.

Rather than special case inputenc, I'd think
trying utf8x.def.ltxml as a copy of utf8
and see if that gets us anywhere.
(We can add further defns to utf8x if needed)

PR #986, opened 2018-05-03
add copy of utf8.def as utf8x.def.ltxml

dginev:
As discussed in #983 , adding a copy of the `utf8.def` binding for `utf8x.def`.

I tried the example from stackexchange (https://tex.stackexchange.com/a/184628/78008) and it seemed to look correct:

utf8x

example source:
```tex
\documentclass[welsh]{article}
\usepackage[T1]{fontenc}
\usepackage[utf8x]{inputenc}
\usepackage{babel}
\begin{document}

 Ââ Êê Îî Ôô Ŵŵ Ŷŷ Ïï

\end{document}
```

### DIFF
--- a/lib/LaTeXML/Package/utf8x.def.ltxml
+++ b/lib/LaTeXML/Package/utf8x.def.ltxml
@@ -1,0 +1,36 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# |  utf8x.def                                                          | #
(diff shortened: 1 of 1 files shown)


brucemiller:
Looks fine, except for the manifest :>

dginev:
I'm really starting to wisht the MANIFEST auto-updated on commit :>

dginev:
Done!

PR #1008, opened 2018-06-19
Add Dockerfile for use with Docker

tkw1536:
This commit adds a Dockerfile to LaTeXML to be able to use latexml easily within a Docker container.

### DIFF
--- a/release/Dockerfile
+++ b/release/Dockerfile
@@ -1,0 +1,67 @@
+
+# /=====================================================================\ #
+# | LaTeXML Dockerfile                                                  | #
+# | A Dockerfile to create a Docker Image with LaTeXML preinstalled.    | #
+# |=====================================================================| #
+# | Thanks to Tom Wiesing <tom.wiesing@gmail.com>                       | #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
+# |  United States Government & not subject to copyright in the US.     | #
+# |---------------------------------------------------------------------| #
+# | Bruce Miller <bruce.miller@nist.gov>                        #_#     | #
+# | http://dlmf.nist.gov/LaTeXML/                              (o o)    | #
+# \=========================================================ooo==U==ooo=/ #
+
+# This Dockerfile expects the root directory of LaTeXML as a build context. 
+# To achieve this run the following command from the root directory:
+#
+# > docker build -t latexml -f release/Dockerfile .
+
+# We start from alpine linux 3.7
+FROM alpine:3.7
+
+# Install the dependencies
+RUN apk add --no-cache \
+    db-dev \
+    gcc \
+    libc-dev \
+    libgcrypt \
+    libgcrypt-dev \
+    libxml2 \
+    libxml2-dev \
+    libxslt \
+    libxslt-dev \
+    make \
+    perl \
+    perl-dev \
+    perl-utils \
+    wget \
+    zlib \
+    zlib-dev
+
+# Install cpanminus from edge
+RUN apk add --no-cache -U --repository http://dl-3.alpinelinux.org/alpine/edge/community perl-app-cpanminus
+
+# Make a directory for latexml
+RUN mkdir -p /opt/latexml
+
+# Add all of the source files
+ADD bin/            /opt/latexml/bin
+#ADD doc/            /opt/latexml/doc/
+ADD lib/            /opt/latexml/lib 
+#ADD release/        /opt/latexml/release/
+ADD t/              /opt/latexml/t/
+ADD tools/          /opt/latexml/tools/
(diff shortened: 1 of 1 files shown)


dginev:
Does this need to be separately maintained every time the Makefile.PL dependencies change? Why not use `cpanm .` instead, which will dynamically ensure all Perl dependencies are installed, based on Makefile.PL?

tkw1536:
That sounds like a good idea, I will update to use cpanm. Although this will need maintenance when the system dependencies (e.g. libxml etc) change.

tkw1536:
Alright, I have updated the PR.

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    make \
+    perl \
+    perl-dev \
+    perl-utils \
+    wget \
+    zlib \
+    zlib-dev
+
+# Install cpanmnius
+RUN curl -L https://cpanmin.us | perl - App::cpanminus
```
dginev: same as in the other PR, you could add the `cpanminus` dependency to the `apk` section, instead of installing from curl, then it feels a bit more robust for a deployment setup.

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+ADD LICENSE         /opt/latexml/LICENSE
+ADD Makefile.PL     /opt/latexml/Makefile.PL
+#ADD MANIFEST        /opt/latexml/MANIFEST
+#ADD MANIFEST.SKIP   /opt/latexml/MANIFEST.SKIP
+#ADD manual.pdf      /opt/latexml/manual.pdf
+#ADD README.pod      /opt/README.pod
+
+# Installing  via cpanm
+WORKDIR /opt/latexml
+RUN cpanm .
```
dginev: this last line is very cool to read!

brucemiller:
One question: Is this file in the top-level LaTeXML directory? If so, is it needed to be there?
FWIW, I have a subdirectory "release", where I've parked whatever special build files I've collected (rpm specfile, macports Portfile, ...), so maybe that's a better place, to reduce top-level clutter?

dginev:
For the sake of argument, would you ever have latexml dockerized on its own, without a web server or similar deployment that hosts it? Since you could always make the dockerizable aspects a plugin to the LaTeXML project. Now that I've merged the Dockerfile PR for `LaTeXML::Plugin::ltxmojo`, you could argue that suffices for practical purposes?

I suspect Bruce doesn't want random files parasitizing the Perl library repository.

tkw1536:
I have moved the Dockerfile to the `release` sub-directory and added instructions on how to build it at the top of the file.

bfirsh:
Dockerfile looks good to me. One extra thing we have in [the Engrafo Dockerfile](https://github.com/arxiv-vanity/engrafo/blob/master/Dockerfile) is installing all of Texlive. I forget what breaks now if you don't have that installed, but stuff did break for us. `--includestyles` styles obviously won't work.

Texlive is _enormous_ though, so perhaps we could add that as another version of the image. `Dockerfile.texlive` or something. :)

@dginev Yes, it would be useful to use standalone. For example, this one-liner would allow you to convert documents without having to install anything on your machine:

```
$ docker run -v "$(pwd)":/workdir -w /workdir brucemiller/latexml latexmc example.tex
```

I don't have that strong feelings about it, but things in Docker-land tend to work better if the `Dockerfile` is in the root of the directory. It's also a neat little advert to people reading the code that "hey - you can run this in Docker!"

Once the Dockerfile is in, an automated build could be set up on Docker Hub so the Docker image gets build on every commit. Then, `docker run brucemiller/latexml` will work. I'm happy to help if you want to set this up.

dginev:
Thanks for chiming in @bfirsh ! Now that you added that one-liner, you've reminded me we should also update the respective manual pages if we decide to include the docker file -- after all the latexml users should have examples and guidance. And the PR looks twice as legitimate a need now cc @brucemiller

dginev:
As to texlive, I think I remember @tkw1536 was telling me the other day that he wanted to add that as a flag to the command and make it configurable? Definitely helpful to have the choice.

tkw1536:
I have updated the Dockerfile to allow to optionally include TeXLive using build arguments. This works as following. 

By default, no TeXLive is included. One would build the Docker Image like so:

```bash
docker build -t latexml -f release/Dockerfile .
```

To include TeXLive, the `WITH_TEXLIVE` build argument can be set to yes:

```bash
docker build -t latexml --build-arg WITH_TEXLIVE=yes -f release/Dockerfile .
```

In my tests, the docker image without TeXLIve is 215 MB in size, whereas the image with TeXLive is 2.81 GB in size. 

Sadly DockerHub (for automated builds that @bfirsh suggested above) does not seem to support build arguments, so it is non-trivial to automate builds for both variants of the image.

PR #1013, opened 2018-07-12
Cleanup and Refactor Locators

tkw1536:
This pull request refactors the behaviour of locators. In particular, it turns locators into a new class `LaTeXML::Common::Locator`, that is then only `ToString`ed when displayed to the user. 

The locator class also improves upon the old locators and is internally a quintuple consisting of
* the source filename, or `undef` in case of an anonymous string source, or `''` in case of a literal string source
* the starting line and column numbers of the locator
* the ending line and column numbers of the locator (or `undef` if not applicable)

This means that an instance of a locator class can refer either to a point, or an entire range in the source file. When requesting a Locator from the Mouth, the code tries to estimate the location and length of the last token, thereby returning a range. 

The new locator class also has a toAttribute method, that turns an instance of a locator into an XPointer and thereby supersedes #739.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -125,6 +125,7 @@ lib/LaTeXML/Common/Error.pm
 lib/LaTeXML/Common/Float.pm
 lib/LaTeXML/Common/Font.pm
 lib/LaTeXML/Common/Glue.pm
+lib/LaTeXML/Common/Locator.pm
 lib/LaTeXML/Common/Model.pm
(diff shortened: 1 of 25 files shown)


dginev:
Travis doesn't like your PR? Also, I am happy there is no new command-line option proposed, or I would need to link to the "ultimate PR response video", which Bruce countered me with some time ago.

Here it is for reference, and artistic enjoyment:
https://youtu.be/xHash5takWU

PR #1025, opened 2018-08-01
Add csquotes binding

tkw1536:
This PR adds bindings for the csquotes package. 

The bindings reads `csquotes.sty` and hooks into appropriate macros.

This PR also adds two test cases for this package, one in `t/structure/csquotes` (for mono-lingual csquotes) and one in `t/babel/csquotes` (for multi-lingual csquotes). 

The latter of these does not work in TeXLive 2013 (which is run on the Travis Tests), because of some issues with the old version of babel, so it is skipped until this can be fixed or the Travis Tests are updated accordingly.


t/81_babel.t
```diff
@@ -6,6 +6,7 @@ use LaTeXML::Util::Test;
 
 latexml_tests("t/babel",
 	      requires=>{'*'=>['babel.sty','babel.def'],
+		     csquotes=>['skipbecauseofoldtexliveintravis', 'csquotes.sty', 'frenchb.ldf', 'germanb.ldf'],
```
dginev: `skipbecauseofoldtexliveintravis` ?

brucemiller: yeah, a missing dependency.  Should be almost self-explanatory :>

PR #1027, opened 2018-08-02
Texlive more tests

tkw1536:
This commit completly overhauls the way Travis tests are done. Previously, these were used by manually installing appropriate versions of perl and tex into the testing environment and then running a script.

This commit introduces a scheme for building appropriate docker test runtime images. 
This allows for a lot more flexibility and reliable when running tests.
It also makes the tests run faster, as it is no longer neccessary to download, install and configure several dependencies.
These images are then pulled and used by Travis at testing time.

### DIFF
--- a/t/81_babel.t
+++ b/t/81_babel.t
@@ -6,7 +6,7 @@ use LaTeXML::Util::Test;
 
 latexml_tests("t/babel",
 	      requires=>{'*'=>['babel.sty','babel.def'],
-		     csquotes=>['skipbecauseofoldtexliveintravis', 'csquotes.sty', 'frenchb.ldf', 'germanb.ldf'],
+		     csquotes=>['csquotes.sty', 'frenchb.ldf', 'germanb.ldf'],
 			 numprints=>'numprint.sty',
 			 german=>'germanb.ldf',
 			 greek=>['greek.ldf','lgrenc.def'],
--- a/tools/travis/README.md
+++ b/tools/travis/README.md
@@ -1,0 +1,82 @@
+# Travis Test README
+
+This directory contains tools to prepare and use inside the Travis Tests. 
+
+The Travis Tests are intended to test the behaviour of LaTeXML under different environments using multiple TexLive and Perl versions. 
+
+The different environments can be split into two main categories:
+
+* those with a TeXLive installation
+* those without a TeXLive installation
+
+Both scenarios internally run the same test suite (`make test`), however because of the way it is set up the latter one will skip a number of tests. 
+
+## The Testing Docker Images
+
+The different testing environments are defined in `.travis.yml` using the `env` key. 
+Each testing environment corresponds to a single docker image. 
+
+Each docker image is based on `Ubuntu Xenial`. 
+These come preinstalled with the respective perl version (by using `perlbrew`) and LaTeXML versions (via the package manager). 
+
+At testing time travis then:
+
+* pulls each of the docker images
+* mounts the version of LaTeXML to be tested using a volume and finally
+* executes the standard test suite inside the resulting container
+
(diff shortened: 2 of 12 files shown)


t/81_babel.t
```diff
@@ -6,7 +6,7 @@ use LaTeXML::Util::Test;
 
 latexml_tests("t/babel",
 	      requires=>{'*'=>['babel.sty','babel.def'],
-		     csquotes=>['skipbecauseofoldtexliveintravis', 'csquotes.sty', 'frenchb.ldf', 'germanb.ldf'],
```
dginev: i like this change :joy:

tools/travis/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+For example, to generate a test image for perl `5.22.4` and `texlive-2015` it can be invoked like so:
+
+    src/build-test-image.sh 5.22.4 texlive-2015
+
+This will build the image `latexml/latexml-test-runtime:5.22.4_texlive-2015`. 
+
+
+To build all test images required by the tests, the `src/build-all.sh` can be used. 
+After building, this script will upload each image to [latexml/latexml-test-runtime](https://hub.docker.com/r/latexml/latexml-test-runtime/) on DockerHub.
```
dginev: cool, :+1: for the DockerHub approach here, very manageable

tools/travis/README.md
```diff
@@ -0,0 +1,82 @@
...
+If instead you would liek to run these command manually (e.g. to debug a failed test), run:
+
+    docker run --rm -t -i -v /path/to/latexml/repository:/root/latexml latexml/latexml-test-runtime:$PERL_$TEX /bin/bash
+
+This will start a bash shell inside the Docker container. 
+To load the approrpiate version from perlbrew, it is then required to load:
+
+    source /usr/local/perlbrew/etc/bashrc
+
+Afterwards, you can run any test command manually.
```
dginev: Great Readme, makes perfect sense to me.

tools/travis/scripts/build-test-image.sh
```diff
@@ -0,0 +1,94 @@
...
+ENV PATH /usr/local/perlbrew/bin:$PATH
+ENV PERLBREW_PATH /usr/local/perlbrew/bin
+RUN perlbrew install-cpanm && \
+    perlbrew info
+
+# set perl version
+ENV PERL $PERL
+
+# run the installation of the given version of perl
+RUN /bin/bash -c "source /usr/local/perlbrew/etc/bashrc; perlbrew --notest install $PERL; perlbrew clean; perlbrew switch $PERL"
```
dginev: Hm, `perlbrew --notest` is a wise approach to speed things along, I am just jotting down that I have had perlbrew's Perl tests fail on old OS-es (Hulk in particular), so it's not always foolproof... but if the tests pass after, it must have worked well anyway?

tools/travis/src/installdeps.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Update apt
+apt-get update
+
+# Install system dependencies
+apt-get install -y \
+    libdb-dev \
+    libxml2-dev \
+    libxslt1-dev \
+    zlib1g-dev
```
dginev: One tricky bit is to remember to keep this file up-to-date as we keep developing if dependencies are added. I guess the good news is that the travis tests will fail when it is out of date, so we can't quite forget to :+1:

tools/travis/scripts/build-test-image.sh
tkw1536: I mostly added it to speed up the repetitive docker builds (when building all runtime images at once), and it has worked so far.

PR #1029, opened 2018-08-06
pass @INC to latexmlc invocation in tests

dginev:
Ideally this helps with CPANTesters, should be just redundant information for a regular Perl installation.

One key question is whether a call to `perl` will use the correct version of Perl on CPANTesters. If this fails, there seem to be alternatives that can be tried without a system call, such as:
```
{
    local @ARGV = qw<param1 param2 param3>;
    do "$latexmlc";
}

```

but that would need a slightly larger rewrite of the Test.pm setup


brucemiller:
Sounds good, let's give it a try

PR #1037, opened 2018-08-12
non-zero exit code on fatal conversions for latexml+latexmlc

### LINKED ISSUE #1036: Exit code 0 returned on fatal error
Minimal test case, using latest master:

```
$ cat > broken.tex
\documentclass{article}
\end{document}
^D
$ latexml broken.tex
...
Conversion complete: 1 fatal error.
$ echo $?
0
```

Also seems to affect `latexmlc`.

I'm not sure if this is a regression. This doesn't actually affect Engrafo, because it then fails to open a non-existent `index.html`, but I could have sworn it worked at some point...

dginev:
Fixes #1036

### DIFF
--- a/lib/LaTeXML/Util/Test.pm
+++ b/lib/LaTeXML/Util/Test.pm
@@ -235,7 +235,11 @@ sub daemon_ok {
   $invocation .= " 2>$localname.test.status ";
   if (!$generate) {
     pathname_chdir($dir);
-    is(system($invocation), 0, "latexmlc invocation for test $localname: $invocation yeilded $!");
+    my $exit_code = system($invocation);
+    if ($exit_code != 0) {
+      $exit_code = $exit_code >> 8;
+    }
+    is($exit_code, 0, "latexmlc invocation for test $localname: $invocation yeilded $!");
     pathname_chdir($current_dir);
     # Compare the just generated $base.test.xml to the previous $base.xml
     if (my $teststrings = process_xmlfile("$base.test.xml", $base)) {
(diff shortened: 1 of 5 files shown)


lib/LaTeXML/Util/Test.pm
```diff
@@ -235,7 +235,11 @@ sub daemon_ok {
   $invocation .= " 2>$localname.test.status ";
   if (!$generate) {
     pathname_chdir($dir);
-    is(system($invocation), 0, "latexmlc invocation for test $localname: $invocation yeilded $!");
+    my $exit_code = system($invocation);
+    if ($exit_code != 0) {
+      $exit_code = $exit_code >> 8;
```
dginev: This is an interesting change I noticed was needed - sometimes the return code on successful runs was shown as 255 or 256 instead of 0 and 1, also something we have seen in some of the CPANTester runs, and I believe it has something to do with the inner mechanics of how `waitpid` passes children exit values (as for the `system` call in the daemon tests). A link I half-understand:

https://stackoverflow.com/a/3659648

In any case, the only case this affects in this part of the code is recognizing 255 as a 0 exit, so that the test can succeed in that setup.

dginev:
I also added a small test with Ben's example, to guarantee we keep supporting the behavior.

brucemiller:
I'm guessing that this patch is responsible for the extra messages during make test:
```
t/95_complex_config.t .. ok   
t/96_fatal.t ........... 1 fatal error
Error! Did not write file broken.html
t/96_fatal.t ........... ok   
All tests successful.
Files=28, Tests=333, 133 wallclock secs ( 0.06 usr  0.01 sys + 129.77 cusr  2.89 csys = 132.73 CPU)
Result: PASS
```
It does pass, but the "fatal" is somewhat disconcerting.  Is there a way to avoid it?

dginev:
Sure, hadn't noticed that. See #1051 for a quick fix

PR #1042, opened 2018-08-17
Make abstract, references, etc <h2>

bfirsh:
Abstract, references, acknowledgements, keywords, classification, and TOC headings were all previously `<h6>`. I think they all appear a level directly under the title, which is `<h1>`. So, to be semantically correct, they should be `<h2>`.

I imagine this was originally done to make them smaller. This should be styled with CSS instead of using incorrect heading levels. I have converted some documents with the default stylesheet and the abstract and references headings still look fine.

### DIFF
--- a/lib/LaTeXML/resources/CSS/ltx-apj.css
+++ b/lib/LaTeXML/resources/CSS/ltx-apj.css
@@ -6,12 +6,12 @@
 .ltx_role_author .ltx_personname {
     font-family:serif; font-weight:normal; font-variant:small-caps; font-size: 100%; }
 .ltx_abstract { margin-left:4em; margin-right:4em; }
-.ltx_abstract h6 {
+.ltx_abstract h2 {
     text-transform:uppercase; font-family:serif; font-weight:normal; font-size: 100%; 
     display:block; text-align:center; margin-bottom:0.1ex; }
 .ltx_keywords    {
     margin-top:0.5em; margin-left:4em; margin-right:4em; }
-.ltx_keywords h6 {
+.ltx_keywords h2 {
     font-weight:normal; font-style:italic; }
 
 .ltx_title_bibliography,
@@ -31,4 +31,3 @@
     font-family:serif; font-variant:small-caps;
     display:block; text-align:center; }
 .ltx_table .ltx_caption { font-variant:small-caps; } /* ! */
-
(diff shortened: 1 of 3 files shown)


bfirsh:
Ah, I see! Yes, I do understand your reasoning.

I suppose there are two use-cases we could use to approach HTML semantics:

1. The most important: accessibility. I am not an expert on screen readers, but I presume special meaning is given to an `<hN>` element. What element it is doesn't really matter, I am guessing, so the `<h6>` works fine in this case. Just using a `<span>` would be worse for this use-case, I think. [MDN has a pretty good section explaining this.](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML#Text_content)

2. Another use-case is some machine that needs to understand the semantic content of the article. For headings, that would be a browser generating an outline of a page, say.

    For this use-case, [MDN again has some pretty good resources explaining how outlines are generated](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines#Implicit_sectioning). From what I can gather from that, the heading numbers don't really matter, because LaTeXML uses `<section>` everywhere. The abstract, keywords, etc headings aren't in a section, but presumably they fall in some implicit top-level section.

    All it has to say about not using adjacent numbers is this:

    > Note that the rank of the heading element (in the example `<h1>` for the first top-level section, `<h2>` for the subsection and `<h3>` for the second top-level section) is not important. (Any rank can be used as the heading of an explicitly-defined section, although this practice is not recommended.)

So, I don't have any strong feelings really. How it works at the moment isn't breaking anything. I understand your reasoning, which is entirely reasonable. The only thing that leans me towards changing it is MDN saying "this practice is not recommended".

bfirsh:
I’m pretty sure some kind of heading element is better than none, otherwise a screen reader would just smush the words together as if it were a run on sentence. And document outlines won’t recognise it, I think.

Tell you what - I’m going to do some proper testing of the documents with screen readers and so on, so how about I do that and then report back whether this is worth doing or not. Feel free to close in the meantime. :)

brucemiller:
OK, so I guess we'll keep things as-is for now, but I'm certainly open to alternatives if we can figure out what would be more accessible.  Thanks for keeping accessibility in mind!

Issue #1035, opened 2018-08-10
"\uppercase{äöü}" results in "äöü" instead of "ÄÖÜ"


siegfriedzoetzsche:
It seems like `lookupUCcode()` in Core/State.pm does not return a code for "ä", "ö" or "ü" and so `ucToken()` in Package/TeX.pool.ltxml just maps "ä", "ö" and "ü" to "ä", "ö" and "ü" respectively.

#### `example.tex`
```tex
\documentclass{article}
\usepackage[utf8]{inputenc}
\begin{document}
\uppercase{(äöü)(ÄÖÜ)}
\end{document}
```

#### output of `latexml --strict --nocomments --destination latexml.xml example.tex`
```xml
[...]
<p>(äöü)(ÄÖÜ)</p>
[...]
```

dginev:
This is simply missing from the `utf8.def.ltxml` binding. Looking into it.

PR #1049, opened 2018-08-31
Add unicode uccode/lccode pairs for diacritic letters

dginev:
- good catch in #1035 with the umlaut letters
- turns out utf8.def introduces the uccode/lccode pairs for these characters
- hence extend utf8.def.ltxml with the definitions for 
- all [combining diacritical marks](https://www.fileformat.info/info/unicode/block/combining_diacritical_marks/list.htm) which have a single unicode char in [Normalization Form C](https://perldoc.perl.org/Unicode/Normalize.html#Normalization-Forms) that is known to Perl
- I precomputed the map into an array, so that the load time overhead for utf8.def.ltxml is minimal
- Added test for issue example. Works! (and surprisingly pdflatex fails on the \lowercase ? latexml passes, and i think it's worth keeping the upgrade)


dginev:
As I suspected, `xelatex` generated the correct PDF - much more satisfying, I added a commit with it. pdflatex disappoints yet again with its unpredictable unicode bugs.

dginev:
I have no idea how you manage to remember every time :> Added!

brucemiller:
Cool! Your fix is better than the one I was fumbling with...

PR #1050, opened 2018-08-31
Include fixltx2e by default

bfirsh:
It is now in LaTeX by default:

https://www.latex-project.org/news/latex2e-news/ltnews22.pdf


brucemiller:
Looks good; Thanks!!

PR #1051, opened 2018-09-04
Polite fatal failing test

dginev:


dginev:
It is interesting this has never been an issue, but I sort of recall certain platforms do not support the `2>` redirect. I already am using it for the latexmlc calls in Test.pm, and it's been there since the start of that feature being in master.

Happy to wait for an actual report complaining about that before I a refactoring route though.

PR #1053, opened 2018-09-06
use robust text_content_aux sub instead of ->textContent

dginev:
Addresses the new perl dies introduced by my slightly careless use of `->textContent` on what end up a wide range of different Perl objects. 

The existing `text_content_aux` method is exactly what one would like instead, so this should avoid any fragility while keeping the clean fallback patch.


brucemiller:
solved other ways... but thanks!

Issue #1066, opened 2018-10-23
Make imagemagick DPI configurable


bfirsh:
Most of the images that LaTeXML outputs look something like this on a large, high resolution screen:

<img width="1059" alt="screenshot 2018-10-23 12 21 36" src="https://user-images.githubusercontent.com/40906/47339947-95ca6600-d6be-11e8-9fb9-0e498c808577.png">

(I am assuming you are viewing this issue on a large, high resolution screen. ;)

This one is barely readable, but there are some images which are completely unreadable. Particularly graphs and so on.

The ideal fix would be #902, but in the meantime a quick fix would be to let us control the DPI used in `Post/Graphics.pm` and Post/LaTeXImages.pm` and so on. I would set it to something high at the expense of filesize, because I assume the images may be viewed on large screens with high DPI, but I can't predict the exact size they will be displayed at.

dginev:
Now that we can vary the DPI, which helps in a whole lot of cases for PDF source images, I should add a note that it won't entirely fix the specific document with which the issue was opened. 

The images included in the paper in question [1803.07574](https://arxiv.org/abs/1803.07574), use further graphicx manipulations:
```tex
\includegraphics[clip,trim=2.3cm 1.7cm 2.7cm 1.8cm,scale=0.4]{Graphs/1000_iSNR_5_series_89_lambda_10_3.pdf}
```

Studying the document, images which only use the `trim` and `scale` directives immediately benefit from boosting the DPI and come out looking great pretty quickly. But it appears that all cases which use `clip` do not -- they remain at very low resolutions, as if the DPI setting was ignored. Writing this as a report for now, will take a look if I can also add a quick patch.

Issue #979, opened 2018-04-29
HTML subfigures via flexbox rather than tables?


dginev:
I was inspecting an arXiv document (1802.06832) which has 4 subfigures a)-d) in its Results section and noticed they don't reflow in the HTML.

They are currently marked as 4 `td` table cells with class `ltx_subfigure` in the same `tr` row. 

The fixed single-row display looks quite bad even on large displays, as it is significantly wider than the main article width. And on small displays you end up with a huge horizontal scroll, which is just as bad. 

As I recently did some flexbox work with the moderncv binding, I'm wondering if @brucemiller would see that as a better mechanism for presenting the subfigures. That would give us decent reflow, and we can contain the max-width to a reasonable responsive width.

dginev:
Here is a screenshot from the arXiv PDF, as viewed in my 27'' screen:
![pdf](https://user-images.githubusercontent.com/348975/39405334-50097950-4bac-11e8-9261-56f178deb883.png)

and the Firefox rendering of the subfigure table by the current latexml master (with the etoolbox PR):
![firefox](https://user-images.githubusercontent.com/348975/39405346-738cfd8e-4bac-11e8-8bbd-0ddaa34e56ae.png)

They manage to overflow even my 27'' screen :wilted_flower:

dginev:
TeX snippet:
```tex
\begin{figure*}[t!]
	\centering
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{bps_word_error}
    \caption{Word error as bits per sentence changes for 0.05 bit erasure probability. \label{fig:bps_error}}
    \end{subfigure}%
    ~
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{bdr_word_error}
    \caption{Word error as erasure or bit-drop rate increases for 400 bit encoding. \label{fig:bdr_error}}
    \end{subfigure}%
    
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{len_sentence_word_error}
    \caption{Effect of sentences of different sizes with 400 bit encoding, 0.05 drop rate. \label{fig:sen_len_error}}
    \end{subfigure}%
    ~
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{example_embedding}
[...]

PR #1067, opened 2018-10-24
[Experiment] Convert PDF graphics to SVG with dvisvgm

### LINKED ISSUE #902: PDF, EPS, etc graphics should be converted to scalable SVGs
When using `--graphicsmap=pdf.svg`, it converts the graphic to a SVG with a raster rendering of the PDF. I would expect it to convert it to a vector graphic. The same presumably applies for EPS, AI, and PS.

For [Engrafo](https://github.com/arxiv-vanity/engrafo), we had success using [pdf2svg](https://github.com/dawbarton/pdf2svg). Presumably the same result can be achieved by piping it through the same LaTeX rendering system that renders math/tikzpicture as SVG.

bfirsh:
This is the start of an attempt to fix #902. It seems like the main problem in that issue was adding an extra dependency. Turns out dvisvgm, which is already used by LaTeXML to convert pictures into SVGs, can also convert PDF and EPS files to SVG!

This makes the figures in [this paper](https://arxiv.org/abs/1803.07574) actually readable:

screenshot 2018-10-24 18 01 50

(Compare with example in #1066.)

An aside: The built-in LaTeXML CSS doesn't do a good job of laying out the subfigures, but I think the root cause is #979 so it's probably worth implementing that instead of trying to get the flawed table layout working.

Anyway! I'm just throwing this out here as a discussion-starter. Before I go any further with implementation, any objections the general idea?

If there are no objections, how do you think it should be implemented? `Graphics.pm` revolves pretty heavily around imagemagick, so I'm not sure the best way to add this without making a mess.

### DIFF
--- a/lib/LaTeXML/Post/Graphics.pm
+++ b/lib/LaTeXML/Post/Graphics.pm
@@ -59,9 +59,8 @@ sub new {
     ai => { destination_type => 'png',
       transparent => 1,
       prescale => 1, ncolors => '400%', quality => 90, unit => 'point' },
-    pdf => { destination_type => 'png',
-      transparent => 1,
-      prescale => 1, ncolors => '400%', quality => 90, unit => 'point' },
+    pdf => { destination_type => 'svg',
+      raster => 0, desirability => 11 },
     ps => { destination_type => 'png', transparent => 1,
       prescale => 1, ncolors => '400%', quality => 90, unit => 'point' },
     eps => { destination_type => 'png', transparent => 1,
@@ -212,6 +211,20 @@ sub transformGraphic {
         if (pathname_timestamp($source) <= pathname_timestamp($absdest)) {
           NoteProgressDetailed(" [Reuse $cached @ " . ($width || '?') . " x " . ($height || '?') . "]");
           return ($cached, $width, $height); } } } }
+  
+  # HACK: convert PDFs with dvisvgm
+  if ($srctype eq "pdf" && $type eq "svg") {
+    $dest = $self->generateResourcePathname($doc, $node, $source, $type);
+    my $absdest = $doc->checkDestination($dest);
+    my $dvicommand = "dvisvgm --stdout --pdf '$source' > '$absdest'";
+    my $dvierr     = system($dvicommand);
+    if ($dvierr != 0) {
+      Error('shell', 'dvisvgm', undef,
+        "Shell command '$dvicommand' (for dvi conversion) failed",
+        "Response was: $!");
+      return; }
+    return ($dest, undef, undef); }
+
   # Trivial scaling case: Use original image with (at most) different width & height.
   my $triv_scaling = $$self{trivial_scaling} && ($type eq $srctype)
     && image_graphicx_is_trivial($transform);


brucemiller:
Wow, Cool! I didn't notice divsvgm had that option!

Yeah, dvisvgm is used, currently, but with reservations; I'm not too fond of the way it deals with fonts, so the math svgs come out really fat.  Probably should look for a solution with web fonts, or something.  But that issue probably doesn't apply here (or is unavoidable anyway).

One limitation with your experiment is that it omits the various scaling, clipping, rotation options.  Of course, with SVG it should be trivial, if you're able to do a bit of surgery on the XML. Still, might be better than the current situation?

When I saw #1066, I started wondering whether inkscape had a usable command line interface, and could be used for this (and some other cases).  inkscape & dvisvgm are sorta "soft" dependencies; they can be used if available, so long as we have a fallback when they're not.

brucemiller:
Oh, and I commend your bravery in looking at Graphics.pm; that code ain't pretty!  (but in my defence, the task of gluing `\includegraphics` (and it's crazy options) to Image::Magick (and it's crazy options) is not so easy).

bfirsh:
Seems like dvisvgm has a `--transform` option that all those various options could be translated to: https://dvisvgm.de/Manpage/

And indeed -- the SVGs are large. Each of those graphs turns into a ~200KB SVG from a ~20KB PDF. Not unreasonable though -- particularly considering the current ~50KB PNGs are almost useless.

I haven't investigated the font options in dvisvgm yet, but perhaps there is some clever way of using a shared web font...

brucemiller:
I've also done a bit of experimentation, but it's sort of a good-news/bad-news story.  I had bad luck with any versions of pdfsvgm dropping parts of the images in the arXiv paper you cited. And also the transformation options are insufficient to match graphicx. However, inkscape does a good, but slower, job converting. 

See the branch pdfsvg (can I link to that?); I've set up the possibility to get either of those to transform pdf to svg, and then wrote a small transformer that wraps the appropriate svg:g elements to carry out the transform/clip/etc. Seems to work ok, within limits.

None of this is the default (even in the branch); you've got to add the commandline option ```--graphicsmap=pdf.svg```
to activate it. Unfortunately that *replaces* rather than adds to the graphics map, so *no* other graphics conversions occur! Probably we'll want to adjust the option to support something like ```--graphicsmap=+pdf.svg```? Not done yet.
And if you set the environment variable ```LATEXML_INKSCAPE``` to point to inkscape, it will use inkscape instead of pdfsvgm.   All round, needs a better configuration setup.

dginev:
You can always copy the link and write a markdown link: [pdfsvg branch](https://github.com/brucemiller/LaTeXML/tree/pdfsvg)

dginev:
Just to have a clear comparison in this PR, this is how these images looked when I last converted them with the version of latexml from January 2021:
   1    | 2
:---- | :----
image | image

They're really not of sufficient quality to be left as-is. So this is very much a problem worth solving.

dginev:
I'd rather we close here for now (without forgetting/dismissing the proposal) and keep tracking this at #902 . I still think inkscape needs further investigation as a possible delegate, since it tries to preserve more of the textual content unharmed. Here are the respective examples from the document here:

```
$ time inkscape 3_bestlambda_1000000.pdf -o test1.svg
real	0m0.403s

$ time inkscape 2_bestlambda_379269.pdf -o test2.svg
real	0m0.406s
```

Results: https://gist.github.com/dginev/b0465d2ebcf972d6808537643dde45bd

PR #1069, opened 2018-10-26
Allow \newpage breaks in epub targets

### LINKED ISSUE #1068: Page break compatibility?
Hi there, this is a great program!

In case the authors are not aware, LaTeXML can also be used as part of the process to produce ebooks! Since it produces HTML, it is just a matter of converting the HTML into EPUB or other format using tools like Calibre. 

Anyway, my question relates to commands like `\pagebreak` and `\newpage`. It seems LaTeXML takes no notice of these commands, as they do not produce HTML commands related to them (no surprise, since a page break in an HTML document makes little sense!). An example of such command would be the one presented in [this page](https://blog.epubandebookhelp.com/2011/12/21/epub-formatting-how-to-add-a-page-break-to-an-epub-document-using-html-or-css/). What LaTeXML can do is to produce separate files in terms of sections (e.g. based on `\chapter{}`). However, when producing an ebook, you just would like to add a `\pagebreak` command rather than adding a section for the title page, copyright page, acknowledgment and other minor content. Adding a section introduces them into the table of contents of the ebook, which is not desired.

At the moment I'm using sections for everything and modifying the TOC manually in Calibre, so this is not a big issue. But I wonder if there is a plan for LaTeXML to allow for page breaks (perhaps as suggested in the linked page)? 

Best regards.

dginev:
Fixes #1068 

Styles the page break elements produced already by latexml only for the ePub targets, via a dedicated CSS file.

ePub once again strains the customization paradigms we have in place, and while I did the "most direct" implementation, the approach is closer to the joint latexmlc executable, than latexmlpost, as it uses the `RequireResource` binding routine to designate the need for copying over the new custom CSS file.

Switching the technique to something compatible with latexmlpost would require replicating the `RequireResource` functionality into the post-processing flow... which is redundant for its own reasons.

In any case, this PR is fully functional and creates ePub documents with properly enforced breaks.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -285,6 +285,7 @@ lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
 lib/LaTeXML/resources/XSLT/LaTeXML-xhtml.xsl
 lib/LaTeXML/resources/XSLT/LaTeXML-tei.xsl
 lib/LaTeXML/resources/CSS/LaTeXML-blue.css
+lib/LaTeXML/resources/CSS/LaTeXML-epub.css
 lib/LaTeXML/resources/CSS/LaTeXML-marginpar.css
 lib/LaTeXML/resources/CSS/LaTeXML-navbar-left.css
 lib/LaTeXML/resources/CSS/LaTeXML-navbar-right.css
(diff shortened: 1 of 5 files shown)


dginev:
@brucemiller I switched to an opt-in model here, and showed the use in the test - though we ought to add some ePub generation documentation as well, maybe in the [Demos](https://github.com/brucemiller/LaTeXML/wiki/Demos) wiki page... 

Default remains unchanged.

PR #1072, opened 2018-11-09
[minor] perl-critic over LaTeXML.pm

dginev:
Unsure if this PR is worth the trouble...

I got a scary `severity: 5` error from perl critic when i tried to change LaTeXML.pm and decided to at least fix that. The downside is that it requires a new dependency `IO::Interactive`, and lead to reformatting a couple more files.

Upside is that Critic is happier. Should be functionally equivalent in all usual cases I can think of.


dginev:
@brucemiller I backed out of the unnecessary (albeit critic-suggested) new dependency, the rest of the changes are quite minor. Maybe mergeable in this shape?

Makefile.PL
```diff
@@ -32,7 +32,7 @@ our $TEXMF;
 my ($texmfspec) = grep { /^TEXMF/ } @ARGV;
 if ($texmfspec && $texmfspec =~ /^TEXMF\s*=(.*)$/) {
   $TEXMF = $1;
-  @ARGV = grep { $_ ne $texmfspec } @ARGV;    # Remove so MakeMaker doesn't fret.
+  local @ARGV = grep { $_ ne $texmfspec } @ARGV;    # Remove so MakeMaker doesn't fret.
```
brucemiller: doesn't local here restrict the effect to the body of the if? If so, it's pretty much a NOOP (& doesn't have desired effect)

dginev: Indeed, this was a harmful change, apologies. It bugs me that removing it once again makes any change to Makefile.PL requires `--no-verify` on commit, but I shouldn't have broken it. Critic's insisting messages pushed me too far.

brucemiller:
I guess this all looks right now. Lotsa gratuitous spacing changes, though.  But Thanks!

dginev:
Please redirect all whitespace gratitude to perltidy, I take no responsibility for its decisions :>

PR #1078, opened 2018-11-20
Add support for quantumarticle.cls

bfirsh:
https://quantum-journal.org/instructions/authors/
https://github.com/quantum-journal/quantum-journal

I think this is everything needed from the .cls file. I have tested with the template and a real paper from the journal and both render without errors.

I wanted to include the template as a test, but I don't think I can because it is licensed under LPPL. Is that right? I don't think I can even write my own test because the .cls file is not in texlive, so I'd have to commit it to the repo to get the PDF to render.

ref https://github.com/arxiv-vanity/engrafo/issues/316

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -574,6 +574,7 @@ lib/LaTeXML/Package/pst-node.sty.ltxml
 lib/LaTeXML/Package/pstricks.sty.ltxml
 lib/LaTeXML/Package/pxfonts.sty.ltxml
 lib/LaTeXML/Package/pzd.fontmap.ltxml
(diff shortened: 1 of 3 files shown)


brucemiller:
Looks good! The only problem is the annoying requirement to add the new binding to the MANIFEST file.

dginev:
I'm _this_ close to writing a manifest bot for the repo PRs. It is quite easy to do actually ...

bfirsh:
Ah! Didn't realise that was needed, sorry. Added it.

brucemiller:
Yay; Thanks!!

PR #1079, opened 2018-11-20
Baseline styling for raw .cls loads

dginev:
Simple upgrade to solve the scenario described in https://github.com/arxiv-vanity/engrafo/issues/270

In essence, when loading a `.cls` file as raw TeX, also define a baseline CSS, similarly to the OmniBus.cls.ltxml approach.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1916,11 +1916,7 @@ sub loadTeXDefinitions {
     # It's probably even the ltxml version is asking for it!!
     # Of course, now it will be marked and wont get reloaded!
     return if LookupValue($request . '_loaded');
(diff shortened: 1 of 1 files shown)


bfirsh:
Oh _nice_! I didn't look into it much -- I didn't realise it was just a styling issue.

A few thoughts:

- [quantum-template.tex](https://github.com/quantum-journal/quantum-journal/blob/master/quantum-template.tex) seems to render with more errors when using `--includestyles` compared to #1078, so perhaps that is still needed as a stopgap. Also, the author affiliations, emails, etc, don't display.
- [Another paper I was testing] has a fatal error with `--includestyles`

brucemiller:
Looks pretty reasonable. Thanks!

PR #1081, opened 2018-11-24
More careful location range heuristic

dginev:
* standardized variable names to Tom's from/to pairs, which are more readable
 * removed the rather bizarre column arithmetic in the locator heuristic
 * replaced with a simple "zero or here" rule

More work can be done, but this provides slightly more meaningful locator ranges for the stex tests.


brucemiller:
Looks reasonable; Thanks!

PR #1084, opened 2018-11-27
Add pre-commit check for MANIFEST membership

dginev:
Example effect, when adding a new binding and forgetting to update the MANIFEST:

```
Some files are missing from MANIFEST, please add them: 
	lib/LaTeXML/Package/new.sty.ltxml;
COMMIT ABORTED
```

Hopefully can make the @brucemiller 's reviewer life a little more pleasant!

### DIFF
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -39,7 +39,23 @@ if (my @modified_files =
       push(@failed_files, $modified_file);    # Note the original file name
       $exit_code = $code; } }
 
-  print STDERR "Some files had issues: " . join(',', @failed_files)
-    . "; To correct these, run:\n tools/latexmllint <files>.\n"
-    . colored("COMMIT ABORTED", 'red') . "\n" if @failed_files;
+  if (@failed_files) {
+    print STDERR "Some files had issues: " . join(',', @failed_files)
+      . "; To correct these, run:\n tools/latexmllint <files>.\n"
+      . colored("COMMIT ABORTED", 'red') . "\n"; }
+  else {
+    # Check if files are missing from the manifest
+    my %manifest = ();
+    open(my $fh, "<", "MANIFEST") or die "Missing MANIFEST file?\n";
+    while (<$fh>) {
+      next if /^[#]/;
+      chomp;
+      $manifest{$_} = 1; }
+    close($fh);
+    # Not in MANIFEST, (or in tools)
+    if (my @missing_files = grep { !($_ =~ /^tools/ || $manifest{$_}) } @modified_files) {
+      print STDERR "Some files are missing from MANIFEST, please add them: \n\t"
+        . join(',\n\t', @missing_files) . ";\n"
+        . colored("COMMIT ABORTED", 'red') . "\n";
+      exit 1; } }
   exit $exit_code; }
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -38,24 +38,25 @@ if (my @modified_files =
     if ($code) {
       push(@failed_files, $modified_file);    # Note the original file name
       $exit_code = $code; } }
-
   if (@failed_files) {
     print STDERR "Some files had issues: " . join(',', @failed_files)
-      . "; To correct these, run:\n tools/latexmllint <files>.\n"
-      . colored("COMMIT ABORTED", 'red') . "\n"; }
-  else {
-    # Check if files are missing from the manifest
-    my %manifest = ();
-    open(my $fh, "<", "MANIFEST") or die "Missing MANIFEST file?\n";
-    while (<$fh>) {
-      next if /^[#]/;
-      chomp;
-      $manifest{$_} = 1; }
-    close($fh);
-    # Not in MANIFEST, (or in tools)
-    if (my @missing_files = grep { !($_ =~ /^tools/ || $manifest{$_}) } @modified_files) {
-      print STDERR "Some files are missing from MANIFEST, please add them: \n\t"
-        . join(',\n\t', @missing_files) . ";\n"
-        . colored("COMMIT ABORTED", 'red') . "\n";
-      exit 1; } }
+      . "; To correct these, run:\n tools/latexmllint <files>.\n"; }
+
+  # Also, Check if files are missing from the manifest
+  my %manifest = ();
+  open(my $fh, "<", "MANIFEST") or die "Missing MANIFEST file?\n";
+  while (<$fh>) {
+    next if /^[#]/;
+    chomp;
+    $manifest{$_} = 1; }
+  close($fh);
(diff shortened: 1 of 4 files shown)


dginev:
I was considering making this a GitHub PR bot, but realized the "change suggestions" flow isn't that pleasant, at best the bot can leave a comment, and that's already too late and inconvenient.

Since I have the pre-commit hook always on when committing nowadays, adding the manifest check in there is very natural for the dev flow, and saves the public embarrassment aspect of it all :>

bfirsh:
Might be worth running this check as part of Travis too, if for whatever reason the pre-commit hook doesn't run. That would have made me fix it in #1078.

dginev:
Oh, that's a curious idea. I could move it under Util::Test and add a simple test entry for the check.

tools/pre-commit
```diff
@@ -39,7 +39,23 @@ if (my @modified_files =
...
       $exit_code = $code; } }
 
-  print STDERR "Some files had issues: " . join(',', @failed_files)
-    . "; To correct these, run:\n tools/latexmllint <files>.\n"
-    . colored("COMMIT ABORTED", 'red') . "\n" if @failed_files;
+  if (@failed_files) {
+    print STDERR "Some files had issues: " . join(',', @failed_files)
+      . "; To correct these, run:\n tools/latexmllint <files>.\n"
+      . colored("COMMIT ABORTED", 'red') . "\n"; }
+  else {
```
brucemiller: Probably better not to else here, but print any manifest messages in addition to any lint messages.

tools/pre-commit
```diff
@@ -39,7 +39,23 @@ if (my @modified_files =
...
+    # Check if files are missing from the manifest
+    my %manifest = ();
+    open(my $fh, "<", "MANIFEST") or die "Missing MANIFEST file?\n";
+    while (<$fh>) {
+      next if /^[#]/;
+      chomp;
+      $manifest{$_} = 1; }
+    close($fh);
+    # Not in MANIFEST, (or in tools)
+    if (my @missing_files = grep { !($_ =~ /^tools/ || $manifest{$_}) } @modified_files) {
```
brucemiller: As I was pondering whether this should be white or black listing, I noticed that Perl already has such a mechanism: MANIFEST.SKIP, which is a sequence of regexps for files that needn't be in the MANIFEST.

dginev:
@brucemiller if you merge this PR next, I can quickly run through all my other open PRs and ensure new files are properly added to the manifest. Cool incentive? :>

brucemiller:
Hmm...tempting offer!

But I'm wondering; shouldn't this be more whitelist than blacklist? Ie. except for a few top-level files, only files in ```t``` and ```lib``` should appear in the MANIFEST (if committed, of course). Isn't the MANIFEST primarily for the benefit of  the tarball?

dginev:
@brucemiller you'd have to take that particular point up with the ExtUtils developers. I relied on the mature tooling of `ExtUtils::Manifest` and it only offers a [SKIP](https://perldoc.perl.org/ExtUtils/Manifest.html#MANIFEST.SKIP) blacklist option for refining the manifest filter.

I could switch to a whitelist, but then I think I would have to rely on my own code to post-filter the results of `fullcheck()`. But the bigger problem is that this PR is supposed to tell you when you *miss* a file that was meant to be added to the manifest -- and if you're adding a new subdirectory to latexml, a whitelist will just ignore it silently. The blacklist will pick it up and ask you to either add a rule for it to the SKIP or the MANIFEST file.

brucemiller:
Indeed, the exclusions are already covered in SKIP.  Thanks!!!

brucemiller:
Hmm... this makes sense as a pre-commit hook, if it's checking committed files. But I'm having second thoughts about being part of make test: it's pretty radical if you have any scratch work lying around.

dginev:
@brucemiller it is indeed very strict, but only as long as you haven't added your scratch examples to SKIP. 

If you're at a point of running `make test` on a branch, one would think the files in it are ready to be mentioned in the manifest?

dginev:
Another thing to keep in mind is that you may avoid this entirely by using a more precise (and faster) make test invocation. For instance, if you're only extending grammar coverage, you could run:
```
make test TEST_FILES=t/70_parse.t 
```

brucemiller:
As Dirk Gently says: Everything is Connected. So, I always run a full make test *before* I commit, branch or not.  And having to maintain SKIP over local, temporary changes seems more obtrusive to being obliged to maintain MANIFEST over intentional, permanent ones, no?

dginev:
@brucemiller I understand it looks obtrusive, especially if you already have a lot of scratch files at the local checkout. Or if you end up moving the same file several times and you need to rename it in the manifest each time.

That said, those ought to be rare events for most branches, so I am personally OK with having the extra strictness. If you hate it, feel free to remove it.

I recall the main reason I loved the idea of having a standalone test is that Travis-CI (and now Appveyor) will fail if we forget to update the manifest. So even if you do both of:
 * not run make test
 * commit with `--no-verify` and push

You still won't trick the maintainer, because Travis will immediately go red and complain. I have pushed "small changes" with --no-verify, and not testing in my life, though I would claim a long time ago when I was young and careless, but I'd rather trust a machine to double-check any PR open for latexml.

dginev:
Indeed, the real main use case is that if a brand new contributor shows up and files a new pull request for their new binding, Travis will educate them about our manifest issues, and you don't have to double-check the pull request every time.

dginev:
OK, double-checked, both Travis and Appveyor set environmental variables, notably `CI` is set to True, so I can make the manifest test only execute if we're in a continuous integration setting. Does that sound reasonable?

brucemiller:
Hmmm... let me see how cleanupable skip (and development habits) are...

brucemiller:
Since I'd still like to do things like add files & testcases and run make test before committing, I'm still likely to accept the other PR. But in the interest of general hygiene, it seemed like a good time to cleanup things, and improve ```.gitignore``` and ```MANIFEST.SKIP``` at which point I noticed... or rather failed to notice...

So, does the pre-commit hook actually use the skip file? I don't see it, and it's complaining about committing ```.gitignore```

dginev:
The pre-commit hook doesn't use the SKIP file in master, and I am oddly confused where my change to it disappeared into. I recall updating it at a later point, when I discovered `fullcheck()` but it's probably buried in my stash or a side branch or ... I will update that in the travis PR if that's ok with you.

dginev:
A bit saner now, here is the commit https://github.com/brucemiller/LaTeXML/pull/1122/commits/77155b031bf5b4a9219bfa9dbcfb0aa96bb9ed84

PR #1086, opened 2018-11-29
Add support for glossaries.sty

bfirsh:
Seems to work pretty well. There's one weird bug where LaTeXML eats the trailing space when `\gls` is used within a sentence (compare PDF and XML).

This fixes all the remaining issues seen in https://github.com/arxiv-vanity/engrafo-issues/issues/102.


brucemiller:
Sorry for the delay in responding. It's nice that the simple binding works, but it seems a shame not to leverage the mechanisms used in the ```acronym``` binding; there's underlying support in the schema and stylesheets.

brucemiller:
I was looking at whether I could tweak the binding to leverage the acronym work, but this patch doesn't actually work for me, using texlive 2016 (and I was seduced by Travis, since test cases in packages don't run by default).  What version of TeX did you use that worked with this binding?

brucemiller:
Took a bit more work to get this to work, and then a bit more to intercept glossaries's calls to wrap/add the appropriate XML, but it seems to work well.  We'll still need to add a PostProcessor to collect up the glossary definitions (which are potentially scattered through the document) to fill-in the glossary itself, but the terms are displayed correctly.   Thanks for the proposed patch!

bfirsh:
Nice!! :tada:

PR #1110, opened 2019-03-20
\dospecials and form-feed

dginev:
This PR gets latexml to "survive" Chapter 1 of the TeXbook, somewhat on a whim.

It makes good headway down the content as well, but there seems to be some deep issues with plain TeX alignments as one jumps into absorbing chapter 2. I may wait until the alignment refresh lands in the latexml master before I look into that.

For chapter 1, two simple fixes:
 - add the `\dospecials` macro, as described in the TeXbook, to TeX.pool. We already reference in some of the bindings (verbatim.sty.ltxml) and had a version of it defined in LaTeX.pool
 - add an active catcode for the form-feed character, which appears often in the TeXbook sources, as well as its TeX.pool definition via `^^L` (as per the plain.tex format, p. 343).
 - Perltidy demanded to lint the files I touched, so there are also some whitespace diffs

### DIFF
--- a/lib/LaTeXML/Core/State.pm
+++ b/lib/LaTeXML/Core/State.pm
@@ -101,7 +101,7 @@ sub new {
     my %std = ("\\" => CC_ESCAPE, "{" => CC_BEGIN, "}" => CC_END, "\$" => CC_MATH,
       "\&" => CC_ALIGN, "\r" => CC_EOL,   "#"  => CC_PARAM, "^" => CC_SUPER,
       "_"  => CC_SUB,   " "  => CC_SPACE, "\t" => CC_SPACE, "%" => CC_COMMENT,
-      "~" => CC_ACTIVE, chr(0) => CC_IGNORE);
+      "~" => CC_ACTIVE, chr(0) => CC_IGNORE, "\f" => CC_ACTIVE);
     map { $$self{catcode}{$_} = [$std{$_}] } keys %std;
     for (my $c = ord('A') ; $c <= ord('Z') ; $c++) {
       $$self{catcode}{ chr($c) } = [CC_LETTER];
@@ -726,7 +726,7 @@ sub getStatusCode {
 
 __END__
 
-=pod 
+=pod
 
 =head1 NAME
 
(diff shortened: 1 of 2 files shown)


brucemiller:
Aren't all the changes to TeX.pool whitespace? (And after somebody went and got me to use perltidy all the time!)

dginev:
@brucemiller they are, and they are even automatic - I wonder if my local setup has a differently configured perltidy? 

```
$ perltidy --version
This is perltidy, v20180101
```

dginev:
Oh, btw, you can go to the `File changes` tab, then `Diff Settings` at the upper right and tick `Hide whitespace changes` + apply, for the current viewing session. Then you see the actual payload of the PR without the supposedly "tidied up" whitespace.

Here it is as a screenshot for now:
all

dginev:
I rebase on master just in case, though nothing much changed. Here is the exact output I see when linting TeX.pool:

```
$ perl tools/latexmllint lib/LaTeXML/Package/TeX.pool.ltxml

lib/LaTeXML/Package/TeX.pool.ltxml  has syntax issues
  Name "LaTeXML::Core::Document::ID_SUFFIX" used only once: possible typo at lib/LaTeXML/Package/TeX.pool.ltxml line 3465.

lib/LaTeXML/Package/TeX.pool.ltxml  has policy issues
  [BuiltinFunctions::ProhibitComplexMappings] Map blocks should have a single statement at line 6072, column 32.  (Severity: 3)
  [Subroutines::RequireArgUnpacking] Always unpack @_ first at line 6721, column 1.  (Severity: 3)

Examined 1 file(s): 0 were reformatted; 1 needs code revision; 0 fail requirements for commit.
```

So for one reason or another my local perltidy considers this the optimal whitespace arrangement.

dginev:
Hi @brucemiller , I have reformatted with the most recent perltidy, `v20181120` and the space changes seem to be relevant?

brucemiller:
Actually, the code formatting is now even nicer than before. Sigh. I guess I'll just have to accept a local cpan install of perltidy, so we'll get the current best formatting, and more importantly, we're on the same formatting page!  Thanks for the patch!

PR #1113, opened 2019-04-04
[Experiment] Pandoc AST output format

dginev:
This is a draft pull request for an ongoing work item.

The goal here is to have latexml output the Pandoc AST JSON format. This would allow using latexml as an advanced `tex reader` for pandoc, increasing drastically the understood subset of TeX. Symmetrically, it allows using pandoc as a second stage post-processor for latexml, unlocking the full range of supported pandoc formats as direct latexml targets.

The current PR contains a basic skeleton that uses XSLT to map the latexml XML format into JSON, as part of the usual post-processing pipeline. Example use would be:
```
latexmlc example.tex --format=pandoc  |pandoc -f json -o example.md
```

As just one example of converting advanced latex to markdown.

The big remaining challenge of this PR is entirely in writing excellent XSLT transformations. We need to align the XML schema of latexml to the AST of pandoc, covering reasonably as much as possible. As this is a side-project for me, I welcome contributions from any interested parties. Coverage may appear here incrementally meanwhile.


dginev:
An example of a practical benefit here is being able to deprecate plugins that are too complex to properly maintain such as [LaTeXML-Plugin-Doc](https://github.com/KWARC/LaTeXML-Plugin-Doc) and instead rely on pandoc's active development team for reliably producing `.docx` , `.odt` etc

brucemiller:
Looks promising.

dginev:
Update: rebased with master.

dginev:
Can be continued after #1153 is implemented and lands into master, which is a very reasonable first step.

dginev:
Closing this PR for now since we're not actively working on this, but if anyone stumbles on it and wants to contribute - do drop us a line!

PR #1114, opened 2019-04-07
Add math lexeme test

dginev:
PR is ready for review and merge.

I found a pretty good example in a DL textbook I am reading for a semantic difference for a bold-x (a vector-valued random variable) and a bold-italic-x -- one of the individual values of bold-x.

So I decided to turn that into a test, also so that we exercise the option and avoid regressions.

That led me back into one of my most frustrating bits of programming, at the end of the convert call in LaTeXML.pm, where serializing into a UTF-8 marked string becomes an outright ordeal. I reread `XML::LibXML` carefully, and left detailed notes on my updated understanding of the circumstances. I think the code is now extra-robust, but I really dislike the design of the serialization in `XML::LibXML`. Not only are the defaults inconsistent between classes (Document doesn't encode, Node does), but they are also inconsistent between platform installations. If the `HAVE_UTF8` compile flag is used as false, the default behavior of Node's `toString` method *flips* to using byte strings. Which makes latexml's output unpredictable, if we use Node's `toString` anywhere. There is an explicit escape hatch, where we request to encode using the document's encoding, which is what I refactored into. Which lead to all tests passing without the "wide character" warnings, which lead me here in the first place.

So hopefully this PR provides both a useful test, and some useful cleanup on the serialization for the latexmlc-near API.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -828,6 +828,9 @@ t/daemon/formats/citationraw.xml
 t/daemon/formats/jats.spec
 t/daemon/formats/jats.status
 t/daemon/formats/jats.xml
+t/daemon/formats/lexmath.spec
+t/daemon/formats/lexmath.status
+t/daemon/formats/lexmath.xml
 t/daemon/formats/makebib.spec
 t/daemon/formats/makebib.status
 t/daemon/formats/makebib.xml
@@ -852,6 +855,7 @@ t/daemon/formats/parallel-math-xmath.xml
 t/daemon/formats/tei.spec
 t/daemon/formats/tei.status
 t/daemon/formats/tei.xml
+t/daemon/lexmath.tex
 t/daemon/fragment.tex
 t/daemon/mixedmath.latexml
 t/daemon/mixedmath.tex
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -317,28 +317,44 @@ sub convert {
     if (delete $$opts{placeholder_destination}) {
       delete $$opts{destination}; } }
 
-  # 5 Output
+  # 5 Output -- if not using Post::Writer, which never considers this serialization logic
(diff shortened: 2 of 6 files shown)


brucemiller:
Firstly, I'm not too convinced that the bold-italic is all that conventional for matrix elements; usually I've just seen plain old italic.  But that "semantic" doesn't seem to be carried into the test, so no matter.

Most of the PR is just adding that test case, but I'm a bit puzzled by the changes to LaTeXML.pm; it looks almost like it's from another PR to deal with encoding issues?

dginev:
Well, extensionally speaking, if it is *used* in published literature, it is a *real* notation, as far as that piece of literature goes. I just found it to be a very convenient example for a test case, no opinion on whether this is The One True Way of writing matrix elements. If anything I wonder how many notational dialects there are for writing matrix elements, could vary based on both decade and community (or even publisher/editor!).

As to the encoding -- the test case hit me with one of those gnarly "Warning: wide character in print" warnings, which indicated the final encoding to unicode was missing, hence the patch. I will rebase the branch on master and double-check that the patch is both still needed and still correct, and maybe add a more detailed technical comment for you to review.

dginev:
Rebased on master, still operational and correct. @brucemiller feel free to read the new comment block in LaTeXML.pm that you can see in the changed files in the PR and let me know if it is understandible.

For the new test, the serialization used is the fragment mode one, which uses `->toString` on a `XML::LibXML::Element`. What I disocvered this time, is that depending on how `libxml2` was *compiled* on the local machine, the *default behavior* of that method can flip from using the encoding, to not using it. Which is why I changed to passing the explicit flag for "use the document encoding when serializing", which looks like (the rather odd) `->toString(1,1)`. More on this in the LibXML.pm comment, but happy to elaborate further.

brucemiller:
Nice; makes sense (and suggests we may want to look deeper at other ```toString``` usages).  Thanks!

PR #1115, opened 2019-04-07
TeX-near \indent, \noindent and \parindent

### LINKED ISSUE #1098: parindent is ignored
Thank you for this package. I have a few problems with it.

Since I don't want left space when a new paragraph is created, I put the command `\setlength{\parindent}{0pt}` in my preamble. But LatexML seems to ignore it: new paragraphs have left space nonetheless

Example:
```
\documentclass[12pt,a4paper]{book}
\usepackage[utf8x]{inputenc}
\setlength{\parindent}{0pt}
\begin{document}
\part{Ciao}
Hello again

Hello hello
\end{document} 
```

dginev:
Fixes #1098 .

The PR is a little on the brave side and may warrant some review and discussion. The B book suggests that `\indent` and `\noindent` explicitly start a new paragraph in the right contexts (e.g. when invoked in vertical mode), and are completely ignored in the wrong contexts (e.g. when already inside a paragraph). 

Similarly, but separately, setting the length of `\parindent` to `0pt` is identical to a global `\noindent` for all following paragraphs.

I have ventured to emulate the behavior in the latexml setting as: 
 * Treat `\indent` and `\noindent` as constructors that open (or use an already open) `ltx:para` element, when possible, with the right class attribute, or be no-ops if that's impossible.
 * Have `\normal@par` be uniquely concerned with tracking the register value of `\parindent`, and use it for the *following* `ltx:para` if `0` and no class has been previously set (as `\noindent` and \indent` take precedence).
 * Most bravely, I noticed that indents don't work at all on the last paragraph of a latex document, in master, as it relies on an explicit `\par` break to indent. So I added an explicit call to `\normal@par` at the end of `\end{document}`. Since all the close calls are with `maybe`, it should be a silent no-op in all cases where requesting a paragraph break isn't appropriate.
 * I have added a new test `t/expansion/parindent.tex` which covers the tricky cases of usage, and have updated the old tests where needed - a couple of indents were wrong/missing, and `\noindent` now seems to show up more reliably in reversions. But I could've taken a wrong turn here, definitely would appreciate a review!


t/structure/options.xml
```diff
@@ -8,7 +8,7 @@
   <para class="ltx_noindent" xml:id="p1">
     <p>Class options: , acommonoption, aclassoption</p>
   </para>
-  <para xml:id="p2">
+  <para class="ltx_noindent" xml:id="p2">
```
dginev: This test is a simple example that we never indent the last para in a document.

brucemiller:
Interesting, and yeah, a bit scary.  I'm apprehensive about making ```\indent``` into a constructor, rather than just a primitive for effect (but presumably the ```\par``` behaviour is consistent with that).  Further, it further propagates the "error" that we have to look for ```ltx:para``` (during construction, which is actually too late), rather than actually understanding mode correctly.  The upside is that this introduces a test case which we'll have to track if we ever do it correctly!

dginev:
I thought introducing vertical mode just to solve this contained problem may be a bit overkill :> It's an incremental upgrade, and I agree more once we have vertical modes we can be "more correct" and return to the DefPrimitive implementation. I only went the constructor route since I didn't have enough rope to fish for an open paragraph at the `afterDigest` stage -- or at least didn't see a good way of making a knot with the rope I saw :>

brucemiller:
Well, it's interesting that it fixes a few mis-indented paragraphs in the test suite. So, what could possibly go wrong? Let's do it! Thanks!!

Issue #1116, opened 2019-04-08
quote substitution and utf-8 input


jkukul:
My input file initially looks like this:

```
\documentclass[12pt,a4paper]{report}

\begin{document}
I like Titanic very müch.
\end{document}
```

All good for now, `latexml` produces the expected output:
```
I like Titanic very müch.
```

As soon as I add any of `'`, `"`, `` ` `` , anywhere in the document, e.g:
```
I like `Titanic' very müch.
```

all `utf-8` characters from the input are not recognised anymore and the output looks like this:

```
I like ‘Titanic’ very mÃ¼ch.
```

I can fix it by adding `\usepackage[utf8]{inputenc}` in the document or using `--inputencoding=utf-8` option. Nonetheless, this behaviour is a little bit confusing, because:

- non-ASCII characters in the input are handled correctly until I insert a backqoute, apostrophe, etc anywhere in the document
- `pdflatex` always produces the expected output, also without `\usepackage[utf8]{inputenc}`

EDIT: Looks like  `\usepackage[utf8]{inputenc}` is also compulsory in other cases, e.g. when a backslash is present anywhere inside a comment, e.g. `% foo \ bar`, so looks like my issue is not strictly related to quote substitution.

dginev:
HI @jkukul . I actually can not reproduce the first example where you claim latexml just produces the expected unicode in the final HTML. As you mention, because we are trying to emulate the old school tex approach to encodings, you need either an explicit use of inputenc, or the command-line option:
```
--inputencoding=utf-8
```

to enable unicode support. Here is a test run over your example:
```
$ latexml 'literal: very müch '
latexml (LaTeXML version 0.8.3; revision 874fa6a3)
processing started Mon Apr  8 10:40:01 2019

(Digesting TeX Anonymous String...
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/TeX.pool.ltxml...
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/eTeX.pool.ltxml... 0.00 sec)
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/pdfTeX.pool.ltxml... 0.00 sec) 0.07 sec)
(Processing content Literal String... 0.00 sec) 0.08 sec)
(Building...
(Loading compiled schema /home/deyan/perl5/lib/perl5/LaTeXML/resources/RelaxNG/LaTeXML.model... 0.00 sec). 0.02 sec)
[...]

PR #1119, opened 2019-04-10
Default inputencoding (utf-8)

dginev:
Foundation for #1116 

A big part of the fragility and double-encoding in the various unicode tests one can try with the executables comes from there being *no* default input encoding in play. That implies that latexml treats any input as a perl byte string and an author of a TeX document unfamiliar with Perl finds the behavior hard to predict.

From a usability standpoint, it seems more direct that when we say "latexml is unicode native" we also imply a default input encoding of UTF-8. This PR changes that, and all tests continue to pass without need for extra intervention (luckily the ascii origins are invariant to unicode decoding).

I also have ensured that string mouths also use the inputencoding logic, as so far we only employed it for file mouths.

This PR can be followed up with my partial restructuring on the output side at #1114 . 

I think enforcing a default input encoding allows for additional predictability for a lay user. For advanced use cases (such as arXiv), the latexml build systems already set their own input encoding 
(`--inputencoding='iso-8859-1'`). Additionally, web service uses of latexml have worked seamlessly so far, as the frontend code prepared correctly decoded utf-8 strings which were passed through the
 `LaTeXML::convert` API, and hence avoid the command-line underspecification of #1116 .

Happy to discuss further, but it feels like an elegant way of reducing the encoding complexity.

### DIFF
--- a/lib/LaTeXML/Common/Config.pm
+++ b/lib/LaTeXML/Common/Config.pm
@@ -67,7 +67,7 @@ sub getopt_specification {
     "mode=s"      => \$$opts{profile},
     "source=s"    => \$$opts{source},
     # Output framing
-    "embed"      => sub { $$opts{whatsout} = 'fragment'; },
+    "embed" => sub { $$opts{whatsout} = 'fragment'; },
     "whatsin=s"  => \$$opts{whatsin},
     "whatsout=s" => \$$opts{whatsout},
     # Daemon options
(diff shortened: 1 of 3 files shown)


dginev:
now reformatted with latest perltidy

brucemiller:
Cool; sounds like a solid approach. Hopefully fixes more mystery arXiv problems than open up new ones! :>

PR #1121, opened 2019-04-11
more rules for SKIP

dginev:
A few quick rules that get the most recent daemon tests to pass without complaints -- the failures only become visible if you run `make test` on a multi-threaded setup.

 I run 32 tests in parallel so a lot of the temporary files are visible when the manifest check is run.


dginev:
I'm also happy to report all other PRs have had their manifests updated, using the new manifest tests :100:

PR #1123, opened 2019-04-25
Add complete set of ICML styles

bfirsh:
I've seen quite a lot of these in the wild: https://github.com/arxiv-vanity/engrafo/issues/241

Judging by the changelog, it seems like it hasn't changed much, so I'm guessing it'll work all the way back to 2000.

This seems like a pretty ridiculous implementation, so open to better ideas if you don't like the verbosity!


dginev:
We now auto-check for MANIFEST inclusion of new files in CI, as seen here:
```
t/97_manifest.t ........ 1/2 
#   Failed test 'Files missing from MANIFEST: 
# 	lib/LaTeXML/Package/icml2001.sty.ltxml
# 	lib/LaTeXML/Package/icml2002.sty.ltxml
# 	lib/LaTeXML/Package/icml2003.sty.ltxml
# 	lib/LaTeXML/Package/icml2004.sty.ltxml
# 	lib/LaTeXML/Package/icml2005.sty.ltxml
# 	lib/LaTeXML/Package/icml2006.sty.ltxml
# 	lib/LaTeXML/Package/icml2007.sty.ltxml
# 	lib/LaTeXML/Package/icml2008.sty.ltxml
# 	lib/LaTeXML/Package/icml2009.sty.ltxml
# 	lib/LaTeXML/Package/icml2010.sty.ltxml
# 	lib/LaTeXML/Package/icml2011.sty.ltxml
# 	lib/LaTeXML/Package/icml2012.sty.ltxml
# 	lib/LaTeXML/Package/icml2013.sty.ltxml
# 	lib/LaTeXML/Package/icml2014.sty.ltxml
# 	lib/LaTeXML/Package/icml2015.sty.ltxml
# 	lib/LaTeXML/Package/icml2019.sty.ltxml
# 	lib/LaTeXML/Package/icmlstymod.sty.ltxml
# 	lib/LaTeXML/Package/mk2k.sty.ltxml'
#   at t/97_manifest.t line 9.
# Looks like you failed 1 test of 2.
```

so feel free to update with those entries. 

As to the verbosity, up to Bruce's stylistic judgement, as we already have three of those binding files, but I agree it starts to feel as if we're missing a feature. I wonder if we should make a new fallback that in the case of a `name\d+.sty.ltxml`, in cases where a binding can't be found, fall back onto `name.sty.ltxml` with only an Info severity message akin to "assuming versioned style/class file, using base implementation" ? Could there be unwanted side-effects in casual use? (I can't think of any, and it's clearly helpful for arXiv).

bfirsh:
Oops yes I asked for that didn't I so I should have remembered. :)

brucemiller:
Thanks for tracking those down. In the past, when thinking over this problem, I've been wary of hiding such behaviour in the code somewhere (whether a regexp, or hash, or whatever), and usually there were only few equivalents, and sometimes they weren't even exactly equivalent. But this case pushes the envelope and cries out for a better solution, doesn't it?

But I also recall from arXiv quite a few modified styles with fairly random, less regexp friendly, extensions (rarely used, but still). Also, a regexp could lead to surprises like getting, oh, tikz3 instead of tikz2? (though in most cases, you're weighing one failure against a different one).

Still, I wonder if a more explicit approach would be better: include a hash of style aliases, that can be assigned (State), initialized somewhere, can be extended in bindings (eg. preloaded) etc.  Presumably these would be consulted only after the main request was not found?

bfirsh:
I always err on the side of explicitness and simplicity. Agreed that a regexp could cause complexity and breakage.

A hash seems like a better solution than lots of files. But, on the other hand, when I'm wondering whether a package has been implemented in LaTeXML, I always look in `lib/LaTeXML/Package/` so if it's not there, it might be hard to track down where a package is actually implemented. (It took me a while to realise a lot of things were being defined in "omnibus" for example.) Perhaps the `lib/LaTeXML/Package/` directory could have a readme explaining how packages are resolved, and where to look to find the aliases etc.

Either way I haven't got time to implement a hash system, so if you haven't got time either, perhaps this would work as an interim solution. ;)

dginev:
@brucemiller I am about to push a PR that does the fallback as a general measure, and prints an `Info` message explaining what happened. 

I would argue misreading tikz3 as a generic tikz is a more graceful degradation of behavior than fully bailing on any tikz-related primitives (which usually leads to a Fatal error).

I would also argue the generic fallback allows us to achieve "unexpected gains" over arXiv, as there may be a number of cases where this behavior achieves a smooth conversion of otherwise erring documents. The examples of this PR are exactly in this category.

bfirsh:
FWIW I don't have any strong feelings about this, so happy to go with @dginev's PR if that's the right course of action. :)

dginev:
Hi @bfirsh , we talked this over with Bruce yesterday, and we're likely to merge a variant of my PR pretty soon (just updated it there), so closing this one. Thanks for initiating this, we may see some additional gains over the arxiv conversion as a consequence.

bfirsh:
Cool! Sounds good to me. :)

PR #1124, opened 2019-04-25
Fallbacks for derivative binding pathnames

dginev:
Alternative (and more general) fix for #1123 .

The hope is that this provides unexpected gains over arXiv, and is a more graceful "degradation of behavior" in cases where latexml hasn't provided support for bindings of a given version, but does provide generic support for the binding name.

The fallback is only triggered when searching for `.ltxml` files, and only for `.sty` and `.cls` suffixes.


dginev:
Following our discussion with @brucemiller I took this PR a step further, after we reviewed the custom suffixes seen in the latest testbench run:
https://corpora.mathweb.org/corpus/arxiv_testbench/tex_to_html/warning/missing_file?all=false

We could keep extending the `FindFile_fallback` logic as we keep improving over arXiv, now that there is a place for the logic. I have added a `t/002_unit_findfile.t` test to keep track of supported patterns, which can also be extended as we go forward.

Ready for review/merge.

dginev:
Now that the full arXiv corpus is being rerun with latexml 0.8.4, another good bookmark to add here is:

https://corpora.mathweb.org/corpus/arxmliv/tex_to_html/info/fallback?all=true

You can see all successful uses of this fallback feature, when converting the entirety of arXiv.

brucemiller:
yep, ya done good! :>  I see you have "all=true", but it doesn't seem to make a difference; can you distinguish the cases which succeed from those that still fail (for some reason)?

dginev:
Ah, right -- I double-checked and it is as I remember - the info reports are a bit of a "quick add-on" implementation, and lack the `all=true|false` flag internally at the backend. What cortex currently does is that it *always* provides all tasks.

Also, `info` is not - currently - treated as a severity inside cortex. I guess the logical approach would be to unify that report with `no_problems` in some sense, maybe even entirely replace the `no_problem` report with the `info` level report... Yep, that would make sense! The current no_problem report is a simple task list, which was grandfathered in from Heinrich's build system, but now that we have info messages, it makes more sense to adapt those as the main `no_problem` report, and add the `all` query parameter properly. Maybe worth making myself an issue if that sounds appealing...

brucemiller:
I don't think it's worth changing anything (at this point),
since there's not much debugging value (yet) from info msgs.
It just that while it's gratifying to see that more docs
are getting *some* bindings to work with, it would have been
even moreso to see them "succeed" with the alternative binding.

PR #1126, opened 2019-05-01
Freshen up: tools/maketest for spec tests

dginev:
I was advertising the `tools/maketest` utility script today [here](https://github.com/MathWebSearch/LaTeXML-Plugin-MathWebSearch/pull/7) and realized it is missing a piece of default logic that makes it useful with the simplest invocation.

 Just matching the `doxml` default logic in the regular testing path actually.

So that:
```
perl tools/maketests t/daemon/profiles/fragment.spec
```
automatically creates the .xml file when it is missing or stale.


brucemiller:
Very useful! Thanks!!

PR #1130, opened 2019-05-16
Robust NCNames in LaTeX counter-based xml:ids

dginev:
So it turns out TeX can get away with doing nasty hacks related to "namespacing" stepped counters, while latexml can't, mostly due to the requirement for xml:ids to be NCNames.

I traced down an example of this class of issues from the arXiv reports, a good paper being: https://arxiv.org/abs/1103.4198

which was dying with a Fatal when cross-referencing an equation, as its id was not valid for the XML parser. This is otherwise a very well-converting warning-level (math parsing) document, which really shouldn't have to suffer a Fatal.

The offending line of TeX that lead to all the drama was:
```tex
%                TO NUMBER EQUATIONS BY SECTION
%
\makeatletter \@addtoreset{equation}{chapter} \makeatother
```

Which, when used in this `\documentclass{article}` document, implied the leading prefix to the equation xml:ids was going to be missing - there are no chapters in the document. This is - I believe - the exact internal behavior in latex, but latex doesn't have the NCName constraint, and also likely doesn't use the information in a similar manner.

After spending most of the time diagnozing, I am offering a simple patch - which could be improved with a more high-level robustness NCName strategy - which gets this document to convert without issue. Should get us some fractions of a percept up in the arXiv conversion?

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -472,6 +472,11 @@ sub CleanID {
   $key =~ s/&/-amp-/g;
   $key = unidecode($key);
   $key =~ s/[^\w\_\-.]//g;                 # remove everything else.
+
+  if ($key =~ /^[^a-zA-Z\_]/) {
+    # in case we are missing a prefix and have a leading ., default to X
+    $key = "X$key";
+  }
   return $key; }
 
 sub CleanLabel {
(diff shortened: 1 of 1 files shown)


dginev:
Updated to consistently using `CleanID` (thanks for the pointer @brucemiller ), and retested with the arXiv document, which is now warning status. Ids look on the lines of:
```xml
<table id="X.Ex2" class="ltx_equation ltx_eqn_table">
<td class="ltx_eqn_cell ltx_align_center"><math id="X.Ex2.m1" 
```

PR #1132, opened 2019-05-19
\fmtversion and empty mouthstack fallback

dginev:
I was checking the too_many_fatals page, as it has been a long while since I looked at it, and the first entry already showed two patch opportunities that upgraded the task to a usable "error" status:
https://corpora.mathweb.org/corpus/arxiv_testbench/tex_to_html/fatal/too_many_errors/100?all=false

1. `\fmtversion` is apparently parsed at times, so it should really contain a real date to make the parse succeed. I used the date in my local `latex.ltx` file.

2. Curiously, when a `mouthstack` is empty during Gullet's `readingFromMouth`, it currently loops indefinitely, firing an error on each step and hence always Fatal-ing with "100 errors reached". So I just added a `last;` escape hatch, which lets the conversion continue and recovers at least part of the document. In the document I am testing on, all of the content actually appears in the final document, so I assume it may have been a macro termination/argument error due to missing definitions.

I'll check a couple more files, will file separate PRs if I spot easy fixes.

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -79,8 +79,8 @@ sub mouthIsOpen {
 # Corresponds (I think) to TeX's \endinput
 sub flushMouth {
   my ($self) = @_;
-  $$self{mouth}->finish;;    # but not close!
-  $$self{pushback}  = [];    # And don't read anytyhing more from it.
+  $$self{mouth}->finish;    # but not close!
+  $$self{pushback}  = [];   # And don't read anytyhing more from it.
   $$self{autoclose} = 1;
   return; }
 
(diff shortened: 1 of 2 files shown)


dginev:
Here is a link to the file diff without the perltidy whitespace changes:

https://github.com/brucemiller/LaTeXML/pull/1132/files?w=1

dginev:
Conversion status with patches applied:
```
Conversion complete: 2 warnings; 2 errors; 1 undefined macro[\layoutstyle]; 1 missing file[aipproc.cls].
```

dginev:
I think I won't be opening new PRs, just jotting down a quick observation that a big portion of the "too many error" fatals I opened were due to `\xymatrix` going over 100 alignment `&` characters and breaking the conversion. So one more reason to consider xy our biggest binding target for improving arXiv coverage with one swing.

dginev:
Update: rebased with master

PR #1133, opened 2019-05-19
Allow pre-comma spaces in package option lists

dginev:
This was a satisfying fix, as it may be responsible for a range of hard-to-trace Fatal errors in arXiv. It may also be a very rare error, as it is a subtle edge case in splitting out the package options, combined with advanced option logic for revtex4.

A rather innocent looking minimal tex snippet is:
```tex
\documentclass[twocolumn,superscriptaddress,showpacs,preprintnumbers,amsmath
,amssymb,aps,prd]{revtex4}
\begin{document}
minimal
\end{document}
```

As it turns out, the master branch of latexml splits by `,\s*`, hence obtaining the `amsmath\n` name and failing to load amsmath, which it kindly told me via:
```
Info:unexpected:options Unused global options: amsmath
	at paper.tex; line 62 col 0 - line 62 col 16
```

In the example I was looking at, that in turn was followed by an error:
```
Error:undefined:{align} The environment {align} is not defined.
```

then followed by 100 alignment errors, and thus a Fatal. Once the regex was fixed and amsmath got loaded correctly again, `{align}` was available and the conversion succeeded (with 1 minor error).

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -52,7 +52,7 @@ DefConstructor('\documentclass OptionalSemiverbatim SkipSpaces Semiverbatim []',
     my ($stomach, $whatsit) = @_;
     my $options = $whatsit->getArg(1);
     LoadClass(ToString($whatsit->getArg(2)),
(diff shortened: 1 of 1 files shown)


dginev:
As to the regex I used `/\s*,\s*/`, it was already uniformly used in Package.pm, and was also my first instinct to do a correction.

PR #1134, opened 2019-05-19
revtex4 option of emulateapj and aastex should load the binding

dginev:
Another quick fix turning a Fatal into a warning (100 alignment errors again). Example was:
```tex
\documentclass[iop,revtex4]{emulateapj}
```

which (in pdflatex) lead to loading revtex4.cls , then longtable.sty, which defined `{longtable*}` which is where the 100 alignment errors were being triggered. Simply adding the loading behavior to the binding allowed for an error-free conversion run with a beautiful output, quite neat.

### DIFF
--- a/lib/LaTeXML/Package/aastex.cls.ltxml
+++ b/lib/LaTeXML/Package/aastex.cls.ltxml
@@ -32,6 +32,9 @@ foreach my $option (qw(10pt 11pt 12pt
 # Number equations within sections
 DeclareOption('eqsecnum', sub { Digest(Tokenize('\AtEndOfClass{\eqsecnum}')); });
 
+# revtex4 option from emulateapj.cls loads revtex4.cls
+DeclareOption('revtex4', sub { LoadClass('revtex4'); return; });
+
 DeclareOption(undef, sub {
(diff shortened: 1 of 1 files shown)


dginev:
From `emulateapj.cls` raw cls sources:
```tex
%% [revtex4]   - force using revtex4
%%
%% NOTE 2: Starting from version 11/10/2009, emulateapj tries to load
%% revtex4-1 if present on the system. This may cause small changes in
%% typesetting for old documents. Also, revtex4-1 does not interact
%% well with the lscape package which is needed for typesetting of
%% long rotated tables. If you want full compatibility, please
%% download the last version based on revtex4,
%% http://hea-www.harvard.edu/~alexey/emulateapj/emulateapj-rtx4.cls
%% and bundle it with your TeX source code. Starting with the
%% 8/17/2012 version, you can also use the [revtx4] option. 
%% emulateapj requires revtex4.cls. If you don't have it already, it can be
%% downloaded from http://publish.aps.org/revtex4/ (it's a small package). 
%% Other extrnal packages used are latexsym, graphicx, amssymb, longtable,
%% epsf. They should already be present in the modern TeX distributions; if
%% not, download them from www.ctan.org. 
%%
% ... 
```

PR #1135, opened 2019-05-19
symbolscan maintenance refresh

dginev:
While I was fishing for `texscan` runs, I ran a `symbolscan` run just out of curiosity and realized that script hasn't been updated to the latest LaTeXML API.

I did the minimal amount of changes to make it runnable on my machine (including using kpsewhich to find the unicode data file, which for me is only available in the texlive installation).


brucemiller:
Can hardly work worse :> Thanks!

PR #1136, opened 2019-05-19
Assert less for \unit  mode (units.sty)

dginev:
Recovering another Fatal document for arXiv, where latexml encountered 100 errors in seeing math-only characters outside of math mode, due to entering the wrong mode at the wrong time.

In particular, a minimal example that works with pdflatex and shows the wrong math mode turn is:
```tex
\documentclass{article}
\usepackage{units}
\usepackage{graphicx}
\begin{document}

\begin{figure}
  \begin{picture}(1,0)
    \put(0,0){\makebox(0,0)[lb]{\smash{Dose (\unit[10$^{14}$]{ions/cm$^2$)}}}}%
  \end{picture}%
\end{figure}

\end{document}
```

latexml master branch log:
```
Error:unexpected:^ Script ^ can only appear in math mode
	at minimal.tex; line 8 col 78 - line 8 col 78
	In Core::Definition::Primitive[Superscript] /home/deyan/perl5/lib/perl5/LaTeXML/Package/TeX.pool.ltxml; line 3796
	 <= Core::Stomach[@0x56205a97cab8] <= Core::Gullet[@0x56205a97d028] <= Core::Stomach[@0x56205a97cab8] <= ...
```

The units.sty binding even had a comment noting that the intent of the original sty file was to remain in text mode, but attempted to do auto-switch to math for consistency. That is indeed the case when I double-checked the original `units.sty` file from CTAN. It turns out that breaks some of the arXiv real world examples, so my proposal in this PR is to simply back off from making the assertion. 

The arXiv document converted to a "1 warning" status from a "Fatal", which is a very healthy upgrade.


dginev:
If the original enhancement is considered crucial for latexml, I could rethink the PR as doing the slightly stranger wrapper of ` $\nf@mathcopytextfont \text{#2}$, which may actually work correctly with both interleaved math and regular text...

dginev:
Ok, checked in an approach a little closer to the original attempt to auto-math the units. And added the arXiv usage to the test file so that we keep track of it. The Fatal-ing test case I had continues to convert properly.

dginev:
Update: I kept feeling unhappy about losing the previous functionality of the binding, so I went and tried to add a new set of rules that recognize the current approach and recover the "single unit in text mode" semantics. I think it worked out nicely for the examples we had in the tests.

brucemiller:
So, I'm thinking something simpler (not changing TeX or unitfrac), but since the culprit is unit, deal with it there. Maybe just read the arg, scan for "$" and wrap as math if none found? (timing issues, or if a macro expands to $ problems)

PR #1138, opened 2019-05-19
amsmath \And; DefMath macros allowed in text mode

dginev:
Along my debugging in #1137 , another `ltx:XMTok` not allowed in text, can be seen when authors decide to use amsmath's `\And` macro in text mode, e.g. to separate out authors (as seen in arXiv).

Potentially makes me wonder if there should be a more general error-free fallback, where any DefMath that has a simple unicode replacement could be interpreted error-free in text mode, as its literal text content ... maybe with a warning? This PR is just one solution idea, staying very specific to the `\And` macro. It is defined via 
```tex
\def\And{\DOTSB\;\mathchar"3026 \;}
```

in the sty file, and latex is lenient enough to allow misusing that in text mode...


dginev:
Since i started discussing DefMath here, let me continue by adding one more example (not fixed in the PR, but discussed in this comment). A minimal example derived from `0803.3486` in arXiv is:
```tex
\documentclass{amsart}

\begin{document}

Let $t_1 = s$, \dots, $t_{M}$ be a numeration of $O$ and let $g_i\in G$

\end{document}
```

Where `\dots` was mistakenly used outside of math mode and produced the error:
```
Error:malformed:ltx:XMTok <ltx:XMTok> isn't allowed in <ltx:p>
```

meanwhile, the conversion is completely error-free in latex. Digging in, I noticed that TeX.pool makes `\dots` flexible via using `?#isMath`  in the constructor replacement, which is actually a better fix for `\And` in my PR. I will add a commit. 

Back to the `\dots` example, the definition in `amsmath.sty.ltxml` is less flexible, already assuming we're in math mode. So upgrading there would also eliminate this example of a malformed XMTok error.  Maybe I tuck it into the current PR after all...

dginev:
Thinking out loud a bit further, the TeX.pool constructor for `\ldots` really seems to be a generic enough pattern to be moved into `DefMath` as an attribute - it is already copy-pasted for `\vdots` in the pool. And as this PR explores, `\And` and `\ldots` from amsmath are also in need of this error-free flexible fallback.

Maybe it is worth thinking of a mechanism akin to:
 - `DefMath(... , in_text_mode=>"content")` 
 - `DefMath(... , in_text_mode=>"error")` 
 - `DefMath(... , in_text_mode=>sub {...})` 

Marking the PR as a work-in-progress draft, since I think more work is needed here.

dginev:
Ok, there may actually be a great trade-off patch here. The case for `\And`, which may be useful to extrapolate to any other `DefMath` definitions of this form, was handled by the `defmath_prim` experimental handler, that installs a Primitive for the mathematical macro.

I simply made that installed primitive aware of the current tex mode, and deposited the same content Box, with the mode switch flipped. That seems to work nicely, and I can approximate it in the `\dots` definition, so that the content is in the final XML, and the errors disappear.

One issue with make test is that a text attribute broke:
```
#       got : '      <p><Math mode="inline" tex="AB\dots" text="A * B * […]" xml:id="S2.p9.m1">'
#  expected : '      <p><Math mode="inline" tex="AB\dots" text="A * B * dots" xml:id="S2.p9.m1">'
```

which I am looking into next. But this may actually work, and potentially fix a number of other `malformed` errors over arXiv.

dginev:
Update: some semi-unpleasant bookkeeping later, the `\dots` tests pass again, and I feel happy with the PR -- may be worth a review+merge in fact!

dginev:
Update: rebased with master, backing off from explicitly changing `\dots` in amsmath, now that master has a patch for it.

The general point of this PR remains, but maybe we can spend some more time thinking of the best programmatic solution -- most constructs defined via `DefMath` wouldn't raise an explicit error during a `pdflatex` pass of the document, so latexml may be better off allowing for a safe fallback whenever a primitive defined via DefMath is invoked in text mode.

dginev:
I added a test for the `\And` case I was trying to handle and am realizing it's likely more complicated than what I bargained for. In the produced PDF, one can see that even though you can use the macro without errors raised, it *also* acts as a switch from text into math mode... Hm.

dginev:
Here is the `\And` example from arXiv in full, to give maximal context:
```tex
\author{
  \bf{Ryan Prescott Adams}\\
  Cavendish Laboratory\\
  Cambridge CB3 0HE\\
  United Kingdom
  \And
  \bf{David J.C. MacKay}\\
  Cavendish Laboratory\\
  Cambridge CB3 0HE\\
  United Kingdom
}
```
Document 0710.3742 [Preview link](https://corpora.mathweb.org/preview/arxiv_testbench/tex_to_html/0710.3742)

dginev:
Ok, this PR is the wrong solution, since my diagnostic equating the `\And` problem to `\dots` was incorrect.

What really happens in the article from arXiv, is that a second, custom, sty file is loaded, called `proceed2e.sty`, which overrides `\And` to be a tabular-related construct, and hence typeset the names better. DefMath is just the wrong thing to focused on for this, so closing this PR entirely, and will keep  thinking about the correct framing in the background.

PR #1140, opened 2019-05-20
Localized search for `.bib` paths

dginev:
Currently we search the entirety of `kpsewhich` for .bib paths. While that may sound perfectly legal for an ordinary document, it is not "careful enough" for the arXiv conversion.

In particular, I have an example document which depends on a `biblio.bib`, but in reality using `paper.bbl` for its arXiv compilation. The paper does not even include the .bib file in the bundle. However, in this case latexml finds a biblio.bib if you have a texlive installation, in particular:
```
$ kpsewhich biblio.bib
/usr/share/texlive/texmf-dist/bibtex/bib/msc/biblio.bib
```

Which leads to a couple of macro errors, as that bibliography contains macros. This PR is based on my claim that looking for `.bib` files inside the texlive installation should never be latexml's intent. Instead, we may want to look into the user-defined `TEXINPUTS` and `BIBINPUTS` variables, which may contain custom bib paths.

This is more frequent of a situation than one might suspect -- all cases of the `\mscversion` undefined error seem to be caused by loading the wrong `biblio.bib` : 

https://corpora.mathweb.org/corpus/arxiv_testbench/tex_to_html/error/undefined

Using the PR to convert the first report (0705.2800) moves the document from Error to a "no problem" entry, and attaches the right bibliography.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1731,7 +1731,7 @@ sub pathname_is_raw {
   return ($pathname =~ /\.(tex|pool|sty|cls|clo|cnf|cfg|ldf|def|dfu)$/); }
 
 my $findfile_options = {    # [CONSTANT]
-  type => 1, notex => 1, noltxml => 1, nokpsewhich => 1 };
+  type => 1, notex => 1, noltxml => 1 };
 
 sub FindFile {
   my ($file, %options) = @_;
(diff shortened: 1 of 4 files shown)


dginev:
Update: rebased with master.

dginev:
A different / supplemental approach could also be to consider the arXiv's explicit statement that they do not run bibtex, but simply ingest the bbl with pdflatex.

So, an alternative patch to this document case would be to introduce some customization switch that disables entirely bibtex processing, while continuing to allow .bbl processing...

dginev:
Thanks!

Example use (for cortex) can be seen here:
https://github.com/dginev/LaTeXML-Plugin-Cortex/commit/f12bdf24c688a138528135bf9f115f7aa65336d4

PR #1141, opened 2019-05-20
keep empty groups as script anchors during scriptHandler

dginev:
It looks like `scriptHandler` was erroneously dropping empty `{}` groups that acted as script anchors, leading to a wrong message:
```
Error:unexpected:double-superscript Double superscript
```

for an otherwise reasonable formula: 
```tex
$\xi'{}_i^j$
```

This PR adds a minor patch keeping the `{}` for the grammar to try and parse on the "scripted decorated symbol". The grammar fails with a warning (cool extension opportunity!), but the document is upgraded to a "warning" status from the previous "error" status. Test was `0906.1692`.

I was originally interested in that file to check why the `U` fontmap wasn't being found, and added a couple of undef checks to make perl happy on the route to that, but that requires more infrastructure to improve on.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1359,11 +1359,12 @@ sub DefMathI {
   my $csname  = $cs->getString;
   my $meaning = $options{meaning};
   my $name    = $options{alias} || $csname;
+  $presentation = '' unless $presentation;
   $name =~ s/^\\//;
   $name = $options{name} if defined $options{name};
   $name = undef          if (defined $name)
     && (($name eq $presentation) || ($name eq '')
-    || ((defined $meaning) && ($meaning eq $name)));
+    || ($meaning && ($meaning eq $name)));
   $options{name} = $name;
   $options{role} = 'UNKNOWN'
     if ($nargs == 0) && !defined $options{role};
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1359,12 +1359,14 @@ sub DefMathI {
   my $csname  = $cs->getString;
   my $meaning = $options{meaning};
   my $name    = $options{alias} || $csname;
-  $presentation = '' unless $presentation;
+  # Avoid undefs specifically, we'll be doing string comparisons
+  $presentation = '' unless defined $presentation;
+  $meaning      = '' unless defined $meaning;
   $name =~ s/^\\//;
   $name = $options{name} if defined $options{name};
   $name = undef          if (defined $name)
     && (($name eq $presentation) || ($name eq '')
-    || ($meaning && ($meaning eq $name)));
+    || ($meaning eq $name));
   $options{name} = $name;
   $options{role} = 'UNKNOWN'
     if ($nargs == 0) && !defined $options{role};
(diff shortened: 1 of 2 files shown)


dginev:
As the test suite illustrates, there are at least two complications here:
 * The `0` string value case has to be handled better in my fix
  * The text attributes seem to now have two empty `{}` groups in the XML, the reason for which I still need to study better

dginev:
The xmath tree after the patch looks reasonable to a point. Before parsing starts:
```xml
<XMath>
  <XMTok name="xi" role="UNKNOWN">ξ</XMTok>
  <XMApp role="POSTSUPERSCRIPT" scriptpos="1">
    <XMArg rule="Superscript">
      <XMTok name="prime" role="SUPOP">′</XMTok>
    </XMArg>
  </XMApp>
  <XMApp role="FLOATSUBSCRIPT" scriptpos="1">
    <XMArg rule="Subscript">
      <XMTok role="UNKNOWN">i</XMTok>
    </XMArg>
  </XMApp>
  <XMApp role="POSTSUPERSCRIPT" scriptpos="1">
    <XMArg rule="Superscript">
      <XMTok role="UNKNOWN">j</XMTok>
    </XMArg>
  </XMApp>
</XMath>
```

After parsing ends (fails):
```xml
<Math mode="inline" tex="\xi^{\prime}{}{}_{i}^{j}">
  <XMath>
    <XMApp>
      <XMTok role="SUPERSCRIPTOP" scriptpos="post1"/>
      <XMTok font="italic" name="xi" role="UNKNOWN">ξ</XMTok>
      <XMTok fontsize="70%" name="prime" role="SUPOP">′</XMTok>
    </XMApp>
    <XMApp>
      <XMTok role="SUBSCRIPTOP" scriptpos="post1"/>
      <XMApp>
        <XMTok role="SUPERSCRIPTOP" scriptpos="post1"/>
        <XMTok meaning="absent"/>
        <XMTok font="italic" fontsize="70%" role="UNKNOWN">j</XMTok>
      </XMApp>
      <XMTok font="italic" fontsize="70%" role="UNKNOWN">i</XMTok>
    </XMApp>
  </XMath>
</Math>
```

dginev:
I have become grown to better understand the logic in `scriptHandler` and its surroundings. So I have gone back and restored the behavior when empty groups anchoring scripts are encountered, as a lot of the logic takes that as a foundation.

What is instead closer to the spirit of the `scriptHandler` processing is that once a previous script is compared against, and all checks take place, we add a final check whether that prior token was a FLOATING case. If so, that implies it was _preceded_ by an empty anchoring group, and we should hence terminate the loop, as no further information would be available.

The difference is only truly visible in cases such as:
 * `$\xi'{}_i^j$` - no error, but only because the j superscript is stopped by the floating i subscript
 * $`\xi'{}_i^j_d$`- error encountered, as the d subscript collides with the i subscript, since no group anchors it as a separate construct.

All tests pass, and the original document is once again Error-free. We get a failed parse warning, and the grammar can be asked the "hard" question of what the tree ought to be next.

Ready for review/merge on my part!

dginev:
Double-checked the test arXiv document, all good with the better conditional!

PR #1142, opened 2019-05-20
Better math mode recognition (aa eqn; para-type cells)

dginev:
Two examples for errors caused by wrongly switching from/to math mode, both somewhat specialized.

1. aa.cls equations allow error-free inline math fragments in latex. Aim: error-free + good markup in latexml
```tex
\documentclass{aa}
\begin{document}
\begin{equation} % usual markup
  A B,
  \label{eq1}
\end{equation}

\begin{equation} % aa-only permissible markup
  A $ B,$
  \label{eq2}
\end{equation}
```
I assume people are tempted by the second version as a silly shortcut to doing `\mathrm`, or just by mistake or... In any case pdflatex accepts this happily.

2. paragraph-type cells should treat all lines of their content as text mode
```tex
\documentclass{article}
\begin{document}
\[ \begin{array}{cp{4.5cm}}
4 &  $x_1$ \newline $x_2$ \\
\end{array} \]
\end{document}
```

latexml already treats the first line, as text mode, and wraps it as `\tabularcell@hbox{$x_1$}`. However, the second line is handled as if it were outside of the `p{Dimension}` cell definition, receives no wrapper, and thus produces an error for subscript being only allowed in math mode. I assume the fix for this case would be to improve the handling so that the wrapper takes the entire content, i.e. `\tabularcell@hbox{$x_1$ \newline $x_2$}`.

Both examples are distilled minimal cases of real-world equations in arXiv papers.


dginev:
The PR is still a work in progress (and help is welcome, since my alignment knowledge is very limited for the moment). I have a somewhat naive patch for the first example, and am still fishing for a good patch for the second.

dginev:
Travis caught a silly typo, rebuilding...

PR #1146, opened 2019-05-24
\bezier and \dashbox argument patches

dginev:
I was just quickly skimming another Fatal-status doc, and it encountered problems related to:

```tex
\bezier{200}(30,30)(40,25)(50,30)
```
and
```tex
\put(10,10){\dashbox(40,0)}
```

So, just as a test, I patched the signatures of those two macros to be more lenient and... well, the document converted with just Warnings? So this may be a fix? I don't know enough about the `{picture}` environment to be certain this is a robust patch, but in both cases all "mandatory" argument cases should still work, while also allowing the new "optional"/"flexible" patterns.

arXiv document was `quant-ph0506026`

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4476,7 +4476,7 @@ DefConstructor('\qbezier [Number] Pair Pair Pair',
   alias      => '\qbezier',
   properties => sub {
     picProperties(pt => PairList(picScale($_[2]), picScale($_[3]), picScale($_[4]))); });
-DefConstructor('\bezier Number Pair Pair Pair',
+DefConstructor('\bezier {Number} Pair Pair Pair',
   "<ltx:bezier ?#1(displayedpoints='#1') points='&ptValue(#pt)' stroke-width='#thick' />",
   alias      => '\bezier',
   properties => sub {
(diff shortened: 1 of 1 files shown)


dginev:
Actually, the switch to optional arguments in `\dashbox` will not work if given the mandatory argument pattern, that was blatantly false. But I don't have examples that use it with the mandatory arguments. Is this another version mismatch? The paper seems to be from 2005: https://arxiv.org/abs/quant-ph/0506026

dginev:
Also, I independently verified this fix also upgrades `math-ph0303031` from Fatal to Warning, I was looking at both from via the same cortex report page.

dginev:
Hm hm, latex is cheating. Both of these documents compile perfectly:
```tex
\documentclass{article}
\begin{document}
\begin{figure}

\begin{picture}(80,40)(0,-20)

  \put(10,10){\dashbox(40,0)}

\end{picture}
\end{figure}
\end{document}
```

as well as:
```tex
\documentclass{article}
\begin{document}
\begin{figure}

\begin{picture}(80,40)(0,-20)

  \put(10,10){\dashbox{200}(40,0)}

\end{picture}
\end{figure}
\end{document}
```

dginev:
Oh. Patterned arguments are optional / allowed to be empty? Here's the distilled essence of this example:
```tex
\documentclass{article}
\begin{document}

\def\test#1(#2,#3){one #1 two #2 three #3}

\test{9}(8,7)

\test(6,5)

\end{document}
```

dginev:
Well, alright, I got stubborn and implemented the argument check procedure, so that we can continue using the more "high level" signature pattern. It just took one extra level of indirection. I am not sure if the resulting SVG is (and "was") of high quality, but at least we don't get avoidable fatals.

brucemiller:
You're being way too complicated --- thinking like a programmer, rather than a TeXy :>  The catch is that latex defines dashbox like ```\def\dashbox#1(...``` which effectively makes the 1st arg optional.  Simplest approach is like
```
DefMacro('\dashbox Until:(','\ifx.#1.\lx@dashbox{0}(\else\lx@dashbox{#1}(\fi');
DefConstructor('\lx@dashbox {Float} Pair [] {}',...
```
But what puzzles me is the apparent optionalness of the last arg (the box contents).  I actually think that ```\put``` is putting something in that tricks it?  Still puzzling over that...

brucemiller:
ok, got it...should talk...

dginev:
Fixed via f3f99b83dcde16208cb48d62f499bde7db59c210

PR #1148, opened 2019-05-28
Simple fix for travis

dginev:
"Forget and forgive" approach: Drop old version without regenerating docker image.


dginev:
Just in case it gets too frustrating to fish around and you want a fallback. Feel free to close if you do the legwork for regenerating.

brucemiller:
Ended up doing the same, myself, but adding some comments so I don't go through this again...  But thanks!

PR #1154, opened 2019-06-18
Patch infinite loop in some \newtheorem uses

dginev:
Once in a blue moon I take a look at the arXiv timeout Fatals, to check if some easy infinite loop case can be spotted. I ended up hitting one at the frist document I checked from the 1905 testbench, a minimal example being:

```tex
\documentclass{article}
\usepackage{amsmath}
\usepackage{amsthm}

\newtheorem{thm}{\theoremname}
\newtheorem{example}[thm]{\protect\examplename}
\providecommand{\examplename}{Example}

\begin{document}
\begin{example}
	test
\end{example}
\end{document}
```

As it turns out, latexml defines `\examplename` at the point of the `\newtheorem` for `example`, while the actual amsthm.sty definitions don't go as far. I would need to study the exact logic in newtheorem for some more time to be able to understand if a general upgrade is possible, but can already contribute a simple patch for now.

It also makes me wonder how many of the arXiv jobs could be sped up by detecting an infinite expansion chain of the same macro...

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -2707,7 +2707,12 @@ sub defineNewTheorem {
       => { map { ($_ => LookupValue($_)) } keys %parameters }, 'global');
 
   my $thmname = '\\' . $thmset . 'name';
-  DefMacroI($thmname, undef, $type, scope => 'global');
+  # Avoid weird infinite loops, when the $type is already containing a $thmname macro
+  # TODO: What is a better way to stay TeX-near and avoid these issues?
+  my $type_str = ToString($type);
+  if ($type_str !~ /\Q$thmname\E\b/) {
+    DefMacroI($thmname, undef, $type, scope => 'global');
+  }
   my $swap = LookupValue('thm@swap');
 
   DefMacroI('\fnum@' . $thmset, undef,


dginev:
Ran into a second paper with this type of loop, via:
```tex
\providecommand{\algorithmname}{Algorithm}
\floatname{algorithm}{\protect\algorithmname}
```

So the `\newfloat` binding definition in `float.sty.ltxml` would also need the same type of defensive patch.  It already attempts to defend via:
```perl
    if (!LookupDefinition(T_CS('\\' . $type . 'name'))) {
      DefMacroI('\\' . $type . 'name', undef, $type); }
```

But as the examples show, the definition of the name macro happens *after* the environment/float definition is completed. So the general refactor may be a better way to proceed...

brucemiller:
Hmm, I wonder if it's a case of over-generalizing defineNewTheorem (which has analogs in several packages). I suspect the name macro was either defined in some cases, so I did it for all, or that the name was wanted for some of the clever ref packages.  Need to do some tracking...

brucemiller:
Or... it was a generalization of a trick often used in latex (defining a <something>name macro for generating tags, references, etc). Maybe I need another layer of indirection in all the ```\lx@tag``` machinery.
(I do have a recursion patch coming, ,though)

dginev:
That's great! Let me know if I can help for some technicality. Maybe we turn both timeouts into a single new test file? Here is the tex snippet that exhibits both:

```tex
\documentclass{article}
\usepackage{amsmath}
\usepackage{amsthm}
\usepackage{float}

\newtheorem{thm}{\theoremname}
\newtheorem{example}[thm]{\protect\examplename}
\providecommand{\examplename}{Example}

\begin{document}
\begin{example}
	test
\end{example}

\providecommand{\algorithmname}{Algorithm}
\floatstyle{ruled}
\newfloat{algorithm}{thp}{lop}
\floatname{algorithm}{\protect\algorithmname}

\begin{algorithm}
 test
\caption{A test algorithm}
\end{algorithm}

\end{document}
```

brucemiller:
So there are several commits addressing different parts of this, which include: avoiding defining \<type>name macros in favor of \lx@name@<type> (when the original package seems not to define the former); a patch to avoid direct recursions in macro expansions; and a test case which slightly elaborates on your suggestion above.  With all the patches together, the test case does not trigger the recursion, and current tests work as good (or better) than before.

I used bits & pieces of your patch, especially the focus :>

dginev:
The direct recursion guard has the potential to save tons of arXiv documents from being busy timeouts, exciting!

PR #1160, opened 2019-06-24
raw interpretation of blindtext.sty

dginev:
Interesting stumble on a lipsum-like package that latexml can interpret directly (even in texlive 2018!).

A minimal example looks like:
```tex
\documentclass{article}
\usepackage{blindtext}
\begin{document}
\Blindtext{4}
\end{document}
```

which generates 4 lipsum paragraphs. Works identically in pdflatex and latexml's raw sty interpretation. The one interesting catch is that blindtext assumes the modern pdflatex behavior -- which I just learned about -- of preloading parts of the babel support. In particular, it expects the macro `\languagename` to be available, and to my surprise it is. This minimal latex documents works great under pdflatex:

```tex
\documentclass{minimal}
\begin{document}
test \languagename
\end{document}
```

I grep-ed around, but it looks like the availability of that macro happens somewhat "secretely" in the pdflatex bootup.

In any case, this could be an easy merge/win for arXiv coverage, 1.4% of tasks in the 1905 testbench:
https://corpora.mathweb.org/corpus/arxiv_1905/tex_to_html/warning/missing_file?all=true

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -391,6 +391,7 @@ lib/LaTeXML/Package/bbm.sty.ltxml
 lib/LaTeXML/Package/bbold.sty.ltxml
 lib/LaTeXML/Package/beton.sty.ltxml
 lib/LaTeXML/Package/bibunits.sty.ltxml
+lib/LaTeXML/Package/blindtext.sty.ltxml
 lib/LaTeXML/Package/bm.sty.ltxml
 lib/LaTeXML/Package/book.cls.ltxml
 lib/LaTeXML/Package/bookman.sty.ltxml
--- a/lib/LaTeXML/Package/blindtext.sty.ltxml
+++ b/lib/LaTeXML/Package/blindtext.sty.ltxml
@@ -1,0 +1,25 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# | blindtext.sty                                                       | #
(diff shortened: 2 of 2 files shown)


myexampl35-dotcom:
> Interesting stumble on a lipsum-like package that latexml can interpret directly (even in texlive 2018!).
> 
> A minimal example looks like:
> ```tex
> \documentclass{article}
> \usepackage{blindtext}
> \begin{document}
> \Blindtext{4}
> \end{document}
> ```
> 
> which generates 4 lipsum paragraphs. Works identically in pdflatex and latexml's raw sty interpretation. The one interesting catch is that blindtext assumes the modern pdflatex behavior -- which I just learned about -- of preloading parts of the babel support. In particular, it expects the macro `\languagename` to be available, and to my surprise it is. This minimal latex documents works great under pdflatex:
> 
> ```tex
> \documentclass{minimal}
> \begin{document}
> test \languagename
> \end{document}
> ```
> 
> I grep-ed around, but it looks like the availability of that macro happens somewhat "secretely" in the pdflatex bootup.
> 
> In any case, this could be an easy merge/win for arXiv coverage, 1.4% of tasks in the 1905 testbench:
> https://corpora.mathweb.org/corpus/arxiv_1905/tex_to_html/warning/missing_file?all=true

59b1251759cb204e301bc51072c52f2fc10f41f5

PR #1163, opened 2019-06-26
Fix preload options parsing

stamer:
latexml --preload=latexml[nobibtex] will not work without this patch.


dginev:
Hi Heinrich!

I think the command you were looking for is `--preload=[nobibtex]latexml.sty`, which should work as intended? Unsure if we want to allow the optional argument at the end of the package name, since it was originally motivated by the latex syntax e.g. `\uspackage[a4]{article}`

stamer:
Hi Deyan,

Now I understand, a suggestion to use --preload was the other way around, and I was suprised, that this did not work, but did not think about moving the option to the front.

So I'll just close the PR.

PR #1164, opened 2019-06-27
Make epub work under cygwin

dginev:
Cpantesters failed for cygwin with the epub test failing to find the CSS assets.

Reading around, I'm suspecting absolute paths don't agree in that setup, as discussed here:
https://www.cs.unc.edu/~jeffay/dirt/FAQ/cygwin-perl.html

As it turns out, the `-f` check was just "extra safety" but wasn't necessary due to the following asset name checks. So, hoping this was indeed the underlying issue, submitting this patch which simply omits the `-f` check for epub asset packing.


brucemiller:
Worth a try!

Issue #980, opened 2018-05-01
bibtex: macro provisions in one file do not survive to the next. 


kohlhase:
I use 
```
\documentclass{article}
\begin{document}
\cite{KopKohTab:mpdrmm18}
\bibliographystyle{alpha}
\bibliography{preamble,kwarcpubs}
\end{document}
```
where `preamble.bib` contains
```
@PREAMBLE{
{\providecommand\seen{seen }
\providecommand\March{March}
}
```
and `kwarcpubs.bib` contains
```
@Misc{URL:demo:arXMLiv,
  howpublished = {Webpage at\url{http://trac.kwarc.info/arXMLiv/wiki/Demo}},
  note = {\seen {\March} 2011}}
```

This runs well with `pdflatex`, but not with LaTeXML. 
```
3 undefined macros[\seen,\March]
```

dginev:
If you used `latexmlc` to convert them, that is indeed the case, as I am using a separate converter object for each bibliography, and each bib file is processed in isolation.

I could use the main converter object if I make provisions to keep it around during post-processing, but curious to hear if @brucemiller thinks this is the right approach. It's certainly the most tex-like one. An alternative would be to pass a state object around, but I am unsure if that gains anything, and may be more confusing than helpful.

brucemiller:
On 05/01/2018 02:47 AM, Deyan Ginev wrote:
> If you used |latexmlc| to convert them, that is indeed the case, as I am 
> using a separate converter object for each bibliography, and each bib 
> file is processed in isolation.

The situation would be even worse using latexml+latexmlpost;
it points up a fundamental limitation in the model of processing
the bib file separately (and thus xml for bib files) -- you've lost
the context of the main document.

> I could use the main converter object if I make provisions to keep it 
> around during post-processing, but curious to hear if @brucemiller 
> <https://github.com/brucemiller> thinks this is the right approach. It's 
> certainly the most tex-like one. An alternative would be to pass a state 
> object around, but I am unsure if that gains anything, and may be more 
> confusing than helpful.

Pretty much everything needed should be in the state; i guess the
only issue may have to do with start/finish (sub)tasks?

An alternative approach would be to extract all the needed packages
from the main doc's PI's, as is done when latex is invoked to create
images of TeX snippets.  This would allow processing in phases (latexml+
latexmpost (and other applications)), but in the simplest case
would require loading packages twice.  It might be that in the long
run both should be possible?

brucemiller:
[...]

Issue #981, opened 2018-05-01
various bibtex-related problems


kohlhase:
I am using ```bibtest.tex```
```
\documentclass{article}
\usepackage{url}
\begin{document}
\cite{Ginev-11,}
\bibliographystyle{alpha}
\bibliography{test}
\end{document}
```
with ```test.bib```
```
@MastersThesis{Ginev-11,
  author =   {Deyan Ginev},
  title =    {The Structure of Mathematical Expressions},
  school =   {Jacobs University Bremen},
  address =  {Bremen, Germany},
  type =     {Master's Thesis},
  month =    aug,
  url = {http://kwarc.info/people/dginev/publications/DeyanGinev_Mastersthesis.pdf},
  howpublished = {\url{http://kwarc.info/people/dginev/publications/DeyanGinev_MScThesis.pdf}},
  pubs = {dginev,omdoc,openmath},
  keywords = {lamapunbibs,mscthesis},
  year =     {2011}}
```
and get 
```
MiKo:tmp kohlhase$ latexmlc --destination bibtest.html bibtest

(Loading /Users/kohlhase/localmh/ext/perl5lib/lib/perl5/LaTeXML/Package/TeX.pool.ltxml...
(Loading /Users/kohlhase/localmh/ext/perl5lib/lib/perl5/LaTeXML/Package/eTeX.pool.ltxml... 0.00 sec)
(Loading /Users/kohlhase/localmh/ext/perl5lib/lib/perl5/LaTeXML/Package/pdfTeX.pool.ltxml... 0.00 sec) 0.11 sec)

latexmlc (LaTeXML version 0.8.2; revision 38fc034e)
processing started Tue May  1 07:57:50 2018

(Digesting TeX bibtest...
(Processing content /Users/kohlhase/tmp/bibtest.tex...
(Loading /Users/kohlhase/localmh/ext/perl5lib/lib/perl5/LaTeXML/Package/LaTeX.pool.ltxml... 0.11 sec)
(Loading /Users/kohlhase/localmh/ext/perl5lib/lib/perl5/LaTeXML/Package/article.cls.ltxml... 0.01 sec)
(Loading /Users/kohlhase/localmh/ext/perl5lib/lib/perl5/LaTeXML/Package/url.sty.ltxml... 0.00 sec) 1.32 sec) 1.33 sec)
(Building...
[...]

PR #1171, opened 2019-07-21
neurips binding and reliable preloads for bib conversion

dginev:
Starting to take the "make sure your own arXiv submissions convert well!" mantra seriously.

This PR adds a simple binding for a base `neurips.sty` package, which has several variations in use in arXiv (the one I used is `neurips_2019.sty`).

More interestingly, I used `\url` consistently in my bibtex entries this time, so preloading the right macros was crucial to get the right output. Luckily the code was already there it was just being shy(?) and only letting inputenc through. I now made it possible to preload the full dependency tree, which gets a full "no problem" post-processing on my article!

I don't think this PR will have significant impact on arXiv, but it does get my paper error-free and looking good!

### DIFF
--- a/lib/LaTeXML/Post.pm
+++ b/lib/LaTeXML/Post.pm
@@ -203,8 +203,13 @@ sub generateResourcePathname {
   $doc->cacheStore($counter, $n);
   return pathname_make(dir => $subdir, name => $name, type => $type); }
 
-# Get a list [class,classoptions, oldstyle],[package,packageoptions],...]
-# The options are strings
+# Returns a two-part list of the form:
+#
+# [class, classoptions, oldstyle], [package1,package1options], [package2,package2options], ...
+#
+# Where there first element is always the *class* (if none, "article" returned as a default)
+# And all following elements are *packages*
+#
 sub find_documentclass_and_packages {
   my ($self, $doc) = @_;
   my ($class, $classoptions, $oldstyle, @packages);
--- a/lib/LaTeXML/Post.pm
+++ b/lib/LaTeXML/Post.pm
@@ -218,12 +218,10 @@ sub find_documentclass_and_packages {
       $classoptions = $$entry{options} || 'onecolumn';
       $oldstyle     = $$entry{oldstyle}; }
     elsif ($$entry{package}) {
-      push(@packages, [$$entry{package} . ".sty", $$entry{options} || '']); } }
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Post.pm
```diff
@@ -218,11 +218,12 @@ sub find_documentclass_and_packages {
...
       $oldstyle     = $$entry{oldstyle}; }
     elsif ($$entry{package}) {
-      push(@packages, [$$entry{package}, $$entry{options} || '']); }
-  }
+      push(@packages, [$$entry{package} . ".sty", $$entry{options} || '']); } }
   if (!$class) {
     Warn('expected', 'class', undef, "No document class found; using article");
     $class = 'article'; }
+  if ($class !~ /\.cls$/) {
+    $class = $class . ".cls"; }
```
dginev: There was also a weird oversight with the return value of this method, which returns a single list of names, but loses the distinction between classes and packages. When I tested on my bibliography, it had returned `article` as the class, and hence the preload assumed it was referring to a package, loading `article.sty`. Luckily that threw an error during loading so I noticed it. 

I think the safe way out is to append the extensions to the output list and preserve the distinction, which is what my commit does.

dginev:
As it happens, I believe this PR would fix #980  and #981  which coincide with the "`\url` use in bibtex" problem.

lib/LaTeXML/Post.pm
brucemiller: it's a single list, but the class (even undef) is always first. Wouldn't it be better to fix your preload? Did you check the other callers of ```find_documentclass_and_packages```?

dginev: Oh, there is a return convention where the first element is the class, and if no class is present the result is `undef`? Ouch. I could rework to respect that, but it feels a little 'magical' to me. Would have preferred two array refs, one with zero-or-more classes, followed by one with zero-or-more packages, to make the convention more visible (though I do know that hurts performance a little...)

I did check other callers yes, was not able to spot this convention from them on my own reading.

dginev: Wish we had types at points like these... The existing use in MakeBibliography was using the method incorrectly (treating class and packages together), but I started from it and assumed it was correct.

So yes, I can refactor and continue using the existing method structure, by just refactoring the MakeBibliography bits. Thanks for the suggestion!

dginev: Done!

PR #1173, opened 2019-07-24
avoid Mouth time-travel when preparing url from an XUntil context

dginev:
Fixes #1172 

Bug is fixed by simply consistently using the Gullet API from all binding levels. `\url` was being too low-level and was working directly on a Mouth object, not accounting for contexts in which tokens were already added to the Gullet's `{pushback}` stack, such as the `XUntil` parameter type read for `\institute`.

The PR produces the desired behaviour in the issue example.

PR #1174, opened 2019-07-25
Prefer Gullet to Mouth API in binding closures

dginev:
Follow up to #1173 , I hunted for other mentions of mouths in `*.ltxml` files and refactored the similar patterns with clear Gullet analogues. Needs a review, could add some robustness.


brucemiller:
Way to many iffy things going on here; Not sure if you're good or lucky or that there aren't sufficient test cases :>  General issue is that when catcodes are being mangled (verbatim, urls, etc), you don't want tokens to have been read-ahead (and pushed back to gullet), since they'll have the wrong catcodes.  This leads either to errors or unexpected behaviour.  So all this needs some more extensive testing and/or exploring.  I'm particularly concerned with arbitrarily not expanding in calc.

dginev:
Curious objection! The problem is that you could always engineer a tiny TeX test where an XToken parameter type (rare as it is) interacts with these mouth-oriented handlers, and thus concoct a bug where the wrong argument is read (what I called "time-travel in mouth").

So *if* there is legitimate worry about catcodes being wrong when dealing with `{pushback}` gullet tokens, *then* it may warrant a separate upgrade to the way tokens are re-ingested from `{pusback}`, ideally with the TeX B book open on the right pages... 

But I did mention all tests pass right? :> Can add specific tests for the bits of code I am changing if that sounds potentially helpful?

brucemiller:
Oh, I mixed up some patches that were in the other PR (on expansion), particularly calc.

I'll have to look through more carefully to see what is good here and what isn't. Particularly suspect are the cases with readline (such as in verbatim environments). There are peculiar situations where Time Travel is *exactly* what TeX/LaTeX does, like 
```
\begin{mystery} Some ignored text
Not ignored text
\end{mystery}
```

dginev:
Rebased on master and retested, still aces all tests without any issues :+1:

brucemiller:
Just to verify: This isn't addressing any specific bug, but trying to proactively head them off, right?  If so, my thinking for at least some of the changes is to synthesize some weird tests to make sure they work "correctly" (verbatim stuff can be surprising).

dginev:
That is right, the specific bug was in #1173, this one is just a defensive refactor that tries to sanitize our API use

brucemiller:
OK, this will be peculiar, possibly annoying.
Firstly, I was concerned because the Gullet analogues of Mouth methods that you've used are subtly different. They may lead to breakage (or inadvertent correction?), even though all the tests pass (maybe some missing tests?). But in short experiments, I couldn't break either version. In part because the verbatim situations you've fixed here are very restrictive and even pdflatex will always fail before you encounter the kinds of "time travel" issue in the earlier #1173. 

But then the rationale for the patch wasn't to fix anything, but to have a cleaner, more sensible API, or such. While that's a sentiment with which I strongly agree, in general, that's perhaps exactly why we should *not* change it here: Looking at ```\verb``` and friends, you *should* be scared!

dginev:
I don't think having a scary piece of logic necessitates having a scary API, but it certainly would necessitate scary test cases and lots of warning comments... I could try and create a time-travel bug with `\verb` myself next, maybe that helps move the PR along to some resolution.

brucemiller:
Finding (or not) a time triavel bug could help. But what I think is more important is understanding the differences between the Gullet & Mouth methods and synthesizing (if possible) a test that acts differently between them. The version that matches TeX behaviour wins!

dginev:
Alright, here is the test, pretty direct to synthesize:
```tex
\documentclass{article}
\def\hello{Hello to all the verbatim text one could want}
\begin{document}
\expandafter\verb\expandafter|\hello .|
\end{document}
```

In pdflatex you get all the text, and the dot, inside verbatim. In latexml you get an error and misplaced XML:
```
$ latexml verb.tex 
latexml (LaTeXML version 0.8.4; revision d320f96b)
processing started Sun Sep  8 18:05:55 2019

(Digesting TeX verb...
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/TeX.pool.ltxml...
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/eTeX.pool.ltxml... 0.00 sec)
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/pdfTeX.pool.ltxml... 0.00 sec) 0.10 sec)
(Processing content /tmp/verb.tex...
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/LaTeX.pool.ltxml... 0.12 sec)
(Loading /home/deyan/perl5/lib/perl5/LaTeXML/Package/article.cls.ltxml... 0.01 sec) 0.14 sec)
Error:expected:\end{document} Input ended while environment document was open
	at String; line 1 col 0 - line 1 col 0
 0.24 sec)
(Building...
(Loading compiled schema /home/deyan/perl5/lib/perl5/LaTeXML/resources/RelaxNG/LaTeXML.model... 0.01 sec). 0.03 sec)
(Rewriting... 0.00 sec)
(Finalizing... 0.00 sec)
Conversion complete: 1 error.
processing finished Sun Sep  8 18:05:56 2019
```

```xml
<?latexml searchpaths="/tmp"?>
<?latexml class="article"?>
  <para xml:id="p1">
    <p><verbatim font="typewriter">|
\end{document}</verbatim>—Hello to all the verbatim text one could want</p>
  </para>
</document>
```

Now, if I rerun the same test with this PR, I get the expected no_problems and:
```xml
<?latexml searchpaths="/tmp"?>
<?latexml class="article"?>
  <para xml:id="p1">
    <p><verbatim font="typewriter">Hello to all the verbatim text one could want.</verbatim></p>
  </para>
</document>
```

I'll add this as a test to this PR if you'd prefer. Although this is very standard TeX behavior - anything that is reached via an `\expandafter` return to the Gullet via the `{pushback}` and any use of the Mouth API will be wrongly missing those tokens.
```

**Comment by dginev** [comment]
Test added, should be good to go.

**Comment by brucemiller** [comment]
OK. And how about the differences in char/token/line testing?

**Comment by dginev** [comment]
You mean a test that has to ignore pushback and would only work with the Mouth API? I would have to think hard to discover such a thing, since it would mean the binding is written in a suspicious way... Pushback tokens are "real" when the interpretation happens as dictated by TeX

**Comment by brucemiller** [comment]
No, nothing to do with pushback. It has to do with the fact that you've switched Mouth methods for Gullet ones that are nominally, but not exactly, equivalent. There are subtle differences in comparisons, and so on. Maybe they function exactly the same in the cases in question. Maybe not. Maybe the Gullet ones are even more correct (beyond the pushback handling, which apparently is more correct, at least for ```\verb```)

**Comment by dginev** [comment]
In fear of repeating myself, I'll just say I can't follow without an example. Which exact interplay would you like me to add a test for from the changed files? The other specific changed macros are `\URL`, `\verbatim@`, `\newverbtext`.

**Comment by brucemiller** [comment]
Look at the diff: Mouth->readTokens is not the same as Gullet->readUntil, even besides pushback; the two readRawLines are different.

**Comment by dginev** [comment]
1. The gullet `readRawLine` is a pushback-aware version of the Mouth's readRawLine, which it always calls.

2. The `Mouth->readTokens` method called with an `$until` value already felt like it should have been named `readUntil`. It seems that the gullet `readUntil` also allows balanced sub-arguments, which the Mouth method does not, which I am not particularly certain as to the reasoning behind. Would be nice to get some extra information there.

In the cases where there are no "balanced" sub-arguments, the gullet `readUntil` is a pushback-aware version of the mouth method, as it does a `readToken` at a time until the `$until` is met.

**Comment by dginev** [comment]
Now that I started digging, I found an example that fails in both master and this PR w.r.t to the `\verb` behaviour. You can expand a delimiter before `\verb` starts expanding, and it will be accepted as a regular token. Here's the snippet:

```tex
\documentclass{article}
\begin{document}
\def\nested{Nested | is allowed}
\expandafter\verb\expandafter|\nested .|
\end{document}
```

dginev:
Can now pass that test as well, by taking care of the delimiter activation sequence, PR updated!

dginev:
The next difference we haven't emulated is getting an error when `\verb` is terminated by an end-of-file. pdflatex prints `! LaTeX Error: \verb ended by end of line.` in those cases.

dginev:
I think I can now summarize `\verb`'s reading behaviour as "read from pushback and the current mouth, raising an error if the closing delimiter is missing / we reach end of the current mouth". Would be nice if that can be captured by a gullet method call, be it a `readUntil` with a set of parameters, e.g. 
```perl
my $body = $gullet->readUntil(until=>$token, mouth=>'current', balanced_checks=>0);
```

brucemiller:
This is looking promising!  As for the differences between readTokens and readUntil: presumably the weird test in readTokens was a poor attempt to deal with ```\verb``` setting the marker to active?; and the fact that readUntil reads balanced shouldn't matter if {} are set to other.  If we can clean up the other 2 uses (ams_support, verbatim), we can probably get rid of the $until arg in Mouth->readTokens and clear up the code & future confusion.  I suspect the ams case should get a url-like treatment as in url etal.  The verbatim case probably needs some probing.

brucemiller:
So you convinced yourself that those latter 2 cases are good using Gullet?

dginev:
They're certainly at least as good, while also avoiding the pushback errors. As mentioned there is more to emulate (end-of-file error signaling) for some edge cases, but the regular valid use should be solid.

brucemiller:
What's the worst that could...

PR #1177, opened 2019-07-28
Use algorithmic instead of algorithmicx in icml*.sty

mkardas:
All icml style files I've tested (2013-2019) require `algorithmic` package, but `algorithmicx` doesn’t define any `algorithmic` commands. [Here's](https://arxiv.org/pdf/1402.5766v1.pdf) an example of paper (using icml2014.sty) on which LaTeXML hangs during processing when `algorithmicx` is used.


dginev:
Sounds like an oversight? Looks good to merge!

brucemiller:
Or even just a typo! Thanks!!

PR #1181, opened 2019-08-04
Add support for DockerHub auto-builds (fixes #1178)

### LINKED ISSUE #1178: Publishing Docker image to Docker Hub
Hi, do you have any plan for publishing a Docker image built with [release/Dockerfile](https://github.com/brucemiller/LaTeXML/blob/master/release/Dockerfile) to Docker Hub?
There may be a problem mentioned in https://github.com/brucemiller/LaTeXML/pull/1008#issuecomment-399694452

> Sadly DockerHub (for automated builds that @bfirsh suggested above) does not seem to support build arguments, so it is non-trivial to automate builds for both variants of the image.

but an image with default build arguments (`ARG WITH_TEXLIVE="yes"`) would be a great help for us.

tkw1536:
I've re-read the DockerHub documentation and figured out that it does support [custom build hooks](https://docs.docker.com/docker-hub/builds/advanced/). As such, I have to revise https://github.com/brucemiller/LaTeXML/issues/1178#issuecomment-517242737. With this PR I would like to merge a branch that has auto-builds (i.e. docker image build on every push) enabled. The generated images can be found at  [tkw01536/latexml](https://hub.docker.com/r/tkw01536/latexml/builds). 

To speed up the build process I made the following two optimizations:
- do not include TeXLive (i.e. `--build-arg WITH_TEXLIVE=no`)
- do not run perl tests during the build (i.e. installing with `cpanm --notest .`)

With these optimizations, building seems to take between 10-20 minutes for each commit.

Please note that deploying DockerHub auto-builds for the main LaTeXML repository will require someone with Owner rights on the GitHub repository to configure them.

### DIFF
--- a/release/docker/Dockerfile
+++ b/release/docker/Dockerfile
@@ -15,17 +15,24 @@
 # This Dockerfile expects the root directory of LaTeXML as a build context. 
 # To achieve this run the following command from the root directory:
 #
-# > docker build -t latexml -f release/Dockerfile .
+# > docker build -t latexml -f release/docker/Dockerfile .
 
 # This Dockerfile can include a full TeXLive installation. 
 # This is enabled by default however it can be disabled by providing a
 # build argument like so:
 #
-# > docker build -t latexml --build-arg WITH_TEXLIVE=no -f release/Dockerfile .
+# > docker build -t latexml --build-arg WITH_TEXLIVE=no -f release/docker/Dockerfile .
 #
 # Please be aware that including a full TeXLive installation can take a
 # significant amount of time (depending on network connection) and will
 # increase the image size to several Gigabytes. 
+#
+# Futhermore to speed up the build process, it is also possible to
+# tell docker not to run the tests during the build proess. To achieve
+# this, pass --build-arg WITH_TESTS=no to the docker build command, e.g:
+#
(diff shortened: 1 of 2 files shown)


brucemiller:
Cool! I think -- I'm not sure I understand the implications. Just to make sure: Accepting this PR does *not* obligate me to do anything with Docker (yet); is that correct?

dginev:
@tkw1536 maybe you could enumerate which/if any steps are needed here to enable the auto-builds once the PR is merged? I think it's a pretty safe PR to land in master, even if the configure takes a little bit of time to figure out, nothing existing should be affected.

tkw1536:
@brucemiller Correct, yes this PR itself does not obligate you to do anything Docker. The commit mere adds support to do it at all. 

@dginev with this PR merged, to enabled DockerHub autobuilds one has to
1. Create an account on [DockerHub](https://hub.docker.com/) and link it to a GitHub account with admin access to this repo. I would recommend using a dedicated bot account for this (see section "Service users for team autobuilds" of the dockerhub autobuild documentation linked below). 
2. Create an [automated build](https://docs.docker.com/docker-hub/builds/) and select `release/docker/` as the Dockerfile path. 
Then the [build hook](https://github.com/brucemiller/LaTeXML/blob/master/release/docker/hooks/build) should take care of everything else and enable auto-magically creating new docker images every push.

PR #1182, opened 2019-08-06
[WIP] TeX-faitfhul errors on undefined command sequence expansion

### LINKED ISSUE #1180: Discrepancy in expansion of undefined command sequences
This issue goes towards further parity with the native tex interpretation, towards the `1.0` milestone of latexml.

Minimal example:
```tex
\documentclass{article}
\edef\expandedondef{\tricky~\tricky}
\def\tricky{tricky!}

\begin{document}
\expandedondef
\end{document}
```

In pdflatex this produces an empty document with an error mentioning `\tricky` is undefined at time of `\edef` execution. In latexml, no error is ever displayed, and what's worse - the new "late binding" of `\tricky` is used in the final output, questioning if `\edef` ever expanded the macro.

dginev:
Fixes #1180 . 

And likely breaks a range of hard-to-foresee arXiv jobs, and third party raw TeX uses. I think my fixes definitely need some review and discussion/thought, as I touched bindings I am not closely familiar with.

The general concept is simple enough - expansion must *always* have defined command sequence arguments. And all *good* uses of expansion already do. The breakage in latexml happens at the fringes -- e.g. bindings that take very Perl-y approaches and leave certain command sequences undefined, or raw interpretation that silently ends up accepting invalid inputs where it should alert us with an error that latexml needs further improvement. I gave two examples in the issue of error cases, which now work just the same as in pdflatex with this PR. And tests pass with my modifications.

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -182,14 +182,15 @@ sub show_pushback {
 #**********************************************************************
 # Not really 100% sure how this is supposed to work
 # See TeX Ch 20, p216 regarding noexpand, \edef with token list registers, etc.
-# Solution: Duplicate param tokens, stick NOTEXPANDED infront of expandable tokens.
+# Solution: Duplicate param tokens, stick NOTEXPANDED infront of expandable+undefined tokens.
 sub neutralizeTokens {
   my ($self, @tokens) = @_;
   my @result = ();
   foreach my $token (@tokens) {
     if ($$token[1] == CC_PARAM) {    # Inline ->getCatcode!
       push(@result, $token); }
-    elsif (defined(my $defn = LaTeXML::Core::State::lookupDefinition($STATE, $token))) {
+    elsif (!defined(my $meaning = LaTeXML::Core::State::lookupMeaning($STATE, $token)) ||
+      defined(my $defn = LaTeXML::Core::State::lookupDefinition($STATE, $token))) {
       push(@result, Token('\noexpand', CC_NOTEXPANDED)); }
     push(@result, $token); }
   return @result; }
@@ -271,6 +272,9 @@ sub readXToken {
                 : ($r eq 'LaTeXML::Core::Tokens' ? @$_
                   : Fatal('misdefined', $r, undef, "Expected a Token, got " . Stringify($_))))) }
             @{$r}); } }
+    elsif ($cc == CC_CS && !(LaTeXML::Core::State::lookupMeaning($STATE, $token))) {
+      Error('undefined', $token, $self, "The token " . Stringify($token) . " is not defined during expansion. Consuming it and proceeding, expect trouble...");
+      return; }
     else {
       return $token; }    # just return it
   }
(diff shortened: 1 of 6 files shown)


dginev:
Also, @brucemiller , I suspect rebasing on this branch may offer extra insight into debugging expl3 and other related raw interpretation cases, if we're lucky. Could make it very explicit that expansion is incorrect at a certain point in processing.

dginev:
I've verified the showcase Tikz images work with this branch as well, which is reassuring.

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -182,14 +182,15 @@ sub show_pushback {
...
+# Solution: Duplicate param tokens, stick NOTEXPANDED infront of expandable+undefined tokens.
 sub neutralizeTokens {
   my ($self, @tokens) = @_;
   my @result = ();
   foreach my $token (@tokens) {
     if ($$token[1] == CC_PARAM) {    # Inline ->getCatcode!
       push(@result, $token); }
-    elsif (defined(my $defn = LaTeXML::Core::State::lookupDefinition($STATE, $token))) {
+    elsif (!defined(my $meaning = LaTeXML::Core::State::lookupMeaning($STATE, $token)) ||
+      defined(my $defn = LaTeXML::Core::State::lookupDefinition($STATE, $token))) {
```
dginev: Wondering if this is best left without either check, just always add the `\noexpand`. We have an example in the tests that needs `\noexpand` added for undefined CSes.

brucemiller:
OK, I mixed up responses to PR's. There may be some good fixes here, but there's definitely some fishy stuff (esp. calc, siunitx); it seems a regression. And the motivating issue is about "incorrect" documents, nothing indicates that the issue or patches would change the behavior of a correct document (at least not for the better).  It might make debugging easier by failing earlier.

dginev:
Ok, I will try to restore at least calculus back to doing `->readXToken`, but it may entail different parts of the binding need a refactor. The problem is that some of the bindings are silently leveraging the lenience of `->readXToken` to return undefined tokens as a feature, so when I introduce the texy error message, that technique breaks. So some best practice refactor needs to be invented.

Also, I agree this PR shouldn't break existing correct behaviour, not my intention. Was fishing for some feedback :+1:

dginev:
Ok @brucemiller I struggled some more and partially managed to simplify the binding modifications. The calc in particular should not worry you at all anymore. For the siunitx, I found a piece of trickery as well, but got scared for alignments, and allowed them to keep the old `readXToken` behavior for now. To switch to consistently expanding I would need to rewrite a part of `\@@open@inner@column` which is above my pay grade at this instant. That said, it is probably with long-term benefit to get that bit of code closer to the TeX behavior as well (though your `halign` work may be touching on that already...).

Turns out there are two new best practices one can use to save failing bindings that were leveraging the "expanding undefined command sequences" behavior:
 1. Use a noexpand trick: 
    ```perl
    DefMacro('\minof',    '\noexpand\minof');
    ```
 2. In tighter local scopes, let to the other catcode of the name: 
    ```perl
    Let($token, T_OTHER($token->getCSName));
     ````

I've only needed them in the files you can see changed in the PR, but we may be missing tests for some of the other bindings... I don't really feel comfortable merging this very soon, but would definitely consider it relevant for the 1.0 milestone.

dginev:
Note to self: I should add the two TeX files from the issue as tests, to avoid regressing.

dginev:
Closing in favor of #1188

PR #1184, opened 2019-08-07
Unified conversion for multi-bib collections

### LINKED ISSUE #980: bibtex: macro provisions in one file do not survive to the next.

dginev:
Fixes #980 , but begrudgingly so.

We don't really have a clean API for passing a collection of documents for conversion (so that they reuse the same state), and there is no `\input` opportunity when we are in BibTeX mode. Which leaves me with the sole option of detecting a multi-`.bib` setup in post-processing, and concatenating the content of all of those files together, in order to have a `@preamble` from one bib file visible to all following.

This actually gets Michael's example working smoothly, but makes it annoying to debug cases where you've used several bib files and one of them encounters an error -- since the error message you'll get will be about a literal string that has nothing to do with the original files, even the line numbers are different.

In any case, I'm offering it as a PR, at least as a conversation starter.

### DIFF
--- a/lib/LaTeXML/Post/MakeBibliography.pm
+++ b/lib/LaTeXML/Post/MakeBibliography.pm
@@ -149,11 +149,11 @@ sub getBibliographies {
   # Lastly, If we found any raw .bib files/literaldata, convert and include them.
   if (@rawbibs) {
     my $raw;
-    if (scalar(@rawbibs) == 1) {    # Multiple, arrange into a single conversion payload
+    if (scalar(@rawbibs) == 1) {    # Single, just convert as-is
       $raw = $rawbibs[0]; }
     else {
       $raw = 'literal:';
-      for my $rawbib (@rawbibs) {
+      for my $rawbib (@rawbibs) { # Multiple, arrange into a single conversion payload
         if ($rawbib =~ s/literal\://) {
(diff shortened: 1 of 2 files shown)


brucemiller:
As you say, not the prettiest solution. But: in the not too distant future, we'll be basing bib conversions on a better fake bibtex, which will be somewhat more explicitly running "TeX" over a bbl-like file (hopefully with source locators), and will thus be able to inherit (or synthesize) a more appropriate State and solve many of the issues. And since we'll then be doing a major rewrite of MakeBibiliography, no harm in accepting this patch. Thanks!

PR #1188, opened 2019-08-13
Towards functioning expl3

dginev:
Combines #1182 and Bruce's `expl3` branch with additional fixes from my debugging.


dginev:
In need of further work:
 - `XUntil` parameter needs to be more careful not to expand the final match marker
 - Alignments need to be adapted to the more strict `readXToken` and the third parameter should be removed
 - `\lipsum` should work fluently, and a test case added, ideally with basic xparse support

lib/LaTeXML/Core/Definition/Register.pm
```diff
@@ -73,6 +74,16 @@ sub invoke {
...
+  return (defined $other
+      && (ref $self) eq (ref $other)) && Equals($self->getParameters, $other->getParameters)
+###    && Equals($$self{replacement},  $$other{replacement})
+    && ($self->isRegister eq $other->isRegister)
+    && Equals($$self{getter}, $$other{getter})
+##    && Equals($$self{beforeDigest}, $$other{beforeDigest})
+##    && Equals($$self{afterDigest},  $$other{afterDigest})
+    ; }
 #===============================================================================
 1;
```
brucemiller: Can we checkout this back to master (pre *my* pointless commit); it's a no-op.

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -182,14 +182,15 @@ sub show_pushback {
...
-# Solution: Duplicate param tokens, stick NOTEXPANDED infront of expandable tokens.
+# Solution: Duplicate param tokens, stick NOTEXPANDED infront of expandable+undefined tokens.
 sub neutralizeTokens {
   my ($self, @tokens) = @_;
   my @result = ();
   foreach my $token (@tokens) {
     if ($$token[1] == CC_PARAM) {    # Inline ->getCatcode!
       push(@result, $token); }
-    elsif (defined(my $defn = LaTeXML::Core::State::lookupDefinition($STATE, $token))) {
+    elsif (!defined(my $meaning = LaTeXML::Core::State::lookupMeaning($STATE, $token)) ||
```
brucemiller: Isn't both meaning & defintion redundant?

lib/LaTeXML/Core/Definition/Register.pm
dginev: Confused by the phrasing "Check it out back to master". Do you mean remove the code block entirely, or do you mean "it's in master, let's rebase"?

lib/LaTeXML/Core/Gullet.pm
dginev: What I intended to add here, was also neutralizing undefined command sequences, which is what a negative `lookupMeaning` achieves.

If your point is that every time `lookupDefinition` is true, then `lookupMeaning` is also true, that may be.

The question is, are there cases where `lookupMeaning` is true, but `lookupDefinition` is false, where we do not want to add a `\noexpand`. An example of that is e.g. any `T_LETTER` token, and a variety of others. I think adding a `\noexpand` before those is simply a no-op, so it may be a "distinction without a difference"... Unsure what is cleanest.

dginev: If we could get away with adding a `\noexpand` unconditionally, before every element of the input `@tokens`, that would certainly feel most elegant to me.

lib/LaTeXML/Core/Definition/Register.pm
brucemiller: I must meant like ```git checkout ...\Register```: ie, no change to the file at all, since the only changes have no effect.

lib/LaTeXML/Core/Gullet.pm
brucemiller: Doh; missed the "!".

lib/LaTeXML/Core/Definition/Register.pm
dginev: Alright, I've removed the changes. Sadly "reverting" to the previous Register isn't simple this many commits later, I just papered over with a new commit.

dginev:
We can squash, certainly. Would you like to leave comments on the bits that need removing?

brucemiller:
Incidentally, with regard to XUntil, I may have finally come around to your POV: that it is legitimate to leave it as is, and require that if you use it, you should probably \let the delimiting token to \relax or something. As it stands, you'll get an error message (right?) which should be relatively decipherable.

dginev:
As to XUntil: yes to your question, the last token will shoot an error if it's expanded without being defined. And if you let it to relax, or some such, XUntil will succeed properly.

brucemiller:
Mostly I was thinking of removing the commented out debugging code (eg. prints or otherwise dead code)

dginev:
There is still the issue with `readXToken` having 3 arguments in t his PR, which as you mentioned is something to eliminate. So maybe we should make sure alignments work with the new expansion errors and transfer them to the new behaviour before merging?

dginev:
I've removed the debug comments, thanks for mentioning.

brucemiller:
Oh, you mean backing off the extra args to readXToken in ```\@@open@inner@column```? Yeah, if it works without, then we can go ahead & remove that mod to readXToken. Better earlier than later :>

dginev:
Indeed - I am trying that change now, but I have a subtle failure in the siunitx tests, need a few more mins...

brucemiller:
take your time; I gotta run errands :>

dginev:
Alright, feeling very good about the refactor of `readXToken` now! The method signature remains the same, and the only remaining change is the raised error when an undefined token is expanded.

No issues with alignment are visible from the tests - the once encountered with siunits in tables were related to DeclaredUnits not being defined -- let-ing them to relax when needed allowed for smooth passing of the tests. I experimented a little with literal strings and discovered siunitx isn't actually loading some of its dependencies (amsmath and array) so I threw those in for good measure.

All good on my end!

lib/LaTeXML/Package.pm
```diff
@@ -430,9 +430,14 @@ my @rmletters = ('i', 'v', 'x', 'l', 'c', 'd', 'm');    # [CONSTANT]
 
 sub roman_aux {
```
dginev: Side comment, the `roman` functions really feel like `Util`-level pieces of code, rather than `Package.pm` bits.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -148,7 +148,7 @@ DefParameterType('GeneralText', sub {
     my ($gullet) = @_;
     my $open = $gullet->readXToken;
     if ($open->equals(T_BEGIN)) {
-      return $gullet->readBalanced; }
+      return scalar($gullet->readBalanced); }
```
dginev: I am a bit stumped at this line. Returning the number of tokens instead of the tokens themselves? From within a parameter type ?!?!? Is this some sort of intermediate code that slipped by or am I missing something extremely obvious?

dginev: Good news is that all tests pass with the scalar removed. I'm tempted to just push in the change since it looks so horrendous.

dginev: Done.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -248,6 +248,19 @@ DefParameterType('Expanded', sub {
     my ($arg) = @_;
     (T_BEGIN, Revert($arg), T_END); });
 
+DefParameterType('EExpanded', sub {
+    my ($gullet) = @_;
+    local $LaTeXML::NOEXPAND_THE = 1;
+    my $token  = $gullet->readXToken(0);
+    my @tokens = ();
+    if ($token->getCatcode == CC_BEGIN) {
+      return scalar($gullet->readBalanced(1)); }
```
dginev: same scalar here...

dginev:
Please ignore all `scalar` comments. Perl is way too overloaded for my normal mortal mind. After a test finally failed when I removed all `scalar()` calls, I went and re-read wantarray, and it hit me that you've used scalar to force the evaluation of a function to return in scalar context. I had - way too simply - memorized that if you are returning an array and invoke `scalar()` you get back the number of elements, so I was expecting the `readBalanced` calls to return the scalar `2`.

But of course when you apply the scalar on the function call that ends with `wantarray` it would not return the array output and count it, but would return the scalar output... Sigh. It's a headscratcher when you see it unprepared. But all good, I restored the PR to the state before I sunk into the rabbit hole.

brucemiller:
TeeHee! I knew it'd hit you shortly :>

PR #1192, opened 2019-08-19
slight upgrade to noexpand infra

dginev:
Importantly adding a passing test that tries the noexpand bits in isolation.

The expl3 test fails with the `#` behavior we discussed with @brucemiller . I hope the code here is a slight intermediate upgrade.


lib/LaTeXML/Core/Gullet.pm
```diff
@@ -254,15 +255,13 @@ sub readXToken {
...
-      # Should only occur IMMEDIATELY after expanding \noexpand (by readXToken),
-      # so this token should never leak out through an EXTERNAL call to readToken.
-      return $unexpanded; }    # Just return the next token.
     elsif (($cc = $$token[1]) == CC_COMMENT) {    # NOTE: Inlined ->getCatcode
       return $token if $commentsok;
       push(@{ $$self{pending_comments} }, $token); }    # What to do with comments???
     elsif ($cc == CC_MARKER) {
       LaTeXML::Core::Definition::stopProfiling($token, 'expand'); }
+    elsif ($$token[2]) {                                # Inline is_dont_expand
+      return $token; }
```
brucemiller: Kinda bizarre that this still works (if it does); the whole point is if you're forced to expand a dont-expand token, you should get the *normal* token. Otherwise, it will never expand.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -731,8 +731,12 @@ DefConditionalI('\ifinner', undef, sub { 0; });
...
 DefConditionalI('\ifmmode', undef, sub { LookupValue('IN_MATH'); });
 
-DefConditional('\if XToken XToken',    sub { $_[1]->getCharcode == $_[2]->getCharcode; });
-DefConditional('\ifcat XToken XToken', sub { $_[1]->getCatcode == $_[2]->getCatcode; });
+DefConditional('\if XToken XToken', sub {
+    ($_[1]->is_dont_expand == $_[2]->is_dont_expand) &&
+      ($_[1]->getCharcode == $_[2]->getCharcode); });
+DefConditional('\ifcat XToken XToken', sub {
+    ($_[1]->is_dont_expand == $_[2]->is_dont_expand) &&
+      ($_[1]->getCatcode == $_[2]->getCatcode); });
```
brucemiller: Something like this seems necessary, but your testcase below doesn't seem to test it (at least, it passes w/o this change). I think the interesting thing about these ifs is their treatment of active chars.

lib/LaTeXML/Core/Gullet.pm
dginev: Finicky to think and talk about, let me try.

1. I _think_ that we want to get the token returned by `readXToken` marked as `dont_expand` when it was returned by that route, so that the requesting bit of logic, say `\ifcat` can be looking at it flagged appropriately and have the right comparison outcome.

2. A token is just a (string, catcode) pair, so returning `$token` or `$$token[2]` with your technique is functionally equivalent for all other TeX purposes. You still end up looking up the same meaning/definition, having the same effects, etc etc.

lib/LaTeXML/Package/TeX.pool.ltxml
dginev: Indeed, I need to add more tests, I'm also feeling a bit paranoid about `\if` and `\ifcat` interacting with the dont_expand flags.

dginev:
One meta thing to note about this PR - it's target is not master, but your `noexpand` branch, so merging it leaves master unchanged, and would only move the feature branch forward. Starting to feel like proper industry-setting development  :+1:

lib/LaTeXML/Core/Gullet.pm
brucemiller: But readXToken at ```\noexpand\something``` will return the dont-expand token to be seen by the ifs.  OTOH, returning the still-dont-expand token at this point will keep the token from *ever* expanding!

brucemiller: OBTW: Storing the normal token as a secret 3rd element of the dont-expand token is "Magic"; open-coding it makes it "Black".  Jus' sayin'

dginev: Ah, I think you may be right now that I think this through with more token. I'm working on a test case to exercise reusing a noexpand token now...

dginev: I added a test which failed with my `return $token` that passes as in the PDF with returning the inner token. Thanks for catching that. Today I feel like I am probably better of just adding new tests than thinking the details through in my head, keep getting tripped up.

dginev:
I added another handful of tests, improved the presentation of that test file a bit, and removed the "paranoid" checks for `dont_expand` in `\if` and `\ifcat`, indeed they are not needed.

dginev:
Just rebased on the `#` fixes from this afternoon, all tests now pass! Cool! cc @brucemiller . Will attempt adding the lipsum test again.

dginev:
The basic lipsum test is back to operational! Very nice.

image

lib/LaTeXML/Core/Token.pm
```diff
@@ -251,15 +251,15 @@ sub neutralize {
...
   my ($self) = @_;
   return bless [$$self[0], $$self[1], $self], 'LaTeXML::Core::Token'; }
 
 # Return the original token of a not-expanded token,
 # or undef if it isn't marked as such.
-sub is_notexpanded {
+sub is_dont_expand {
   my ($self) = @_;
-  return $$self[2]; }
+  return $$self[2] ? 1 : 0; }
```
brucemiller: I guess I like your renaming generally; more consistent. But I still think I'd rather have an accessor to the token, so how about renaming "is_dont_expand" to "get_dont_expand" and just returning ```$$self[2]```

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -254,15 +255,13 @@ sub readXToken {
...
-      # so this token should never leak out through an EXTERNAL call to readToken.
-      return $unexpanded; }    # Just return the next token.
     elsif (($cc = $$token[1]) == CC_COMMENT) {    # NOTE: Inlined ->getCatcode
       return $token if $commentsok;
       push(@{ $$self{pending_comments} }, $token); }    # What to do with comments???
     elsif ($cc == CC_MARKER) {
       LaTeXML::Core::Definition::stopProfiling($token, 'expand'); }
+    elsif ($$token[2]) {                                # Inline is_dont_expand
+      return $$token[2]; }
     # Note: special-purpose lookup in State, for efficiency
```
brucemiller: since we're grasping for efficiency, isn't ```my $unexpanded = $$token[2]``` infinitesimally better?

dginev: indeed

lib/LaTeXML/Core/Token.pm
dginev: sounds good

brucemiller:
Looks good. Merge to the branch, and maybe do some external testing...

dginev:
Should I next run the `noexpand` branch over the `--includestyles` version of the arXiv testbench?

Actually, I should first run the testbench with `--includestyles` on master, and then rerun it with the `noexpand` branch, so that we can keep track of the history deltas.

I will also only use my home texlive 2019 machine, to ensure some clearcut stress-testing.

PR #1193, opened 2019-08-21
Remove stex aids, now in separate plugin

dginev:
Attempting to be self-contained at https://github.com/slatex/LaTeXML-Plugin-sTeX/


brucemiller:
Thanks!

PR #1194, opened 2019-08-21
Correct latexmlc example

dginev:
It turns out [the web manual](https://dlmf.nist.gov/LaTeXML/manual/commonmodules/LaTeXML_Common_Config.html) has a misleading example for using latexmlc from the Config.pm perldoc. Fixing here, and would also suggest regenerating the officially hosted manual when there is some available time. Tripped up @tkw1536 today (sorry).


brucemiller:
Thanks!

PR #1195, opened 2019-08-23
remove --mathlex and llamapun.sty; replace with [mathlexemes]latexml.sty

dginev:
I was asked about adding a new `--option` switch today, and as I restated my belief that latexml has too many options as it is, I got reminded of a simplification I wanted to get to eventually but hadn't gotten the chance to.

The math lexemes experiment is definitely a success in terms of its final output, as we're about to preprint. But, simultaneously, the way I implemented it added experimental boilerplate which really shouldn't be necessary. This PR removes all of it, replacing it with a minimal `--preload[mathlexemes]latexml.sty`. That switch can only be used via `latexmlc`, which is in hope to also restart the conversation about making `latexmlc` the main executable of the project.

Removed: `--mathlex` and `llamapun.sty`.

I kept the math lexemes test, which continues to pass happily.


brucemiller:
This seems innocuous enough :> Why not?

PR #1197, opened 2019-08-27
Try to obey more of the noexpand directives/interop logic

dginev:
Importantly, added an extension to the `t/expansion/noexpand_conditionals.tex` test, with more of the tricky cases.

The PR changes to latexml itself are experimental and need to be thought over some more. The expl3 test has regressed, but the more targeted noexpand tests are now passing in full.


dginev:
Everything useful from this PR has now entered the `noexpand` branch behind the scenes, closing here!

PR #1198, opened 2019-08-27
Patch finalize_rec for stex ltx:text elements

dginev:
This is another one of those "too many moving parts" debugging streams, as sTeX usually is.

I localized a snippet that experienced a deep recursion on `Document::finalize_rec` when serializing, see `deep_rec.xml` at:
https://gist.github.com/dginev/4a1eaa48ca26b662a0f217da288642ad

It turned out it was looping on a trivial `ltx:text` element containing `foo`, as the `%pending_declarations` were never emptied. They contained a very artificial `{ element => {} }`, and thus the recursive call kept happening, until it bottomed out.

I am not really certain as to why/how sTeX got itself into that position, but here is a trace of what finalize_rec was seeing as arguments when it went into the loop:
```xml
finalize_rec on :
 <omdoc:rendering _box="LaTeXML::Core::Whatsit=HASH(0x55bb206a2290)" _font="Font[serif,medium,upright,10,black,white,1,OT1,*,*,0]">
  <ltx:text _font="Font[serif,medium,upright,10,black,white,1,OT1,*,*,0]" _box="LaTeXML::Core::Whatsit=HASH(0x55bb206af3d0)" class="ltx_markedasmath">foo</ltx:text>
</omdoc:rendering>
finalize_rec on : <ltx:text _font="Font[serif,medium,upright,10,black,white,1,OT1,*,*,0]" _box="LaTeXML::Core::Whatsit=HASH(0x55bb206af3d0)" class="ltx_markedasmath">foo</ltx:text>
finalize_rec on : <ltx:text _font="Font[serif,medium,upright,10,black,white,1,OT1,*,*,0]" _box="LaTeXML::Core::Whatsit=HASH(0x55bb206af3d0)">foo</ltx:text>
finalize_rec on : <ltx:text _font="Font[serif,medium,upright,10,black,white,1,OT1,*,*,0]" _box="LaTeXML::Core::Whatsit=HASH(0x55bb206af3d0)">foo</ltx:text>
finalize_rec on : <ltx:text _font="Font[serif,medium,upright,10,black,white,1,OT1,*,*,0]" _box="LaTeXML::Core::Whatsit=HASH(0x55bb206af3d0)">foo</ltx:text>
...
```

The fix in this PR escapes the loop and terminates with reasonable output. Just hoping it's not even subtler.

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -415,10 +415,9 @@ sub finalize_rec {
     elsif ($type == XML_TEXT_NODE) {
       # Remove any pending declarations that can't be on $FONT_ELEMENT_NAME
       my $elementname = $pending_declaration{element}{value} || $FONT_ELEMENT_NAME;
+      delete $pending_declaration{element};    # If any...
       foreach my $key (keys %pending_declaration) {
-        # make sure we also delete the element key we just used, or one can infinitely recurse on ltx:text
-        if (($key eq 'element') || !$self->canHaveAttribute($elementname, $key)) {
-          delete $pending_declaration{$key}; } }
+        delete $pending_declaration{$key} unless $self->canHaveAttribute($elementname, $key); }
       if ($self->canContain($qname, $elementname)
         && scalar(keys %pending_declaration)) {
         # Too late to do wrapNodes?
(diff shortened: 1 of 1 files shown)


kohlhase:
Thanks Deyan for looking into this. I agree with the "too many moving parts" meme. I am only hoping that when we have a version that moves at all, we will be able to concentrate on tweaking the parts separately.

kohlhase:
@dginev I know that there is an update task I have to keep in sync. But the changes to LaTeXML are not so severe that they should break much. 

The interleaving is the only way I can keep track of the LaTeXML changes at all.

Jazzpirate:
I have checked out and installed @dginev 's fork, and it does in deed get rid of the deep_recursion error with stex files :)

lib/LaTeXML/Core/Document.pm
```diff
@@ -416,7 +416,9 @@ sub finalize_rec {
       # Remove any pending declarations that can't be on $FONT_ELEMENT_NAME
       my $elementname = $pending_declaration{element}{value} || $FONT_ELEMENT_NAME;
```
brucemiller: Could you check if adding immediately after this line
```
      delete $pending_declaration{element};    # If any...
```
solves the problem (instead of the code below)?  Your version makes even more obscure the intention of the "element" NON-attribute.

dginev: Certainly that works too -- just tested on my passing sTeX branch, all looks good. I rebased on master as well for extra safety. Good to merge?

PR #1200, opened 2019-08-30
remove redundant XMDual lexemes; extra safe end lexeme spacing

dginev:
I wish I had arrived at this issue+patch before the full arXiv run, but oh well. I could rerun or guard against it in preprocessing...

I noticed a problem with the end lexeme markers for certain `OPFUNCTION` arguments - by far not all. Upon investigation, a minimal example:

```
latexmlc test.tex  --preload=[mathlexemes]latexml.sty --pmml
```

```tex
\documentclass{article}
\usepackage{amsmath}
\DeclareMathOperator{\dd}    {d\!}     % symbol for integration d
\begin{document}
\begin{equation}
  \dd
\end{equation}
\end{document}
```

Has master branch lexemes:
```
OPFUNCTION:start OPFUNCTION:start d OPFUNCTION:endOPFUNCTION:end
```

this PR upgrades that to:
```
OPFUNCTION:start d OPFUNCTION:end
```

addressing both issues - the wrongly spaced end lexemes, as well as the redundancy of the XMDual role with the inner XMWrap role, for this example. This should remove a wide range of noisy lexemes I am seeing in the arXiv token model.

### DIFF
--- a/lib/LaTeXML/MathParser.pm
+++ b/lib/LaTeXML/MathParser.pm
@@ -737,12 +737,12 @@ sub node_to_lexeme_full {
       # so just ignore the XMDual role attributes
       if ($tag ne 'ltx:XMDual') {
         $mark_start = "$role:start ";
-        $mark_end   = "$role:end";
+        $mark_end   = " $role:end";
       }
     } elsif ($tag =~ '^ltx:XM(Arg|Row|Cell)') {
       my $tag_role = uc($1);
       $mark_start = "$tag_role:start ";
-      $mark_end   = "$tag_role:end";
+      $mark_end   = " $tag_role:end";
     }
   }
   my $lexemes = $mark_start;
@@ -772,7 +772,7 @@ sub node_to_lexeme_full {
       }
     }
   }
-  $lexemes .= ' ' . $mark_end;
+  $lexemes .= $mark_end;
   return $lexemes;
 }
 
--- a/lib/LaTeXML/MathParser.pm
+++ b/lib/LaTeXML/MathParser.pm
@@ -733,8 +733,12 @@ sub node_to_lexeme_full {
   my ($mark_start, $mark_end) = ('', '');
   if ($tag ne 'ltx:XMath') {
     if ($role) {
-      $mark_start = "$role:start ";
-      $mark_end   = "$role:end";
+      # the dual marks are redundant with the marks on their inner XMWrap or other contents
+      # so just ignore the XMDual role attributes
+      if ($tag ne 'ltx:XMDual') {
+        $mark_start = "$role:start ";
+        $mark_end   = "$role:end";
+      }
     } elsif ($tag =~ '^ltx:XM(Arg|Row|Cell)') {
       my $tag_role = uc($1);
       $mark_start = "$tag_role:start ";
@@ -768,7 +772,7 @@ sub node_to_lexeme_full {
       }
     }
   }
-  $lexemes .= $mark_end;
+  $lexemes .= ' ' . $mark_end;
   return $lexemes;
 }
 


dginev:
Usual notice, just as #1195 , this PR only affects the "math lexeme serialization" feature and has no side-effects to the rest of math parsing.

brucemiller:
Hmm, annoying redundancy in the first place.

dginev:
I wasn't sure if both role attributes are used in different pieces of code, so I didn't dare suggest modifying the underlying dual markup. But yes, sneaked up on me here.

brucemiller:
Could you check if commit b8489591 solves the problem? (by avoiding creating redundant XMDuals in the first place)

dginev:
It over-solved it, now I am just seeing a `d` without the OPFUNCTION at all. Hm.

dginev:
The attributes look a bit confused as well, in particular the `tex` and `text` ones:
```xml
      <Math mode="display" xml:id="S0.E1.m1" tex="\dd{\operatorname{d\!}}" lexemes="d" text="dd" fragid="S0.E1.m1">
```

dginev:
Feel free to use the TeX snippet as a test case by the way, might be wise to keep the representation anchored properly.

brucemiller:
The "text" attribute is unchanged, but the tex and content-tex got messed up.  Was the original behaviour, ie.
``` content-tex="\dd" tex="\operatorname{d\!}"
```
appropropriate, from your POV?

dginev:
Ok, I noticed the role is now on the XMTok, so if the symbol appears in isolation, it will be lexematized to a regular `d`, without the opfunction information. That is reasonable.

Changing my example to `\dd x` I can see the `OPFUNCTION` wrapper is still alive and well when actually used:

```xml
      <Math mode="display" xml:id="S0.E1.m1" tex="\dd{\operatorname{d\!}}x" lexemes="OPFUNCTION:start d OPFUNCTION:end italic-x" text="dd@(x)" fragid="S0.E1.m1">
        <m:math alttext="\dd{\operatorname{d\!}}x" display="block">
          <m:semantics>
            <m:mrow>
              <m:mpadded width="-1.7pt">
                <m:mi mathvariant="normal">d</m:mi>
              </m:mpadded>
              <m:mo>â¡</m:mo>
              <m:mi>x</m:mi>
            </m:mrow>
            <m:annotation encoding="application/x-llamapun">OPFUNCTION:start d OPFUNCTION:end italic-x</m:annotation>
          </m:semantics>
        </m:math>
      </Math>
```

dginev:
@brucemiller that looks entirely appropriate yes!

brucemiller:
OK, have a fix. A good place to add this test would seem to be ```t/parse/operator```, but is there anything I need to do to "maintain" the  parse statistics?

brucemiller:
OK, I fixed the reversion and extended the test case.

dginev:
You mean the recently merged test suite upgrade? That is dynamically rechecked against the current grammar, there shouldn't be anything to add to it I believe. And it only checks that the parse routes are used, and that the parses succeed, not the final markup of the formulas. So you shouldn't need to manually revisit that at all.

dginev:
Thanks, closing here!

dginev:
The big question to answer is whether I should rerun the entire arXiv to get all lexemes updated... but it feels like a waste at the same time... Hm.

brucemiller:
I expect that this should be very rare (basically only uses of ```\DeclareMathOperator```). But then you *did* notice it!  I've lost track of how much of our hacking was since the last full run (or how much is in the short pipeline); that might determine the rerun threshold?

dginev:
I don't think the difference is big enough to warrant a full rerun to be honest, but I could heuristically grep for the files with the `OPFUNCTION_endOPFUNCTION_end` markers and queue those for a rerun... That's likely the wise route forward. A bit of a custom SQL job, but doable.

dginev:
Alright, my special grep-and-rerun operation is complete, just over 5500 documents were affected by this type of "lexeme noise". I've marked them all for rerun. Should offer some nice denoizing to the GloVe embeddings prepared by llamapun.

I will start bundling the 2019 version of the dataset after this partial rerun is complete. Thanks!

dginev:
Just as a cautionary update for anyone tracking here, I revisited the 150,000 articles in the currently packing 08.2019 arXMLiv dataset, and sadly there are more cases of mangled lexemes than the `OPFUNCTION` one. I don't yet know if the new changes to master fixed all of those, as that would require a full rerun, something to investigate moving forward.

So a grep for `\w+:end\w+` revealed 33,000 formulas with "glued" end lexemes in the no_problem set, so I could only imagine a 10x for the entire corpus. Luckily I can use the same regex in llamapun to guard against this type of problem, so it doesn't in itself mean the 2019 set needs to be regenerated. But something to look out for at the next big rerun, when it would be good to pre-check a handful of these documents work well in master, before the rerun is initiated.

dginev:
Example [showcase link](https://latexml.mathweb.org/editor?tex=%5Cusepackage%7Baasms4%7D%0A%5Cbegin%7Bdocument%7D%0A%5Cfarcs%0A%5Cend%7Bdocument%7D)
```tex
\usepackage{aasms4}
\begin{document}
\farcs
\end{document}
```

Has the lexemes:
```
ID:start POSTFIX:start SUPERSCRIPTOP:SUPERSCRIPTOP . SUPOP:prime SUPOP:prime POSTFIX:endID:end
```

PR #1202, opened 2019-09-04
Noexpand

brucemiller:
Turn ```\noexpand```-ed tokens into a ```\relax``` (usually) with the original token appended. Fix behaviour of dont-expand tokens with respect to if/when they get expanded, and how they act in let and in the comparisons (if, ifcat and ifx).


dginev:
Looks good to merge! Tests pass locally for me, and the small lipsum example I have succeeds with the full raw interpretation pipeline of expl3.sty, xparse.sty and lipsum.sty, fantastic!

The changeset to the codebase looks quite reasonable, given what it took to untangle everything.

PR #1203, opened 2019-09-04
xparse.sty binding

dginev:
Quick freebie from merging #1202 , we can now support xparse.sty via raw interpretation.

I added a small test from the documentation as a sanity check.


brucemiller:
Cool! Thanks!

dginev:
Ah you were too fast! Turns out passing on texlive 2019 (my machine) is quite different from passing on all other texlives out there. This binding is broken on 2016-2018, as Travis told us. Oops.

dginev:
There are two _different_ failures, one for the 2015,2016 tests and a more innocuous one (spacing) for the 2018 test.

brucemiller:
Yeah; Whoops!

dginev:
Ugh, it gets worse really. I copied over the xparse from texlive 2015 and pdflatex won't even run, because the expl3.sty it expects is different from the 2019 one (macros remain undefined etc.) so it's a expl3+xparse toolchain issue.

Makes me wonder if the wisest course of action would be to simply remove the binding+test for now, until we have as part of our dev workflow a way to easily develop against different texlive versions.

brucemiller:
Well, you could write the test case more defensively, omitting all the extra space (especially before the %) (which is often problematic anyway). Then it reduces to figuring out why 2015 fails. Hopefully the packages are consistent within any given texlive.

PR #1204, opened 2019-09-04
xparse fixes for older texlives

dginev:
Using this PR to debug the older texlive issues with xparse, with the help of Travis.


dginev:
Ok, my test outright fails in the 2015 texlive under `pdflatex`. So I think the package itself has changed in significant ways in the last few years, which makes sense given it is now using latex3.

Again I am unsure what is the right strategy for latexml in these cases. One idea is to back away from raw bindings that require a specific version of texlive, but advise latexml users to use the latest texlive with `--includestyles`.

Another idea is to fine-tune the test suite to allow detecting the texlive version, and specify the version as part of the test description. Maybe there are other viable ideas.

This is particularly scare for arXiv documents, since it is conceivable that there is no single version of texlive that can compile all of them. So there are "inevitable" errors, when aiming for generic coverage... Oh well, at least we know they can't be _that_ many cases.

dginev:
Ok, let's see if this works. I added a test-time check for the texlive version (if any) and will only execute the xparse test specifically when there is evidence this is a texlive 2018 or newer.

That still should get us one travis test that runs the xparse code (for texlive 2018) and helps with regressions, while avoiding having to look into the different implementation and test that would be needed for 2015.

dginev:
Alright, this is ready to merge with the "texlive 2018 or newer" testing strategy.

t/831_xparse.t
```diff
@@ -0,0 +1,10 @@
+# -*- CPERL -*-
+#**********************************************************************
+# Test cases for LaTeXML
+#**********************************************************************
+use LaTeXML::Util::Test;
+use LaTeXML::Util::Pathname;
+
+latexml_tests("t/xparse",
+   texlive_min => 2018,
+   requires => {xparse => ['expl3.sty','xparse.sty'] });
```
brucemiller: But, but... I thought I was joking. Or maybe I wasn't? Could actually crop up in the future.
However, it's kind of a shame that it is at the test-set level, rather than at the test case level (ie. it requires a separate directory of tests).That said, a natural evolution of the requires spec isn't immediately obvious.  Maybe something like:
```
requires => {xparse => {packages=>['expl3.sty','xparse.sty'], texlive=>2015} }
```
?

dginev: Certainly doable. Should I refactor to that state?

dginev: That said, why not have the option on both levels? The way I added you allows you to quickly skip an entire test set. Your suggestion is very granular to individual test files, but if you have say 10 different `.tex` tests for xparse, that means writing the `texlive=>` option ten times, once per file.

It's also the case that I really need `texlive_min` since both 2018 and 2019 are allowed at the moment. So maybe mergeable as-is ? It can be extended later.

brucemiller: Sure, both levels works for me. But I'd still like to refactor rather than split off a new test directory (already too many). Actually, your top-level comment makes me think one might want to require a package for a whole directory (eg. babel), but the current terminology hijacks "requires".  Also, given the usage pattern, writing test sets/cases should be Easy; while Test.pm can be complex as it's seldom touched (and incomprehensible anyway)

dginev: Alright, tests pass, left the global switch, but the xparse test now should look exactly as you want it:

```perl
latexml_tests("t/expl3",
   requires => {
      tilde_tricks => 'expl3.sty',
      xparse => {
         texlive_min => 2018,
         packages => ['expl3.sty','xparse.sty']
      } } );
```

brucemiller: Beautiful! Thanks for accommodating :>

aragont:
I tried to install LaTexML on Centos 8 with TeX Live 2018 and got the error when running test expl3/xparse

#   Failed test 't/expl3/xparse'
#   at /home/user/LaTeXML-0.8.5/blib/lib/LaTeXML/Util/Test.pm line 112.
# Difference at line 11 for t/expl3/xparse
#       got : '    <p>(find,finded,finded) [found]</p>'
#  expected : '    <p>(find,found,found)</p>'
# Looks like you failed 1 test of 3.

My environment: pdfTeX 3.14159265-2.6-1.40.19 (TeX Live 2018)

Diff file for my working version of  this test is attached
[xparse.txt](https://github.com/brucemiller/LaTeXML/files/5879846/xparse.txt)

Issue #723, opened 2016-02-23
XMDual question


dginev:
I am using a GH issue for the nice highlighting. I am trying to refresh my alternative [math parsing plugin](https://github.com/dginev/LaTeXML-Plugin-MathSyntax), and some of the previously passing tests are now passing due to the renovations using XMDual. If you could help me understand a basic example, I could take it from there.

An example input is ```$(a+b)^2$```, which produces the following XMath:

```xml
<Math mode="inline" tex="(a+b)^{2}" text="(a + b) ^ 2" xml:id="p1.m1">
  <XMath>
    <XMApp>
      <XMTok role="SUPERSCRIPTOP" scriptpos="post2"/>
      <XMDual>
        <XMRef idref="p1.m1.1"/>
        <XMWrap>
          <XMTok role="OPEN" stretchy="false">(</XMTok>
          <XMApp xml:id="p1.m1.1">
            <XMTok meaning="plus" role="ADDOP">+</XMTok>
            <XMTok font="italic" role="UNKNOWN">a</XMTok>
            <XMTok font="italic" role="UNKNOWN">b</XMTok>
          </XMApp>
          <XMTok role="CLOSE" stretchy="false">)</XMTok>
        </XMWrap>
      </XMDual>
      <XMTok fontsize="70%" meaning="2" role="NUMBER">2</XMTok>
    </XMApp>
  </XMath>
</Math>
```

My question here is - could you please explain why the XMDual was generated, and what it is trying to capture? In my math grammar test suite, I would be comparing to two trees generated from the "ground truth", which I am writing down as a lispy dialect for Content MathML:

```
(:arith1:power (+:arith1:plus a:ci b:ci) 2:cn)
```

The first comparison being a shallow skeleton comparison (does the tree have the right constituents?), which I would call an "operator tree test", while the second tests also check the meaning and role attributes, which I would consider a "semantics construction test". The stripped down, minimal operator tree in question for this example is

```json
[
  "ltx:XMApp",
  {},
  [
    "ltx:XMTok",
    {}
  ],
  [
    "ltx:XMApp",
    {},
    [
       "ltx:XMTok",
       {},
       "+"
    ],
    [
       "ltx:XMTok",
       {},
       "a"
    ],
    [
       "ltx:XMTok",
       {},
       "b"
    ]
  ],
  [
    "ltx:XMTok",
    {},
    "2"
  ]
]
```

[...]

PR #1205, opened 2019-09-11
End lexeme marker always preceded by space

dginev:
As reported in #723 , the XMDual markup  being addressed, the lexeme serialization still need extra guarantees for spacing out the end-lexemes. So this fix is really needed.


dginev:
The `\farcs` example improves to:
```
ID:start POSTFIX:start SUPERSCRIPTOP:SUPERSCRIPTOP . SUPOP:prime SUPOP:prime POSTFIX:end ID:end
```

brucemiller:
Seems like a good safeguard (although there may be a better fundamental way to do it)

PR #1207, opened 2019-10-01
[WIP] Tight lock/unlock for DB_File cache, for parallel postproc jobs

dginev:
An experiment allowing to do parallel jobs on large+paginated projects which reuse the same cache infrastructure.

Currently these types of jobs are not possible, as concurrent write access to the cache easily creates inconsistent states, causing all kinds of errors. This PR is not all that is needed for a parallel pipeline to be constructed, but is a definite prerequisite.

Implementation:
 * utilizes the `flock` mechanism, which is a lightweight file lock primitive
 * added very tightly scoped lock/unlock groups  to operations that touch the external cache file. Tightly, in order to avoid strange deadlocks, as well as locking for too long in a way that inhibits other processes from getting access. For example, if a lock lives as long as the ObjectDB instance, it is effectively allowing for just a single process at a time.
 * Made sure that the low-level `lock`/`unlock` are no-ops when the locking conditions aren't met, or when the input is malformed, to avoid any risk of edge cases causing deadlocks.

This is still risky abstraction in theory, and it takes only a moment of inattention to write code that deadlocks. So not my top solution in theory -- I would prefer having a cache that handles atomic locks in its own infrastructure, such as berkeleydb/postgresql. But in practice, this is definitely the smallest refactor, and adds no new dependencies.

I am still working on a fully successful parallel build for DLMF, so marking this work-in-progress until then.

### DIFF
--- a/lib/LaTeXML/Post.pm
+++ b/lib/LaTeXML/Post.pm
@@ -21,6 +21,7 @@ use base qw(LaTeXML::Common::Object);
 use LaTeXML::Global;
 use LaTeXML::Common::Error;
 use LaTeXML::Core::State;
+use LaTeXML::Util::ObjectDB;
 our @EXPORT = (@LaTeXML::Common::Error::EXPORT);
 
 sub new {
@@ -742,7 +743,7 @@ sub setDocument_internal {
       $$self{idcache}{ $node->getAttribute('xml:id') } = $node; }
     # Fetch any additional namespaces from the root
 
-    if(my $docroot = $root->documentElement){
+    if (my $docroot = $root->documentElement) {
       foreach my $ns ($docroot->getNamespaces) {
         my ($prefix, $uri) = ($ns->getLocalName, $ns->getData);
         if ($prefix) {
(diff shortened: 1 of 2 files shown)


brucemiller:
Ouch!
I guess I wasted my time verifying that moving the open/close to 
new/finalize was much faster.  And, I guess that if we tackle
this we'll have to switch to a much less lightweight db.

dginev:
Indeed! The PR works great for speeding up post-processing when you are completely ignoring the .cache file (e.g. when you are collecting math lexemes), but there is no path forward for any official use. It may have been easier to just remove the --crossref switch for my purpose, rather than optimizing it :>

brucemiller:
The cache in Post.pm is a bit different than ObjectDB. The latter would 
definitely need to support multiprocessing for parallel, but the former 
really should be restricted to a single process (but no guarantees).
It's for caching info about image conversions for a single (source) page.

dginev:
That is a great clarification, as I have not yet been a power user of the featureset. Since both the Post.pm `LaTeXML.cache` and the `--dbfile` provided file for ObjectDB are realized via the DB_File backend, I had implicitly referred to all caching as a single piece of infrastructure. We might as well continue have a single caching backend for all necessary operations, concurrent or not, simpler to maintain.

PR #1210, opened 2019-10-08
Enable custom LaTeXML::Core options in 'latexml_tests'

tkw1536:
When using the `latexml_tests` method to create test cases, LaTeXML is instantiated using `LaTeXML::Core->new` with appropriate default options.
This works fine for tests by LaTeXML itself, however it prevents the function from being used in test cases for custom plugins (as none of the profiles, bindings, etc.) can be loaded.

This PR adds a new option `core_options` to `latexml_tests`. When provided, this should be a hash of options passed to `LaTeXML::Core->new`. This allows plugin authors to customize the options used to instantiate LaTeXML for each test case.

### DIFF
--- a/lib/LaTeXML/Util/Test.pm
+++ b/lib/LaTeXML/Util/Test.pm
@@ -118,7 +118,7 @@ sub do_fail {
 # NOTE: This assumes you will have successfully loaded LaTeXML.
 sub latexml_ok {
   my ($texpath, $xmlpath, $name, $core_generator, $compare_kind) = @_;
-  if (my $texstrings = process_texfile($texpath, $name, $core_generator, $compare_kind)) {
+  if (my $texstrings = process_texfile(texpath => $texpath, name => $name, core_generator => $core_generator, compare_kind => $compare_kind)) {
     if (my $xmlstrings = process_xmlfile($xmlpath, $name, $compare_kind)) {
       return is_strings($texstrings, $xmlstrings, $name); } } }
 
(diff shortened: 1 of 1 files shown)


tkw1536:
Thanks for your feedback. 

- The main function (`latexml_tests`) already takes a hash, but I will update `process_texfile` and another intermediate function to also take hashes. 
- I'm not sure if an option hash is sufficient -- can every sensible test customization be expressed via parameters for LaTeXML::Core or do some plugins need more than that -- like setting up dependent code? Also if we pass in a hash, do we have to worry about mutablitlity if it is used for multiple core instances?
- I also couldn't think of a good name for the parameter, I'll go with your suggestion.

tkw1536:
I updated `process_texfile` to take a hash. `latexml_ok` appears to be exported, hence I am hesitant to change it to taking a hash (who knows which external code might depend on the current parameters). I also added a commit to ensure that calls to it are backwards compatible by moving the new `core_generator` argument to the back.

dginev:
The more I think about it, the more certain I am you only want to allow the core options as new customization here.

You can definitely have plugins that would test more than the latexml Core conversion, and they should most certainly *not* use the `latexml_tests` code path. That bit of logic has a very concrete purpose - and it is ensuring that a Core conversion results in the expected XML.

Even in the core latexml suite, there are the daemon tests which piggy back on the `latexml_tests` to do something _completely_ different, which is to run latexmlc on `.spec` specifications. It was a decent short-term solution, but I would certainly rename those to a new function `latexmlc_tests` if I started them now. We could make the underlying logic in Test.pm more reusable from external testing modules, but I don't think we should allow arbitrary closures to be passed in.

This also rekindles an old point of frustration for latexmlc remaining named as it is, awkwardly placed in-between the other executables. We had at one point agreed with Bruce that latexmlc should really be the main `latexml` executable, and the `latexmlcore` and `latexmlpost` executables could remain as deprecated/secondary ones. Would allow to name things a little better, we could then have `latexml_tests`, `core_tests`, `post_tests`, etc.

tkw1536:
I have pushed an appropriate commit, should I squash and update the PR description?

dginev:
Thanks! LGTM to me, left to @brucemiller  for final squash+merge.

If you'd like to update title/description of the PR - go ahead, I usually do so without asking whenever appropriate. As to the commit history, no need to worry about that on your end, that's what "squash and merge" was made for :+1:

tkw1536:
Description updated. Ready to be merged ☺

brucemiller:
When I saw this PR opened, I was a bit concerned that it'd get too far ahead of the hopefully-not-too-far-off config setup which should make plugins and extensions more pleasing (although they'll of course want to provide their own tests!).  But this looks simple enough and clean, so let's go.  Thanks!!

PR #1218, opened 2019-11-26
More robust xpath search for \lxRDFa

### LINKED ISSUE #1217: Unexpected behaviour of \lxRDFa using xpath
I'm trying to attach an attribute to a specific children of the current tag using `\lxRDFa`.

This is my LaTeX document:

```latex
\documentclass{article}
\usepackage{lxRDFa}
\begin{document}
  \section{Heading}
  \lxRDFa[.]{foo=bar}
\end{document}
```

`latexml document.tex` produces

```
latexml (LaTeXML version 0.8.4; revision 9a0e7dc5)
processing started Tue Nov 26 09:42:03 2019

(Digesting TeX test...
(Loading /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/TeX.pool.ltxml...
(Loading /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/eTeX.pool.ltxml... 0.00 sec)
(Loading /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/pdfTeX.pool.ltxml... 0.00 sec) 0.15 sec)
(Processing content /Users/paolobrasolin/rb/tex2mn/test.tex...
(Loading /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/LaTeX.pool.ltxml... 0.20 sec)
(Loading /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/article.cls.ltxml... 0.01 sec)
(Loading /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/lxRDFa.sty.ltxml... 0.00 sec) 0.25 sec) 0.41 sec)
(Building...
(Loading compiled schema /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/resources/RelaxNG/LaTeXML.model... 0.01 sec). 0.02 sec)
(Rewriting... 0.00 sec)
(Finalizing... 0.00 sec)
Conversion complete: No obvious problems.
processing finished Tue Nov 26 09:42:03 2019
<?latexml searchpaths="/Users/paolobrasolin/rb/tex2mn"?>
<?latexml class="article"?>
<?latexml package="lxRDFa"?>
  <section foo="bar" inlist="toc" xml:id="S1">
    <tags>
      <tag>1</tag>
      <tag role="refnum">1</tag>
      <tag role="typerefnum">§1</tag>
    </tags>
    <title><tag close=" ">1</tag>Heading</title>
  </section>
</document>
```

and this seems perfectly reasonable.

Any way I tried to match the `ltx:title` using the xpath failed either ignoring me (the attribute would still be attached to `ltx:section` or throwing an exception. E.g,

```
  \lxRDFa[ltx:title]{foo=bar}
```

throws

```
Fatal:misdefined:LaTeXML::Package::Pool::__ANON__ Can't call method 'setAttribute' on an undefined value
	at at /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/lxRDFa.sty.ltxml line 164.
	In Core::Definition::Constructor[\lxRDFa... /Users/paolobrasolin/perl5/lib/perl5/LaTeXML/Package/lxRDFa.sty.ltxml; line 165
	 <= Whatsit[\lxRDFa,Box[l],Core::KeyVals[... <= Core::Document[@0x7fb7faaba620]
1 fatal error
```

I'm not sure how the xpath is evaluated and whether there is any limitation, but this behaviour is definitely unexpected to me. Am I missing something obvious?

dginev:
Fixes #1217 

There are other approaches here, but this seemed simple enough. 

- when the xpath fails to find a node, go up the ancestry tree and retry, as long as there is some document available. It may be slow if the author intended a distant element, but provides predictable markup, as one tends to expect the metadata on the node that is the "closest meaningful match" to the current insertion context

- when a node is not found even after the more thorough search, handle that with a graceful warning, and default to the `floatToAttribute('property')` case. 

One bit I didn't spend time remembering was that the result of ToString on the `[Semiverbatim]` argument returned only the first letter, which is why I switched to the simpler `[]` plain optional argument.

### DIFF
--- a/lib/LaTeXML/Package/lxRDFa.sty.ltxml
+++ b/lib/LaTeXML/Package/lxRDFa.sty.ltxml
@@ -149,7 +149,7 @@ sub RDFTriple {
 # You are responsible for how the attributes relate to those
 # in both parent and children nodes, such as establishing
 # of default subject or objects, chaining and so forth.
-DefConstructor('\lxRDFa [] RequiredKeyVals:RDFa', sub {
+DefConstructor('\lxRDFa OptionalSemiverbatim RequiredKeyVals:RDFa', sub {
     my ($document, $xpath, $kv) = @_;
     my ($save, $node);
     if ($xpath) {
--- a/lib/LaTeXML/Package/lxRDFa.sty.ltxml
+++ b/lib/LaTeXML/Package/lxRDFa.sty.ltxml
@@ -149,13 +149,20 @@ sub RDFTriple {
 # You are responsible for how the attributes relate to those
 # in both parent and children nodes, such as establishing
 # of default subject or objects, chaining and so forth.
-DefConstructor('\lxRDFa [Semiverbatim] RequiredKeyVals:RDFa', sub {
+DefConstructor('\lxRDFa [] RequiredKeyVals:RDFa', sub {
     my ($document, $xpath, $kv) = @_;
     my ($save, $node);
     if ($xpath) {
-      $save = $document->getNode;
-      $node = $document->findnode(ToString($xpath), $save); }
-    else {
+      $xpath = ToString($xpath);
+      $save  = $document->getNode;
+      # Try searching up the ancestry,
+      # whenever the node is not local to the current Documnent insertion point
+      my $parent = $save;
+      while (!$node && $parent) {
+        $node   = $document->findnode($xpath, $parent);
+        $parent = $parent->parentNode; }
+      Warn('expected', 'node', $xpath, "Expected to find a node for RDFa attributes at the xpath: $xpath") if !$node; }
+    if (!$node) {
       $save = $document->floatToAttribute('property');    # pic arbitrary rdf attribute
       $node = $document->getElement; }
     my $attr = RDFAttributes($kv);


paolobrasolin:
Thanks for this @dginev!

I'm a bit perplexed by the idea of climbing the tree until the xpath matches.
I feel that's counterintuitive since when using xpath it's usually evaluated relative only to the "current" node.

Furthermore, I played around a bit having fixed only the argument type of `xpath` and the xpath engine seems able to evaluate complex relative expressions; this works perfectly:

```latex
\documentclass{article}
\begin{document}
  \begin{itemize}
    \item This would fail matching a Whatsit and calling setAttribute %\lxRDFa[.]{match=true}.
    \item This matches item/para/p \lxRDFa{match_1=true} (because it floats above the Whatsit automatically)
    \item This matches item/para/p \lxRDFa[..]{match_2=true}.
    \item This matches item/para \lxRDFa[../..]{match_3=true}.
    \item This matches item \lxRDFa[../../..]{match_4=true}.
    \item This globally matches the sixth item \lxRDFa[{//ltx:item[6]}]{match_5=true}.
    \item This matches the second tag of the seventh item \lxRDFa[{ancestor::ltx:itemize/ltx:item[7]/ltx:tags/ltx:tag[2]}]{match_6=true}.
  \end{itemize}
\end{document}
```

I think the only missing feature is adding the attributes to *all* the matching nodes, as one would expect from an xpath expression, instead of just the first one. E.g. to add an attribute to all `ltx:item`s of the document:

```latex
\documentclass{article}
\begin{document}
  \begin{itemize}
    \item foo
    \item bar
    \item baz
  \end{itemize}
  \lxRDFa[//ltx:item]{match=bar}
\end{document}
```

That looks like a breaking change. However, if I'm not mistaken, the `xpath` argument was not working to begin with so the change would not be breaking anything.

I'm fiddling with the code myself. I'll submit a PR soon.

dginev:
It is objectively a confusing problem to think about, because the machinery involved has several moving parts at different levels. It is intuitive to think if you add a metadata macro under a `\section{..}` it will be able to see any XML element related to that section's markup. However, that is not immediately the case, since `ltx:title` is often a _sibling_ to the "current insertion point" in the document. For example:

```tex
\section{Heading}
some content and btw \lxRDFa[ltx:title]{foo=bar}
```

has the XML:
```xml
  <section inlist="toc" xml:id="S1">
    <tags>
      <tag>1</tag>
      <tag role="refnum">1</tag>
      <tag role="typerefnum">§1</tag>
    </tags>
    <title foo="bar"><tag close=" ">1</tag>Heading</title>
    <para xml:id="S1.p1">
      <p>some content and btw</p>
    </para>
  </section>
```

So when the `\lxRDFa` macro is invoked, the active insertion point is `<p>`, but you are really talking about the title at `->parent->parent->previous_sibling`. Which is why to me the closest algorithm is to gradually climb up the ancestry tree and re-execute the xpath, until it finds an element that might have been "intuitively" intended by the author.

Traversal details like these are a big part of the motivation behind Bruce's `floatToX` methods in Document.pm, and I wonder if there is a more "proper" implementation to what I am describing here.

While you are correct that using the `../..` and similar relative paths with xpath will work (even without this PR!), that is a terrible user experience for a casual latex author, who simply wants an intuitive default behaviour. Requiring people to go inspect the XML tree of a large document and construct by hand a relative xpath is something to be avoided if possible, I'm pained just thinking about the workflow. The care-free UX is to just say "ltx:title" and have the binding figure out the details for you.

I also would hope most of the time authors can get by  using `\lxRDFa` without the optional argument, with latexml providing a good default guess with productive outcomes.

dginev:
And I like your remark about "this is not something one would expect from an xpath". It leads me to think we shouldn't be calling the optional argument to `\lxRDFa` an "xpath" at all, because realistically it should be the "target node"/"anchor node" in simple uses. And as I explained above I am quite eager to support the intuitive uses first, and keep things low-difficulty. We can always add a new macro `\lxRDFaXPath{xpath}` with a mandatory xpath argument which executes "the xpath way" annotating all nodes that are matched. But even then you have a question of timing - the macro is invoked during the intermediate construction of the document, so it can't realistically annotate the fully constructed doc. If you wanted to do global annotations, you'd want a hook at `\end{document}` which then executes global metadata insertions. But this is heading into very advanced use case territory, and we would need to think of the publication pipeline examples that require it. Not the current intent of `lxRDFa.sty` to my understanding.

brucemiller:
Certainly falling through to a Fatal is a fail, but...

It appears that the main problem is that that ```[Semiverbatim]``` (unlike ```OptionalSemiverbatim```) is loosing (most of) the argument, so the xpath fails. I'll have to track that down because they should (?) be the same thing.  Using OptionalSemiverbatim gives the (mostly) expected result.

I do think the arg should be, and be thought of as, an xpath, and probably shouldn't go searching for matches. Of course, that obligates the user to think about where LaTeXML "is".
 OTOH, I'm not yet sure about whether it should apply to all matches. I'd implicitly been thinking of first-match, ie. implicitly appending a [1].  I suppose it would provide more user possibilties; it's a change, but probably affects few use cases.

brucemiller:
Ah, no: ```[Semiverbatim]``` reads an optional argument, then reads a Semiverbatim inside, which is only the 1st token, unless there are braces. So, first bugfix is replace parameter with OptionalSemiverbatim.

dginev:
@brucemiller alright, updated to `OptionalSemiverbatim`. I have to say `[Semiverbatim]` is quite confusing - makes me wonder if we want it removed or changed to be identical to `OptionalSemiverbatim`, or if there are legitimate cases where it is useful.

I would still maintain it's better usability to go a level up and repeat the `->findnode()` search on the xpath argument, so that authors can provide an element name and get the intuitive result. Obviously if you provide a path specifier - be it absolute or relative - and you get the first `->findnode()` call to succeed, the climbing up the ancestors won't ever happen, so it's not a conflict w.r.t the "full xpath" behaviour.

brucemiller:
I think searching around is more confusing. Too lazy to make a PR, but what about:
```
DefConstructor('\lxRDFa OptionalSemiverbatim RequiredKeyVals:RDFa', sub {
    my ($document, $xpath, $kv) = @_;
    my ($save, @nodes);
    $xpath = ToString($xpath);
    if ($xpath) {
      $save = $document->getNode;
      @nodes = $document->findnodes(ToString($xpath), $save);
      Warn('expected','node',$document,
           "No node matched xpath $xpath for RDFa properties") unless scalar(@nodes); }
    else {
      $save = $document->floatToAttribute('property');    # pic arbitrary rdf attribute
      @nodes = ($document->getElement);
      Warn('expected','node',$document,
           "No node matched accepts RDFa properties") unless scalar(@nodes); }
    my $attr = RDFAttributes($kv);
    foreach my $node (@nodes){
      foreach my $k (sort keys %$attr) {
        # use direct method ($doc method doesn't skips about="", which we need!)
        $node->setAttribute($k => ToString($$attr{$k})); } }
    $document->setNode($save); });
```

dginev:
I don't follow. I think we need to get some clarity on what the use of `\lxRDFa` is _ought to be_ first. I opined in #1219 that the focus on "assign this metadata to all nodes matching the xpath" appears to me to be centered on XPath, rather than centered on the annotation workflow of the TeX author. How and when would one use `\lxRDFa` with the optional argument provided?

One answer is that they may want to adjust the default annotation target. Say in the blog test we have [here](https://github.com/brucemiller/LaTeXML/blob/fb71c30ff7674fc62441ce6187a98240d19e279a/t/complex/aliceblog.tex#L35)

We see:
```tex
\paragraph{The Trouble with Bob\lxRDFa{property={dcterms:title}}}\label{blog:trouble}
\lxRDFa{resource={/alice/posts/trouble_with_bob},typeof={schema:BlogPosting}}
```

which coincides with my thinking - you use the macro to deposit a _single_ metadata triple, on a node of choice, or the "nearest node that can contain a property". The xpath aspect is largely incidental - if you're casually working on a blog, you really don't want to think of latexml insertion points, so you would prefer not using the optional argument at all. And when you get to use it, you'd want to do so in passing if possible, e.g. quickly choose the `ltx:title` node, instead of the `ltx:paragraph` node.

brucemiller:
Well, I don't know what "centered on the annotation workflow of the TeX author" means, other than "do what I mean".  If you just want the annotation to appear "around here", omit the optional argument; Simple. If you want to annotate a specific node, you've got to specify more carefully and the argument ought to mean something more than "somewhere"; xpath is fairly precise, isn't it?

The code I proposed satisfies the original case for ```\ltxRDFa[ltx:title]{...}``` and the additional cases @paolobrasolin gave (at least I think they do). That's at least one example of "works as expected".

dginev:
To me (as I said before, personal view), "centered on the annotation workflow of the TeX author", is the perspective "is this easy to use when I write TeX", rather than "what is the most powerful way to leverage xpath".

To repeat my [comment example](https://github.com/brucemiller/LaTeXML/pull/1218#issuecomment-559280016):
```tex
\section{Heading}
some content and btw \lxRDFa[ltx:title]{foo=bar}
```

has the intention for the XML
```xml
  <section inlist="toc" xml:id="S1">
    <tags>
      <tag>1</tag>
      <tag role="refnum">1</tag>
      <tag role="typerefnum">§1</tag>
    </tags>
    <title foo="bar"><tag close=" ">1</tag>Heading</title>
    <para xml:id="S1.p1">
      <p>some content and btw</p>
    </para>
  </section>
```

The current insertion point at the time `\lxRDFa` is invoked is a `<p>`, while the intended title - while _obvious_ when writing the TeX, is _unknown_ without reading the target XML. The explicit xpath would be:
```tex
\section{Heading}
some content and btw \lxRDFa[../../../ltx:title]{foo=bar}
```

And you can't get that xpath down without significant trial and error - you need to read the XML, then realize the insertion context was the `<p>`, then write `../../ltx:title`, fail again, and realize the insertion context was really the text element inside the `<p>`. Feel free to decide any which way. In practice, I have only had the need to use document-level metadata for e.g. my personal blog, so I'm happy to seat out the final resolution. Closing here to avoid confusion.

brucemiller:
Instead of "what is the most powerful way to leverage xpath", how about "what is the most powerful way to express which node I mean, preferably using a language I already know"? If you're thinking in a pure TeX mode, then perhaps XPath doesn't make sense --- but neither does RDF. If you concede that you're working with XML, and have a picture of the XML being produced, you can put the RDF exactly where you want with no trial and error at all:
```
\lxRDFa[ancestor::ltx:section/ltx:title]{foo=bar}
```

paolobrasolin:
Thanks @dginev and @brucemiller for the insights!

My use case is annotating very specific nodes in the XML tree for further parsing, so like @brucemiller I think that if a user explicitly gives an xpath then it should behave exactly like one.

I think that's especially true since xpaths are rich enough to express ideas like "get me to the title of the current section" very clearly (`ancestor::ltx:section/ltx:title`) with no guessing at all. Just specifying `ltx:title` and trusting the floating algorithm is terser, but I feel less expert users would just do `\section{My section}\lxRDFa[ltx:title]{foo=bar}` anyways, so that would make no difference.

Using xpath canonically when given and otherwise defaulting to the floating algorithm sounds like a fine macro API to me. Perhaps separating the behaviours into two distinct macros would make that even clearer, but I think it's mostly a matter of documenting that clearly.

As for adding attributes to all nodes matching the xpath, I'm strongly in favour as that's what xpath canonically matches. While that may seem a complication I think it's for the best:
* when there are multiple nodes one should have the freedom to either match them all `ancestor::ltx:itemize/ltx:item` or match a specific one `ancestor::ltx:itemize/ltx:item[2]`
* when there is a single matching node like in `ancestor::ltx:section/ltx:title` then no index is required.

The only possible pitfall of xpath usage is what @dginev pointed out: timing issues. I.e.:

```
\section{First section}
You can't reach the future: \lxRDFa[{//ltx:section[2]/ltx:title}]{matchedfromfuture=true}
\section{Second section}
You can reach the past: \lxRDFa[{//ltx:section[1]/ltx:title}]{matchedfrompast=true}
```

That's also simply a matter to be clearly pointed out in the documentation, I think.

Issue #1214, opened 2019-11-06
Regressions on Windows 10


VB24:
Hi! I am looking forward to using LaTeXML, however I am unable to install it. 
I ran: `choco install latexml` from PowerShell and lot of prerequisites got installed. 
However it seems that few tests failed (all related to daemon). Here is one such warning that I get: 

```
#   Failed test 't/daemon/formats/citation'
#   at C:\Users\Vedran\.cpanm\work\1572992197.7568\LaTeXML-0.8.4\blib\lib/LaTeXML/Util/Test.pm line 93.
# Difference at line 1 for t/daemon/formats/citation
#       got : 'kpsewhich: warning: running with administrator privileges'
#  expected : 'No obvious problems'
```

Maybe I am missing something obvious, but I am new to LaTeXML and this type of installations...

dginev:
Hi @VB24 , I think you may need to open PowerShell with admin rights here, and then the install should go smoothly.

VB24:
Hi, I ran PowerShell with admin rights, without that I am not able to install anything.

dginev:
Ah, apologies, I misread the error. So `kpsewhich` fires a warning _because_ the test is running with admin rights.

I recall I added a [special flag](https://github.com/brucemiller/LaTeXML/blob/f99018db924e855a31441b298f13d5f12abc96aa/lib/LaTeXML/Util/Pathname.pm#L378) to get the tests to pass when running under Appveyor, but curious I haven't seen the issue when I tested on my local Windows. I would have to check more carefully and get back to you, apologies.

The good news is that it's not a significant error, rather a toolchain warning message, so forcing the install will likely give you a working latexml if only this case is failing. Thanks for letting us know!

VB24:
Thanks for answering promptly. I also tried `choco install latexml --force` and `cpan latexml` but the same issue always gets in the way.

dginev:
Ok, I can reproduce on my Windows machine with a usual MikTeX installation and the master branch. Also, it seems like the choco packaging effort [was aware](https://github.com/metanorma/chocolatey-latexml/issues/6) of this issue at some point, but it got punted on. Will try to see if I can quickly find a patch, given that the warning is not related to the core latexml functionality.

dginev:
[...]

PR #1221, opened 2019-12-01
add \@ifl@t@r helper to LaTeX.pool

dginev:
Part of #1214 .

This PR makes the following TeX snippet produce identical output with pdflatex and latexml:
```tex
\documentclass{article}
\makeatletter
\edef\versioned{\@ifl@t@r\fmtversion{2019/10/01}{earlier}{later}}
\begin{document}
\versioned
\end{document}
```

The code is transferred over from a texlive 2019 latex.ltx.


brucemiller:
Thanks!

PR #1226, opened 2020-01-08
Proposed internal support for fixed refnum, id keyed to \label

brucemiller:
For consideration @dginev, @tkw1536 : The basic idea is that you assign to LABEL_MAPPER a function ```sub($label,$counter,$norefnum)``` which returns 2 strings: the desired refnum and id.  Not that you're expected to do that: More relevant is that i has low run/maintenance overhead for virtually everyone who doesn't use it.


lib/LaTeXML/Package.pm
```diff
@@ -717,6 +719,70 @@ sub RefStepCounter {
...
 
+# Internal: Use a label-derived reference number and/or ID
+# instead of the traditional counter based ones.
+# Since the \label{} determins the reference number and ID,
+# we MUST sniff out the label BEFORE we call RefStepCounter/RefStepID !!!!!
+# (see MaybePeekLabel below; and also MaybeNoteLabel for use within
+# captions & certain equation environments)
+# Assign a sub to LABEL_MAPPER: &sub($label,$counter,$dorefnum,$doid)
+# to return the desired refnum and id for a given object.
+sub maybeHighjackRefnum {
```
dginev: `highjack` -> `hijack` (3 occurrences)

dginev:
It's certainly a clean abstraction! 

Seems like the overhead when it isn't used is 2 subroutine calls, with one state lookup, and nothing else. It's not the end of the world if it solves a general problem.

In terms of naming, it feels more official if you provide slots for say `customMapperHook`, which execute when a mapper is present, rather than `maybe` `hijack`-ing the code, which gives me a bank robbery vibe. I think I like it!

PR #1228, opened 2020-01-21
Refactor test suite behaviour

tkw1536:
This PR refactors the behavior of the test suite. The aim of this is to:

- have a reasonably fast "consistency" test suite, that checks if anything major is broken
- have a more extensive test suite that can perform more thorough checks
- be able to add tests to either suite easily
- run the 'consistency' suite with 'make test' by default, run the extensive test suite on Travis by default, but also allow running the extensive test suite locally. 

Previously this was achieved in two different ways:
- by having '*.tt' files which are run by a custom 'Makefile.PL' command `make extratests`
- by checking for a '$CI' environment variable, and skipping tests if it is not set. 

In this PR I suggest using only the first approach, as it results in less spam on the console when running a normal 'make test' (no "skipped because of CI"). Furthermore it also results in less boilerplate being added to each test file (no ` if $ENV{CI})``` ...). 

Primarily, this PR renames the test files in accordance with this schema, and removes now-dead code that checks if we are on the CI. 

Additionally, this PR updates the Travis Tests to run `make fulltest` instead of `make test`  (Note: This will require me to rebuild the Docker Test Images and push them. This change is fully backwards-compatible, as there are no '*.tt' files in master at the time of writing. I have already started the building process for the images, and will push those to DockerHub once completed. )

Furthermore, this PR allow tests either directly in the 't/' directory or one level deep -- this allows easily grouping a particular set of tests, for instance a group of unit tests. Doing this, we need to pay some extra attention to the order of tests. For example we want tests in 't/000_unit/' to run before '005_tokenize.t'. By default, this would not be the case -- first all tests directly in `t/`, and then all the tests in all the sub directories would be run. To fix this, we also add extra code to sort all files alphabetically. 

Finally, all test files in this PR are now prefixed with a three-digit-number (as opposed to a two-digit-number).


dginev:
I think this entire PR comes down to a discussion of taste, so I'd be happy to discuss it in person next week - it's too easy to not say anything constructive in the comments here. In fact, might have been better to build a team consensus on taste before you went and did all the work, but hey - could be a learning opportunity :+1: 

The one bit that I can say entirely rooted in performance considerations is that I wouldn't run any Tikz tests in CI, as they are - since they are too slow + considered somewhat optional (you need texlive to use it). It may be a moot point now that we have latex 3 tests, which are in the same general magnitude of slowness, but waiting for over 30 minutes for the CI to finish could be a bit on the heavy side.  There is a larger discussion (or two) to have around these points as well - latex 3 is starting to get used by default in newer texlives, which poses somewhat of a hard requirement on latexml to make it a core supported component (the way LaTeX.pool is supported even when no texlive is around).

tkw1536:
After offline discussions I assume this will not be merged (for now).

PR #1232, opened 2020-01-31
Move Travis Test Commands into .travis.yml

tkw1536:
Previously, when running Travis Tests, the Docker Images determined which commands (like 'make test' and friends) were run during the build.

This PR changes the behavior, and instead moves the commands to be run into `.travis.yml`. This will allow customizing travis behavior in the future (to e.g. run 'make fulltest' instead of 'make  test') without needing to rebuild the Docker Images.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: generic
-sudo: required
+os: linux
 dist: trusty
 
 services:
@@ -22,7 +22,7 @@ env:
   ## Testing with texlive 2015 -- perl 5.22
   - PERL=5.22.4 TEX=texlive-2015
 
-matrix:
+jobs:
   # We allow failures for all the TeXLives excepts the 2016 ones
   allow_failures:
     - env: PERL=5.28.0 TEX=texlive-2018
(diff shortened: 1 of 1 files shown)


dginev:
While I don't mind the ever more precise dockerized travis, we can also consider the alternative route - only testing the most recent texlive from the apt packages, and returning to a minimal travis file with no docker in it at all. The choice is pretty much a consequence of where we're headed with long-term maintenance goals for latexml - do we support any and all perl and texlive, or do we focus on the last few recent perls and the most recent texlive.

The situation in 2020 and 2021 is poised to be different than 2010-2020 since latex is once again on the move with the LaTeX 3 effort, and there are likely to be very significant differences between texlives. It's something for Bruce to consider, since it may end up the trade-off between partially supporting a large number of toolchains, vs ensuring great support on a single toolchain (running ever more texlive-using tests, such as tikz)

tkw1536:
The entire reasoning for using Docker in the first place was that 'the most recent' texlive on Travis was constantly outdated and proved unreliable during installation.

tkw1536:
What is the status on this being merged?

brucemiller:
This all looks guessable, if not completely readable;  Does it require new Docker images, or does it continue to work with the existing ones?

tkw1536:
No, this (intentionally) does not require new docker images.

brucemiller:
Cool, well it will add some flexibility so let's do it. Thanks!!

dginev:
Alright, seems to be working as expected:
https://travis-ci.org/brucemiller/LaTeXML/jobs/660794050

PR #1241, opened 2020-03-08
Fatal-avoiding patches for a handful of arXiv cases

dginev:
The only purpose of this PR is to "plug" needless Fatals, and allow the processing to proceed further. I tested ~10 papers from the latest arXiv run, and ~5 of them switched to Error status, the others switching to `100 error` Fatals. 

I claim both are better outcomes, as the `Perl:die` fatal category is rather opaque to comb through, and we might as well gain some robustness for odd cases.

Changes:
 - Adding stubs for Number/Pair methods to Box, so that cases where the arguments are misinterpreted can fail more gracefully. No need to Fatal a document because a List snuck into the number arithmetic. Though of course no sensible content will come out for that particular piece of the document.

 - Be a little more careful with unblessed refs sneaking into `AddToMacro`, which is something I've seen in the examples. Checking for `blessed` instead of `ref` does the trick.

- Lastly, escape the search string provided to `\patchcmd`, one arXiv example had used `\par` as the input there, and the regex was failing with "`\pa` is an unrecognized Unicode class".

We've had the discussion about whether reducing Fatals is a good thing on its own, for now I will keep rooting on team "extra robustness is good, especially for arXiv jobs".

### DIFF
--- a/lib/LaTeXML/Core/Box.pm
+++ b/lib/LaTeXML/Core/Box.pm
@@ -236,12 +236,21 @@ sub computeSize {
 #**********************************************************************
 # What about Kern, Glue, Penalty ...
 
+# stubs to avoid "shallow" Fatal errors on broken documents
+sub multiply { my ($self) = @_; return $self; }
+sub subtract { my ($self) = @_; return $self; }
+sub add      { my ($self) = @_; return $self; }
+sub negate   { my ($self) = @_; return $self; }
+sub divide   { my ($self) = @_; return $self; }
+sub getX     { my ($self) = @_; return $self; }
+sub getY     { my ($self) = @_; return $self; }
+sub ptValue  { return; }
 #======================================================================
 1;
 
 __END__
 
-=pod 
+=pod
 
 =head1 NAME
 
--- a/lib/LaTeXML/Core/Box.pm
+++ b/lib/LaTeXML/Core/Box.pm
@@ -13,6 +13,7 @@ package LaTeXML::Core::Box;
 use strict;
 use warnings;
 use LaTeXML::Global;
+use LaTeXML::Common::Dimension qw(Dimension);
 use LaTeXML::Common::Object;
 use base qw(LaTeXML::Common::Object);
 use base qw(Exporter);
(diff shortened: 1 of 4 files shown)


dginev:
I should also add these are minor fixes in terms of arXiv impact, the main culprit w.r.t Fatals is as described in #1240 .

lib/LaTeXML/Core/Box.pm
```diff
@@ -236,12 +236,21 @@ sub computeSize {
...
 
+# stubs to avoid "shallow" Fatal errors on broken documents
+sub multiply { my ($self) = @_; return $self; }
+sub subtract { my ($self) = @_; return $self; }
+sub add      { my ($self) = @_; return $self; }
+sub negate   { my ($self) = @_; return $self; }
+sub divide   { my ($self) = @_; return $self; }
+sub getX     { my ($self) = @_; return $self; }
+sub getY     { my ($self) = @_; return $self; }
+sub ptValue  { return; }
```
brucemiller: I see it avoiding Fatals, but probably next to impossible to find the problem if we ever try to debug the underlying problem.  Perhaps better to signal an Error (or Warning?) first?

Perhaps many such errors are due to something like "1.1em" getting treated as straight text but should have been dimensions (or numbers or...)

dginev: Most of the ones I debugged already had both warning and error traces before the Fatal, e.g. "missing package X" followed by "undefined" macro \y, and then "missing argument for macro \z". And it is exactly in cases where interpretation went awry and arguments don't arrive at the right places.

So I'm unsure I see the "getting lost" argument if the Fatal is missing - hitting the "100 errors" fatal is a fine upper bound, and in the cases where we're at fewer errors, an HTML doc can be created. And often reading through the HTML file is more helpful than the error messages themselves, for a TeX author.

brucemiller: Sure, in the specific cases you're looking at, yeah there will be lots of other errors. But if you're in a situation with just a few errors, you've hidden an important symptom.  If you want to take that approach, better would be to return Dimension(0) rather than $self (which will likely just generate even more errors to paper over)

dginev:
Dimension(0) it is!

There are some very rewarding wins actually, e.g. `hep-th0001047` swtiches from a Fatal to:
```
Status:conversion:2 
318 warnings; 88 errors; 15 undefined macros[{opening}, \ellipticalarc, \setdashes, \setplotsymbol, \setsolid, \putrectangle, \circulararc, \putrule, \beginpicture, \setlinear, \setshadesymbol, \tencirc, \subtitle, \setcoordinatesystem, \plot]; 3 missing files[crckapb.cls, bezier.sty, pictex.sty]
```

lib/LaTeXML/Core/Box.pm
brucemiller: Could even check the box contents for whether it looks like a number :>....  For another time

brucemiller:
Oh yeah, that's a real nice improvement! :>

PR #1242, opened 2020-03-08
paranoid xml:id thawing from XML chunks

### LINKED ISSUE #1240: Perl die: bad ns attribute
We seem to have a regression, judging by the latest arXiv run. A timed link (will expire on next rerun) for the reports is:

https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/perl/die?offset=2700&page_size=100

Example message:
```
Perl died bad ns attribute! at /usr/lib/x86_64-linux-gnu/perl5/5.22/XML/LibXML.pm line 1690. 
  In Post::Document[1909.00250.html] ->addNodes <= Post::CrossRef[@0x542de50]
```

It's possible there are other die-related regressions, but it seems significant enough, as Fatals rose by 1%

dginev:
Fixes #1240 .

Turns out, after an afternoon of gradual debugging and tracing (= learning the CrossRef ropes), I tracked the issue for the new arXiv regression down to a problem that was already known and addressed in old versions of libxml.

In fact, the problem is so identical to the "pre 1.59" situation of libxml, that dropping that guard leads to all my test conversions succeeding cleanly. This is quite strange given that I have some of the most recent versions of libxml (C and Perl)... But it is a working fix!

Submitting it as a temporary patch, if desired. I tried to dig around the XML::LibXML documentation for a flag that would allow the chunk parser to understand the `xml:` namespace, but to no avail. The only "correct" idea I have is to add a dedicated namespace declaration to each chunk, but that is quite wasteful in serialization time -- the existing patch is quite adequate in comparison.

I should also mention that this regression was exhibited by the two remote worker servers as well, which run slightly older OS-es and libxml versions, but run close to bleeding edge XML::LibXML versions. Might be a perl wrapper change, if we're to continue digging...

### DIFF
--- a/lib/LaTeXML/Common/XML/Parser.pm
+++ b/lib/LaTeXML/Common/XML/Parser.pm
@@ -45,7 +45,7 @@ sub parseChunk {
   # In 1.58, the prefix for the XML_NS, which should be DEFINED to be "xml"
   # is sometimes unbound, leading to mysterious segfaults!!!
 ###  if (($xml_libxml_version < 1.59) && $hasxmlns) {
-  if (($LaTeXML::Common::XML::xml_libxml_version < 1.59) && $hasxmlns) {
+  if ($hasxmlns) {
     #print STDERR "Patchup...\n";
     # Re-create all xml:id entrys, hopefully with correct NS!
     # We assume all id are, in fact, xml:id,


brucemiller:
Yeah, well, ugh! It ends up replacing all "id" attributes, but then it passes Travis, so that's good.  Based on your comments, I concocted this test program which we might send to the XML::LibXML maintainers; What does it give on your broken systems?  (BTW: I assume this will fix #1229, as well)
[xmltest.txt](https://github.com/brucemiller/LaTeXML/files/4307112/xmltest.txt)

dginev:
```
$ perl test.pl 
VERSIONS:
  Perl        v5.28.1
  XML::LibXML 2.0201
  libxml      20904
INPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum"><text xml:id="S3.E2.2a">2</text></tag>
OUTPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum">
  <text :id="S3.E2.2a">2</text>
</tag>
```

dginev:
The key is running it a few times, then you see a lot of alterations for the id line:
run 2:
```
  <text 0���"V:id="S3.E2.2a">2</text>
```
run 3:
```
  <text �#��U:id="S3.E2.2a">2</text>
```

brucemiller:
The key thing is that it's getting noise for the prefix.  I get 
```VERSIONS:
  Perl        v5.28.2
  XML::LibXML 2.0134
  libxml      20909
INPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum"><text xml:id="S3.E2.2a">2</text></tag>
OUTPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum">
  <text xml:id="S3.E2.2a">2</text>
</tag>
```
All my versions are newer, except XML::LibXML.

dginev:
Decided to also update to the latest XML::LibXML:
```
$ cpanm XML::LibXML
$ perl test.pl 
VERSIONS:
  Perl        v5.28.1
  XML::LibXML 2.0202
  libxml      20904
INPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum"><text xml:id="S3.E2.2a">2</text></tag>
Wide character in print at test.pl line 19.
OUTPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum">
  <text P*?�U:id="S3.E2.2a">2</text>
</tag>
```

I worry that the actual libxml releases are somewhat sluggish however, unsure if they won't take multiple months to get a patch in, assuming they decide to place a patch...

dginev:
cpanm allows me to also downgrade:
```
$ cpanm https://cpan.metacpan.org/authors/id/S/SH/SHLOMIF/XML-LibXML-2.0134.tar.gz
$ perl test.pl 
VERSIONS:
  Perl        v5.28.1
  XML::LibXML 2.0134
  libxml      20904
INPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum"><text xml:id="S3.E2.2a">2</text></tag>
Wide character in print at test.pl line 19.
OUTPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum">
  <text �'_�U:id="S3.E2.2a">2</text>
</tag>
```

Seems to also fail with your version

dginev:
From our `monster` server:
```
VERSIONS:
  Perl        v5.22.1
  XML::LibXML 2.0123
  libxml      20903
INPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum"><text xml:id="S3.E2.2a">2</text></tag>
Wide character in print at test.pl line 19.
OUTPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum">
  <text ���:id="S3.E2.2a">2</text>
</tag>
```

From our `mercury` server:
```
VERSIONS:
  Perl        v5.28.1
  XML::LibXML 2.0201
  libxml      20904
INPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum"><text xml:id="S3.E2.2a">2</text></tag>
OUTPUT: <tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum">
  <text xml:id="S3.E2.2a">2</text>
</tag>
```

Nice! mercury has the reliably correct behavior, I reran a few times to make sure.

dginev:
Somewhat eye-popping. Exactly the same version numbers on a Debian 10 machine work fine, but on my Ubuntu 19.10 fail. Getting even subtler... ?

dginev:
Ha! I found something cool. It turns out that you can actually use the main `parse_string` methods for chunks! So changing the minimal file to:

```perl
my $xml = $parser->parse_string($chunk);
```

Reliably outputs:
```
OUTPUT: <?xml version="1.0"?>
<tag xmlns="http://dlmf.nist.gov/LaTeXML" role="refnum">
  <text xml:id="S3.E2.2a">2</text>
</tag>

```

On both my ubuntu desktop and the monster server.  And one can get rid of the pesky xml declaration with preceding the `toString` with:
```
$XML::LibXML::skipXMLDeclaration = 1;
my $serialized = $xml->toString(1);
```

I will close here and make a new PR with that solution, if it sounds acceptable?

brucemiller:
Sadly, not good, because a "chunk" includes document fragments, whereas parse_string only accepts a single document (or node).

However, I noticed in the perldoc that ```parse_xml_chunk``` is deprecated in favor of ```parse_balanced_chunk```.  The reason given seemingly is about confusion with the name, but the code is actually slightly different (non-SAX oriented).  Maybe try changing that in the test script on the offending machines?

dginev:
Sadly, `parse_balanced_chunk` exhibits the exact same problem as `parse_xml_chunk`. Could we claim that all cases that have an `xml:id` are single node cases? I could use `parse_string` when an `xml:id` is present, and keep using the `xml_chunk` approach otherwise.

dginev:
Moving to the next PR, also easier to follow without the huge comment trace.

Issue #1227, opened 2020-01-15
TikZ output lacks text on Fedora 31


flyn-org:
Something seems to have broken TikZ output on Fedora 31, but I am not yet sure what. When I view LaTeXML's output using a browser, none of the text appears in the SVG diagrams that result from TikZ. Here is an minimal example:

```
\documentclass{article}
\usepackage{fullpage}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
\node {XXX};
\end{tikzpicture}
\end{document}
```

This produces a page containing "XXX" when processed the following commands on Fedora 30:

```
latexml --destination foo.xml foo.tikz
latexmlpost --destination foo.html --format=html5 foo.xml
```

The same commands on Fedora 31 produce an empty page, other than the LaTeXML footer.

I am the maintainer of the Fedora LaTeXML package. I understand that this might be a packaging/distribution issue, but I am not yet sure how to debug it.

dginev:
Thanks @flyn-org ! 

Could you quickly provide the output of `tex --version` on that machine, so that we know what we're debugging against? The tikz bindings rely on the native texlive distribution, and changes to texlive that latexml can not yet handle are the likely culprit here.

flyn-org:
```
# tex --version
TeX 3.14159265 (TeX Live 2019)
kpathsea version 6.3.1
Copyright 2019 D.E. Knuth.
There is NO warranty.  Redistribution of this software is
covered by the terms of both the TeX copyright and
the Lesser GNU General Public License.
For more information about these matters, see the file
named COPYING and the TeX source.
Primary author of TeX: D.E. Knuth.
# rpm -q texlive
texlive-2019-13.fc31.noarch
```

[foo.tex.txt](https://github.com/brucemiller/LaTeXML/files/4068356/foo.tex.txt)
[foo.xml.txt](https://github.com/brucemiller/LaTeXML/files/4068357/foo.xml.txt)
[foo.html.txt](https://github.com/brucemiller/LaTeXML/files/4068358/foo.html.txt)
[LaTeXML.css.txt](https://github.com/brucemiller/LaTeXML/files/4068367/LaTeXML.css.txt)
[ltx-article.css.txt](https://github.com/brucemiller/LaTeXML/files/4068368/ltx-article.css.txt)

pksweby:
Hi,
I have a similar problem on windows using MiKTeX:
```
tex --version
[...]

PR #1243, opened 2020-03-09
Add missing xcolor primitive for basic Tikz load in texlive 2019

dginev:
All tikz-related examples on texlive 2019 currently lead to an Error status under the latexml master, as the latest pgf kernel also invokes `\XC@tgt@mod` from xcolor.sty, which is not yet added to that binding.

I've added the basic version of the macro, which gets my basic showcase Tikz examples working. There is a separate regression to address with #1227 , which is unrelated to this basic fix.


brucemiller:
Seems helpful; I can't even tell what it's supposed to do, but often ends up defined as you've defined  it. Thanks!

PR #1245, opened 2020-03-09
Use parse_string for XML chunks with xml:id

### LINKED ISSUE #1240: Perl die: bad ns attribute

dginev:
Fixes #1240 , as a better approach to #1242 .

In essence, I noticed that `parse_string` has a reliably correct behaviour over `xml:id`-carrying XML chunks. While it can not handle all chunk cases, it appears that our use case for CrossRef in particular is to store a single XML node, which `parse_string` handles correctly.

Thus this PR, which should be a targeted improvement. I tested two of the arXiv Fatals and the issue is resolved entirely also with this changeset. 

I have also tested the `parse_string` approach on 3 separate machines - my desktop, the mercury and monster servers, on the minimal example Bruce prepared. Looks consistently solid.

### DIFF
--- a/lib/LaTeXML/Common/XML/Parser.pm
+++ b/lib/LaTeXML/Common/XML/Parser.pm
@@ -34,7 +34,13 @@ sub parseChunk {
   my ($self, $string) = @_;
   my $hasxmlns = $string =~ /\Wxml:id\W/;
   # print STDERR "\nFISHY!!\n" if $hasxmlns;
-  my $xml = $$self{parser}->parse_xml_chunk($string);
+  my $xml;
+  if ($hasxmlns) {
+    local $XML::LibXML::skipXMLDeclaration = 1;
+    $xml = $$self{parser}->parse_string($string);
+  } else {
+    $xml = $$self{parser}->parse_xml_chunk($string);
+  }
   # Simplify, if we get a single node Document Fragment.
   #[which we, apparently, always do]
   if ($xml && (ref $xml eq 'XML::LibXML::DocumentFragment')) {


brucemiller:
As I said, this doesn't work in general: fragments can be stored in the database[*]


[*] Well, it certainly has been true in the past. And yes, they are annoying. Perhaps would be better design to assure that there aren't any? Maybe there currently aren't any?  You'd have to track down every ```findnode`` and make sure it selects only a single node. Not pleasant.

dginev:
Indeed, you raise valid points.

It would not be pleasant to track it down, but may be quite pleasant to forbid directly from the ObjectDB interface? All we need is a change in `LaTeXML::Util::ObjectDB::Entry`'s `encodeValue` and `decodeValue`, where the XML case is checked for having "multiple nodes", in which case it is mapped to an array of `->toString(1)` strings for each node, and then deserialized respectively.

I'd much rather not use advanced XML features when the wrapper is bearing its teeth at us...

dginev:
And indeed, looking closer into `encodeValue`, I see:
```perl
# The node is cloned so as to copy any inherited namespace nodes.
  elsif ($ref =~ /^XML::/) {
    return "XML::" . $value->cloneNode(1)->toString; }
```

`cloneNode` is a method only on `XML::LibXML::Node` (and inheriting classes such as Element), so we seem to be encoding/decoding Node-level data when it comes to using `parseChunk`. And the array behaviour is already supported, which is what I assume will happen for `findvalues` results, which ought to be packaged as a perl array of libxml nodes?

brucemiller:
You're apparently correct;  I tested DLMF and indeed there seems to be no multi-node fragments going into/out of the database (or at least there no longer are).  Which is good since you were gambling on only getting fragments that didn't also have xml:id! :>  Which then obscured the fact that you were returning a Document some times, and (possibly) a DocumentFragment others. But actually, the callers are expecting an Element, .... 

Anyway, enough harping. As long as we're going ```parse_string```, let's go all in.  See #1247 for a simpler alternative to parseChunk, which also has a bit of a guard in case we ever do try to store a DocumentFragment.

dginev:
Very cool! The simplified version is quite an improvement -- I got scared of the tests passing when I first tried swapping in `parse_string` and didn't want to rock the diff-boat too much, so kept it as a fallback. But I am very much in favour of your clean refactor, so closing here.

And excitingly - arXiv is about to get sane again!

PR #1246, opened 2020-03-10
Allow for raw definitions in bibliography post-processing

dginev:
Today @stamer alerted me by email about another weakness in our attempt to faithfully convert bibliographies in the MakeBibliography post-processor.

Namely, he had an example which interpreted `some.cls` raw via `--includestyles` in the main processing, which then failed miserably when its `some.bib` file was being converted in post-processing. The failure is easy to explain - `--includestyles` is never made visible to the post-processor, so the search for the `.cls.ltxml` file fails and the entire conversion of the bib file surrenders.

Now, I had a very happy thought while understanding the error trace. During post-processing, all packages and classes that have been copied over from the Core processing have essentially been "vetted". Namely, if the core conversion managed to "succeed" with them, we should assume the post-processor could be less cautious with re-vetting them. In particular, I could have the includestyles switch **always on** in MakeBibliography, as:
 - sty/cls files in the XML PIs with bindings will still have the bindings read first
 - sty/cls files in the XML PIs files without bindings, must have been (~successfully) interpreted raw by Core, so they can be interpreted raw by MakeBibliography.

Thus, I am offering a minimal change that flips `includestyles` to on for the MakeBibliography `->convert` calls.

While testing that I also realized a side flaw that also could be easily improved (thanks to babel.sty, as usual). When a class is preloaded via `--preload=some.cls --includestyles`, it does not get loaded by LoadClass, but via InputDefinitions. This leads to `\@classoptionslist` not being set for the preloaded class, and could result in additional trouble - which it does when that class loads babel internally. To fix that, I moved the logic introducing that macro into `InputDefinitions`, while still checking for the `cls` type, so that using that API call is sufficient.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -2251,7 +2251,12 @@ sub InputDefinitions {
     && grep { $prevname eq $_ } @{ LookupValue('@masquerading@as@class') || [] };
 
   $options{raw} = 1 if $options{noltxml};    # so it will be read as raw by Gullet.!L!
-  my $astype = ($options{as_class} ? 'cls' : $options{type});
+  my $astype = ($options{as_class} ? 'cls' : $options{type} || 'sty');
+
+  if ($astype eq 'cls' and $options{options}) {
+    PushValue(class_options => @{ $options{options} });
+    # ? Expand {\zap@space#2 \@empty}%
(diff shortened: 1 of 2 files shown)


brucemiller:
I wonder what the optimal division between the two should be; or whether preloading should even call LoadClass... but probably can't (shouldn't?) process options quite the same way before there's even TeX being read?

PR #1247, opened 2020-03-10
Simplify storage of XML in ObjectDB to only be single nodes

brucemiller:
Simplify storage of XML in ObjectDB to only be single nodes (not DocumentFragment); this avoids the buggy parse_xml_chunk method of XML::LibXML which fails (sometimes) to bind the xml prefix to the namespace


dginev:
Very neat refactor, LGTM !

PR #1248, opened 2020-03-12
faster GenerateID for [ids]latexml.sty workflows (arXiv)

dginev:
Working on reducing the rate of arXiv timeouts. One of them, a document with 1600 formulas, is the stress test behind the PR. Converting it with latexml-master, took **37min**. Here's the local command to get that:

```
 time latexmlc 2002.10839.tex --dest=test.html --preload=[ids]latexml.sty \
                              --pmml --cmml --timeout=600000
```

Time in master:
```
real	37m0.826s
user	36m57.642s
sys	0m0.907s
```

I ran the conversion with a huge timeout setting, while making dinner. Meanwhile,

Time with this PR is **10min**:
```
real	10m16.427s
user	10m15.048s
sys	0m0.807s
```

The main workload reduction is from now also ignoring `XMArg` when running GenerateID. I was inspired by the rule to ignore XMWrap and tried adding arg in, and the improvement was geometric (to my surprise).

A smaller speedup is from using the `->parentNode` loop instead of running the XPath, motivated by the observation that the parent node almost always has an id, and most calls execute in constant time, rather than having to setup an xpath context.

With this PR, The first expansion/building phase takes 3 minutes, the math parsing (with associated GenerateIDs) is now 2 min and 40 seconds. Post-processing spends 1min 30sec on Scan and 2 min 40sec on the MathML post-processing. The rest of the phases take smaller amounts of 10 or less seconds. (In master, 28 mins were in math parsing,  the post-proc times were comparably the same).

I took the liberty of moving the `Tag('ltx:*',..` rule for id generation out of the sizeable TeX.pool and relocated it to the only piece of code that required it - `latexml.sty.ltxml`. I also tried to straighten out some bits of logic in GenerateID, given that it is a very hot path for the arXiv runs, which have thousands of formulas, each with hundreds of xml:id attributes.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -828,26 +828,38 @@ sub ResetCounter {
 # <number> for <prefix> amongst its descendents.
 sub GenerateID {
   my ($document, $node, $whatsit, $prefix) = @_;
-  # If node doesn't already have an id, and can
-  if (!$node->hasAttribute('xml:id') && $document->canHaveAttribute($node, 'xml:id')
-    # but isn't a _Capture_ node (which ultimately should disappear)
-    && ($document->getNodeQName($node) ne 'ltx:_Capture_')) {
-    my $ancestor = $document->findnode('ancestor::*[@xml:id][1]', $node)
-      || $document->getDocument->documentElement;
-    ## Old versions don't like $ancestor->getAttribute('xml:id');
-    my $ancestor_id = $ancestor && $ancestor->getAttributeNS($LaTeXML::Common::XML::XML_NS, 'id');
-    # If we've got no $ancestor_id, then we've got no $ancestor (no document yet!),
-    # or $ancestor IS the root element (but without an id);
-    # If we also have no $prefix, we'll end up with an illegal id (just digits)!!!
-    # We'll use "id" for an id prefix; this will work whether or not we have an $ancestor.
-    $prefix = 'id' unless $prefix || $ancestor_id;
+  # Nothing to do if xml:id already set, or not allowable by schema
+  if ($node->hasAttribute('xml:id') || !$document->canHaveAttribute($node, 'xml:id')) {
+    return; }
+  # Also, we want to be avoid adding ids to otherwise valid elements, which are not inherently content
+  # e.g. ltx:document which only gets an id from an EXPLICIT \thedocument@id.
+  # e.g. isn't a _Capture_ node, or XMWrap (which ultimately should disappear)
+  my $name = $node->nodeName || '';
+  if ($name =~ /^(?:_Capture_|document|XM(?:Wrap|Arg))$/) {
+    return; }
+  ## Old versions don't like ->getAttribute('xml:id');
+  # This is hot code when [ids]latexml.sty is active, optimized for performance
+  my $at_node = $node->parentNode;
+  my ($ancestor, $ancestor_id);
+  while ($at_node && !$ancestor_id && ($at_node->nodeType == XML_ELEMENT_NODE)) {
+    $ancestor_id = $at_node->getAttributeNS($LaTeXML::Common::XML::XML_NS, 'id');
+    $ancestor    = $at_node;
+    $at_node     = $at_node->parentNode;
+  }
+  # If we've got no $ancestor_id, then we've got no $ancestor (no document yet!),
+  # or $ancestor IS the root element (but without an id);
+  # If we also have no $prefix, we'll end up with an illegal id (just digits)!!!
+  # We'll use "id" for an id prefix; this will work whether or not we have an $ancestor.
+  $prefix = 'id' unless $prefix || $ancestor_id;
 
+  my $ctr = 1;
+  if ($ancestor) {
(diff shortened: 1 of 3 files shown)


dginev:
I'm a little concerned that the tests passed originally when I avoided XMArg _entirely_ from GenerateID, I had to back off and ignore it only for the `[ids]` option call, so that both the tests pass and the arXiv article gets the speedup.

May be worth a careful review - but it may get us a huge win in the current 1.19% of jobs that are timing out.

dginev:
Here's the source link if you feel like reproducing (though you should give it some time):
https://arxiv.org/e-print/2002.10839

brucemiller:
interesting. Nice (potential) speedup!!!  I'm a little concerned about the change of behaviour (though I'm not sure how "advertised" it was). In moving the Tag, it's purely latexml.sty feature and no longer respects ```GENERATE_IDS```; also noids has no effect (think the awkward latexml[ids noids])

dginev:
Quick add that `1001.0618` is also mitigated via this patch, and another couple I tried with lots of math.

I think the `GENERATE_IDS` is something only latexml developers currently know about - what is advertised is the `[ids]latexml.sty` option. And given that the `noids` is the default behaviour, I am ready to place some bets that no one anywhere has ever used that switch :> `[ids noids]` is very artificial - but if you worry about it, I could reach in and re-enable it. I'd much rather completely obsolete "noids" as a concept, given that it's the default behaviour.

dginev:
Also, `noids` is a bad name -- it implies there will be _no_ `xml:id` attributes at all, rather than it won't generate ids. Should've been `someids` vs `allids` really.

brucemiller:
Yeah, well you can rationalize in several directions. Such as whatever feature you're talking about, <feature> enables it and <nofeature> disables it; anything else is counterintuitive.

The big dramatic, 21min down to 6 (for me) is simply excluding XMArg.  How can that possibly be so much! There can't be so many XMArg?!?! (Commenting out the Tag completely drops to under 3min). We must be Doing Something Wrong!

dginev:
Something pretty special about this example is that a very large number of the parse attempts fail. There is a chance that failed parses are just really slow compared to successful parses.

(A lot of them also fail for silly reasons, such as trailing comas in superscripts)

brucemiller:
Actually, it's not so much the cost of generating the ids, but the cost of *having* ids!  The two phases scan and parsing significantly increase runtime; the latter dramatically; which makes sense, since they've got significantly more bookkeeping to deal with: Scan has more ids to record; Since parsing does a lot of rearrangement of XML, it has to check that modified nodes with ids are have any references maintained (it can't assume there aren't references).  Apparently, the big offender is at line 355 in MathParser.pm, which interestingly enough is primarily concerned with XMArg's that may have moved or disappeared.

Now, LaTeXML places ids whereever it *needs* them, so the ids option is icing; you want "everything" to have an id. Why not XMArg? (they hopefully go away in parsing, but I don't think that's guaranteed).  But since it's optional icing, I suspect the best approach is to add the (unneeded) ids' at the latest stage, at least after parsing.

dginev:
Thanks for troubleshooting! It certainly sounds logical that if MathParser is doing lots of metadata bookkeeping, adding the ids on the notable nodes will run up the runtime cost.

Taking a step back, the main motivation behind `[ids]latexml.sty` originally came from the preference of MathWebSearch to have easy pointers to all math nodes, ideally being able to point to an id for any content symbol or application tree. This has remained the main use of them (e.g. llamapun doesn't use the id information at all), and I am not really aware of other projects relying on that feature (e.g. we also never used them in Authorea, and sTeX had its own id-generation logic).

All that to say, that in my mind the `XMArg` elements are as immaterial as the `XMWrap` elements - they are useful internal XMath constructs, but are not crucial for the applications we have in mind. Similarly, it is not the case today that all elements get an id, even with the `[ids]` option - e.g. `<tag>` and `<tags>` do not, possibly others. Which is perfectly OK, as the original use case has always been to have an easy id handle on all pieces of _content_ in the final HTML document (or XML, either way).

Why not simplify a little, and maybe clarify the intent of the option in the documentation? I think that would increase usability.

brucemiller:
Maybe you're right, but I recall the request being somewhat in the context of locators and other such; that is, it was more general-purpose motivated.  If it were really so narrow, it may not even be needed; if you are only interested in math, and request pmml and cmml, all the math is going to end up with ids anyway (and of course, cross-referenced).

In any case, I don't think you can say that XMArg or XMWrap are any less material than any other XMath element; they *should* disappear, but if they don't they'll be reflected in the generated mml and presumably their ids get carried over.  All in all, as it gets more selective, it may be harder to document, rather than easier.

Anyway, I pursued the approach of deferring ids as late as possible. As usual, it wasn't *quite* as easy as expected, but take a look at #1254 and see what you think.

dginev:
Alright, I just thought and prod to see if anything would give for simplifying the interface right now, but it's definitely not the motivation behind working on the PR. If we can remove the performance overhead, problem=solved. I'd probably raise the `[noids]` being a weird name again at some point, but it's nothing urgent in itself. Thanks for taking my stab and making it much better!

PR #1249, opened 2020-03-12
[infinite loop patch] wrapped beginBibliography in omnibus

dginev:
Maybe not the highest-impact timeout, but since I've spotted it... The minimal example is:
```tex
\documentstyle[aas2pp4]{article}
\begin{document}
\begin{references}

\reference{key} A

\end{references}

\end{document}
```

Which leads to:
 - `Info:note:OmniBus Loading OmniBus class to attempt to cover missing options`
 - `PATCHING with \begin{thebibliography}`
 - `Error:undefined:\reference The token T_CS[\reference] is not defined.`

at which point an infinite loop takes place at `\end{references}`. I don't exactly follow, but there seems to be some unfortunate ladder of re-leting macros to values that form a cycle. I read around and noticed that `\lx@bibliography` in LaTeX.pool has a special begingroup/endgroup wrapped around its `beginBibliography`, so that "redefinitions don't take effect", which sounded very topical. Adding that guard to OmniBus eliminates the loop, hence this PR.

Would be great if Bruce could double-check whether this is sensible given the (admittedly Error-level) example, and maybe even whether `beginBibliography` should have the groups enforced internally (and always).

### DIFF
--- a/lib/LaTeXML/Package/OmniBus.cls.ltxml
+++ b/lib/LaTeXML/Package/OmniBus.cls.ltxml
@@ -129,7 +129,10 @@ DefConstructor('\references',
     . "bibstyle='#bibstyle' citestyle='#citestyle' sort='#sort'>"
     . "<ltx:title font='#titlefont' _force_font='true'>#title</ltx:title>"
     . "<ltx:biblist>",
-  afterDigest => sub { beginBibliography($_[1]); });
+  afterDigest => sub {
+    $_[0]->begingroup;    # wrapped so redefns don't take effect!
+    beginBibliography($_[1]);
+    $_[0]->endgroup; });
 
 DefConstructor('\endreferences',
   "</ltx:biblist></ltx:bibliography>");


brucemiller:
This doesn't quite cover it, since there's a surprising unexpected ```\end{document}```, and an extra bibliography generated.  I think the problem is that the LaTeX.pool bibliography handing evolved and OmniBus (and probably some others) haven't tracked it.  But it's kinda messy. Still trying to sort out what's actually going on and missing.

dginev:
Data comment: Spotted another instance of the infinite loop in another arXiv doc, where it minimally looked as:
```tex
\documentclass[a4paper,twoside,10pt]{letter}
\references

Woeger, F., von der Luehe, O., Reardon, K. 2008, 75W

\endreferences

\end{document}
```

Almost identical, but thought I'd note it.

brucemiller:
Have you figured out a failure "signature"? What seems to be the issue (if my interpretation is correct) is that bibliography-like environments that contain actual bib info *may* need a bit of extra prep. LaTeX.pool has it, but OmniBus didn't have it and seems to need it.  Also revtex, amsrefs, aas don't, but would like to verify if they need it.

dginev:
Sure, the pattern is from LaTeX.pool's `setupPseudoBibitem`:
```perl
Let('\save@bibitem', '\bibitem');
Let('\bibitem',      '\restoring@bibitem');
DefMacro('\restoring@bibitem',
  '\let\bibitem\save@bibitem\let\par\save@par\bibitem');
```

The issue arises when in the example I gave in the previous comments, this subroutine is entered while `\bibitem` has a current meaning of `\restoring@bibitem`. At that time, an infinite loop occurs, as `\restoring@bibitem` keeps invoking itself, now that `\save@bibitem` has been set to it. I think the recursion check fails to activate due to things getting re-\let again and again?

Not sure what to make of it in terms of improving the PR, but those are the concrete loop mechanics I see.

dginev:
Now that I think about it, this could indeed be a good case study for improving further the `deep_recursion` failsafe that you recently added, if possible. As an add-on to fixing the actual loop.

brucemiller:
Sorry I wasn't clear; the pattern I was looking for was: How to find arXiv docs that fail due to this?
(Part of the problem is testing ```\if@lx@inbibliography``` which doesn't get set true in all cases)

dginev:
No reliable pattern, because the timeouts are rather opaque / look roughly the same. 

The way I found the examples above was by going through a number of timeouts by hand and spotting the similarities - but nothing in the message reports guided me to this case in particular.

dginev:
But let me give you at least one specific paper to examine. As you can tell this was the first PR I opened, so indeed one of the first timeouts to be spotted from cortex:
https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/timeout/timedout?all=false

second entry - `astro-ph0001480`, which happens to have the details string:

```
Conversion timed out at x0614_timing.tex; line 790 col 4 - line 790 col 4 ALRM In Core::Definition::Primitive[\let Toke... /home/deyan/perl5/lib/perl5/LaTeXML/Package/TeX.pool.ltxml; line 1270
```

If you want to reverse-engineer a pattern to look for, trying to find timeouts with `\let` or `\bibitem` or any other of the related macros in the details, may be a lead. But they are not "required" to time out inside a macro invocation, so a lot of these cases could go unnoticed / have opaque details strings. Timeouts are delivered by Perl at a somewhat arbitrary execution point, so they're the worst category in which to look for systematic patterns (at least for now).

brucemiller:
I think #1256 may be a more targetted fix.

dginev:
Looks great!

PR #1250, opened 2020-03-14
PGF processing on a tighter compute budget

dginev:
Another mini-project motivated by arXiv conversion Fatal:timeouts. The minimal snippet was:
```tex
% source: arXiv 1002.3757
\documentclass{article}
\usepackage{tikz}

\newcommand\dw[2]{\draw[#1!#2,fill=#1!#2]}

\begin{document}
\begin{tikzpicture}
\def\LL{10}
\pgfmathsetmacro{\LLQ}{200/(\LL*\LL)}
\def\list{0,.2,...,\LL}
\foreach \x in \list {
    \foreach \y in \list {
        \pgfmathsetmacro{\d}{(\LLQ*(\LL*(\x + \y) - \x*\x - \y*\y))}
        \dw{black}{\d} (\x, \y) +(-.1, -.1) rectangle ++(.1, .1);
    }   
}
\end{tikzpicture}
\end{document}
```

In order to keep things fast I developed my speedups with `\def\LL{2}`. The times in question are on my x230 thinkpad laptop, so a little slower than usual:

| LL | converter | time sec |
|----:|----------:|-----------:|
| 2   | master | 95 |
| 10 | master | 1860|
| 2   | PR       |    30 | 
| 10 | PR       |  420 |   
|   2| pdflatex | 0.7   |
| 10 | pdflatex| 5.6   |

Here is the PDF with LL 10 for intuition's sake:
[itertikz.pdf](https://github.com/brucemiller/LaTeXML/files/4332266/itertikz.pdf)
And the zip-ed HTML+SVG
[testpr.zip](https://github.com/brucemiller/LaTeXML/files/4332306/testpr.zip)

The intuition is the usual -- nothing is _particularly_ slow about Tikz, there is just a lot of it getting processed due to running the full TeX machinery. I spent more time that I'd like to admit hunting for easy wins, and I found one big one and a couple of smaller ones.

1. The big winner is actually a cheat in `pgfmathparse`. It turns out even in the cases where the expression is simple arithmetic, the RecDescent machinery takes quite some time. And an advanced drawing would have thousands of expressions. In this example the nested loops samples a total of 2500 (x,y) coordinates, and computes an expression for each of them, only then proceeding to draw. Turns out parsing 2500 expressions with RecDescent just takes too long for a 30 min timeout. However, it is near-instant when we let `perl` give them a go -- just an `eval $string` call is a geometric speedup, even if most of the expressions end up failing (which I doubt). I also tried some improvements with regexes, but didn't see as noticeable a speedup over the grammar approach. One could even take it further and have a quick pre-scan turning all units into perl floats, and then passing the resulting expression into an "eval" call.

2. 10% speedup. Error-only uses of locators were calling getLocator in tens of thousands of instances where there was no error. In fact NYTProf reported `getLocator` as a top 15 subroutine in my tikz tests, which was surprising -- especially since the locators were never preserved in the two cases I refactored. The changes here only call the subroutine when it is actually used, losing some accuracy as it now reports on the end rather than the start position of the error. Something to consider, maybe the old behavior could be kept when a debug flag is on.

3. Opted in for `$stomach->digest` over `Digest` to save on overhead in pgf "hot paths" that get visited frequently.

4. Added perl bindings for `\pgfmathsetmacro` and `\pgfutil@in@`, which were both simple and often used enough to try. It had marginal success, a second or two shaved here or there, but seems reasonable enough to keep.

Still quite far from pdflatex, but promising in terms of eliminating arXiv fatals, at least to an extent.

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -29,9 +29,10 @@ use LaTeXML::Core::MuGlue;
 use base qw(LaTeXML::Common::Object);
 #**********************************************************************
 sub new {
-  my ($class) = @_;
+  my ($class, %options) = @_;
   return bless {
-    mouth => undef, mouthstack => [], pushback => [], autoclose => 1, pending_comments => []
+    mouth     => undef, mouthstack => [], pushback => [], autoclose => 1, pending_comments => [],
+    verbosity => $options{verbosity} || 0
   }, $class; }
 
 #**********************************************************************
@@ -367,6 +368,7 @@ sub readBalanced {
   my ($self, $expanded) = @_;
   my @tokens = ();
   my ($token, $level) = (undef, 1);
+  my $startloc = ($$self{verbosity} > 0) && $self->getLocator;
   # Inlined readToken (we'll keep comments in the result)
   while ($token = ($expanded ? $self->readXToken(0, 1) : $self->readToken())) {
     my $cc = $$token[1];
@@ -384,9 +386,9 @@ sub readBalanced {
   if ($level > 0) {
  # TODO: The current implementation has a limitation where if the balancing end is in a different mouth,
  #       it will not be recognized.
+    my $loc_message = $startloc ? ("Started at " . ToString($startloc)) : ("Ended at " . ToString($self->getLocator));
     Error('expected', "}", $self, "Gullet->readBalanced ran out of input in an unbalanced state.",
-      "Ended at " . ToString($self->getLocator));
-  }
+      $loc_message); }
   return (wantarray ? (Tokens(@tokens), $token) : Tokens(@tokens)); }
 
 sub ifNext {
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -367,7 +367,6 @@ sub readBalanced {
   my ($self, $expanded) = @_;
   my @tokens = ();
   my ($token, $level) = (undef, 1);
-  my $startloc = $self->getLocator;
   # Inlined readToken (we'll keep comments in the result)
   while ($token = ($expanded ? $self->readXToken(0, 1) : $self->readToken())) {
     my $cc = $$token[1];
@@ -386,7 +385,7 @@ sub readBalanced {
  # TODO: The current implementation has a limitation where if the balancing end is in a different mouth,
  #       it will not be recognized.
     Error('expected', "}", $self, "Gullet->readBalanced ran out of input in an unbalanced state.",
-      "started at " . ToString($startloc));
+      "Ended at " . ToString($self->getLocator));
   }
   return (wantarray ? (Tokens(@tokens), $token) : Tokens(@tokens)); }
 
--- a/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
+++ b/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
@@ -297,8 +297,18 @@ sub pgfmathparse {
   my $input = $string;
 ###    if ($string =~/^+/){        # Don't parse as expression, just as glue. ???
   #    print STDERR "\nPGF Math Parse: $string\n";
-  $PGFMATHGrammar = Parse::RecDescent->new($PGFMATHGrammarSpec) unless $PGFMATHGrammar;
-  my $result = $PGFMATHGrammar->expr(\$string) || 0.0;
+  my $result;
+  {
+    local $@;
+    $result = eval $string;
+    if (!$@) {
+      # need to erase string when perl eval works, to keep it consistent with recdescent
+      $string = ''; }
+    else {    # failed to eval, use tikz grammar
+      $PGFMATHGrammar = Parse::RecDescent->new($PGFMATHGrammarSpec) unless $PGFMATHGrammar;
+      $result         = $PGFMATHGrammar->expr(\$string); }
+  }
+  $result = $result || 0.0;
   #  $result = "0.0" if $result eq "0";
   #    print STDERR "  GOT " . (defined $result ? $result : '<fail>') . "\n";
   if ($string) {
--- a/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
+++ b/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
@@ -300,6 +300,7 @@ sub pgfmathparse {
   my $result;
   {
     local $@;
+    no warnings;
     $result = eval $string;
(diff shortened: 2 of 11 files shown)


dginev:
Edited my original description, and forgive the `infinite loop` misnomer - there is in fact a perfectly reasonable and finite loop, which I found in the course of working with the branch+example. It just takes 31 minutes to convert to SVG.

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -367,7 +367,6 @@ sub readBalanced {
   my ($self, $expanded) = @_;
   my @tokens = ();
   my ($token, $level) = (undef, 1);
-  my $startloc = $self->getLocator;
```
dginev: I am not too happy about losing the accuracy here... maybe we could use the locator of the first read token instead...

brucemiller: Knowing where a failed balanced sequence started is very helpful for debugging. My first thought was maybe could depend on verbosity, but likely the cost of repeatedly looking up the switch may cancel the benefit.

dginev: Actually the verbosity flag is completely workable, it is much much cheaper to execute a conditional, compared to calling `getLocator`.

dginev:
I have rebased on master and added a object-level verbosity for Gullet -- since whether we're doing things verbosely or not in the gullet calls to Error() shouldn't be dependent on the TeX state (at least not in this case).

Measuring the difference in times on the big LL 10 with the final PR state took a while as usual. 
  - Running latexml normally (verbosity=0), took 5min and 21 seconds. 
  - Running latexml `--verbose`, took 5min and 36 seconds.

So the difference is not massive, 100*(14/336) = 4% speedup for a large-ish tikz figure. Still noticeable.

lib/LaTeXML/Package/pgfmath.code.tex.ltxml
```diff
@@ -297,8 +297,19 @@ sub pgfmathparse {
   my $input = $string;
 ###    if ($string =~/^+/){        # Don't parse as expression, just as glue. ???
   #    print STDERR "\nPGF Math Parse: $string\n";
-  $PGFMATHGrammar = Parse::RecDescent->new($PGFMATHGrammarSpec) unless $PGFMATHGrammar;
-  my $result = $PGFMATHGrammar->expr(\$string) || 0.0;
+  my $result;
+  {
+    local $@;
```
brucemiller: You're going to want 
```
local $LaTeXML::IGNORE_ERRORS = 1;
```
here, otherwise we're still registering errors. They don't get reported, but if they exceed 100, you get a Fatal! which shows up in the final status!

brucemiller:
Very interesting... and almost too tricky? I figured this needed some hands-on, so I dusted off my TeXamples framework. That led to the comment about needing to bind ```$LaTeXML::IGNORE_ERRORS```. And while debugging, I discovered that most perl parse failures were trivial things like ```1.0em```! I'm experimenting with a regexp as phase 0, which usually leads to further speadup (but, occasionally slightly slower). Also, a number times a register. Still experimenting....

dginev:
Certainly a little tricky, and one could argue the only really valuable part of the PR is using `eval` as early as possible for pgfmath. The speedup is just too significant to ignore. Let me know how the experimentation goes, the direction you headed into for further optimizations is something that came to mind to me as well - could lead to another set of big speedups!

brucemiller:
Parsing simple dimensions with a regexp helps a little, if done right since pgf has some curious units management. And the perl seems safe if you don't get overloaded by internal errors, and if you note that pgf's trig use degrees by default.

Since I've mangled your fork by merging in other LaTeXML changes, maybe the simplest approach is to paste my diff:
```
--- a/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
+++ b/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
@@ -288,6 +288,7 @@ RawTeX(<<'EoTeX');
 \@ifundefined{pgfmathmathunitsdeclaredtrue}{\newif\ifpgfmathmathunitsdeclared}{}
 EoTeX
 
+our $PGMATH_UNITS_REGEXP = undef;
 sub pgfmathparse {
   my ($gullet, $tokens) = @_;
   SetCondition(T_CS('\ifpgfmathunitsdeclared'),     0, 'global');
@@ -295,20 +296,28 @@ sub pgfmathparse {
   my $string = (ref $tokens ? ToString(Expand($tokens)->stripBraces) : $tokens);
   $string =~ s/^\s+//; $string =~ s/\s+$//;
   my $input = $string;
-###    if ($string =~/^+/){        # Don't parse as expression, just as glue. ???
-  #    print STDERR "\nPGF Math Parse: $string\n";
   my $result;
-  {
+  $PGMATH_UNITS_REGEXP
+    = join('|', qw(em ex mu),keys %{$STATE->lookupValue('UNITS')})
+    unless $PGMATH_UNITS_REGEXP;
+  # Also common would be unit=\pgflinewidth (height?)
+  if($string =~ /^([+-]?[\d\.]+)($PGMATH_UNITS_REGEXP)$/){
+    $result = pgfmath_convert($1,$2);
+    $string = ''; }
+  if($string && !$result
+   && ($string !~ /\W(?:sin|cos|tan|cot|sec|cosec)\W/)){  {
+    local $LaTeXML::IGNORE_ERRORS = 1;
     local $@;
     no warnings;
     $result = eval $string;
     if (!$@) {
       # need to erase string when perl eval works, to keep it consistent with recdescent
       $string = ''; }
-    else {    # failed to eval, use tikz grammar
-      $PGFMATHGrammar = Parse::RecDescent->new($PGFMATHGrammarSpec) unless $PGFMATHGrammar;
-      $result         = $PGFMATHGrammar->expr(\$string); }
-  }
+    } }
+  if($string && !$result){
+    $PGFMATHGrammar = Parse::RecDescent->new($PGFMATHGrammarSpec) unless $PGFMATHGrammar;
+    $result         = $PGFMATHGrammar->expr(\$string); }
+
   $result = $result || 0.0;
   #  $result = "0.0" if $result eq "0";
   #    print STDERR "  GOT " . (defined $result ? $result : '<fail>') . "\n";
(END)
```

dginev:
Alright, thanks for the improvements! Now I am getting very curious to try these over arXiv. I rebased on master, which also solved the warnings in my tikz example from the issue description, as expected. And quite nicely the performance is now 305 seconds for the LL 10 example, which was originally 6 times slower! Not bad!

Ready for a merge when you are :+1:

brucemiller:
let's do it! :>

PR #1252, opened 2020-03-15
\documentstyle for .sty targets

dginev:
Motivated by [this arxiv paper](https://arxiv.org/abs/cond-mat/0003484), while looking for what turned out to be an incidental [ltx:XMTok in ltx:p](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/malformed/ltx%3AXMTok?all=false) malformed error.

The main issue with the document is an old use of `\documentstyle` that really targets a `.sty` file, and one that is distributed together with the paper sources - a custom `jpsj.sty`.

I added a check for a finding `.cls` file, which when OK proceeds, but when failed tries to load a `.sty` file of the same name, with a preloaded OmniBus.cls. It improved the status of the document substantially (even threw in a `\@options` helper definition for it, but since alignments were involved it still failed flat after a certain point. I thought I would pass the work along as a PR, whether or not it's deemed impactful in the long run.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -80,7 +80,13 @@ DefConstructor('\documentstyle OptionalSemiverbatim SkipSpaces Semiverbatim []',
     # So, we'll just try for a .cls, punting to OmniBus if needed.
     # If we start wanting to read style files by default, we'll still need to handle this
     # specially, since class (or sty files pretending to be) cover so much more.
-    LoadClass($class, options => $options, after => Tokens(T_CS('\compat@loadpackages')));
+    if (FindFile($class, type => 'cls', notex => !LookupValue('INCLUDE_STYLES'))) {
+      LoadClass($class, options => $options,
+        after => Tokens(T_CS('\compat@loadpackages'))); }
+    else {
+      InputDefinitions('OmniBus', type => 'cls', noerror => 1, handleoptions => 1);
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -80,7 +80,13 @@ DefConstructor('\documentstyle OptionalSemiverbatim SkipSpaces Semiverbatim []',
     # So, we'll just try for a .cls, punting to OmniBus if needed.
     # If we start wanting to read style files by default, we'll still need to handle this
     # specially, since class (or sty files pretending to be) cover so much more.
-    LoadClass($class, options => $options, after => Tokens(T_CS('\compat@loadpackages')));
+    if (FindFile($class, type => 'cls', notex => !LookupValue('INCLUDE_STYLES'))) {
+      LoadClass($class, options => $options,
+        after => Tokens(T_CS('\compat@loadpackages'))); }
+    else {
+      InputDefinitions('OmniBus', type => 'cls', noerror => 1, handleoptions => 1);
+      RequirePackage($class, options => $options,
```
brucemiller: I'm wondering if this should have the ```as_class=>1``` option? I think it is sorta intended for this purpose.

dginev: Sounds reasonable, added!

brucemiller:
Doh! silly me; of course this doesn't completely fix the test case w/o some sort of includestyles trick, like #1253

PR #1253, opened 2020-03-15
Restricted inclusion of raw styles to source directory

dginev:
A proposal pull request for the arXiv conversion, motivated by the hope that the recent timeout-resolving PRs buy us more time to do raw interpretation.

A wide range of arXiv articles contain custom `.sty` and at times `.cls` files, which are often easy for latexml to interpret. Until now we have ignored them, under the rightful fear that `--includestyles` will interpret texlive packages that latexml fails badly on, leading to otherwise avoidable Fatals and Errors. We've talked on a number of occasions that it would be informative to do an experiment that only loads raw files from the source directory of a conversion job, which should bring in more positives than negatives (in theory).

So, this PR offers one approach to start such an experiment, via a new experimental latexml.sty option. As a running example, take the minimal:
```tex
\documentclass{memoir}
\usepackage{mysty}
\begin{document}
hello
\end{document}
```

With some trivial/empty `mysty.sty` in the same directory. Running a simple `latexml test.tex` yields, among other messages:
```
Warning:missing_file:memoir Can't find binding for class memoir (using OmniBus)
...
Warning:missing_file:mysty Can't find binding for package mysty
```

Turning on `--includestyles` allows latexml to find both files, and to fail quite badly on memoir, with 40 errors just for reading the `.cls` file. Now, this PR allows to run:
```
latexml test.tex --preload=[rawsourcestyles]latexml.sty
```

Which only yields:
```
Warning:missing_file:memoir Can't find binding for class memoir (using OmniBus)
```

while successfully loading `mysty.sty`.

Definitely open to renaming any/all new bits of code here, it's quite odd to name within the existing scheme, without majorly refactoring other intuitions. For now I intended this option to be used in isolation / apart from `--includestyles`, to avoid the mixed intuitions of using them together. At least good names don't come to me intuitively, I end up with:
```
--includestyles --preload=[toplevelincludesonlyforsourcedir]latexml.sty
```
Similarly for the `restricted` option, maybe it should be `restricted_raw_paths` or some such?

In any case, would love to run this setup over arXiv, and see how the percentages compare, especially after the rest of the upgrade PRs land.

### DIFF
--- a/lib/LaTeXML/Core.pm
+++ b/lib/LaTeXML/Core.pm
@@ -51,7 +51,9 @@ sub new {
     'global');
   $state->assignValue(GRAPHICSPATHS => [map { pathname_absolute(pathname_canonical($_)) }
         @{ $options{graphicspaths} || [] }], 'global');
-  $state->assignValue(INCLUDE_STYLES => $options{includestyles} || 0, 'global');
+ # For now the "includestyles" option passed from --includestyles will accept both classes and styles?
+  $state->assignValue(INCLUDE_STYLES  => $options{includestyles} || 0, 'global');
+  $state->assignValue(INCLUDE_CLASSES => $options{includestyles} || 0, 'global');
   # Core has to ensure a default input encoding, and we default towards modern utf-8 documents
   # This can be removed when all executables rely on LaTeXML::Common::Config
   $options{inputencoding} = "utf-8" unless $options{inputencoding};
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1859,7 +1859,7 @@ sub pathname_is_raw {
   return ($pathname =~ /\.(tex|pool|sty|cls|clo|cnf|cfg|ldf|def|dfu)$/); }
 
 my $findfile_options = {    # [CONSTANT]
-  type => 1, notex => 1, noltxml => 1, no_kpsewhich => 1 };
+  type => 1, notex => 1, noltxml => 1, searchpaths_only => 1 };
 
 sub FindFile {
   my ($file, %options) = @_;
@@ -1927,7 +1927,7 @@ sub FindFile_aux {
     (!$options{notex} ? ($file) : ()));
   local $ENV{TEXINPUTS} = join($Config::Config{'path_sep'},
     @$paths, $ENV{TEXINPUTS} || $Config::Config{'path_sep'});
-  if (my $result = (!$options{no_kpsewhich}) && pathname_kpsewhich(@candidates)) {
+  if (my $result = (!$options{searchpaths_only}) && pathname_kpsewhich(@candidates)) {
     return (-f $result ? $result : undef); }
   if ($urlbase && ($path = url_find($file, urlbase => $urlbase))) {
     return $path; }
@@ -2251,7 +2251,7 @@ sub AddToMacro {
 my $inputdefinitions_options = {    # [CONSTANT]
   options => 1, withoptions => 1, handleoptions => 1,
   type    => 1, as_class    => 1, noltxml       => 1, notex => 1, noerror => 1, after => 1,
-  no_kpsewhich => 1 };
+  searchpaths_only => 1 };
 #   options=>[options...]
 #   withoptions=>boolean : pass options from calling class/package
 #   after=>code or tokens or string as $name.$type-h@@k macro. (executed after the package is loaded)
@@ -2288,7 +2288,7 @@ sub InputDefinitions {
         "Option clash for file $filename with options '$curroptions'",
         "previously loaded with '$prevoptions'") unless $curroptions eq $prevoptions; } }
   if (my $file = FindFile($filename, type => $options{type},
-      notex => $options{notex}, noltxml => $options{noltxml}, no_kpsewhich => $options{no_kpsewhich})) {
+      notex => $options{notex}, noltxml => $options{noltxml}, searchpaths_only => $options{searchpaths_only})) {
     if ($options{handleoptions}) {
       Digest(T_CS('\@pushfilename'));
       # For \RequirePackageWithOptions, pass the options from the outer class/style to the inner one.
@@ -2354,7 +2354,7 @@ sub InputDefinitions {
 
 my $require_options = {    # [CONSTANT]
   options => 1, withoptions => 1, type => 1, as_class => 1,
-  noltxml => 1, notex       => 1, raw  => 1, after    => 1, no_kpsewhich => 1 };
+  noltxml => 1, notex       => 1, raw  => 1, after    => 1, searchpaths_only => 1 };
 # This (& FindFile) needs to evolve a bit to support reading raw .sty (.def, etc) files from
 # the standard texmf directories.  Maybe even use kpsewhich itself (INSTEAD of pathname_find ???)
 # Another potentially useful option might be that if we are reading a raw file,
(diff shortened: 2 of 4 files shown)


brucemiller:
Cool that you took a look at this! The code is surprisingly simple; I guess switching kpsewhich on/off is really the main issue; I think I always wanted to make it too complicated!  I do think the naming could be improved. If it really boils down to "no kpsewhich", perhaps all the switches should say so, instead of "restricted" (which is unclear) and rawresources (since "resources" potentially means too many things).  Hmm...  Let me think (obl movie quote https://www.youtube.com/watch?v=AZn9ghG9nRw)

dginev:
Definitely can agree with the direction of renaming you're going toward. Also, you've reminded me of a remix of that great quote, which you may find absolutely appalling, depending on your tolerance for auto-tune. But proper link etiquette must be followed: https://youtu.be/0EygqL--RW4

dginev:
Well, here's a stab at slightly better names. The tricky bit is that saying "no_kpsewhich" is only true+intuitive at the FindFile level, where it's simply being disabled.

At the top level (say latexml.sty), it is not true that we're turning off all kpsewhich - we are turning it off for "top-level" loads, i.e. for LoadClass and RequirePackage. But if we have something like tikz, kpsewhich will be happily used to load all sub-package dependencies (which get loaded via InputDefinitions). 

Hence, I used `no_kpsewhich` at the Package.pm level, but kept the longer and more intentional `INCLUDE_STYLES_NO_KPSEWHICH` for the state variable name.

Similarly, I think it's a completely different consideration how to name the option for a user's eyes in latexml.sty, I tried to upgrade that again, currently at `[rawsourcestyles]latexml.sty`. I really wanted it to capture the entire use case in the option name.

brucemiller:
Indeed, thanks for updating, but as soon as I posted I started thinking that kpsewhich is too specific to the internal method that happens to be used at the moment.  Perhaps "local" or "local_only" is more meaningful in the relevant contexts?

And then I was wondering whether having two ```INCLUDE_STYLES``` switches that interacted in a possibly non-intuitive way was right. And whether it was worth a distinction between reading style and class files. I haven't managed to turn those questions into a backward compatible, intuitive proposal yet, however.

dginev:
How about keeping it specific with `searchpaths_only`, for the Package methods? Then it's still immediately clear "this will only use the search paths to load resources of the desired types".

I was trying to stay away from refactoring `includestyles`, which was my first actual instinct:
* `--includestyles=no`
* `--includestyles=local`
* `--includestyles=all`

brucemiller:
That's close to my thinking, although I hadn't gotten as far as commandline or options; focusing on the logic.  Namely, as you say, 3 choices on interpreting raw: not at all, only local sources, or any found.  And one conceivably might want to differentiate  style files from class files (eg. interpret all style files, but only local class files); Possibly a good arXiv choice would be interpret local style files, but not class files at all.  And probably want a good shorthand like ```--includestyles``` that says read 'em all from whereever :>

brucemiller:
On re-reading your previous comment, ```searchpaths_only``` seems to be only standalone sensible at the FindFile level; at RequirePackage/LoadClass it seems to be a potential value (or alternative to) ```notex```, since it affects what we're looking for as well as where.

dginev:
It sounds like we're growing to two different considerations
 1. What do we include? Nothing (i.e. only .ltxml), styles, classes, all
 2. Where do we look for assets? Searchpaths (always, at minimum), searchpaths + tex installation

dginev:
So maybe the internal options ought to be a boolean `search_tex_installation`, as well as `allowed_raw_assets` which could be an open-ended array of extensions - `tex`, `sty`, `cls`, `def`, `ltx`, ...

brucemiller:
Hmm, I think we're starting to skew orthogonal :>  We always accept .ltxml (ignoring notex, noltxml which are somewhat different purpose); if we don't find, we'll look for raw either everwhere, locally (or searchpaths only, as you say), or not at all, ie. fail.  And we might want to apply a different choice for different categories of files.

dginev:
Certainly, I wrote my last 2 comments with that already in mind.

brucemiller:
Hmm, yeah, ```searchpaths_only``` has the right feel for the option to FindFile, RequirePackage, LoadClass. What I'm not yet clear about is the control variables and latexml.sty option names & structure...

dginev:
Well, I'm not too sure how to keep the ball rolling constructively, since my instinct continues to be too major a change -- I would be tempted to either redo `--includestyles` entirely (maybe even changing its name, among adding values), or to keep things minimal and just "patch them on" the includestyles naming intuitions - which is what I did in this PR as a stopgap.

brucemiller:
I've got the same problem. I'm assuming we're on the same page w.r.t desirability of distinguishing styles and classes --- at least for some limited arXiv testing?

Here's a proposal that tries to slip between counterintuitively convoluted and backward incompatible.
* Keep what you have for FindFile, InputDefinitions, but rename ```no_kpsewhich``` to ```searchpaths_only```
* ```INCLUDE_STYLES``` variable can be: 1 (include all); ```searchpaths```; 0 or undef meaning don't interpret.  This applies only to style files
* New ```INCLUDE_CLASSES``` same as above but applies to class files;  Optionally for backward compatibility: if undef,  copy ```INCLUDE_STYLES```

So, for RequirePackage, you'd want to do:
```
   $options{searchpaths_only} //= (!$options{notex}) && ((LookupValue('INCLUDE_STYLES') || 0) eq 'searchpaths');
```
For LoadClass, essentially the same expect STYLES => CLASS.

Finally, latexml.sty could get two new options: ```includestyles``` and ```includeclasses``` which, if no value provided, sets them to 1; otherwise takes something like ```(true|false|searchpaths)```.

Does that make sense and seem appropriate?

dginev:
Sounds great! I'll implement and send an update in a jiffy.

dginev:
One question is now that I am separating `INCLUDE_STYLES` from `INCLUDE_CLASSES` internally, whether the default `--includestyles` behaviour will still include both `.cls` and `.sty` or I should limit it to only `.sty`. It's a single bit of weirdness to be handled In `LaTeXML::Core::new`, but maybe worth raising...

dginev:
Alright, better names are here. For arXiv, I'd imagine all we'd like is a `--preload=[localrawstyles]latexml.sty`.

I am more and more fond of using the word `raw` in the user-facing options, and not using the word `include`. Currently the implication is that the inclusion is raw. I'd rather switch to an implication that the `raw` asset will be included.

dginev:
Also, since the latexml.sty package options are not key-values, I had to be a little creative with the names. 
 * rawstyles - "load all styles found anywhere"
 * localrawstyles - "load all styles found in searchpaths"
 * norawstyles - "do not load any raw styles" 

analogous, styles -> classes for the classes.

dginev:
Once we're all set, I am planning to rerun the ~10,000 articles from the [testbench corpus](https://corpora.mathweb.org/corpus/arxiv_testbench/tex_to_html) with all three configurations - no, local and all raw "styles". I am assuming turning on raw classes is something we can do later, when we feel it makes sense.

brucemiller:
Good job sorting out that "specification"! Actually, in the end pretty close to what you started with.

There's one bit left, where we just recently modified ```\documentstyle```. I suspect line 83 in LaTeX.pool should now use ```INCLUDE_CLASSES```

dginev:
rebase with push+force, and done!

brucemiller:
Fantastic! Looks good.

PR #1254, opened 2020-03-16
Defer generating full ids until finalizing the document.

brucemiller:
Defer generating full ids until finalizing the document. This avoids expensive verification of cross-references while parsing.  ```GenerateID``` becomes a bit contrived to try to keep ids "sensible" both when full ids are requested, and when not. Since this generates ids in different order, some are changed, thus regenerated daemon test cases

See #1248 for more details;   This reduces the run time for the referenced document to under 4 minutes, from about 22 ! (the previous patch was about 6 minutes).


dginev:
I like it! And you're saying it's faster than removing XMArg? That's great news! I like the optimization with gathering the nodes in the parentNode traversal and id-ing them together, that's neat&efficient!

This PR along should have a substantial impact on the arXiv conversion runtime, so I am quite excited to merge the fastest improvement. We can refactor the public interface at a later time, if needed, I just thought I'd prod and see how it is received. LGTM!

dginev:
Also, on my machine my PR took 10 minutes, while this improved implementation only took 6.5 min! Convincing numbers.

brucemiller:
The precomputing ids was more of a (mostly unsuccessful) attempt to match up the old order of generating ids, rather than efficiency, but whatev'.  Thanks for the review & even more for pointing me in the right direction!

PR #1255, opened 2020-03-17
Cleanup generated svg:g and svg:switch

### LINKED ISSUE #1227: TikZ output lacks text on Fedora 31

dginev:
Fixes #1227 with a patch.

I think there are deeper issues with the Tikz->SVG support in a specific subset of cases, but the concrete example could be addressed with some basic mitigation patching.

The key being that there is a `svg:switch` element generated holding (a foreignObject with) an empty `ltx:p` element, which is entirely unnecessary. I added a cleanup step at the end of that bit of construction code, which removes such needless elements, and also modified the Tag for `svg:switch` to unwrap the element if unnecessary (i.e. if only a single child is present).

Similarly, I noticed we generate a whole lot of redundant `svg:g` markup without any attributes, so refactored to use a dedicated subroutine which would unwrap such needless g elements. I'm now wondering whether I did that wrong, and maybe I should have it auto-unwrap also as an `afterClose` Tag rule... 

All that said, I haven't actually addressed the root cause that generates the incorrect markup in the first place, in fact I haven't even understood it yet. Reading the [official switch documentation](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/switch) I don't even think we should be using that element -- rather, we should be throwing Errors and dropping content that was "unexpected" or lead to "malformed" SVG. At least in principle.

An example that also improves with this patch is the [showcase](https://latexml.mathweb.org/editor) "Tikz AC Voltage", which has no text in master, and has all of the text visible after the PR (though the axes legends still have alignment problems). I should mention that I ran all examples under texlive 2019, since it is correctly converting on the showcase server (texlive 2015). The issues are likely all 2019 Tikz regressions.

To contrast, the PR generates this relatively compact snippet for the #1227 tex (texlive 2019):
```xml
<svg id="p1.pic1" class="ltx_picture" style="width:48.6pt;height:20.1pt;" height="27.77" overflow="visible" version="1.1" viewBox="-33.6 -13.88 67.2 27.77" width="67.2">
   <g transform="matrix(1 0 0 -1 0 0.00999999999999801)">
      <g>
         <g stroke="#000000">
            <g fill="#000000">
               <g stroke-width="0.4pt" color="#000000">
                  <g>
                     <g>
                        <path d="M -30.56 -10.84 h 61.11 v 21.68 h -61.11 Z" style="fill:none" />
                        <g>
                           <g transform="matrix(1 0 0 1 -25.94 -3.46)">
                              <g class="ltx_svg_fog" transform="matrix(1 0 0 -1 0 12.45)">
                                 <g stroke="#000000">
                                    <g fill="#000000">
                                       <foreignObject width="7.5pt" height="9.0pt">Text</foreignObject>
                                    </g>
                                 </g>
                              </g>
                           </g>
                        </g>
                     </g>
                  </g>
               </g>
            </g>
         </g>
      </g>
   </g>
</svg>
```

while master generates the longer, and failing to render its text (texlive 2019):
```xml
<svg id="p1.pic1" class="ltx_picture" style="width:48.6pt;height:20.1pt;" height="27.77" overflow="visible" version="1.1" viewBox="-33.6 -13.88 67.2 27.77" width="67.2">
   <g transform="matrix(1 0 0 -1 0 0.00999999999999801)">
      <g>
         <g stroke="#000000">
            <g fill="#000000">
               <g stroke-width="0.4pt" color="#000000">
                  <g>
                     <g>
                        <path d="M -30.56 -10.84 h 61.11 v 21.68 h -61.11 Z" style="fill:none" />
                        <g>
                           <g transform="matrix(1 0 0 1 -25.94 -3.46)">
                              <g class="ltx_svg_fog" transform="matrix(1 0 0 -1 0 12.45)">
                                 <switch>
                                    <foreignObject color="#000000" height="100%" overflow="visible" width="51.89">
                                       <p class="ltx_p" />
                                    </foreignObject>
                                    <g stroke="#000000">
                                       <g fill="#000000">
                                          <switch>
                                             <foreignObject width="7.5pt" height="9.0pt">Text</foreignObject>
                                          </switch>
                                       </g>
                                    </g>
                                 </switch>
                              </g>
                           </g>
                        </g>
                     </g>
                  </g>
               </g>
            </g>
         </g>
      </g>
   </g>
</svg>
```

### DIFF
--- a/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
+++ b/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
@@ -96,7 +96,7 @@ DefConstructor('\lxSVG@insertpicture{}',
     $document->openElement('svg:g',
       transform => "matrix(1 0 0 -1 0 $props{flipnmove})");
     $document->absorb($content);
-    while ($document->maybeCloseElement('svg:g')) { }
+    while (maybeCloseG($document)) { }
     #                 $document->closeElement('svg:g');
     $document->closeElement('svg:svg');
     $document->closeElement('ltx:picture');
@@ -144,7 +144,15 @@ DefParameterType('SVGMoveableBox', sub {
     $box; });
 
 # Try to avoid errors inserting random boxes (esp. ltx:text)(Is this safe?)
-Tag('svg:switch', autoOpen => 1, autoClose => 1);    # ?
+Tag('svg:switch', autoOpen => 1, autoClose => 1, afterClose => sub {
+    my ($document, $sw) = @_;
+    my @swc = element_nodes($sw);
+    if (!scalar(@swc)) {    # empty? remove.
+      $document->removeNode($sw); }
+    elsif (scalar(@swc) == 1) {    # single child? unwrap
+      $document->replaceNode($sw, $swc[0]); }
+    # otherwise: keep as-is.
+    return; });
 Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
   afterClose => sub {
     my ($document, $node, $whatsit) = @_;
@@ -285,7 +293,7 @@ DefConstructor('\pgfsys@hbox{Number}', sub {
       if ($n->isSameNode($fo)) {
         $doc->closeElement('svg:foreignObject');
         $doc->closeElement('svg:switch');
-        $doc->closeElement('svg:g');
+        maybeCloseG($doc);
         # Check the children of the inserted ltx:p
         # We should at least try to denest svg ?
         # If svg:switch/svg:foreignObject/ltx:p/ltx:picture/svg:svg/svg:g
@@ -298,6 +306,16 @@ DefConstructor('\pgfsys@hbox{Number}', sub {
             $doc->replaceNode($sw, $g);
     } } } }
     # Could conceivably be simplifying ltx:text ???
+
+# What follow are a series of crutch measures that combat inappropriately generated markup.
+#      they are sanitzers that mostly work, but we ought to also work on improving the entire svg:switch-related
+#      machinery.
+# Empty p? Remove, and unwind up the parent hierarchy
+    if (!element_nodes($p)) {
+      my $node_to_remove = $p;
+      while (scalar(element_nodes($node_to_remove->parentNode)) == 1) {
+        $node_to_remove = $node_to_remove->parentNode; }
+      $doc->removeNode($node_to_remove); }
   },
   reversion => sub {
     my $box = $_[0]->getProperty('thebox');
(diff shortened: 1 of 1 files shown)


dginev:
I am not quite certain if I am even suggesting this is merged - it's more of a step towards understanding the regression, maybe pointing to where to look next.

dginev:
The specific `ToString($box)` that is getting absorbed with `$doc->absorb($box);` of `\pgfsys@hbox`, and causes the empty ltx:p,  is (again, texlive 2019):

```
\hbox{{\definecolor{pgfstrokecolor}{rgb}{0,0,0}\pgfsys@color@rgb@stroke{0}{0}{0}{}\pgfsys@color@rgb@fill{0}{0}{0}{}\hbox{{Text}}
```

brucemiller:
There's a couple of things going on here: the redundant ```svg:g``` and the whole foreign object bit. I'm fuzzy on the required structure of the latter; I thought it needed ```svg:switch```, but maybe not. And it doesn't seem needed here anway.

The redundant ```svg:g``` is trickier; they're getting opened & closed by pgf commands, and I worry that by optionally opening & closing, they may end up closing the "wrong" one. Probably what is called for is a pass at the end to elliminate, and even combine, redundant ones (watching for compatible attributes, of course).

brucemiller:
```svg:switch``` does seem crutchy, but without it, uses of ```svg:foreignObject``` don't validate (using jing on the LaTeXML schema).

And there are certainly many redundant and unnecessary ```svg:g``` elements, but what remains still has to nest the same way. In many of my tikz samples, lines and things are disappearing.

dginev:
Ugh. Looks like that needless restriction has been removed in SVG2.0, which even has an explicit mention to `MathML` in the foreignObject documentation. Took me a while to find the 2.0 schema to validate against, it's here (`SVG2.0.rng`) in case anyone needs it:
https://dev.w3.org/SVG/profiles/1.2T/master/relaxng/

But it looks like we're some distance away from properly handling these advanced examples (with multiple `\hbox` calls interplaying), so maybe I start the small arXiv reruns before we resolve here. Thanks for the feedback!

Another note is that the unlike the PNG images the SVG graphics end up inline in the HTML, so they get bundled in the redistributed arXMLiv datasets. I try to keep those small, so eventually/gradually dispensing with the unnecessary elements will be helpful. Should also make the rendering simpler/more reliable.

PR #1256, opened 2020-03-17
Update users of beginBibliography

brucemiller:
Update users of beginBibliography to call beforeDigestBibliography when actually digesting bibitems so appropriate switches get set.  Particularly offensive was potential uses of ```\if@lx@inbibliography```.


dginev:
Tested and works great - I still don't personally follow the exact dynamics with `\if@lx@inbibliography`, but I will get there eventually. LGTM!

PR #1257, opened 2020-03-17
Random robustness fixes

brucemiller:
Random robustness fixes; added some new definitions in LaTeX; more forgiving about registers.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1073,8 +1073,11 @@ sub DefConditionalI {
     $STATE->installDefinition(LaTeXML::Core::Definition::Conditional->new($cs, $paramlist, $test,
         conditional_type => 'unless', %options),
       $options{scope}); }
-  elsif ($csname =~ /^\\if(.*)$/) {
-    my $name = $1;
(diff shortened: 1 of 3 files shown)


dginev:
Cool batch!

brucemiller:
Oh; I meant to add; it will read memoit.cls now without errors; at least if you don't do any memoir stuff (your sample file)

brucemiller:
I did randomly check a couple of arXiv articles, hoping that the register fixes would solve some. It mostly doesn't completely solve them, but at least reduces a long cascade of Errors to one. So it's probably not a significant arXiv win, but likely helps cut down the noise. Mostly it was seeing the memoir.cls fail and being surprised that they weren't much worse. For once a "this should be easy" actually was.

dginev:
Might not seem like much to reduce a stream of errors to one, but (provided enough articles are using the missing bits), such gains tend to compound with robustness fixes like these. Each individual patch PR may not be too impactful on its own, but getting 10 to 20 of these little guys positioned may lead to a dramatic improvement in the CorTeX runs, at least in performance, eventually with the % rates.

PR #1258, opened 2020-03-19
add \@cite macro from latex.ltx

dginev:
This is a minimal/piecemeal PR, motivated from an arXiv article (hep-lat0006026) that had Fataled with  >100 errors, but which passes with a single error (undefined `\@cite`) when local .sty files are loaded.

It's piecemeal because the PR doesn't actually provide working TeX-level cite support, it just emulates the macro to the point of depositing placeholders akin to `[?,?]` -- still better than an error, and neat to see a Fatal-to-Warning transition.

We may or may not consider locking certain "core" macros in LaTeX.pool (such as `\cite`) when also interpreting local .sty files...


brucemiller:
sure, why not. that is how it's defined, but not clear how it should be used...

PR #1261, opened 2020-03-20
avoid dividing by zero in pgfmath

dginev:
Continuing my work on the line between a Fatal and an Error document, here is a PR that carefully avoids dividing by zero in pgfmath -- which happens at times in Error-level arXiv documents, instead using a pragmatic epsilon for such cases.

Will likely produce deeply broken output, but if you have one broken figure in a large document, you still get the rest of the document out.

### DIFF
--- a/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
+++ b/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
@@ -40,10 +40,11 @@ InputDefinitions('pgfmath.code', type => 'tex', noltxml => 1)
 # Note that these macros typically get a CS passed as argument whose expansion is the number
 # and that they assign the result, as a token list to \pgfmathresult.
 # Hopefully the savings in doing the math in Perl isn't overwhelmed by string conversion?
-our $PI    = Math::Trig::pi;
-our $LOG2  = log(2);
-our $LOG10 = log(10);
-our $E     = exp(1);
+our $PI      = Math::Trig::pi;
+our $LOG2    = log(2);
+our $LOG10   = log(10);
+our $E       = exp(1);
+our $epsilon = 0.00001;
 
 # Note: We need to lookup /pgf/trig format/deg/ or /rad/ !!!  (default is deg?)
 sub pgfmathargradians {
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/pgfmath.code.tex.ltxml
```diff
@@ -468,23 +470,23 @@ BEGIN {
...
-    pi         => sub { $PI; },
-    abs        => sub { abs($_[0]); },
-    acos       => sub { acos($_[0]); },
-    array      => sub { },
-    asin       => sub { rad2deg(asin($_[0])); },
-    atan       => sub { rad2deg(atan($_[0])); },
-    atan2      => sub { rad2deg(atan2($_[0], $_[1])); },
+    '/'        => sub { my $divisor = (0.0 + $_[1]) || $epsilon; $_[0] / $divisor; },
+    divide     => sub { my $divisor = (0.0 + $_[1]) || $epsilon; $_[0] / $divisor; },
+    div   => sub { my $divisor = (0.0 + $_[1]) || $epsilon; int($_[0] / $divisor); },
```
brucemiller: I'm all for avoiding fatals, but, again, skipping over it silently seems dangerous for non-arXiv usecases.

dginev: Well, it won't be silent - the cases where a zero sneaks in almost invariably have a register warning of `expected:<number>` with `Missing number, treated as zero` details string.

brucemiller: In arXiv, where you only find valid TeX (cough), sure; divide by zero is probably due to some LaTeXML register mishandling that "assumed 0". But in general usage, it might show up while somebody's tweaking code for a graphic.

dginev: Here's a baseline TeX mistake where division by 0 happens:
```tex
\documentclass[tikz]{standalone}
\begin{document}
\begin{tikzpicture}
  \draw (0,1) node {\pgfmathparse{1/0}\pgfmathresult\ result prints ok} ;
\end{tikzpicture}
\end{document}
```

pdflatex will print:
```
! Package PGF Math Error: You'veaskedmetodivide`1'by`0',butIcannotdivideanynumb
erby`0' (in '1/0').
```

latexml master will Fatal. This PR will produce a broken rendering with "no problem" status. I suppose I can shoot off a warning akin to this PGF Math Error, so that a trace is preserved...

dginev:
Alright, now a 0 divisor also emits a warning.

brucemiller:
Cool!

PR #1262, opened 2020-03-20
Elevate fallback interface to all LaTeXML::Common::Object

dginev:
A follow-up to my great line of "robustness at all cost" reasoning that we explored in #1241 .

It turns out in the arXiv Fatals we also see cases of `getX`, `ptValue` or `valueOf` getting called on `Token` and `Tokens` objects, not just boxes. Luckily, there is a common thread -- all of these inherit from `Common::Object`!

Hence, this PR elevates the "fallback interface" up to Object, and I'm also throwing in `valueOf` which I noticed breaking one doc. Clawing my way to less Fatals, one fix at a time...

### DIFF
--- a/lib/LaTeXML/Common/Object.pm
+++ b/lib/LaTeXML/Common/Object.pm
@@ -165,11 +165,25 @@ sub unlist {
   return $self; }
 
 #**********************************************************************
+# What about Kern, Glue, Penalty ...
+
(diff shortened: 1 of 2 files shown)


brucemiller:
Indeed, this looks better and covers more cases. Thanks!

PR #1263, opened 2020-03-20
Dont fatal when ExpandedIfToken is missing; expanded \accent arg

dginev:
an arXiv fatal (`alg-geom9503004`) caused by an attempted macro improvement:
```tex
% intelligent \"; one can type	\"iiii instead of \"\i iii
%
\def\"#1{{\accent"7F \if#1i\i\else#1\fi}}
```

Looks easy enough to avoid with a small upgrade - `Expanded` letter argument for `\accent` fixes the specific issue. And I added a lenient fallback to `T_CS('\@empty')` when a token argument for a conditional is missing, so that the follow-up checks can go through without a perl die. Also shooting off an error in those cases.

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -321,7 +321,7 @@ DefParameterType('Variable', sub {
         Error('expected', '<variable>', $gullet,
           "A <variable> was supposed to be here", "Got " . Stringify($token),
           "Defining it now.");
-        DefRegisterI($token, undef, Dimension(0)); # Dimension, or what?
+        DefRegisterI($token, undef, Dimension(0));    # Dimension, or what?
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -5091,7 +5094,7 @@ DefPrimitiveI('\pounds',    undef, UTF(0xA3));           # POUND SIGN
 # Then we can apply combining accents to it.
 sub applyAccent {
   my ($stomach, $letter, $combiningchar, $standalonechar, $reversion) = @_;
-  my $box     = Digest($letter);
+  my $box     = $stomach->digest($letter);
```
dginev: always faster to call digest directly in the stomach...

brucemiller:
Hmm....tricky, but looks about right, I think! Thanks.

PR #1265, opened 2020-03-22
use latex.ltx definition of \@makeother to properly neutralize \^

dginev:
Debugging an arXiv doc error, here is a minimal example:
```tex
\documentclass{article}
\makeatletter
% verbatimlisting for computer code
\def\verbatimlisting#1{%
  \begingroup \@verbatim
    \input#1
    \endgroup}
\makeatother

\begin{document}
\verbatimlisting{snippet}
\end{document}
```

where the snippet file is just a single line:
```tex
test ^ here
```

pdflatex happily neutralizes and provides a verbatim snippet, while latexml chokes on the `^` sign. It turned out there were two bits needed for this paper to work:
1. Switch to a `\@verbatim` definition closer to the one in latex.ltx, in particular one that uses `\@makeother` internally.

2. More importantly, switch to using the direct latex.ltx macro definition of `\@makeother`, as the perl code for it was missing on subtle trickery that makes it work in raw tex. Namely, the difference between `^` and `\^` is significant, but when one uses `\catcode`, the `\^` form is a stand-in for the bare hat. 

I won't claim I've mastered the nuance in the interpretation, but I will rely on the fact that I copy-pasted the relevant bits from latex.ltx directly, and they seem to be producing the exact right output in XML.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -1639,7 +1639,11 @@ DefRegister('\@itemdepth' => Number(0));
 # NOTE: how's the best way to get verbatim material through?
 DefEnvironment('{verbatim}',  '<ltx:verbatim>#body</ltx:verbatim>');
 DefEnvironment('{verbatim*}', '<ltx:verbatim>#body</ltx:verbatim>');
-Let('\@verbatim', '\verbatim');    # Close enough?
+# Let('\@verbatim', '\verbatim');    # Close enough?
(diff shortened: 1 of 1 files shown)


brucemiller:
Take a look at #1267; Since I ended up needing to correct and enhance some test cases, I included this case as well; so I ended up hijacking your debugging (tho' I ended up with a slightly different patch so it would create an ```ltx:verbatim``` element).

dginev:
Not too crazy about simply removing the leading backslash, but I can't come up with a counter-example so... might as well. Closing happily in favour of the more general upgrade PR.

PR #1267, opened 2020-03-24
verbatim fixes cascade

brucemiller:
Surprising usage like "{\verb x }" led to realizing that
the skipping of whitespace after a control sequence doesn't
happen till reading the next token.  "\verb" immediately
changes the catcodes of space!
But isEOL queries whether the next token read would
be an EOL, so has to be skipping_spaces aware.
This led to rethinking how "\read" determines end of line;
noticing that "\futurelet" is NOT expandable

* Defer skipping spaces until next readToken
* Peak ahead in isEOL (!)
* fix how \read determines EOL
* make \futurelet a primitive, NOT expandable
* fixup \verb handling of space,
* in \verb defer EndSemiverbatim till after digestion, so visible space works
* fix \@makeother to deal with slashed chars
* fix \@verbatim to work standalone (but in group)
* remake test cases to match latex more closely
* enhance test cases to cover more usecases

The peak ahead in isEOL is bothersome!


lib/LaTeXML/Core/Mouth.pm
```diff
@@ -201,16 +201,9 @@ sub handle_escape {    # Read control sequence
   if ((defined $cc) && ($cc == CC_LETTER)) {    # For letter, read more letters for csname.
     while ((($ch, $cc) = getNextChar($self)) && $ch && ($cc == CC_LETTER)) {
       $cs .= $ch; }
+    # We WILL skip spaces, but not till next token is read (in case catcode changes!!!!)
+    $$self{skipping_spaces} = 1;
```
dginev: Ah, crucial difference, nicely spotted.

dginev: Question: is this a Mouth attribute or a Gullet attribute? Could there be a situation where one mouth sets the skipping spaces mode, but they go on skipping into a second mouth?

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -1689,37 +1704,37 @@ DefPrimitiveI('\@vobeyspaces', undef, sub {
...
 
 DefConstructor('\@text@verb{}{}', "<ltx:verbatim font='#font'>#2</ltx:verbatim>",
-  beforeDigest => [sub { $_[0]->bgroup; MergeFont(family => 'typewriter'); }],
-  afterDigest  => sub { $_[0]->egroup; },
+  afterDigest => sub { EndSemiverbatim(); },
   # Since ltx:verbatim is both inline & block, we have to fudge inline mode
   beforeConstruct => sub {
     $_[0]->canContain($_[0]->getElement, '#PCDATA')
       || $_[0]->openElement('ltx:p'); },
   reversion => '\verb#1#2#1');
```
dginev: while reviewing here I just noticed that the `\verb*` variant will also be reverted into `\verb`. Not critical for anything today, but noting in case you feel like patching.

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -4981,7 +4996,10 @@ DefMacro('\@secondoftwo{}{}',    sub { $_[2]; });
 DefMacro('\@thirdofthree{}{}{}', sub { $_[3]; });
 DefMacro('\@expandtwoargs{}{}{}', sub {
     ($_[1]->unlist, T_BEGIN, Expand($_[2])->unlist, T_END, T_BEGIN, Expand($_[3])->unlist, T_END); });
-DefMacro('\@makeother{}', sub { AssignCatcode($_[1] => CC_OTHER, 'local'); });
+DefMacro('\@makeother{}', sub {
+    my ($gullet, $char) = @_;
+    $char = ToString($char); $char =~ s/\\//;
```
dginev: I guess removing the leading backslash is really all that is needed... Fingers crossed

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -2238,9 +2239,7 @@ DefPrimitive('\read Number SkipKeyword:to SkipSpaces Token', sub {
         push(@tokens, $t);
         $level++ if $cc == CC_BEGIN;
         $level-- if $cc == CC_END;
-        last if ((($cc == CC_SPACE) && ($t->getString eq "\n"))
-          || ($cc == CC_COMMENT)
-          || ($t->equals(T_CS('\par')))) && !$level; }
+        last if $mouth->isEOL && !$level; }
```
dginev: you may find this to be a silly optimization but `!$level && $mouth->isEOL` is bound to be faster in the cases $level isn't  set -- no need to call the subroutine.

dginev:
I tried hard to critique, but really couldn't -- all changes seem like very solid upgrades! Good to merge as-is, feel free to cherry pick any of my minor comments to taste :>

dginev:
Also, it speaks volumes about TeX's complexity when even something as simple sounding as "neutralize everything verbatim" ends up filled with sneaky edge cases.

lib/LaTeXML/Core/Mouth.pm
brucemiller: In TeX, it's a global variable. But given how it took so long to notice that the spaces hadn't yet been skipped, I'm figuring the more hidden it is, the better.

lib/LaTeXML/Package/LaTeX.pool.ltxml
brucemiller: Good catch; better fix now, else it'll show up later.

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: and certainly can't hurt; will fix.

lib/LaTeXML/Package/LaTeX.pool.ltxml
brucemiller: The TeX book is rather confusing. One form of "normal integer" is backquote followed by a "character token". The latter (which is not expanded) is defined as  " a (character code, category code) pair, or an active character, or a control sequence whose name consists of a single character."  I have no idea what the pair thing is about (no syntax is specified). The rest sounds like you should just give the character if it's active, otherwise precede it with a backslash.  However, in simple tests using pdflatex, I get the same thing for active and non-active both with and without the backslash.

brucemiller:
Surprised there's no comments about ```isEOL```. Honestly, that just *can't* be the right way! (but it works more  better than before)

dginev:
Well, my comment about whether `skipping_spaces` is an attribute of Mouth or Gullet was in a way fishing for some bedrock on which to anchor any following commentary about the API design. Since you told me you want it "as hidden as possible", I suppose hiding it there is as good as hiding it anywhere else :> I do feel sad for the previous elegance of `isEOL`, would've been great to keep.

lib/LaTeXML/Package/LaTeX.pool.ltxml
dginev: It brings me no pleasure to do this, but I sadly managed to get a torture case which would fail this perly implementation. Here it is:

```tex
\documentclass{article}
\begin{document}
\makeatletter
\catcode`\%=0
\@makeother\\
test \\ here
%end{document}
```

dginev: And now that I've taken a step back -- I am wondering whether this calls for a new `DefParameterType` that deals with what is supposedly a common situation in a number of macros?

brucemiller: latexml & pdflatex are same here.

brucemiller:
If we try to improve isEOL (or read or whatever), at least we now have more test cases to verify behaviour

lib/LaTeXML/Package/LaTeX.pool.ltxml
dginev: Ha! I should've tested against the right branch... It does indeed work. We'll see if arXiv can break it next :> Good to see you've merged already.

PR #1268, opened 2020-03-27
Fixup places where afterassignment takes place; enhance test cases

brucemiller:
Avoid using top-level ```Let``` (which invokes afterassignment), when making internal assignments, especially of the "set to relax" type.


lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -898,7 +898,7 @@ DefParameterType('CSName', sub {
 
 DefMacro('\csname CSName', sub {
     my ($gullet, $token) = @_;
-    Let($token, '\relax') unless defined LookupMeaning($token);
+    $STATE->assignMeaning($token, $STATE->lookupMeaning(T_CS('\relax'))) unless defined LookupMeaning($token);
```
dginev: Cool, I see you found a short-enough $STATE invocation! Looks very satisfying to me personally.

dginev:
Nice catch and nice upgrade! All looks good to me 👍

Issue #1196, opened 2019-08-23
TikZ output distorted


flyn-org:
First, thank you for LaTeXML!

I find that the output generated from TikZ by LaTeXML ends up distorted.

Here is an example input:

```
\begin{tikzpicture}
        \node [draw,shape=rectangle] (server) at (0, 0) {SNPC Server $S$};
        \node [draw,fill=yellow,very thick,shape=rectangle] (worker) at (0, -2) {worker $W$};
        \node [draw,fill=yellow,very thick,shape=rectangle,text width=1.25cm,align=center] (proxy) at (-3, -2) {proxy~$P$ ($I_1$)};
        \node [draw,shape=rectangle] (dependent) at (-6, -2) {dependent $D$};

        \node at (0.3, -1.55) {$C_S$};
        \node at (-1.1, -1.8) {$C_P$};
        \node at (-1.9, -1.8) {$C_W$};
        \node at (-1.9, -2.2) {($P_1$)};
        \node at (-4, -1.8) {$C_D$};
        \node at (-4.05, -2.2) {($P_2$)};

        \draw (server) -- (worker);
        \draw[dotted] (worker) -- (proxy);
        \draw (proxy) -- (dependent);
\end{tikzpicture}
```

Here is a document containing this picture, processed by pdflatex, and rendered as it should be by evince:

![pdfviewer](https://user-images.githubusercontent.com/6109247/63616350-6540aa80-c5b5-11e9-8b90-8614c85e227e.png)

Here is the same document, processed with `latexml instructions.tex >instructions.xml` and `latexmlpost --format=html5 instructions.xml >instructions.html`, and rendered by a browser with distortions:

![browser](https://user-images.githubusercontent.com/6109247/63616398-7a1d3e00-c5b5-11e9-99f4-2b3cf11c2686.png)

Here is the entire document:

[instructions.txt](https://github.com/brucemiller/LaTeXML/files/3535800/instructions.txt)

It is not clear to me how accurate the TikZ processing in LaTeXML is expected to be. I am also not sure if this is a browser issue, although I have tried Chrome, Firefox, and GNOME Epiphany (Web).

dginev:
This example is even further regressed on the latest master and texlive 2019, so there may indeed be some need to revisit the tikz interpretation internals on our part. Here is what I am seeing for this conversion:

![image](https://user-images.githubusercontent.com/348975/64267070-bb8de180-cf03-11e9-8fea-79b79fa87e0a.png)

[...]

PR #1270, opened 2020-04-07
Improve on svg:g scaffolding bindings

### LINKED ISSUE #1227: TikZ output lacks text on Fedora 31

dginev:
Fixes #1227 , and improves towards also fixing #1196 .

Second attempt after the previous #1255  PR hit a dead end. This time I decided to approach things a little closer to the TeX machinery. In particular, by making the `\lxSVG@begingroup@` , which is a constructor solely focused on the output XML markup, a little smarter.

To reiterate my previous realization, since TeX makes no inherent distinction between "drawing" and outputting "content", the pgf/tikz machinery intertwines the two at will without any negative consequences. Meanwhile, at least in the cases of `\hbox` generating `svg:foreignObject` we hit cases where the current open element in the document is in the `ltx:` namespace, but the insertion coming in is for an `svg:` construct. There are a wide range of configurations that could trigger different edge cases of this interplay, _unless_ we can provide for the general situation of "incoming SVG while inside a foreignObject's LTX".

This is what this PR attempts -- when a new `svg:g` attempts to open, it checks if the current insertion context is an `ltx:`, and if so, it "floats up" to its previous `svg:g` ancestor. I have so far only seen a need to do this with the `g` element, as it is a mostly "transparent envelope", hosting an attribute payload.

It would of course be a big problem if the ancestor g had actual drawn content, such as an `svg:path`, which is why the code explicitly targets the g's "last child" which is bound to host the current insertion context, where the foreignObject would be. Works in the #1227 example, and keeps the showcase examples in tact - in fact it _improves_ significantly on some previous regressions, including the one behind [missing diagram text](https://github.com/brucemiller/LaTeXML/issues/1196#issuecomment-527942852) for #1196. 

To wrap up, a reminder that all of these issues have only been observed (and debugged against) in texlive 2019. Older texlives seemed to get by without all of this caution, and it may be worth double-checking they still work with the PR. I assume @brucemiller could verify/contradict that :+1:

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -4034,7 +4034,7 @@ are added.
 
 C<afterOpen:early> or C<afterOpen:late> can be used in
 place of C<afterOpen>; these will be run as a group
-bfore, or after (respectively) the unmodified blocks.
+before, or after (respectively) the unmodified blocks.
 
 =item C<afterClose=E<gt>I<code>($document,$box)>
 
--- a/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
+++ b/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
@@ -474,8 +474,28 @@ DefMacro('\lxSVG@endgroup',
 DefMacro('\lxSVG@endgroups', sub {
     map { T_CS('\lxSVG@endgroup') } 1 .. LookupValue('pgf_scopecount'); });
 
-DefConstructor('\lxSVG@begingroup@ RequiredKeyVals',
-  '<svg:g %&GetKeyVals(#1)>',
+DefConstructor('\lxSVG@begingroup@ RequiredKeyVals', sub {
+    my ($doc, $kv) = @_;
+    my $g_attrs        = $kv->getKeyVals;
+    my @keys           = keys %$g_attrs;
+    my $this           = $doc->getElement;
+    my $orig_this_name = $doc->getNodeQName($this);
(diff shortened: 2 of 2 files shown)


dginev:
And I may also benefit from having more tikz examples. Maybe a "donation" of TeX snippets could be made towards the showcase at https://latexml.mathweb.org/editor ? I can add them to the plugin repository.

flyn-org:
I tested this, and it does seem to fix #1227.

brucemiller:
I still worry that ```\lxSVG@endgroup@``` will close the wrong ```svg:g``` (even with maybeCloseElement), but it does seem to work (no worse) on my test cases, and if it fixes a bug, let's go with it!  Thanks!!

dginev:
At least it will now close the wrong element of the right kind :>

PR #1272, opened 2020-04-21
close text elements when floating to their parent element

### LINKED ISSUE #1271: algorithm2e line number regression
Sometime between 195eadf4b82f361e5b434668f97751145335ead8.. 55cde51cbc635cbdd2be60584f8bb50e41bd0d80 I'm seeing an odd regression in `algorithm2e`, perhaps best explained by this diff in our integration test:

https://github.com/arxiv-vanity/engrafo/pull/973/files#diff-2ec3aacd8f525ebff7df6bf4e51c944e (click "load diff" under `algorithm2e.sty.tex/index.html`)

Seems like the line numbers and line content have been swapped in some cases. We've also got tests for `algorithm` and `listings` and they're unaffected.

dginev:
Fixes #1271 (see discussion in issue for problem diagnostic)

The Document `floatToElement` method is getting a bit subtle for me, and this PR is just the patch that happened to come to mind when poking around the example. Unsure if it's the right general treatment however.

In short, there is a special case where `floatToElement` is returned almost immediately, when the first candidate is acceptable. The return is direct under the assumption that the first candidate is always the current Document node, which is _almost_ always exactly true. There is an exception in the case where the current node is a text node, rather than a "regular" node, in which case the master `floatToElement` would return the parent element, while leaving the text node as the currently open node.

If this reads a little confusing, you see why I am sensing things feel a little subtle :>

This PR takes the approach of simply recognizing this special case, closing the current Document node when returning its parent element (or equivalently, setting its parent element as the current Document node). Fixes the reported example, tests pass, so thought I'd see how it fairs under @brucemiller review.

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -998,7 +998,10 @@ sub floatToElement {
   my @candidates = getInsertionCandidates($$self{node});
   my $closeable  = 1;
   if (@candidates && $self->canContain($candidates[0], $qname)) {
-    return $candidates[0]; }    # Fine right here.
+    # Fine right here, just return
+    # -- special case, if current is a text node, float to its parent element and return
+    $self->setNode($candidates[0]) if $$self{node}->getType == XML_TEXT_NODE;
+    return $candidates[0]; }
   while (@candidates && !$self->canContain($candidates[0], $qname)) {
     $closeable &&= $self->canAutoClose($candidates[0]);
     shift(@candidates); }
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -997,9 +997,9 @@ sub floatToElement {
   my ($self, $qname, $closeifpossible) = @_;
   my @candidates = getInsertionCandidates($$self{node});
   my $closeable  = 1;
+  # If the current node can contain already, we're fine right here - just return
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Core/Document.pm
```diff
@@ -998,7 +998,10 @@ sub floatToElement {
   my @candidates = getInsertionCandidates($$self{node});
   my $closeable  = 1;
   if (@candidates && $self->canContain($candidates[0], $qname)) {
-    return $candidates[0]; }    # Fine right here.
+    # Fine right here, just return
+    # -- special case, if current is a text node, float to its parent element and return
```
brucemiller, line 1002: Actually, this is exactly the problem. The comment could be made to reduce some of the mystery in the future. How about
```
# But don't resume at a text node, if it is current, and append more to it after other insertions.

dginev: Funnily enough, the comment is hard to read for me... Is it:

1. "Don't resume at a text node, if it is current. Don't append more to it after other insertions."
2. "Don't resume at a text node, if it is current. Instead, append more to it after other insertions."

I think you mean 1. ? It's definitely an "Edge case" of sorts, can adapt the comment to that...

brucemiller: you got it!

PR #1275, opened 2020-05-05
Migrate and enhance declaration support from DLMF

brucemiller:
This pull request enhances latexml.sty's ```\lxDeclare``` and ```\lxDefMath```.
The former, as
```
\lxDeclare [keywords]{pattern}
```
 "declares" symbols or simple expression to be of a particular grammatical role (with keyword ```role```), to support math parsing,  or ```meaning```, as well as associating a brief textual description (keyword ```tag```) with the symbol and connecting the symbol to the pllace in the document where it was declared (or "defined").

With this PR, when that tag is given, it propogates to the MathML title attribute on tokens, so it can serve as a tooltip. Also, the definition location is used for an href so the symbol links to it's "definition".

Additionally, the processing of the pattern has been made more consistent when the pattern has multiple symbols (or represents a composite symbol): The attributes are taken to apply to the matched objects as a whole, it being wrapped into a single subtree if necessary.
When ```\WildCard``` is present it matches any single subtree (this occurse *before* parsing!) and the matched expression is converted to an ```ltx:XMDual``` with the wildcards being treated as arguments.  An enhanced form ```\WildCard[subpattern]``` requires that the wildcard matches subpattern, but still treats it as a wildcard for processing.  The wrapping, in both cases, can be inhibited with the ```nowrap``` keyword, in which case attributes are applied only to the *first* symbol in the matched expressions.

Finally,
```
\lxDefMath{cs}[nargs][noptions]{expansion}[keywords]
```
which is similar to ```newcommand```, but creates (potentially semantic) math macros,
now also takes the ```tag``` keyword. If given, it also associates a description and definition location with the use of the macro to provide tooltips and links as described above.


lib/LaTeXML/Core/Rewrite.pm
```diff
@@ -348,20 +470,40 @@ sub domToXPath_rec {
...
+      $qname = '*';
+      unshift(@predicates, 'position()=' . $pos); }
+    my $preds = join(' and ', grep { $_ } @predicates);
+    return ($axis . '::' . $qname . ($preds ? "[" . $preds . "]" : ''), 1, 0, @wilds);
+  }
 
   elsif ($type == XML_TEXT_NODE) {
-###    "text()='".$node->textContent."'"; }}
-    return "*[text()='" . $node->textContent . "']"; } }
+    return ("*[text()='" . $node->textContent . "']", 1, 0); } }
```
dginev, line 482: You probably want to escape any apostrophes in the returned `->textContent` to avoid malformed XPaths, though I can see how that's a rare edge case.

brucemiller: you might be right, but I can't figure out how to get a plain apostrophe in there!!

dginev: Oh, ok. This TeX doc will give you a Fatal error:
```tex
\documentclass{article}
\usepackage{latexml}
\usepackage{textcomp}
\begin{document}

\lxDeclare[role=ID]{$x\textquotesingle$}
\end{document}
```

with:
```

Fatal:perl:die Perl died
	at String; line 0 col 0 - line 0 col 0
	XPath error : Invalid predicate
	descendant-or-self::ltx:XMTok[@role='UNKNOWN' and text()='x' and @_font and contains(@_font,'[math,') and contains(@_font,',medium,') and contains(@_font,',italic,')][following-sibling::*[position()=1 and self::ltx:XMTok and @role='UNKNOWN' and text()=''' and @_font and contains(@_font,'[serif,') and contains(@_font,',medium,') and contains(@_font,',upright,')]][@_pvis and @_cvis]
	                                                                                                                                                                                                                                                              ^
	In Core::Document[@0x56466e6558c8] undef
1 fatal error at /home/deyan/perl5/lib/perl5/LaTeXML/Common/XML/XPath.pm line 36.
```

brucemiller: Ah, very clever.  OK, try this patch.

lib/LaTeXML/Core/Definition/Constructor/Compiler.pm
```diff
@@ -150,6 +150,7 @@ sub translate_constructor {
 sub slashify {
   my ($string) = @_;
   $string =~ s/\\/\\\\/g;
+  $string =~ s/'/\\'/g;     # slashify ', since we're wrapping in ''
```
dginev, line 153: actually, may be this is a good time to remind of [quotemeta](https://perldoc.perl.org/functions/quotemeta.html). I ended up being bitten twice in a separate piece of code ( #1260 ) until I remembered to just defer to Perl's own thorough escapism, and avoid unsuspecting weirdness. It could end up a relief to replace all `slashify` calls (and lookalikes) with that.

lib/LaTeXML/Core/Rewrite.pm
```diff
@@ -479,7 +479,14 @@ sub domToXPath_rec {
...
   elsif ($type == XML_TEXT_NODE) {
-    return ("*[text()='" . $node->textContent . "']", 1, 0); } }
+    return ("*[text()=" . quoteXPathLiteral($node->textContent) . "]", 1, 0); } }
+
+# Return quoted string, but note: XPath doesn't provide sensible way to slashify ' or "
+sub quoteXPathLiteral {
+  my ($string) = @_;
+  if    ($string !~ /'/) { return "'" . $string . "'"; }
+  elsif ($string !~ /"/) { return '"' . $string . '"'; }
+  else { return 'concat(' . join(',"\'",', map { "'" . $_ . "'"; } split(/'/, $string)) . ')'; } }
```
dginev, line 489: I think you could get by without the third case, though this code looks quite workable. If you feel like a slightly simpler idea, you could standardize to always use `"` as the xpath delimiter and have a single replacement call of:

` $string =~ s/([^\\])"/$1\\"/g;`

whenever an xpath string is being evaluated inside the value of a key="value".

brucemiller: quotemeta is interesting, but we're so often doing quoting from one language to another, which screws things up; you have to know the quoting conventions involved.  Case in point: XPath 1 doesn't provide for slashifying single and double quotes. If you have a string with both, you've got to use clumsy code like the above.

brucemiller: I think (hope) we're there.  I'm going to go ahead & merge, since there's some DLMF stuff waiting on it.  Thanks for your observations!

dginev: Great!

PR #1277, opened 2020-05-08
Rework >= and <= ligatures as grammar rules

### LINKED ISSUE #1276: Request: An option to disable and/or warn on ligatures.
I ran into this when trying to compile an expression like the following:
```
\documentclass{article}
\begin{document}
\[<x,y> =x'y\]
\end{document}
```
`latexml` interprets `>=` as a ligature for `\ge`. As far as I can tell, there is no way to disable ligature parsing without disabling all of maths parsing.

I realise that it is a better practice to use `\langle`/`\rangle`, but based on the principle of least astonishment, I think that the program ought to at least warn the user, particularly where it makes a substitution that potentially changes the substance of the expression. (I.e., whether `>>` is rendered as `>>` or as `\gg`, the difference is purely typographic, whereas replacing `<x,y>=` with `<x,y\ge` creates a nonsensical expression.)

dginev:
Fixes #1276 . Relatively direct reworking that removes the ligatures and adds a grammar rule that only matches these two cases creating a lowercase `relop`, as Bruce suggested to me.

Added some related formula examples in `t/math/ambiguous_relations.tex`, also a couple that don't yet parse but eventually might, and the eventually-ambiguous examples I mentioned in the issue.


brucemiller:
Seems you'd want to handle ```<<``` and ```>>``` the same way, since they don't really get combined in TeX. [even though I can't conjure a legitimate ambiguity]

dginev:
@brucemiller , done, hopefully this makes sense in most practical cases. Doing too much with bare `>` and `<` looks quite ugly in the PDFs, so I am hoping authors rarely if ever go into bizarre ambiguous avenues that we haven't thought about in this specific case.

dginev:
Also, we used to have Travis builds enabled for pull requests, but I am not seeing the usual notification. I am wondering if something changed in settings, in our travis integration, or in github itself (I am using their beta issues UI in the moment, since it feels nicer)

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -4229,15 +4229,9 @@ DefMathLigature("!!" => "!!", role => 'POSTFIX', meaning => 'double-factorial');
 # Recognize :=
 DefMathLigature(":=" => ":=", role => 'RELOP', meaning => 'assign');
 
-# Recognize: <=, << and <<<
+# Recognize: <<< and >>>
 # # Note that in <<<, the leading << will already have been converted to \ll !!!
-DefMathLigature("<="        => "\x{2264}", role => 'RELOP', meaning => 'less-than-or-equals');
-DefMathLigature("<<"        => "\x{226A}", role => 'RELOP', meaning => 'much-less-than');
 DefMathLigature("\x{226A}<" => "\x{22D8}", role => 'RELOP', meaning => 'very-much-less-than');
```
brucemiller, line 4234: Ugh. is this even a thing? Maybe it should just be removed, since it most likely can't even fire now? is it worth handling in the grammar at all?

dginev: Indeed, this ligature will never match now. A quick grep through the arXiv data shows 50-100  of `very-much-less-than` matches, but they all seem connected to an explicit `\lll` macro, and one could argue that the `>>>` in math mode TeX looks really unpleasant in the first place... Maybe I just remove the ligature entirely?

t/math/ambiguous_relations.xml
```diff
@@ -0,0 +1,161 @@
...
+            <XMTok font="italic" role="UNKNOWN">a</XMTok>
+            <XMTok font="italic" role="UNKNOWN">b</XMTok>
+          </XMApp>
+        </XMath>
+      </Math>
+    </equation>
+  </para>
+  <para xml:id="p8">
+    <equation xml:id="S0.Ex8">
+      <Math mode="display" tex="0&lt;&lt;a,b&gt;&gt;1" text="formulae@(0 lessless a, b &gt;&gt; 1)" xml:id="S0.Ex8.m1">
```
dginev, line 135: I'm a bit confused why the text attribute for `<` is `less`, while the text attribute for `>` is `&gt;`. The resulting value is kind of broken (`lessless`).

brucemiller:
Odd indeed;  See line 898 of MathParser;  There's some kind
of paper-over-a-bug patch (from 2013!) that replaces "<" by "less",
but not ">" by greater.  Removing it causes all sorts of gratuitous
test fails, but no errors that I can see. (I'll test DLMF).
My suspicion (hope) is that this cruft can simply be removed.

dginev:
Right, I see no reason to keep that -- better face whatever error it's referring to head on, and resolve it. Should I update all test files to avoid the `less`? Is there a batch regeneration command? Seems like most everything will get changed in the test xml...

dginev:
Oh, btw, if you follow the git blame for the line you mention, you find the [original commit](https://github.com/brucemiller/LaTeXML/commit/a174cd036de3a4b4617cbb1077c536c12393e034) that introduced this rewrite dates back to 09.2004, almost 16 years ago. Impressive!

dginev:
On second thought, I suggest we merge this PR for the refactor, and I can start a brand new PR that only does the big regeneration. Or you could if you'd like to be sure nothing tricky is happening, either works for me. May get a bit confusing to add on top of the ligature/grammar touches.

brucemiller:
Sure, we can keep them separate, if you like.
Ready to merge?

dginev:
Yep!

brucemiller:
I can slog through the tests (& verify dlmf), ... if you haven't already started?

dginev:
Haven't started yet, so feel free!

brucemiller:
The spinoff patch & updated test cases is done, too.

Issue #649, opened 2015-08-15
mathsvg option not working in 0.8.1 ?


gpetty:
I just processed a 200+ page book  using the following commands:

latexml --destination=book.xml book.tex
latexmlpost --format=xhtml --mathsvg --destination=book.xhtml book.xml

Both operations completed, but during the latexmlpost phase, there were many, many warnings of the form
"Couldn't find image for 'LaTeXML::Post::MathImages::svg:/begin....'

When I inspect the resulting XHTML file, which is otherwise apparently fine, there is only MathML code, no SVG whatsoever.

Is this a bug, or have I not configured something correctly?  Is there an additional software prerequisite needed for the SVG generation?

I'm running this under Ubuntu 14.04 with LaTeXML 0.8.1

brucemiller:
On 08/15/2015 11:10 AM, gpetty wrote:
> I just processed a 200+ page book using the following commands:
>
> latexml --destination=book.xml book.tex
> latexmlpost --format=xhtml --mathsvg --destination=book.xhtml book.xml
>
> Both operations completed, but during the latexmlpost phase, there were
> many, many warnings of the form
> "Couldn't find image for 'LaTeXML::Post::MathImages::svg:/begin....'
>
> When I inspect the resulting XML file, which is otherwise apparently
> fine, there is only MathML code, no SVG whatsoever.
>
> Is this a bug, or have I not configured something correctly? Is there an
> additional software prerequisite needed for the SVG generation?

Yes, you need dvisvgm to convert the math to svg.
If it isn't present, you should have had an error message
before all the warnings about missing images.

gpetty:
Thanks, I just checked, and dvisvgm is present.  I re-ran latexml and latexmlpost again (same options as above) and captured the diagnostic output.  It shows some errors, but to my untrained eye, they don't seem related to dvisvgm:

https://dl.dropboxusercontent.com/u/1653737/latexml_log.txt  (latexml logged output)
https://dl.dropboxusercontent.com/u/1653737/latexmlpost_log.txt (latexmlpost logged output)

dginev:
```
Error:shell:latex LaTeX command 'latex ltxmlimg > ltxmlimg.ltxoutput' failed
	Postprocessing LaTeXML::Post::MathML::Presentation book.xhtml
	returned code 256 (!= 0): 
	Re-run with --verbose to see TeX log
```

gpetty:
Re-ran with --verbose; here's the revised output:
[...]

PR #1281, opened 2020-05-14
Mathsvg usability upgrades

dginev:
Addresses the usability friction raised in [the discussion](https://github.com/brucemiller/LaTeXML/issues/649#issuecomment-627373957) in #649 .

Namely:
 1. Raise "real" warnings when `--mathsvg` and `--mathimages` are used without `--destination` in both latexmlpost and the LaTeXML.pm API
 2. Change the default single option behavior of `--mathsvg` to *only* provide SVG formulas, rather than implicitly including a presentation MathML base processor
 3. Patch latexmlc to also provide the warning and disable `--mathsvg` without destination, as was always the case in latexmlpost. That features was never thoroughly tested in the latexmlc flow.

### DIFF
--- a/bin/latexmlpost
+++ b/bin/latexmlpost
@@ -18,7 +18,6 @@ use File::Spec;
 use Getopt::Long qw(:config no_ignore_case);
 use Pod::Usage;
 use LaTeXML;    # Currently, just for version information.
-use LaTeXML::Common::Error qw(generateMessage colorizeString);
 use LaTeXML::Post;
 use LaTeXML::Post::Writer;
 use LaTeXML::Post::Scan;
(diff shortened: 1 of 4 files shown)


bin/latexmlpost
```diff
@@ -289,12 +290,14 @@ else {
 if ((!defined $destination)
   && ($dographics || $picimages
     || checkMathFormat('images') || checkMathFormat('svg'))) {
-  print STDERR "WARNING: must supply destination to support auxilliary files;\n"
-    . "  disabling: --nomathimages --nographicimages --nopictureimages\n";
+  print STDERR generateMessage(colorizeString("Warning:expected:options", 'warning'), undef,
```
dginev, line 293: a bit iffy to have to unpack the body of `Warn()` here, but since `Warn()` depends on `$STATE` I decided that is an unnecessary complication to avoid. I basically needed a stateless Warn, could refactor it to a new subroutine e.g. `WarnExt` for "external" to the core conversion warning?

dginev: The reason I wanted a real warning was to get the color highlight, since it's very easy to miss in the regular flow of conversion text.

Issue #1278, opened 2020-05-12
support expl3 extensions in texlive 2019


dginev:
To start positive - here is a new chance to get closer to TeX-level parity!

We have a regression with the expl3 support, when updating to the `2020-02-14` version of that package. The update happened passively on my end, as I simply moved to the 20.04 LTS release of Ubuntu, which brought along the new fun.

A small snippet of the error, cryptic as it is:
```
t/83_expl3.t .............. Dubious, test returned 2 (wstat 512, 0x200) 
```

```
Argh, internal LaTeX3 error!

Module kernel, message name "command-already-defined":
Arguments '\__tl_rescan:NNw' and ' \__tl_rescan:NNw'

This is one for The LaTeX3 Project: bailing out: 

Error:unexpected:<closed> Mouth is unexpectedly already closed
	at String; line 2 col 0 - line 2 col 0
	Reading from Mouth[/usr/share/texlive/texmf-dist/tex/latex/l3kernel/expl3-code.tex@0x0], but it has already been closed.
	Input is empty
	In Core::Gullet[@0x561828dbfc58] Anonymous String; from line 3 col 0 to line 3 col 0
	 <= Core::Definition::Primitive[\@iinput {}] <= Core::Stomach[@0x561828dcf698] <= Core::Gullet[@0x561828dbfc58] <= ...

Error:unexpected:<closed> Mouth is unexpectedly already closed
	at String; line 4 col 0 - line 4 col 0
	Reading from Mouth[/usr/share/texlive/texmf-dist/tex/latex/l3kernel/expl3.sty@0x0], but it has already been closed.
	Input is empty
	In Core::Gullet[@0x561828dbfc58] Anonymous String; from line 5 col 0 to line 5 col 0
	 <= Core::Definition::Constructor[\usepac... <= Core::Stomach[@0x561828dcf698] <= ...

Error:unexpected:<closed> Mouth is unexpectedly already closed
	at String; line 5 col 0 - line 5 col 0
	Reading from Mouth::Binding[/home/deyan/git/my-LaTeXML/blib/lib/LaTeXML/Package/expl3.sty.ltxml], but it has already been closed.
	Input is empty
	In Core::Gullet[@0x561828dbfc58] Anonymous String; from line 6 col 0 to line 6 col 0
	 <= Core::Definition::Constructor[\usepac... <= Core::Stomach[@0x561828dcf698] <= ...
```

[...]

PR #1282, opened 2020-05-20
[WIP] Advanced fine-tuning of expansion for expl3 in texlive 2020

dginev:
Bookkeeping progress towards solving #1278 .

I have so far added the pdftex test behind their introduction of `\expanded` to `t/expansion/pdftex_expanded.tex` and - using one particular set of choices - gotten the first state of this PR to produce the same log output as pdftex does. Their particular tests are via `\show` and `\showtokens`, so I added first stabs for those macros as well. 

expl3 still fails to load with this refactoring, but - somewhat impressively - the `t/expansion/noexpand.tex` test we had from before continues to pass. Some early thoughts so far:
 * The torture test for `\showtokens\expandafter{\expanded{#,\unexpanded{#}}}` seems to suggest if we are to use `neutralizeTokens` for `\unexpanded` it should rather neutralize `#` instead of doubling it to `##` on the spot. Neutralized `#` could be handled as doubling for the cases that require that to be the case... at lest refactoring that seemed to work nicely.
 * More centrally, `\unexpanded` needs to mind the same flag as `\the`
 * And `\expanded` needs to be nestable, i.e. an `Expanded` parameter type in latexml needs to be able to unwrap/remove/purge a `dontexpand` flag from a previously flagged token.
 * I managed to clean up a bit further the prints from `\meaning` (minor, no behavior change) and Token's `stringify` (print the main token's string when dontexpand is set, rather than `\relax`).
 *  It _seems like_ (judging by the pdftex torture test suite) `\expanded` is behaving like `\edef` to a big extent, so it should expect an `EExpanded` argument, akin to a def -- which gets the torture tests to pass in the cross-use with `\unexpanded`, which neutralizes akin to `\the`.

I suspect some of these conclusions are too simple, and there is more machinery I am missing, will try to return to my expl3 tests tomorrow, after codifying the torture test as a proper xml test, rather than using the logs, to keep it passing.


lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -213,13 +213,14 @@ DefParameterType('XUntil', sub {
...
 # and maybe what we should be using for some Digested ??
+# and maybe we have to remove any "dont_expand" tokens here if we are to digest twice?
 DefParameterType('Expanded', sub {
     my ($gullet) = @_;
     my $token    = $gullet->readXToken(0);
     my @tokens   = ();
     if ($token->getCatcode == CC_BEGIN) {
       my $level = 1;
-      while ($token = $gullet->readXToken(0)) {
+      while ($token = $gullet->readXToken(0)->without_dont_expand) {
```
dginev, line 223: This is one of the head-scratchers. Clearly we must not remove any flags in `readXToken` since that token may be used in all kinds of scenarios -- very importantly in a `ExpandedIfToken` parameter type, which then does very precise comparisons on the flagged tokens for the conditional's sake.

However, uses of `\expanded{\expanded{...}}` do indeed expand *twice*, so the flag needs to be removed in whichever parameter type `\expanded` is using (or in some of the gullet calls of said parameter type). This refactor seems to be getting a little closer to the "spirit" of these two particular affairs.

dginev:
Updated. I now have an honest `\expanded` test for latexml in the PR, and wanted to jot down that I appreciated the ergonomics of a decent analogue for console printing (`\show` and `\showtokens`) for the content-oriented metaprogramming macros (`\meaning` and `\detokenize`).

dginev:
One TeXbook nugget that seems very related to some of the machinery I am examining:
image

dginev:
A bit of a gold digging expedition here, but maybe I finally hit it. The tests I've added to the PR seem to capture the few remaining torture cases that need to all pass to move expl3 forward, which is helpful. The master branch passes one of them, fails the other two, while my PR's state passed the other two and regressed on the one passing in master. I keep circling around the exact mismatch with the core TeX engine, and I think I've narrowed it down to not having "match tokens" in latexml yet. The two relevant passages are.

From the TeXbook, chapter 20, page 203:
image


And from the B book, page 178, §476:
image

The key difference being that only when we have a single "match token" scanned for the arguments of a macro call, can we properly propagate the effects of the `\noexpand`/`\expanded` interplay in the exact way that TeX has. The minimal intuition can be gleaned from a call such as:

```
\generate \defcmd \newcsname a#1b#2c { body }
```

Where we call a meta macro `\generate` with all necessary components to instantiate a new macro, particularly the "parameter tokens" such as `#1`. The current implementation can ace most direct cases using the neutralizing tactic of turning the two tokens of `#1` into the three tokens `##1`, but that approach misaligns with TeX proper when the definition happens several expansion levels under (scaffolds of \auxi, \auxii, \auxiii macros in expl3) *and* the addition of explicit expansion/lack of expansion directives.

If we had a single "parameter token" (aka "match token") from the very first argument scan for `\generate`, passing that token down with the various expansion directives should align nicely with TeX. According to the B book §289, these special tokens have their own independent catcode. So we would need at least a new `CC_MATCH` for this special kind of token. TeX also has a special catcode for a token designating the end of a match, but ideally we can postpone adding that until/if absolutely necessary.

dginev:
I seem to be making incremental progress, but it definitely feels like I am navigating a [Greek mythology labyrinth](https://www.greekmyths-greekmythology.com/myth-of-theseus-and-minotaur/) as I gradually tweak and re-tweak the expansion and parameter code snippets in latexml.

I checked in another small test and an incremental commit that once again can pass the `10_expansion` tests, now also with an attempt for a "match token" for parameters, which may or may not have been needed - but it at least makes it easier to reason about the expansion over `#1`-`#9`.

The 2020 texlive still fails the expl3 tests, I even wonder if travis will claim that I am now also failing older texlives, but it should be slightly easier to get a `\tracingall` description of what gets to misalign... May need another distilled test case next.

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -212,7 +211,7 @@ our @hold_token = (
   0, 0, 0, 0,
   0, 0, 0, 0,
   0, 0, 1, 0,
-  0, 0, 1);
+  0, 1, 0);
```
dginev, line 214: this was a cruel typo, and I am still second-guessing myself as to whether it is a typo. The internal `CC_MARKER` has 17 as its catcode, while a couple of these catcode arrays were expecting it to be 18? Bizarre. I think it may be time to rework the arrays as hashes using the `CC_` names, to avoid typos... Btw this type of fixed-element dispatch is something that I have found Rust's `enum`s and `match` statements to be fantastic for.

dginev:
I found a hope for a smaller refactor than the worst case scenario of reimplementing `scan_toks` from the B Book, although I haven't managed to get that working yet.

The key change would be to recognize `\the-like` primitives - the list of 4 names I mentioned above - in `readXToken` and store their output in a global variable analogous to TeX's `the_toks`. My draft version reworks [Gullet::readXToken lines 272-280](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/Core/Gullet.pm#L272-L280) as:
```perl
elsif (defined($defn = LaTeXML::Core::State::lookupExpandable($STATE, $token, $toplevel))) {
      local $LaTeXML::CURRENT_TOKEN = $token;
      my $invoked = $defn->invoke($self) || [];
      my @expansion = ();
      for my $exp_t(@$invoked) {
        my $r = ref $exp_t;
        if ($r eq 'LaTeXML::Core::Token') {
          push @expansion, $exp_t; }
        elsif ($r eq 'LaTeXML::Core::Tokens') {
          push @expansion, @$exp_t; }
        else {
          Fatal('misdefined', $r, undef, "Expected a Token, got " . Stringify($_)); } }
      next unless @expansion;
      if ($$LaTeXML::Token::T_THE{$$token[0]}) { # neutralize if the invocation was from a \the-like macro
        $LaTeXML::THE_TOKS = Tokens(@expansion);
        # and also return ... nothing? We need the caller to handle this case, as in scan_toks in the B Book
        return; }
      else { # add the newly expanded tokens back into the gullet stream, in the ordinary case.
        unshift(@{ $$self{pushback} }, @expansion); } }
```

Which seems to be a faithful emulation of TeX. When a `\the-like` primitive is invoked its output is moved to a global variable behind the scenes, and then the top-level scanner routine (`scan_toks`) moves these secret tokens onto the completed argument list. In latexml these are - unfortunately for the refactoring dev - many diverse cases of parameter types. Any parameter type calling `readBalanced(1)`, or has a loop using `readXToken` would have to be "upgraded" to explicitly stay aware of the global `$LaTeXML::THE_TOKS` variable having content.

I'm making some headway in changing these, but sadly things got worse before they got better, so some debugging ahead... E.g. I also end up having to rewrite `\expandafter` entirely, as I am also fully removing the other crutch global variable we had, `$LaTeXML::NOEXPAND_THE`. More later...

dginev:
I have been twisting and turning all kinds of boilerplate that approximate the `the_toks` magic, and I think splitting the branch up has helped. I am no longer working on this specific branch and am approaching a single example (the [pdftex test](https://github.com/brucemiller/LaTeXML/pull/1282/files#diff-861c653e28761bb9be8e848cf9b9855e) I included) passing with the refactored token handling.

I think something just "clicked" in my head today as I was carrying around boilerplate between places, in particular the Stomach::digestNextBody loop that calls `readXToken` -- it's really unseemly to expose these new magic tokens outside of Gullet, but it is wrong even if one manages, because of this exact loop. Since each step in the digest call wants to run `->invokeToken`, it is imperative that all tokens are *visible* to the Gullet at all times, or certain expansions may fail to find their arguments (or worse, use the wrong tokens for arguments).

What seems to be working a little better, and I will continue refactoring for it tomorrow, is instead of touching all `readXToken` calls, instead create a *second* Gullet pushback, e.g. `pushback_the`. That almost works but then it also gets questionable which pushback to use first. So I think the *real* answer is to add a flag that is *localized to the pushback* (and not to the token itself), which prevents further expansion.

This almost seems like a trivial difference with the existing code, but it's a major change of behavior. In master we can pass around a neutralized token a long distance from its original neutralizing (i.e. deep in nested macro calls), which is incorrect w.r.t to TeX. TeX neutralization impacts one exact call to `readXToken`, and never touches the token itself. I am very optimistic I finally figured the refactor in latexml, fingers crossed I have tests passing tomorrow...

PR #1283, opened 2020-05-28
finish do_def refactor without $expanded

dginev:
Paying off some technical debt, this tripped me up today as I am investigating the expansion details in `\edef`. The leftover `$expanded` flag can now be removed, having standardized towards using an `EExpanded` argument type for `\edef` and `\xdef`.


brucemiller:
seems wise :>

PR #1291, opened 2020-06-24
allow specifying javascript inclusion at the end of <body>

### LINKED ISSUE #1044: It should be possible to insert JavaScript at the end of the <body>
The most efficient way to load JavaScript in older browsers is to put the `<script>` tag directly before `</body>` at the end of the document. That way, downloading and parsing the JS doesn't block the content displaying.

The `--js` flag adds it to `<head>`, which is stopping us from using this option in Engrafo. Perhaps there should be an option or an `--xsltparameter` to put JS in the body rather than the head.

I would argue that this should be the default, because it is more efficient and lets you write JS that assumes the DOM is already loaded (instead of having to wait for onload events, or whatever). But, I don't want to be too controversial. :)

dginev:
Fixes #1044 

Default continues to be javascript placement in head, say:
```
latexmlc --javascript=LaTeXML-maybeMathjax.js --dest=test.html 'literal:$\sqrt{x}$'
```

But you can now add an XSLT parameter `LATEJS` that will tell post-processing to deposit the javascript at the end of the body:
```
latexmlc --xsltparameter=LATEJS:true --javascript=LaTeXML-maybeMathjax.js --dest=test.html 'literal:$\sqrt{x}$'
```

Exact parameter name is open to change, it was the first name that came to mind. Turns out implementing this was way simpler than I was guessing, I can never tell with XSLT

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-/=====================================================================\ 
+/=====================================================================\
 |  LaTeXML-webpage-xhtml.xsl                                          |
 |  General purpose webpage wrapper for LaTeXML documents in xhtml     |
 |=====================================================================|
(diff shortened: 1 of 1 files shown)


brucemiller:
Not (yet) the friendliest API, but it gets the job done, and we can deal with API later. Thanks!

PR #1294, opened 2020-06-26
associateNodeHook with both array and libxml ndoes

### LINKED ISSUE #1293: text subscript error
Hi. The following gives an error on the latest development version, but does not give an error on the release version or earlier versions. (But the release version gives other errors for me). It is called test.tex

`\documentclass{book}`
`\newcommand{\tx}[1]{\mbox{\emph{#1}}}  % text in math`
`\begin{document}
$\tx{position}_2$
\end{document}`

`latexml  --destination=artint.xml test.tex`
`latexmlpost --format=html5    artint`

Note that it does not give an error without html5

The error is 
(MathML::Presentation  processing...
Fatal:perl:die Perl died
	Not an ARRAY reference at /Users/poole/LaTeXML/blib/script/../lib/LaTeXML/Post/MathML/Presentation.pm line 73.
	In Post::MathML::Presentation[@0x7f8d4ae... ->associateNodeHook
1 fatal error

dginev:
Fixes #1293 

In short, a libxml node snuck its way into a subroutine that only expected the arrayref representation of post nodes. I am also hearing bells go off about having too many `getQName`s in the codebase, but that may be a refactor for another day. Fixes the example in the reported issue, and seems reasonable as a patch.

Ideally, especially if we had a language with fast method dispatch, one could imagine a common internal node class that can dispatch equally well to either arrayref nodes or libxml nodes, but that may be itself a problem for another programming language.

### DIFF
--- a/lib/LaTeXML/Post/MathML/Presentation.pm
+++ b/lib/LaTeXML/Post/MathML/Presentation.pm
@@ -70,11 +70,20 @@ sub rawIDSuffix {
 
 sub associateNodeHook {
   my ($self, $node, $sourcenode) = @_;
-  if ($$node[0] =~ /^m:(?:mi|mo|mn)$/) {
+  # TODO: Shouldn't we have a single getQName shared for the entire latexml codebase
+  #  in LaTeXML::Common or LaTeXML::Util ?
+  my $name = LaTeXML::Post::MathML::getQName($node);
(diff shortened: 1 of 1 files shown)


brucemiller:
Indeed. Thanks for catching & fixing that!

PR #1295, opened 2020-06-28
Avoid windows paths in post-processed graphics

### LINKED ISSUE #1187: Windows directory separator
In Windows, `\includegraphics{directory/file}` leads to the xml `<graphics candidates="directory\file.png" graphic="directory/file"/>`,
(possibly from lib/LaTeXML/Package/graphicx.sty.ltxml)
which leads to the html ``.
(possibly from lib/LaTeXML/resources/XSLT/LaTeXML-misc-xhtml.xsl via `f:url(@imagesrc)`).
%5C is the url encoding for \\, but the html needs / if it is going to find the image.  It turns out that a literal \\ also works (at least for local html files), but I'm skeptical that would make for a portable website when I copy everything over to a server.

With
```tex
\documentclass{article}
\usepackage{graphicx}
\begin{document}
\includegraphics{directory/by-nc}
\end{document}
```

`latexmlpost --verbose --verbose` does mention
```
(Graphics windowsPath.html processing... [Using graphicspaths: C:\Users\username\Documents\GitHub\project\lab\windowsPath, C:\Users\username\Documents\GitHub\project\lab\windowsPath]
[Processing C:\Users\username\Documents\GitHub\project\lab\windowsPath\directory/by-nc.png as key=LaTeXML::Post::Graphics:directoryby-nc.png.png] [Destination C:\Users\username\Documents\GitHub\project\lab\windowsPath\directory\by-nc.png] [Copied to directory\by-nc.png @ 403 x 141] [done with LaTeXML::Post::Graphics:directoryby-nc.png.png] 0.11 sec)
```

dginev:
Fixes #1187 in the simplest possible way.

Greetings from my Windows OS. This PR is overly simplistic, but workable in standard cases where people don't use backslashes in Unix filenames. I am opening mostly to state that if the path is sanitized just before it is set as an `imagesrc` attribute, the change is nicely propagated to the result of post-processing, and I happily created an HTML that rendered as expected on a Windows Firefox. So a successful patch attempt over here. 

Happy to test alternatives too. The other bit of good news is that the latest latexml master still has all tests passing on my Windows machine, though that is with the expl3 tests skipped, as we're still avoiding them on Windows.

### DIFF
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -196,7 +196,7 @@ sub pathname_absolute {
 sub pathname_to_url {
   my $relative_pathname = pathname_relative($_[0]);
   if ($SEP ne '/') {
-    $relative_pathname = join('/', split(/$SEP/, $relative_pathname)); }
+    $relative_pathname = join('/', split(/\Q$SEP\E/, $relative_pathname)); }
   return $relative_pathname; }
 
 #======================================================================
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -38,7 +38,7 @@ our @EXPORT = qw( &pathname_find &pathname_findall &pathname_kpsewhich
   &pathname_split &pathname_directory &pathname_name &pathname_type
   &pathname_timestamp
   &pathname_concat
-  &pathname_relative &pathname_absolute
+  &pathname_relative &pathname_absolute &pathname_to_url
   &pathname_is_absolute &pathname_is_contained
   &pathname_is_url &pathname_is_literaldata
   &pathname_protocol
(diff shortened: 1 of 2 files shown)


dginev:
Added a slightly more thoughtful guard, executing the regex only on a Windows OS. Tested, runs as intended.

brucemiller:
Egads, that's way too simple. You've got to complicate it some.  I was thinking it would be better to formalize the concept that we have a path and the attribute should be a url (relative). So, we probably should have a `pathname_to_url` in Util::Pathname. That makes clearer what the src attribute should be, and also localizes the separator/OS stuff to that file.

dginev:
Sure, that's a decent abstraction. I was just trying to pinpoint where and when the sanitization may be needed in post-processing. 

As to a general method, maybe something like:
```perl
sub pathname_to_url {
  my $relative_pathname = pathname_relative($_[0]);
  if ($SEP ne '/') {
   $relative_pathname = join('/', split(/$SEP/,  $relative_pathname)); }
  return $relative_pathname; }
```

dginev:
Alright, tested under Windows with the pathname call, works as intended.

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -545,6 +556,13 @@ The type C<*> matches any extension.
 Like C<pathname_find>,
 but returns I<all> matching (absolute) paths that exist.
 
+=item C<< $path = pathname_kpsewhich(@names); >>
+
+Attempt to find a candidate name via the external C<kpsewhich>
+capability of the system's TeX toolchain. If C<kpsewhich> is
+not available, or the file is not found, returns a
+Perl undefined value.
```
dginev, line 564: bonus documentation item: realized we hadn't included an entry for pathname_kpsewhich

brucemiller:
Seems good; Thanks!!!

PR #1296, opened 2020-06-29
Sanitize date strings when generating xml comments

dginev:
Fixing a rare post-processing fatal error from `arXiv:math0012024`. Simple enough, just being a little careful with double-dashes in creating comments in XSLT.

In this case the author had written `version 3 --- 5 mars 2001` as the date, hence the Fatal of 
```
LaTeXML-common.xsl line 82 element comment xsl:comment : '--' or ending '-' not allowed in comment
```

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
@@ -81,7 +81,7 @@
     <xsl:if test="//ltx:date[@role='creation' or @role='conversion'][1]">
       <xsl:comment>
         <xsl:text>Document created on </xsl:text>
-        <xsl:value-of select='translate(//ltx:date/node(),"--","—")'/>
+        <xsl:value-of select='translate(//ltx:date/node(),"-","‐")'/>
         <xsl:text>.</xsl:text>
       </xsl:comment>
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
```diff
@@ -81,7 +81,7 @@
     <xsl:if test="//ltx:date[@role='creation' or @role='conversion'][1]">
       <xsl:comment>
         <xsl:text>Document created on </xsl:text>
-        <xsl:value-of select='//ltx:date/node()'/>
+        <xsl:value-of select='translate(//ltx:date/node(),"--","—")'/>
         <xsl:text>.</xsl:text>
```
dginev, line 85: the trailing dot ensures we never end a comment with a trailing dash, which is neat, only had to address the double-dash case.

dginev:
Rereading the resulting HTML and XSLT, I really wanted `replace` instead of `translate`, but xslt 1 only has `translate`, which deals with single characters. Thus, I conceded to a simpler, slightly worse take, which maps all ascii `-` to a unicode hyphen `‐`, which isn't "semantically ideal", but alas avoids the Fatal with a simple reassignment.

brucemiller:
Weird, but seems safe. Thanks!

PR #1299, opened 2020-06-29
Add {planotable} alias for {deluxetable} in aaspp4

dginev:
Resolves the Errors in `astro-ph/0003429`, rooted in the [classic planotable construct](https://cds.cern.ch/record/304372/files/9605457.pdf) in AAS. Likely only helps with some handful of old AAS articles, but at least the resolution is simple - planotable seems to be largely identical to deluxetable, and aliasing them works wonders for the reference arXiv article - the three tables typeset well and error-free using the deluxetable internals.


brucemiller:
OK, let's give that a try. Thanks!

PR #1300, opened 2020-06-29
aipproc.sty and extensions to OmniBus

dginev:
Examining some arXiv errors, I am suggesting:
 - add `\volume` and some sibling metadata macros to OmniBus.cls, since they seemed needed in cases where a custom `.cls` file was ignored by latexml when processing arXiv articles
 - add [the arXiv artifact aipproc.sty](https://arxiv.org/list/physics/macros) with a specialized binding. 

The second point is a pain, because we also have a `aipproc.cls.ltxml`, from the arXMLiv Jacobs work, which seems to be developed for the later [aipproc.cls](http://dark.ft.uam.es/dsu2006/proc/AIPPROC/aipguide-6s.pdf), which is more evolved, e.g. expects key=vals for the `\author` macro rather than a simple strict.

It seems like a good default behaviour is to load the essentially revtex3-equivalent `aipproc.sty` when encountering `\documentstyle{aipproc}`, while only loading `aipproc.cls` with `\documentclass{aipproc}`. A case we once never expected to be possible :> But of course it is.

These fixes upgrade two arXiv articles from the latest cortex run, the aipproc one can be e.g. tested against `cond-mat0003014` from the [Error:undefined:\address report](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/undefined/%5Caddress?all=false). (indeed `\address` is another difference between the revtex3 base vs the custom aipproc.cls)

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -83,8 +83,10 @@ DefConstructor('\documentstyle OptionalSemiverbatim SkipSpaces Semiverbatim []',
 #
 # EXCEPTION: of course there is an exception. Older arXiv articles use \documentstyle{aipproc}
 # to load a different aipproc.sty than the later evolved aipproc.cls. And of course we need to support that.
-    if ($class eq 'aipproc') {
-      RequirePackage('aipproc', options => $options); }
+    if (FindFile($class, type => 'sty')) {
+      InputDefinitions('article', type => 'cls', noerror => 1, handleoptions => 1);
+      RequirePackage($class, options => $options, as_class => 1,
+        after => Tokens(T_CS('\compat@loadpackages'))); }
     elsif (FindFile($class, type => 'cls', notex => !LookupValue('INCLUDE_CLASSES'))) {
       LoadClass($class, options => $options,
         after => Tokens(T_CS('\compat@loadpackages'))); }
@@ -577,7 +579,7 @@ DefMath('\mathring{}', "\x{030A}", operator_role => 'OVERACCENT');
 # casual user redefinitions work, too.
 DefMacroI('\chapter', undef, '\@startsection{chapter}{0}{}{}{}{}', locked => 1);
 DefMacroI('\part', undef, '\@startsection{part}{-1}{}{}{}{}'); # not locked since sometimes redefined as partition?
-DefMacroI('\section', undef, '\@startsection{section}{1}{}{}{}{}', locked => 1);
+DefMacroI('\section',       undef, '\@startsection{section}{1}{}{}{}{}',       locked => 1);
 DefMacroI('\subsection',    undef, '\@startsection{subsection}{2}{}{}{}{}',    locked => 1);
 DefMacroI('\subsubsection', undef, '\@startsection{subsubsection}{3}{}{}{}{}', locked => 1);
 DefMacroI('\paragraph',     undef, '\@startsection{paragraph}{4}{}{}{}{}',     locked => 1);
@@ -629,7 +631,7 @@ DefConstructor('\@@numbered@section{} Undigested OptionalUndigested Undigested',
     MaybePeekLabel();
     $type = ToString($type);
     my %props     = RefStepCounter($type);
-    my $xtitle    = Digest(Invocation(T_CS('\lx@format@title@@'), $type, $title));
+    my $xtitle    = Digest(Invocation(T_CS('\lx@format@title@@'),    $type, $title));
     my $xtoctitle = Digest(Invocation(T_CS('\lx@format@toctitle@@'), $type, $toctitle || $title));
     if ($type eq 'appendix') {
       $props{backmatterelement} = LookupMapping('BACKMATTER_ELEMENT', 'ltx:' . $type); }
@@ -1019,7 +1021,7 @@ DefConstructor('\person@thanks{}', "^ <ltx:contact role='thanks'>#1</ltx:contact
   alias => '\thanks', mode => 'text');
 DefConstructor('\@personname{}', "<ltx:personname>#1</ltx:personname>",
   beforeDigest => sub { Let('\thanks', '\person@thanks'); },
-  bounded      => 1, mode => 'text');
+  bounded => 1, mode => 'text');
 
 DefConstructorI('\and', undef, " and ");
 
@@ -1041,7 +1043,7 @@ DefConstructor('\lx@author@prefix', sub {
     my $node       = $document->getElement;
     my $nauthors   = LookupValue('NUMBER_OF_AUTHORS');
     my $i          = scalar(@{ $document->findnodes('//ltx:creator[@role="author"]') });
-    if ($i <= 1) { }
+    if    ($i <= 1) { }
     elsif ($i == $nauthors) {
       $document->setAttribute($node, before => ToString(Digest(T_CS('\lx@author@conj')))); }
     else {
@@ -1162,7 +1164,7 @@ DefEnvironment('{titlepage}', '<ltx:titlepage>#body',
       "When using titlepage, Frontmatter will not be well-structured");
     return; },
   beforeDigestEnd => sub { Digest(T_CS('\maybe@end@title')); },
-  locked          => 1, mode => 'text');
+  locked => 1, mode => 'text');
 
 Tag('ltx:titlepage', autoClose => 1);
 DefConstructorI('\maybe@end@title', undef, sub {
@@ -1193,13 +1195,13 @@ DefMacroI('\center',    undef, '\centering');
 DefMacroI('\endcenter', undef, '');
 
 DefEnvironment('{flushleft}', sub {
-    $_[0]->maybeCloseElement('ltx:p');                        # this starts a new vertical block
+    $_[0]->maybeCloseElement('ltx:p');    # this starts a new vertical block
     aligningEnvironment('left', 'ltx_align_left', @_); },
   beforeDigest => sub {
     Let('\par', '\inner@par');
     Let('\\\\', '\inner@par'); });
 DefEnvironment('{flushright}', sub {
-    $_[0]->maybeCloseElement('ltx:p');                        # this starts a new vertical block
+    $_[0]->maybeCloseElement('ltx:p');    # this starts a new vertical block
     aligningEnvironment('right', 'ltx_align_right', @_); },
   beforeDigest => sub {
     Let('\par', '\inner@par');
@@ -1340,7 +1342,7 @@ sub RefStepItemCounter {
       my $typename  = (IsDefined(T_CS('\\' . $counter . 'name'))
         ? T_CS('\\' . $counter . 'name') : T_CS('\itemtyperefname'));
       my $tags = Digest(T_BEGIN,
-        T_CS('\let'), T_CS('\the' . $counter), T_CS('\@empty'),
+        T_CS('\let'), T_CS('\the' . $counter),   T_CS('\@empty'),
         T_CS('\def'), T_CS('\fnum@' . $counter), T_BEGIN, $formatter, T_BEGIN, Revert($tag), T_END, T_END,
         T_CS('\def'), T_CS('\typerefnum@' . $counter),
         T_BEGIN, $typename, T_SPACE, Revert($tag), T_END,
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -74,13 +74,18 @@ DefConstructor('\documentstyle OptionalSemiverbatim SkipSpaces Semiverbatim []',
...
+# But, we'd prefer .cls, since we'll have better bindings.
+# And in fact, nobody's likely to write a binding for a .sty that wants to be a class anyway.
+# So, we'll just try for a .cls, punting to OmniBus if needed.
+# If we start wanting to read style files by default, we'll still need to handle this
+# specially, since class (or sty files pretending to be) cover so much more.
+#
+# EXCEPTION: of course there is an exception. Older arXiv articles use \documentstyle{aipproc}
+# to load a different aipproc.sty than the later evolved aipproc.cls. And of course we need to support that.
+    if ($class eq 'aipproc') {
+      RequirePackage('aipproc', options => $options); }
```
dginev, line 87: There is certainly better general treatment to be invented here, but which?? I didn't want to get in the way of loading `revtex.cls` when encountering `\documentstyle{revtex}` by reversing the default... And since so far this is our only exception, I excepted it.

dginev:
Speaking of seemingly needed defaults in OmniBus, [the undefined report](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/undefined?all=false) also seems to suggest that the environments `{keywords}`, `{proof}`, `{theorem}` and `{lemma}` are often seen as undefined - a case I checked had `{proof}` defined in a custom `.cls` file for the specific article, which we're of course not loading at the moment.

I guess the challenge with these is defining them in a way that allows the same article to do a follow-up load of the ams* bindings and overload the definitions without any issue... 

Should I go ahead and cargo-cult the binding definitions for these environments into `OmniBus.cls`?

dginev:
I went ahead and did that, and threw in a bonus `\doi` macro. Got a rather ambitious math article looking respectable along the way.

lib/LaTeXML/Package/OmniBus.cls.ltxml
```diff
@@ -97,9 +98,37 @@ DefMacro('\keyword{}',  '\@add@frontmatter{ltx:keywords}{#1}');
...
+\\newtheorem{definition}[theorem]{Definition}
+$token")->unlist; }
+
+DefMacroI(T_CS('\begin{proof}'), undef, sub {
+    return bootstrapAmsThm('\\begin{proof}'); });
+DefMacroI(T_CS('\begin{theorem}'), undef, sub {
+    return bootstrapAmsThm('\\begin{theorem}'); });
+DefMacroI(T_CS('\begin{lemma}'), undef, sub {
+    return bootstrapAmsThm('\\begin{lemma}'); });
+DefMacroI(T_CS('\begin{proposition}'), undef, sub {
```
dginev, line 130: These four amsthm macros act as bootstrapping markers (the way `\begin` bootstraps LaTeX.pool within TeX.pool), and then get redefined upon amsthm loading. Seems like a good deal.

dginev: Actually, found an even better way of defining these. Good for review at this point!

dginev:
Looking forward to discussing/merging here before the next big arXiv rerun :+1:

lib/LaTeXML/Package/LaTeX.pool.ltxml
brucemiller: The general comment about loading .cls was probably more true originally. I suspect that the few cases where we have both a sty and cls (aastex,elsart,emulateapj,revtex,revtex4,subfiles) have gradually been (re)written into a consistent style of both cases loading a support file, and the .cls files also loading article.  So, perhaps the following structure would be better and avoid special casing:
```
if(FindFile($class, type=>'pkg'...){
   InputDefinitions('article', type=>'cls'...
   RequirePackage($class...
elsif(FindFile($class, type=>'cls'...
   LoadClass($class,...
else {
  InputDefinitions('OmniBus'...
```

lib/LaTeXML/Package/aipproc.sty.ltxml
```diff
@@ -0,0 +1,33 @@
...
+# Source: https://arxiv.org/ftp/physics/papers/macros/aipproc.sty
+# aipproc.sty, v1.0 <11 May 95>
+# tested against arXiv:cond-mat/0003014
+package LaTeXML::Package::Pool;
+use strict;
+use warnings;
+use LaTeXML::Package;
+
+PassOptions('revtex', 'cls', 'amssymb', 'amsfonts');
+LoadClass('revtex');
```
brucemiller, line 25: If we adopt the above approach to loading sty vs cls, then this probably shouldn't also LoadClass, but `revtex.cls` is rather simple anyway: basically you'd want `RequirePackage('revtex3_support', withoptions=>1)`, possibly natbib?

lib/LaTeXML/Package/LaTeX.pool.ltxml
dginev: Cool, we can try being more ambitious!
I wonder about the mandatory preload of article.cls though, shouldn't that be the responsibility of whichever "legacy" .sty file we are loading? e.g. in this aipproc.sty case we'd load revtex.cls instead of article, from within that legacy package binding.

In a way, we're either ready to do it in general, or we aren't. I thought we weren't so I made the exception, but if any of the .sty files you enumerated are to follow the same treatment, they should be able to imply a LoadClass call, right? (aastex,elsart,emulateapj,revtex,revtex4,subfiles).

dginev: Taking the case of `aastex`, based on passed options the `aastex.cls` is able to either load article.cls or revtex4.cls. Meanwhile the `aastex.sty` simple loads the `_support` binding without taking care of the class options. 

So changing the dispatch would break the revtex4 portion of aastex support that the current code is able to load, not ideal. Hm...

dginev: The way I see it the current rule is mostly correct, but it has exceptions. I can phrase it as: "whenever latexml has bindings for X.sty and X.cls, as a rule the X.cls is a related superset over X.sty, and can be used directly for \documentstyle".

The issue with aipproc.sty and aipproc.cls is that they are "too distant" and have breaking changes, so they're better thought of as separate entities that aren't as interchangeable as the other cases we have. Maybe we need a way of marking those cases up, but keep the default expectation that "X.cls is compatible with X.sty"?

brucemiller:
This is getting kinda confus(ed|ing). .sty files don't load a class, 
since they might be loaded by `\usepackage` (and none of them currently 
do).   But they do need a base class if loaded by `\documentstyle`. 
Generally, article should be safe, shouldn't it?

aastex.cls chooses which base class to load; why is that an issue?
Apparently aastex.sty shouldn't do that, so no problem.

dginev:
I am unsure where the confusion comes from? .sty and .cls files are just tex files with special conventions. `\documentstyle` uses an older convention, where the `.sty` file doubles as a class file. That is a distinctly different use of a .sty file as the modern one, where your claim that "sty files don't load a class" is accurate.

The `aipproc.sty` this PR is adding was pretty much exclusively meant to be used via `\documentstyle` and is an example of a convention breaker.

brucemiller:
exactly! So if either latex 2 or 2e have `\usepackage{foo}`, you 
probably do *not* want the effect of a load class, but if you have
a latex2 `\documentstyle{foo}`, you do. Easily accomplished with
my suggested convention.

I guess you're trying to convince me that there *never* will be a 
`\usepackage{aiproc}`, so you needn't bother with predictability.
But why not make it predictable?

dginev:
My problem was that I am unsure if users were once allowed to write:
 `\documentstyle[revtex4]{aastex}`

And have aastex load the class revtex4. In which case loading aastex.sty is breaking that feature. In fact, the current aastex.sty doesn't have any option handling added. Until now we've loaded `aastex.cls` in such a setting, so we were predictably getting the class option handling. I am worried we would at least need to vet the existing X.cls/X.sty pairs for options handling, and other potential discrepancies, if we switch the X.cls loading to X.sty loading for `\documentstyle`.

brucemiller:
You don't necessarily need to load the revtex4 *class*, but
(for LaTeXML) `revtex4_support`.  You'd want to copy most of the
`DeclareOption` stuff from cls to sty for whatever is appropriate.
However, it is odd that aastex.cls.ltxml uses `LoadClass('revtex4')`.
I'd think it should be:
```
DeclareOption('revtex4', sub { RequirePackage('revtex4_support'); 
return; });

ProcessOptions();
```

dginev:
Ok, so you're leaning towards vetting every pair of X.sty/X.cls and ensuring they all use a common `_support` interface, then default to always loading the `.sty`. And you'd rather do this before the 0.8.5 release I take it? If so, I'll take a stab refactoring all of them tomorrow in a separate PR.

brucemiller:
But they already are refactored that way, aren't they?
Modulo aiproc and the glitch in aastex.

dginev:
Hard for me to say how correctly they are structured without enough `\documentstyle` tests from the real world. But for example I would expect all `.sty.ltxml` files to be able to pass options into the `_support` code, and move any option handlers from the `.cls.ltxml` files into the `_support`. It's possible there are other bits that may need moving down, or the support files are missing, e.g. `emulateapj` doesn't have a shared support, the sty maps into revtex, while the class defines macros directly.

I'm a bit uneasy about jumping into this, since I don't think we have enough tests, and I may need to disappear into a rabbit hole fishing out templates from journals and running them with the two variants of `\documentclass` and `\documentstyle`...

brucemiller:
As it stands, only latex2 docs, for which we have a separate *.sty file 
are the only ones that would be affected by the proposal. And those will 
mostly be about options. Hard to say if it will work better or worse; 
arXiv *is* our \documentstyle testbed.

In fact, we don't really know that the latex2 style's options would
be the same as the latex2e class's options. Moreover, we don't know
that the a style file used as a latex2 or latex2e package will be
consistent with using the style as a latex2 class.

For example, you talk about a revtex4 option to aastex when used in a 
latex2 `\documentstyle`. That pretty much can't happen as revtex4 came 
*way* later than latex2e. In fact, I kinda doubt there was any revtex 
before latex2e.  I could be wrong, but I think all the ams packages that
we have were designed for latex2e; before that, you would have used 
AMSTex instead of latex.

It sounds like you've found test cases for latex2 & 2e aiproc?
And probably the sty & cls files there, as well, so should be easy
to see what the options are?

dginev:
> Hard to say if it will work better or worse; arXiv *is* our \documentstyle testbed.

This captures the nature of my worry. The PR was adding targeted fixes to examples I examined. The general treatment is a bit of a gamble, since we don't know if it will impact the success rate (and if it does, to what extent). So I'd ideally have to check that before I introduce the general handling, especially since the X.cls.ltxml/X.sty.ltxml pairs we have aren't as well-natured as you'd like me to believe :>

brucemiller:
Ah, yes, but...

dginev:
Oh no, I had one more commit incoming! Will file against master

brucemiller:
On 11/2/20 5:35 PM, Deyan Ginev wrote:
> Oh no, I had one more commit incoming! Will file as new PR against master

Sorry. It looked complete to me. (#1364 is separate, right?  You could 
tack it on there, if you wish)

dginev:
Yep, #1364 is separate. I just added the tiny missing patch from my testing, it is at #1366. 

I was also surprised to discover that `\documentclass[xcolor]{article}` does _not_ load xcolor, while the documentstyle version _does_. So my patch should be complete.

dginev:
Also, before I forget, let me flag that the options-related interface to `InputDefinitions` appears to be begging for a refactor. It is both confusing and easy to misuse (and I was both confused, and initially misused it).

In particular, if you want to get an array of options loaded by `InputDefinitions`, you seem to ideally need to say: 
```perl
withoptions => 1, handleoptions => 1, options => $options
```

which speaks for itself. On a separate note, I wasn't sure what the relation was between the `\documentstyle` of `AmSTeX.pool` and the "official" one of `LaTeX.pool`, but wanted to state for archival that I did not change anything about the definition in `AmSTeX.pool`, which may now be starting to diverge from the "official" behavior.

PR #1301, opened 2020-06-30
[Babel] avoid loading english.sty and default nil.ldf errors-in-name-only

dginev:
1. `\usepackage{english}` creates what is really a deprecation warning ("you should be using babel!") but does it via `\GenericError` so pollutes the cortex reports for arXiv with untrue error cases. I am adding a binding that does what the error suggests -- reroutes to a `\usepackage[english]{babel}`. 
2. There are a number of suspect related mini error reports in loading babel's English on monster, but I'm fixing one that I noticed on my 2019 texlive machine with the default loader (`nil.ldf`) for option-free babel. It turns out the recent `pdflatex` _also_ has babel primitives precompiled in. You can try running pdflatex on a single line of `\show\bbl@languages`, and you'll see it defined. So I read the contents of nil.ldf, saw it is expecting that macro defined, and am adding a specialized binding that just defines it if needed and passes back into the texlive nil.ldf file...

Two arXiv examples to provide here, with a lot of cortex Errors to correspond to each class: `math0006231` for the english.sty case, and `math0209382` for the nil.ldf case. Both can be looked up at the [error:latex:GenericError](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/latex/%5CGenericError?all=false) report page.

I'm considering adding a couple more fixes to this branch that relate to babel, since a couple of the other errors look suspiciously harmless/procedural...

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -425,6 +425,7 @@ lib/LaTeXML/Package/doublespace.sty.ltxml
 lib/LaTeXML/Package/dsfont.sty.ltxml
 lib/LaTeXML/Package/empheq.sty.ltxml
 lib/LaTeXML/Package/endnotes.sty.ltxml
+lib/LaTeXML/Package/english.sty.ltxml
 lib/LaTeXML/Package/eTeX.pool.ltxml
 lib/LaTeXML/Package/ellipsis.sty.ltxml
 lib/LaTeXML/Package/elsart.cls.ltxml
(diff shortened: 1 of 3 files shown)


dginev:
Looking at a number of the `\GenericError` reports related to babel, my main conclusion is starting to be that they would be resolved if converting against a newer version of texlive. Which would be a great conclusion, if not for the expl3 regressions one would also encounter on the newer versions of texlive... Returning to that front.

brucemiller:
Seems to solve (or punt :> ) at least one problem. Thanks!

PR #1302, opened 2020-06-30
Updated enumitem.sty to cover more options and \newlist; added testcase

### LINKED ISSUE #1015: Finish enumitem binding; options and \newlist
A preliminary binding for enumitem was made in response to #580, but it doesn't process the options, and omits `\newlist` (which doesn't make much sense w/o the options).  This issue is a reminder to finish the binding.  Much of what the options do has already been handled in similar cases in LaTeXML, so partly it's just a matter of finding appropriate code snippets to steal.

brucemiller:
Updated enumitem and added testcase.  PR to see if it's close enough with no gotchas.

Fixes #1015


dginev:
Reads good when looking through the code! I found it nice that you have explicitly pointed out which bits remain to be implemented later.

In case you want "examples from the wilderness", I can point to a list of [arXiv articles loading enumitem.sty](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/info/loaded%5Ffile/enumitem%2Esty%2Eltxml?all=false). Will try one or two against this branch to see how things fare.

brucemiller:
OK, let's pretend this works!

0-andy-0:
Just checking: [shortlabels] is not implemented, so the idiom \begin{enumerate}[A)] will fail silently?
Of course the enumerate package will cover it but then I can't [resume].

PR #1307, opened 2020-07-09
[WiP; Experiment] Try a simple semantic annotation

brucemiller:
Try something simpler that leverages the order that things get annotated. Unfortunately it also leaves lots of garbage behind; ties the "meaning" to the non-optimal node, and other weirdness. But if you only look at the parts it gets right, it gets them right :>

### DIFF
--- a/lib/LaTeXML/Post/MathML/Presentation.pm
+++ b/lib/LaTeXML/Post/MathML/Presentation.pm
@@ -117,7 +117,8 @@ sub addAccessibilityAnnotations {
   if(! p_getAttribute($node,'data-arg')){
     my $arg = ++$$self{arg_ctr};
     p_setAttribute($node,'data-arg'=>$arg);
-    $sourcenode->setAttribute('_reffedarg' => $arg); }
+    $sourcenode->setAttribute('_reffedarg' => $arg)
+      unless  $sourcenode->hasAttribute('_reffedarg'); }
   # print STDERR "ANNOTATE "
   #   ."\n".LaTeXML::Post::MathProcessor::shownode($node)
   #   ."\n from ".LaTeXML::Post::MathProcessor::shownode($sourcenode)
--- a/lib/LaTeXML/Post/MathML/Presentation.pm
+++ b/lib/LaTeXML/Post/MathML/Presentation.pm
@@ -14,6 +14,7 @@ package LaTeXML::Post::MathML::Presentation;
 use strict;
 use warnings;
 use base qw(LaTeXML::Post::MathML);
+use LaTeXML::Common::XML;
 
 sub preprocess {
   my ($self, $doc, @maths) = @_;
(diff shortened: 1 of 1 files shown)


dginev:
How distracting, a parallel PR :> And it's cheating, you're setting new custom attributes! I'll steal the `p_` attributes though, definitely makes things cleaner.

brucemiller:
Yeah, well, I wanted to experiment, what can I say. Looking at the order that nodes get visited, it seemed it might be better to carry forward rather than try to look back and tend to overlook things.

brucemiller:
Besides, it's not so much a PR as a place to dump code for you to look at.

lib/LaTeXML/Post/MathML/Presentation.pm
```diff
@@ -84,8 +85,63 @@ sub associateNodeHook {
...
+  $$self{arg_ctr} = 0 unless defined $$self{arg_ctr};
+  my $qname = LaTeXML::Post::MathML::getQName($sourcenode);
+  my $meaning = $sourcenode->getAttribute('meaning');
+  if($meaning){}
+  elsif($qname eq 'ltx:XMApp'){
+#    my($op,@args)=$sourcenode->childNodes;
+    my($op,@args)=map { realize($_) || $_ } element_nodes($sourcenode);
+    my ($opid,@argids) = map { $_->getAttribute('_reffedarg'); } $op,@args;
+    if(! grep { ! $_ } $opid, @argids){
+      $meaning = "#$opid(".join(',', map { "#$_"; } @argids).")"; } }
```
dginev, line 139: Btw, are we sticking with the `#` as the sign for argument refs? I saw Neil use that too recently, even though the original introduction of `arg` used `@` instead. I'm worried that they may conflict with HTML `#id` references, but on the other hand the `#arg` are exactly this type of relationship, but looking at a different place...

dginev:
To prove I've tested and thought about the approach, here is one of my surprisingly slippery examples I've been testing with on the other branch, `2(a+b)` which hides a dual. I'm still contemplating what missing abstraction is needed. Notice that the inner dual for "a+b" didn't get annotated, "mysteriously". I think there is a general judo move that can be applied to duals, but it needs to be performed at just the right time, from just the right angle...

Bruce's branch generates at the moment:

```xml
  <m:math alttext="2(a+b)" data-arg="11" display="inline">
    <m:mrow data-arg="10" data-semantic="#2(#1,#3)">
      <m:mn data-arg="1" data-semantic="2">2</m:mn>
      <m:mo data-arg="2" data-semantic="times">⁢</m:mo>
      <m:mrow data-arg="9">
        <m:mo data-arg="3" stretchy="false">(</m:mo>
        <m:mrow data-arg="7">
          <m:mi data-arg="4">a</m:mi>
          <m:mo data-arg="5" data-semantic="plus">+</m:mo>
          <m:mi data-arg="6">b</m:mi>
        </m:mrow>
        <m:mo data-arg="8" stretchy="false">)</m:mo>
      </m:mrow>
    </m:mrow>
  </m:math>
````

PR #1308, opened 2020-07-12
New binding physics.sty, plus test case

brucemiller:
As it says...


lib/LaTeXML/Package/physics.sty.ltxml
```diff
@@ -0,0 +1,630 @@
...
+sub phys_applyPower {
+  my ($cfunc, $pfunc, $power) = @_;
+  my ($cargs, $pargs) = dualize_arglist('#1', $power);
+  if ($cfunc) {
+    $cfunc = Invocation(T_CS('\@APPLY'),
+      Tokens(Invocation(T_CS('\@CSYMBOL'), T_OTHER('power')),
+        $cfunc, $$cargs[0])); }
+  if ($pfunc) {
+    $pfunc = Tokens($pfunc, Invocation(T_CS('\@@POSTSUPERSCRIPT'), $$pargs[0])); }
+  return ($cfunc, $pfunc); }
```
dginev, line 112: I see how this ties nicely into the XMDual experimentation I am doing at the a11y-experiment branch. Won't repeat any feedback from there, interesting read to see how you're using the dual-related macros in action!

dginev:
High level organization comment - since my ctrl+f is telling me there are 73 instances of `phys_` in the PR code, have you considered making a `package LaTeXML::Package::Pool::Physics;` either in this file or as a parallel `.pm` and then importing all `phys_` auxiliary functions from there?

Seeing how some of those at least appear to be valuable for generally working with duals on the macro level, they may see some valuable reuse later on...

brucemiller:
Indeed; as I developed this binding, I became painfully aware
that I was hunting for or creating both some generally useful
utilities, and some very package-specific ones, along with a
general lack of naming discipline.

Ideally, all these reusable things should be exported from
Package (or from some deeper like Token), rather than scattered
through TeX.pool, LaTeX.pool.  But that's a refactor for
another day (or PR)

kohlhase:
This is a really nice package for semantic math (physics). I have a student, Alpcan Dalga, who is currently looking at using sequence-to-sequence ML to generate content MathML from tokens. He is in need of more "gold standard"  material (siuntitsx.sty being the other one). Do we have numbers on how these two packages are use on arXiv?

brucemiller:
physics is "used" in 0.04% (592) docs in arXiv, but many only
\usepackage, not necessarily actually using the macros.

siunitx is used (in the same sense) in 1.25% of arxiv;
but although I wrote a binding, by the time I had sorted out the
somewhat complex language of it, I put off handling the implied
semantics.
Perhaps since semantics is currently A Thing (again),
it might be worth jumping back in.

kohlhase:
Ah, thanks for the clarification. Then having a more semantic `siunitx` binding would be very welcome indeed.

Issue #432, opened 2013-07-23
Two TeX annotations for Math formulas


dginev:
[Originally Ticket 1756]

We were talking with Howard about the two use cases of copying math sources from formulas:
1. Copy the lowest level of expanded TeX to have 100% guarantee that it will compile in your document (= presentation-oriented copy)
2. Copy the highest level of semantic TeX to have 100% guarantee that it will carry over the semantics to your document (= content-oriented copy)

We refactored the following macro:

```
DefMath('\EulerConstant',"\x{03B3}", role=>'ID',
    meaning=>'Euler-constant'); # GREEK SMALL LETTER GAMMA
```

into:

```
DefMath('\EulerConstant','\gamma', role=>'ID',
    meaning=>'Euler-constant'); # GREEK SMALL LETTER GAMMA
```

Which created the same presentation, but had now changed the annotation from:

```
<annotation encoding="application/x-tex">\left|\EulerConstant\right|</annotation>
```

to:

```
<annotation encoding="application/x-tex">\left|\gamma\right|</annotation>
```

So, it hit us that if we have DefMath statements which have lower-level TeX macros to expand into and at the same time the role and meaning attributes, the conversion could record two-forms of "original TeX", one form stopping at the highest level that has "meaning" (= preserve semantics), the other stopping at the lowest expandable level (= ensure compilable presentation).

brucemiller:
As said, the proper way to store _either_ `@tex` or `@content-tex` in the MathML is with an annotation; If you want _both_, then we'll need some way (a special mime type?) to distinguish them, _and_ likely need another option to choose (currently `--mathtex` adds the `@tex` annotation)

dginev:
Yes, similarly to my "math lexemes" PR. It's kind of strange that math is just a single modality, yet there are all these parallel representations that end up being handy, and similarly strange that each would necessitate its own option.

I still think it would be a useful enhancement, and it's schedule for a very distant milestone as it should be... May be good to keep in mind.

dginev:
I've lost focus on the need for multiple tex attributes, at least for the moment, I think a lot of betterment can be brought along by having great locators instead. And I am itching to close old issues, so this one may be a good first sweep.

PR #1309, opened 2020-07-13
Towards more optimal XMath (compacting XMDual)

dginev:
Hi @brucemiller !  This is a standalone PR that can be merged and discussed independently of anything else.

Following up on #1305 , I discovered a nice nook where I can have my cake and eat it too - namely the `pruneXMDuals` routine in Document.pm. It turns out this is the right timing (after math parsing, so all script attachment magic is finished and settled in, and also all XMArg and XMWrap logic is collapsed as appropriate).

Having found the hook, so far I have only a single case that I need compacted, and I've added it. It arises from scripts, but also seems to commonly be "overdone" when using the dual form of `DefMath`. Namely:
```xml
<XMDual>
<XMApp>
  <XMTok meaning="real-meaning"/>
  <XMRef idref='arg1'/>
  <XMRef idref='arg2'/>
  ...
</XMApp>
<XMApp> <!-- no wrap, because we're looking at it after math parsing, and it parsed into an apply -->
  <XMTok role="some-role">maybe some pres content</XMTok>
  <XM(*) xml:id='arg1'>...</XM(*)>
  <XM(*) xml:id='arg2'>...</XM(*)>
  ...
</XMApp>
</XMDual>
```

This form, in my eyes, is asking to be compacted where the meaning and role of the leading operator token are merged on the same element, and then the realized argument nodes are copied along. Resulting in a single apply node, with no dual. As mentioned in the other thread, it feels more "optimal" for XMath's philosophy, where we try to keep the markup as fine-grained as possible, only creating duals where the presentation and content trees do *not* align. In the "semantically annotated scripts" cases, the two trees *do* align, and the "meaning" of the script can reunite with the script's presentational XMTok.

I had to change a single test case to get this PR running, which I don't fully understand and would appreciate a review + guidance on. Of course, I also added my own new test for using `\power`, which uses duals explicitly directly via the TeX macros (`\DUAL` and friends), but _pleasingly_ does not produce any XMDual elements, as they can be nicely compacted down to an apply.


t/math/compact_dual.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?latexml class="article"?>
+<?latexml package="latexml"?>
+<?latexml RelaxNGSchema="LaTeXML"?>
+<document xmlns="http://dlmf.nist.gov/LaTeXML">
+  <resource src="LaTeXML.css" type="text/css"/>
+  <resource src="ltx-article.css" type="text/css"/>
+  <para xml:id="p1">
+    <p><Math mode="inline" tex="\@CSYMBOL{power}x2+\@CSYMBOL{power}y3" text="power@(x, 2) + power@(y, 3)" xml:id="p1.m1">
```
dginev, line 9: Btw, an unexpected aside, it looks like `\@APPLY` and `\DUAL` don't have a reversion back into the `tex` attribute.

dginev:
I also added an example that does:
```tex
\def\degtwo{\dual@atom{2}{\prime\prime}}
\def\derive2#1{
  \dual@infix{derivative-implicit-variable}{^}{#1}{\degtwo}}
```

Generating the compact XMath:
```xml
<XMApp>
  <XMTok meaning="derivative-implicit-variable" role="SUPERSCRIPTOP" scriptpos="post1"/>
  <XMTok font="italic" role="UNKNOWN" xml:id="p2.m1.1">f</XMTok>
  <XMDual xml:id="p2.m1.2">
    <XMTok meaning="2"/>
    <XMWrap>
      <XMTok fontsize="70%" name="prime" role="SUPOP">′</XMTok>
      <XMTok fontsize="70%" name="prime" role="SUPOP">′</XMTok>
    </XMWrap>
  </XMDual>
</XMApp>
```

Note how the code leaves the inner dual as-is, as the trees diverge - single token meaning for two tokens of presentation.

t/math/compact_dual.xml
dginev: Learning more as I go along, the `tex` attribute is dedicated to the presentation TeX, which I've know obeyed by passing the `hide_content_reversion` option to `\DUAL`. Added a commit.

Feels a little crutchy, should be automatable in the long run... Similarly the `content_tex` attribute needs what a TeX.pool comment referred to as:
```
# NOTE: work through this systematically!
```

and is currently not ideal in my latest XML file. It also reminded me of the old discussion around tex having two annotations for the source TeX ( #432 ), just linking it in case someone's picking up breadcrumbs later on. Not something I want to jump into right now, but I ended up noticing... The improved XMDual pruning is the only focus here, sorry for the diversion.

lib/LaTeXML/Core/Document.pm
```diff
@@ -1439,7 +1439,63 @@ sub pruneXMDuals {
...
+  my @new_args = ();
+  my $single_duality;
+  # walk the corresponding children, and double-check they are referenced in the same order
+  while ((my $c_arg = shift(@content_args)) and (my $p_arg = shift(@pres_args))) {
+    my $c_idref = $c_arg->getAttribute('idref');
+    if ($c_idref && ($c_idref eq ($p_arg->getAttribute('xml:id') || ''))) {
+      push @new_args, $p_arg;
+      next; }    # content-refs-pres, OK
+    my $p_idref = $p_arg->getAttribute('idref');
+    if ($p_idref && ($p_idref eq ($c_arg->getAttribute('xml:id') || ''))) {
```
brucemiller, line 1464: ah, if-elsif implied...

lib/LaTeXML/Core/Document.pm
```diff
@@ -1439,7 +1439,63 @@ sub pruneXMDuals {
...
+      push @new_args, $c_arg;
+      next; }    # pres-refs-content, OK
+
+    # Next up: differences. If:
+    # 1) we saw such a difference beforehand, or
+    # 2) the tree is too complex - give up on compacting and return.
+    # we only handle two XMToks differing for now.
+    if ($single_duality || ($self->getNodeQName($c_arg) ne 'ltx:XMTok') || $self->getNodeQName($p_arg) ne 'ltx:XMTok') {
```
brucemiller, line 1472: why does parg need to be a token? (so long as it's consistently id/ref'd)

lib/LaTeXML/Core/Document.pm
```diff
@@ -1439,7 +1439,63 @@ sub pruneXMDuals {
...
+      # Transfer all c_arg attributes over, it should be primary?
+      for my $attr_key (qw(decl_id meaning name)) {
+        if (my $attr_val = $c_arg->getAttribute($attr_key)) {
+          $c_arg->removeAttribute($attr_key);
+          $p_arg->setAttribute($attr_key, $attr_val); } }
+      $n_arg = $p_arg; }
+    $n_arg->unbindNode;
+    $compact_apply->appendChild($n_arg); }
+  # if the dual has a role/id migrate them to the XMApp
+  for my $attr_key (qw(role xml:id)) {
```
brucemiller, line 1494: This lost an `lpadding` value in my case.  You need to copy more attributes from the XMDual to the new XMApp, not to mention from the old XMApp; probably all attributes (unless already set on the old XMApp).  And you'll want to use Document's setAttribute, since it manages ids.

It smells more&more like we need some sort of `$doc->mergeAttributes($oldnode,$newnode)` that knows which attributes to ignore, which for bookkeeping (id), which to append (class), and who knows else what crops up.

lib/LaTeXML/Core/Document.pm
dginev: It was just a starting point. And I am not sure if passing the meaning attribute to other types of nodes wouldn't confuse the post-processor, so I'd like some examples... E.g. XMArray, XMRow, XMCell, or even another XMDual nested underneath could cause some hiccups. But maybe they're all workable already, haven't tried.

lib/LaTeXML/Core/Document.pm
dginev: Hm, alright. I could also rename the XMDual to an XMApp, that way no attribute moving needs to happen, I'll just replace its children.

lib/LaTeXML/Core/Document.pm
brucemiller: You don't need examples; I need counter-examples!

I think you're looking at it a bit backwards; The attributes *inform* post-processor, if that makes it confused, that's the post-processor's problem.  Or alternatively, assume the attributes are there for a reason, possibly bad, but they don't need to justify themselves.

lib/LaTeXML/Core/Document.pm
brucemiller: No, you'd still want to copy attributes from the old (presentation) XMApp.  (besides renaming nodes w/libxml2 always worries me; there've been problems in the past (and yeah, I know it's used rarely))

With the current approach all the XMRef's are in the content, so you really could just replace the XMDual with the presentation's XMApp (ie. an unwrap), still copying attributes from the XMDual down,  but I can see why you might not want to assume the current arrangement is permanent.

dginev: Indeed. Since the schema is designed to allow XMRefs in either branch, might as well implement with the allowance in mind... I've added a commit that copies all XMDual attributes onto the new "compact" XMApp, unconditionally.

dginev: Btw, I am not sure I can use the Document's `setAttribute`, as it reports an id conflict, since the id was already recorded on the now compacted XMDual. And we don't seem to have a Document `removeAttribute`, which removes the previous bookkeeping. Would this cause problems?

dginev: Nevermind, figured it out, I first had to replace the node to un-document the xml:id, and then I could safely re-record it.

brucemiller: No, but there's a `unRecordID`; it's a bit of dancing around, though.

Hmm, `setAttribute` doesn't use that when it *replaces* the id, though.

lib/LaTeXML/Core/Document.pm
dginev: Alright, I'll transfer to arbitrary nodes then, scary as that seems... Just landed a commit that removes this guard, though I still haven't found a good example of what it does. Should probably rebase the a11y branch on this one and see how things change in the showcase...

dginev:
I made a new branch that combines this dual-compacting PR with the WIP acessibility PR, called `dginev/a11y-deploy` and am now using it for the [showcase](https://dginev.github.io/tiny-mathml-a11y-demo/). All seems in order so far, still examining it.

t/math/declare.xml
```diff
@@ -107,16 +107,10 @@
...
-                  </XMApp>
-                  <XMApp>
-                    <XMTok name="widehat" role="OVERACCENT">^</XMTok>
-                    <XMTok decl_id="S1.XMD4" font="italic" role="ID" xml:id="S1.Ex1.m1.3">x</XMTok>
-                  </XMApp>
-                </XMDual>
+                <XMApp role="ID" xml:id="S1.Ex1.m1.10">
+                  <XMTok decl_id="S1.XMD3" name="widehat" role="OVERACCENT">^</XMTok>
+                  <XMTok decl_id="S1.XMD4" font="italic" role="ID" xml:id="S1.Ex1.m1.3">x</XMTok>
+                </XMApp>
```
dginev, line 113: @brucemiller I rebased to the latest master this morning, and think we should consider merging this dual-cleanup PR. It has been included in the a11y showcase for a week now, all seemed in order, and the declare.xml diff is extra encouraging.

Or let me know what else you'd like me to try and improve, I've moved on from working on this one.

dginev:
Ah, I see travis failed in the PR with something I wanted to ask, namely the reversion of `\@CSYMBOL` in the tex attribute. The failure is:
```
# Difference at line 358 for t/complex/physics

#       got : '            <td align="left"><verbatim font="typewriter">\abs{a}</verbatim> <Math mode="inline" tex="\rightarrow a" text="absent rightarrow absolutevalue@(a)" xml:id="S1.T1.m19">'

#  expected : '            <td align="left"><verbatim font="typewriter">\abs{a}</verbatim> <Math mode="inline" tex="\rightarrow\@CSYMBOL{absolutevalue}a" text="absent rightarrow absolutevalue@(a)" xml:id="S1.T1.m19">'
```

Question being, isn't it cleaner to have the `\@csymbol` reversion make it "disappear" from the tex attribute. Trying not to reopen the big can of worms about semantics vs non-semantic tex attributes for math,  but this is a curious case where the token deposits no visible ink, rather it just feeds the content branch of an XMDual.

brucemiller:
I think making `\@csymbol` disappear is the wrong approach; I just started experimenting with letting `\DUAL` have a reversion keyval option.  That seems like the right level to handle the issue, although physics.sty is kinda scary how to create that reversion.

BTW: you've still got a couple of failing test cases (compact_dual and (cough) physics)

brucemiller:
Interestingly, the physics errors pointed out the New Improved way that Laplacian is handled; namely reducing the `\nabla^2` to an `XMApp` tree, but with `meaning=laplacian`.  Arguably that's correct and good, but the cmml & om postprocessorswill need to be updated to recognize such constructs.

dginev:
Nicely spotted Laplacian example! I've updated all test files, so travis should pass now. But indeed, I should also double-check the post-processor can handle it, here is a link to the code diff you mention:

https://github.com/brucemiller/LaTeXML/pull/1309/commits/e5b2a9eaa0a2acb4c0b94a889a1e9a54686dfd8d#diff-c95bec3fd2f012f6a8dc521538db2fceR1522-R1529

dginev:
Ah, I remembered some secret experimental knowledge from times past. There is a bit of hidden code in MathML.pm which checks if an incoming XMApp has a role `ID`, in which case it treats it as a decorated symbol via `cmml_decoratedSymbol`.

In other words, I can have a very clean rule where "whenever we compact a dual to transfer a meaning attribute onto an XMApp with no role, give it role ID, and treat it as a decorated symbol/embellished operator".

Checked in, and even gets cross-references done for the laplacian, when doing `--pmml --cmml`.

brucemiller:
Well similar, but not that. It doesn't necessarily have role=ID. But if it has meaning, it should be treated as a csymbol (or potentially a built-in for pragmatic).

dginev:
Well, the agreement in MathML.pm experiment is that all XMApp elements with role=ID can be treated as a csymbol. Should we revise and modify that? Maybe introduce a new role?

brucemiller:
From the content pov, it's not embellished --- it simply *is* the laplacian; and it's not an ID, it's an operator in this case. 
You'll  want to add this *before* the embellished operator clause in `cmml_internal`
```
    if(my $meaning = $node->getAttribute('meaning')){
      return &{ lookupContent('Token', $node->getAttribute('role'), $meaning) }($node); }
    elsif (($node->getAttribute('role') || '') eq 'ID') {

**Comment by brucemiller** [comment]
More concretely: the point is that if it has meaning, treat it as if it were a token.  I suspect the same approach works for OM

**Comment by dginev** [comment]
Cool! It is slightly worrying that we have many different handlers for very related things but it's a start. I've pushed this, and verified it has good cross-references and an `m:laplacian` content element (pragmatic!)

**Comment by brucemiller** [comment]
Seems to work; OpenMath.pm
```
  elsif ($tag eq 'ltx:XMApp') {
    if(my $meaning = $node->getAttribute('meaning')){
      my $sub = lookupConverter('Token', $node->getAttribute('role'), $meaning);
      return &$sub($node); }
   ....
```

**Comment by dginev** [comment]
Thanks, added!

File: lib/LaTeXML/Core/Document.pm [hunk 6]
```diff
@@ -1439,7 +1439,66 @@ sub pruneXMDuals {
...
+    $compact_apply->appendChild($n_arg); }
+  # if the dual has any attributes migrate them to the new XMApp
+  my %transfer_attrs = ();
+  for my $attr ($dual->attributes) {
+    if ($attr->nodeType == XML_ATTRIBUTE_NODE) {
+      $transfer_attrs{ $attr->nodeName } = $attr->getValue; } }
+  $self->replaceNode($dual, $compact_apply);
+  # transfer the attributes after replacing, so that the bookkeeping has been undone
+  for my $key (keys %transfer_attrs) {
+    $self->setAttribute($compact_apply, $key, $transfer_attrs{$key}); }
```
brucemiller, line 1501: probably should have a matching `$self->closeElementAt($compact_apply);` which would run any afterClose thingies (since I just tracked down cases where that didn't happen).

dginev: Didn't have a test for that, oops! Done.

lib/LaTeXML/Core/Document.pm
```diff
@@ -1439,7 +1439,67 @@ sub pruneXMDuals {
...
+  return unless $single_duality;
+
+# If we made it here, this is a dual with two mirrored applications and a single XMTok difference, compact it.
+  my $compact_apply = $self->openElementAt($dual->parentNode, 'ltx:XMApp');
+  for my $n_arg (@new_args) {
+    # one of the args has our dual node that needs compacting
+    if (ref $n_arg eq 'ARRAY') {
+      my ($c_arg, $p_arg) = @$n_arg;
+      # Transfer all c_arg attributes over, it should be primary?
+      for my $attr_key (qw(decl_id meaning name)) {
```
brucemiller, line 1486: Should include `omcd` in this list.

brucemiller:
The dlmf diffs are kinda scary to read, although the resulting xmath is nice (occasionally surprising).  But except for missing `omcd` it looks good.

dginev:
Thanks for doing the QA work, much appreciated! omcd has been added.

brucemiller:
WooHoo!! Thanks!

PR #1317, opened 2020-07-29
Declare enhancement

brucemiller:
This PR enhances `ltx:declare` to allow more `ltx:tag`s providing for more labelling/documentation options.  It also better supports duals better.


lib/LaTeXML/Package/latexml.sty.ltxml
```diff
@@ -329,38 +365,34 @@ DefConstructor('\lxDeclare OptionalMatch:* OptionalKeyVals:Declare {}',
...
-      nowrap  => defined $keys->getValue('nowrap'),
+    # Temporary(?) Hack: If no description, bui tag is of form <math>: text
+    # make description = tag, and tag be only the shorter text part
+    $whatsit->setProperties(scope => getDeclarationScope($kv),
+      role    => ToString($kv->getValue('role')),
+      name    => ToString($kv->getValue('name')),
+      meaning => ToString($kv->getValue('meaning')),
+      trace   => defined $kv->getValue('trace'),
+      wrap    => defined $kv->getValue('wrap'),
+      nowrap  => defined $kv->getValue('nowrap'),
```
dginev, line 389: Likely not immediate feedback, but having `wrap` and `nowrap` as two separate booleans is a trap in hiding. In particular, the case where "somehow" we end up with both `wrap=>1, nowrap=>1`. Maybe it's slightly cleaner with a single attribute with 3 values? `undef`, `1`, `0` for "default", "force", "never".

lib/LaTeXML/Core/Rewrite.pm
```diff
@@ -438,6 +439,19 @@ sub domToXPath_rec {
         return ($xpath, $n, $n); }
       else {
         return ($axis . '::*', 1, 1); } }
+    # Also, an XMRef pointing to a wildcard is a wildcard!
+    # (or pointing to an XMArg|XMWrap of a wildcard!)
+    elsif ($qname eq 'ltx:XMRef') {
+      my $id = $node->getAttribute('idref');
```
dginev, line 445: Another tiny question, can't we use the `->realizeXMNode` here as well? May be helpful if XMRef handling is refactored one day.

t/math/declare.xml
```diff
@@ -924,16 +924,16 @@
...
-      <tag>Plus</tag>
+      <text>Plus</text>
     </declare>
     <declare xml:id="S6.XMD3">
-      <tag>Foo</tag>
+      <text>Foo</text>
     </declare>
     <declare xml:id="S6.XMD4">
-      <tag>Oddball</tag>
+      <text>Oddball</text>
```
dginev, line 936: Cool! Is there also a test for the new `<tags>` capability? It may double as a nice example of advanced rewrite output that we keep around to avoid regressions.

lib/LaTeXML/Package/latexml.sty.ltxml
brucemiller: To a certain extent these are left-overs. During development, there kept being cases of automatically wrap this, but don't wrap that, etc.  Was useful to experiment, but I should take a look and cleanup before merging. Between TeX and Perl, reliably maintaining those 3 values can be tricky though!

lib/LaTeXML/Core/Rewrite.pm
brucemiller: That was my first thought when writing, but you don't actually want to do wildcard handling on the referenced node (I think).

t/math/declare.xml
brucemiller: No, actually you'll just get errors (if you used old xml; which is kinda not likely to happen, or?)

dginev: I was referring to the new capability at:
https://github.com/brucemiller/LaTeXML/pull/1317/files#diff-f56ed2130145887d3624288c57713a2dR156

where `ltx:declare` is allowed a new `definiens` attribute and a `ltx:tags` child. Could be nice to have a test for that?

brucemiller: Ah, yesss, well... I'm not sure I want to commit (to coin a phrase) yet to how that attribute gets there...

lib/LaTeXML/Package/latexml.sty.ltxml
brucemiller: I removed `wrap`, since it was dead code.

t/math/declare.xml
brucemiller: But I forgot to copy my dev version of declare.tex over, which at least did exercise tag more!

PR #1318, opened 2020-08-03
Box reuse

brucemiller:
This PR addresses problems that occur when boxes and whatsits (especially those that create `ltx:XMDual`) are absorbed more than once into the XML.  Although this wasn't originally an intended usecase, it's easy to do by accident and hard to forbid.

For your amusement...


dginev:
Very cool, I will do a detailed read through tomorrow morning, thanks!

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -6758,28 +6756,32 @@ DefConstructor('\@XMRef{}', "<ltx:XMRef _xmkey='#1'/>",
...
-        if (my $referrers = LookupValue('xmkey_referrers:' . $key)) {
-          map { $document->setAttribute($_, idref => $id) } @$referrers;
-          AssignValue('xmkey_referrers:' . $key, undef, 'global'); } }
-} });
+      if (($qname ne 'ltx:XMRef') && ($qname =~ /^ltx:XM/) && !$node->hasAttribute('xml:id')) {
+        GenerateID($document, $node, undef, ''); } } });
+
+Tag('ltx:XMDual', 'afterClose:late' => sub {
+    my ($document, $node) = @_;
+    my %ids  = ();
```
dginev, line 6770: It is much nicer to see this lexical `%ids` compared to the previous State-based assignment/lookup. Seems like a win.

t/math/declare.xml
```diff
@@ -46,51 +46,51 @@
...
                   <XMTok font="italic" role="UNKNOWN">x</XMTok>
                   <XMTok fontsize="70%" meaning="1" role="NUMBER">1</XMTok>
                 </XMApp>
                 <XMTok role="PUNCT">,</XMTok>
-                <XMDual role="ID" xml:id="S1.Ex1.m1.4">
+                <XMDual role="ID" xml:id="S1.Ex1.m1.7">
                   <XMApp>
                     <XMTok decl_id="S1.XMD6"/>
-                    <XMRef idref="XM41"/>
+                    <XMRef idref="S1.Ex1.m1.1"/>
```
dginev, line 76: Ah, I see, very nice test, `n` gets reused in a number of locations in the same top-level formula.

t/math/declare.xml
```diff
@@ -397,57 +397,57 @@
...
-                <XMDual role="FUNCTION" xml:id="S2.Ex5.m1.4">
+                <XMDual role="FUNCTION" xml:id="S2.Ex5.m1.10">
                   <XMApp>
                     <XMTok decl_id="S2.XMD10"/>
-                    <XMRef idref="XM30"/>
-                    <XMRef idref="XM31"/>
-                    <XMRef idref="XM32"/>
+                    <XMRef idref="S2.Ex5.m1.1"/>
+                    <XMRef idref="S2.Ex5.m1.2"/>
+                    <XMRef idref="S2.Ex5.m1.3"/>
```
dginev, line 412: Really happy to see the `XM\d+` idrefs disappearing, I had seen them while working but hadn't gotten to asking what is up with them.

dginev:
LGTM! Very sensible upgrade.

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: I believe the What Was I Thinking rationale was thinking of XMRef as for more general sharing than only within XMDual. Only over time did it become clear that any non-dual usage needed different treatment (and probably a different element).

t/math/declare.xml
brucemiller: Same here: I kept forgetting to track down where they came from. Basically a misnaming `getXMArgID` which really shouldn't be ID, but a stand-in for an id.  I'd change it if I came up with a good name!

PR #1325, opened 2020-08-12
disable ApplyNary over XMApp with meaning attribute

dginev:
Most of the discussion here happened via email. I added a minimal test case showing what I tried to achieve, the important aspects being the two declarations for `Na` and `Cl`. The molecule markup is a bit of an add-on that wasn't related to the patch.

Patch itself was realizing that ApplyNary in MathParser would ignore the `meaning` attribute on an XMApp element itself, and always assume the operator (first child of XMApp) is meaningful. With some cases of `\lxDeclare` the operator will not be useful at all, if a parse is even possible, so might as well avoid running that code entirely. Feedback welcome, but should be safe to merge?


t/math/declare_chemistry.tex
```diff
@@ -0,0 +1,18 @@
...
+\makeatother
+
+\begin{document}
+
+\lxDeclare[meaning=sodium]{$Na$}
+\lxDeclare[meaning=chlorine]{$Cl$}
+
+% No markup, rewritten:
+
+\[ Na\,Cl \]
```
dginev, line 16: This is left intentionally in the worst possible "plain ascii typing" input, as an exploratory test of mine for what `\lxDeclare` can and can not handle. It is easier if you have to say `\mathrm{Na}` since then one might as well add a custom macro `\atom{Na}` and avoid the declaration entirely.

brucemiller: I can see the logic in the patch:  If you had A*B*C and A*B means something special, you perhaps don't want it to combine with *C, right?  But InvisibleTimes of "N" and "a" not being multiplied with "C" and "l" isn't very compelling.

In fact, although I appreciate the experimentation, I really don't want to solidify it into a test case (at least at this stage); we end up struggling to keep tests working, or adapt them, as we develop & enhance the rest of LaTeXML.

dginev: Well, that's fine and good, and you are welcome to close the PR.

I am not likely to stop reminding that "invisible times" is a strong assumption, so the n-ary rewrite based on it is also a strong assumption. The very need for `\lxDeclare` here is to have a mechanism to override the grammar, and if there is no way to have the patterns match, one hits a dead-end.

I could also make my case with unmarked capital-letter relation symbols. Say:

```tex
\[ xRyQz \]
```

intended to mean `Q(R(x,y),z)`. I also won't be able to untangle that with \lxDeclares redefining `R` and `Q`, as the n-ary case will trigger analogously to the `NaCl` case.

brucemiller:
> \[ xRyQz \]
> 
> intended to mean |Q(R(x,y),z)|. I also won't be able to untangle 

why not? ezpz
```
\lxDeclare[role=RELOP,meaning=relatively]{$R$}
\lxDeclare[role=RELOP,meaning=generally]{$Q$}
```

dginev:
Not that important for now, either way, closing.

PR #1327, opened 2020-08-20
prune XMDuals with mirrored single tokens

dginev:
Continues the work of #1309 to now also compress duals with single mirrored tokens. 
The changes to the test suite are a very good illustration of the exact effect here.

I stumbled on this while working on the a11y PR #1305 , and wanted to detach it into a standalone PR, to make merging the various pieces easier.

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -1443,11 +1443,31 @@ sub pruneXMDuals {
       $self->compactXMDual($dual, $content, $presentation); } }
   return; }
 
+our @CONTENT_TRANSFER_ATTRS = qw(decl_id meaning name omcd);
+
 sub compactXMDual {
   my ($self, $dual, $content, $presentation) = @_;
-  # For now only handle compacting mirror applications
-  return if ($self->getNodeQName($content) ne 'ltx:XMApp') ||
-    ($self->getNodeQName($presentation) ne 'ltx:XMApp');
+  my $c_name = $self->getNodeQName($content);
+  my $p_name = $self->getNodeQName($presentation);
+  # 1.Quick fix: merge two tokens
+  if (($c_name eq 'ltx:XMTok') && ($p_name eq 'ltx:XMTok')) {
+    for my $attr_key (@CONTENT_TRANSFER_ATTRS) {
+      if (my $attr_val = $content->getAttribute($attr_key)) {
+        $content->removeAttribute($attr_key);
+        $presentation->setAttribute($attr_key, $attr_val); } }
+    # if the dual has any attributes migrate them to the new XMTok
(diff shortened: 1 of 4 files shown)


t/complex/physics.xml
```diff
@@ -1117,10 +1117,7 @@
                   <XMApp>
                     <XMTok name="rightarrow" role="ARROW">→</XMTok>
                     <XMTok meaning="absent"/>
-                    <XMDual>
-                      <XMTok meaning="gradient"/>
-                      <XMTok name="nabla" role="OPERATOR">∇</XMTok>
-                    </XMDual>
+                    <XMTok font="italic" meaning="gradient" name="nabla" role="OPERATOR">∇</XMTok>
```
dginev, line 1120: I noticed that `font="italic"` gets added to the tokens when merged, I was wondering what causes that and how to avoid it - since I am pretty sure we don't want it added for the cases in the tests.

dginev: Ok, figured it out, simply do not overwrite the presentation XMTok's attributes with attributes of the parent XMDual, in the cases where both have the same attribute. With a special exception where `xml:id` gets overwritten to be the XMDual id, in case it was being referenced from somewhere else.

dginev:
Should be good for a review.

lib/LaTeXML/Core/Document.pm
```diff
@@ -1443,11 +1443,32 @@ sub pruneXMDuals {
...
+        $presentation->setAttribute($attr_key, $attr_val); } }
+    # if the dual has any attributes migrate them to the new XMTok
+    my %transfer_attrs = ();
+    for my $attr ($dual->attributes) {
+      if ($attr->nodeType == XML_ATTRIBUTE_NODE) {
+        $transfer_attrs{ $attr->nodeName } = $attr->getValue; } }
+    $self->replaceNode($dual, $presentation);
+    # transfer the attributes after replacing, so that the bookkeeping has been undone
+    for my $key (keys %transfer_attrs) {
+      if (($key eq 'xml:id') || !$presentation->getAttribute($key)) {
```
brucemiller, line 1466: Some of the dual's attributes should override the presentation's attributes, certainly `role`, `meaning`, probably anything from `@CONTENT_TRANSFER_ATTRS`.

Also, theoretically if more than 1 of (dual,presentation,content)have ids, all referrers to the extras should get updated to use the one you keep.

dginev: Thanks, hadn't thought that part through to be honest, and the changes to the tests are Good - `\Langle` from mathbbol and `\Ydown` from stmaryrd now once again have the correct roles.

dginev: Ready for re-review

PR #1330, opened 2020-08-24
Fixes to catcode tables and minor lints

dginev:
Starting to put small piecemeal PRs to disentangle my expl3 work into acceptable chunks...

The first bit is correcting the catcode tables - we currently have 18 catcodes in Token.pm, indexed 0 through 17, and any catcode table should have exactly 18 elements, ideally chunked 4 per line, with a last line of 2 elements. I believe two of the Gullet tables had typos, as they trated Marker as array element 18 instead of 17, which spurred this. It became painfully clear when I added a new catcode and things were breaking. 

I also eliminated the use of raw integers in Token.pm for the `T_` constructors and relied on the `CC_` constants instead, to protect from possible typos.

Lastly, the linter warned me that the escapes `\'` and ``` \` ``` do not exist, so I dropped those backslashes from Gullet.pm

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -155,7 +155,7 @@ sub getSource {
 
 sub getSourceMouth {
   my ($self) = @_;
-  my $mouth = $$self{mouth};
+  my $mouth  = $$self{mouth};
   my $source = defined $mouth && $mouth->getSource;
   if (!defined($source)) {
     foreach my $frame (@{ $$self{mouthstack} }) {
@@ -212,7 +212,7 @@ our @hold_token = (
   0, 0, 0, 0,
   0, 0, 0, 0,
   0, 0, 1, 0,
-  0, 0, 1);
+  0, 1);
 
 sub readToken {
   my ($self) = @_;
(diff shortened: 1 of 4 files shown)


brucemiller:
This should be scary, but looks reasonable. Are we still getting automatic Travis tests on these?

dginev:
We are for all of them, but they are not showing in the github PR interface at the moment?

My "internal PR" branches are here: https://travis-ci.com/github/dginev/LaTeXML/pull_requests

while the "latexml PR" builds are here: https://travis-ci.org/github/brucemiller/LaTeXML/pull_requests

Btw, I notice you haven't yet migrated to the newer `.com` version of Travis, which is done by hand, and is mildly (almost too mildly) painful, but straightforward. They are unifying their infrastructure, and I am indeed seeing the badges for the PRs on my repository, which is already migrated to `.com`. They have a tutorial (or is it a Q&A?) on the topic here:

https://docs.travis-ci.com/user/migrate/open-source-repository-migration

brucemiller:
It appears that I'm migrated at the travis-ci end, but I updated the webhook on the github side from .org to .com. Hopefully that does it.

Otherwise, looks good.

dginev:
Thanks! I am wondering if the `.com` migration was successful, given that I would expect the builds to show up at the standard URL:
https://travis-ci.com/github/brucemiller/LaTeXML

which currently they aren't?

Also, if/when this migration is successful, we'd need to change the README badge to point to
https://api.travis-ci.com/brucemiller/LaTeXML.svg?branch=master

PR #1332, opened 2020-08-24
update mathjax url to recommended latest CDN

matthuszagh:
Updates mathjax to latest CDN recommended by mathjax documentation [here](http://docs.mathjax.org/en/latest/web/start.html#using-mathjax-from-a-content-delivery-network-cdn).


brucemiller:
Cool! Looks reasonable; Thanks for the patch!

PR #1333, opened 2020-08-25
upgrades to \meaning, \show-like macros, writableTokens

dginev:
Detaching more of the (hopefully) independently reasonable upgrades that set the stage for the core upgrades needed for expl3 in texlive2020. Focus is on `\meaning`, `writableTokens` and some of the `\show`-like macros, which I think can be upgraded in isolation.

Curious to see if the Travis expl3 tests will pass.

(This PR was originally opened at [#5](https://github.com/dginev/LaTeXML/pull/5) and I am migrating it by hand, now that the main fork's master has moved)

### DIFF
--- a/lib/LaTeXML/Package/eTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/eTeX.pool.ltxml
@@ -26,7 +26,7 @@ DefPrimitiveI('\protected', undef, sub {
     $STATE->setPrefix('protected'); return; }, isPrefix => 1);
 
 # \detokenize
-DefMacro('\detokenize GeneralText', sub { Explode(writableTokens($_[1])); });
+DefMacro('\detokenize GeneralText', sub { Explode(writableTokens($_[1], 1)); });
 
 # \unexpanded
 # This is like \noexpand, but acts on <general text>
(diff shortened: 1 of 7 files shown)


brucemiller:
Now, the only caller of `writableTokens` that does *not* pass in the 2nd arg (`$addspaces`) is in eTeX's `\scantokens`.  Can you verify that it should not add the space?  If it does need to add the space, we probably should get rid of that extra argument and always add. (simplify!)

dginev:
I just checked via the `t/tokenize/hashes.tex` test by adding some trailing content on the line that uses `\scantokens`, and indeed there is a trailing space in the PDF, and *no* trailing space in the XML.

In other words, your suggestion is correct, `writableTokens` should always produce the spaces and not have this option. I will change to that and commit now. Thanks!

lib/LaTeXML/Package/eTeX.pool.ltxml
```diff
@@ -45,6 +45,11 @@ DefMacro('\readline Number SkipKeyword:to SkipSpaces Token', sub {
       DefMacroI($token, undef, Tokens(Explode(($mouth->readRawLine || '') . "\r"))); }
     return; });
 
+# https://tex.loria.fr/moteurs/etex_ref.html#scantokens
+# TODO: "Parentheses and a single space representing the pseudo-file will be displayed
+#        if \tracingscantokens (q.v.) is positive and non-zero."
+#       We are currently missing a final trailing space in some uses
+#           c.f. t/tokenize/hashes.tex
```
dginev, line 52: Of course this is more complicated for additional reasons, but won't address that in this PR if possible. Left a guiding comment.

brucemiller:
It all seems plausible :>

PR #1335, opened 2020-08-29
Faitfhful handling of the_toks

### LINKED ISSUE #1278: support expl3 extensions in texlive 2019

dginev:
Fixes #1278 . Succeeds #1282 .

This PR contains what I hope (fingers crossed!) is the most viable approach we have to realign the latexml codebase with the B Book handling of `the_toks`, containing the differences to as few pieces as possible.

I need to propagate the change to pushback to more files, this PR has validated 1) the main tests work and 2) the new `expansion/pdftex_expanded.tex` test works.

In particular the diff with master's XML is included here. You can validate against the PDF, also in the PR.
<details>

```
31c31
< ##,“relax,“aaa</p>
---
> ##,“aaa ,“aaa</p>
35c35
< ##,####</p>
---
> ##,##</p>
39c39
< “relax</p>
---
> “aaa</p>
43c43
< “aaa</p>
---
> x</p>
47c47
< x“relax“relax“relax</p>
---
> x“aaa “noexpand “aaa</p>
51c51
< x“relax“relax“relax</p>
---
> xx“aaa</p>
56c56
< x</p>
---
> “aaa</p>
```

</details>

The expl3 interpretation itself is not ready yet, but I have some more test files to include en route to it, one of which contains the distilled problem. Before I do that, I would like to regroup with @brucemiller and complete the work here targeting just `the_toks` in isolation.

Also, to restate my conclusion from the various digging in #1282 , if we can realign this particular piece of logic, I think the rest of the issues with expl3 will be contained to small patches, _if I am to assume_ that my test case is the only remaining blocker in the 2019/2020 texlives. I am quite confident that test requires only a faithful treatment of TeX's `#` CC_PARAM tokens, and respectively `\the`-like expansions.


lib/LaTeXML/Core/Gullet.pm
```diff
@@ -271,13 +275,26 @@ sub readXToken {
...
+        else {
+          Fatal('misdefined', $r, undef, "Expected a Token, got " . Stringify($_)); } }
+      next unless @expansion;
+      if ($$LaTeXML::Core::Token::X_THE{ $$token[0] }) {
+        # magic THE_TOKS handling, add to pushback with a single-use noexpand flag only valid
+        #    at the exact time
+        # the token leaves the pushback.
+        # This is *required to be different* from the noexpand flag, as per the B Book
+        # Let's choose something that will immediately break latexml if we overlook it...
+        @expansion = map { ['the_token', $_] } @expansion; }
```
dginev, line 295: This comment and line of code are the crux of the "solution", as ugly as it feels. That's why I am calling it "faithful to the B Book", maximizing positive attitude.

TeX prescribes a magical global array where tokens from expansions of `\the-like primitives` are smuggled to, and then reinserted a level up at `scan_toks`. Since we don't have scan_toks, a naive approach on my part had to require rewriting all loops with `readXToken`, at the least. As it turned out, it required even more - and ultimately felt unfeasible, as the refactor was going to be huge, and possibly very fragile.

So instead of all that, I had an Eureka moment while cargo-culting boilerplate and refactoring, that this can be handled fully internally, as we initially tried with the "no_expand flag" in `$$token[2]`. But the key reason we failed is that we added the flag to the token and hoped both cases are treated similarly. They are not. I saw expl3 expansion misbehave for this reason.

The `\the`-related flag is one connected to the very first time the expanded tokens are being "consumed" from the global variable holding them. And since we always keep such tokens (coming from an invocation) into the Gullet's `{pushback}`, the "aha!" moment was that what we really want is some way to add a flag "for the duration of presence in the pushback". Crucially, that flag must always and unconditionally expire the exact moment that the token is removed from the gullet. Luckily, at least for this part, virtually all code paths end up in `readToken` and `readXToken` for removing the token from the pushback, so I can add two lines in those two subroutines and have support in the entire codebase. Whew.

The annoying bit is that I don't have a great way to avoid paying _some_ performance penalty on every pop from the pushback, to check for this case. For the first stab I used a "magical array", which is impossible to use as a Token, and has to be handled immediately (or Perl will die violently, which is good in this case - it alerts where to add more support).

lib/LaTeXML/Core/Token.pm
```diff
@@ -191,6 +191,13 @@ our @CC_SHORT_NAME =           #[CONSTANT]
   T_CS
 );
 
+our $X_THE = {
```
dginev, line 194: this is a stupid name, should be `TREATED_AS_THE` or some such...

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -218,18 +218,9 @@ DefParameterType('Expanded', sub {
...
-          $level--;
-          last unless $level; }
-        elsif ($cc == CC_BEGIN) {
-          $level++; }
-        push(@tokens, $token); }
-      return Tokens(@tokens); }
+      return scalar($gullet->readBalanced(1))->without_dont_expand; }
     else {
-      return $token; } },
+      return $token->without_dont_expand; } },
```
dginev, line 223: I feel pretty good that this argument type should remove the noexpand flag conceptually. And I also feel somewhat good about suggesting that we erase `EExpanded` entirely and move all of those to the regular `Expanded`, as the uniform treatment should be able to handle both.

dginev:
Ah, I made a mistake opening the PR against master, which now made the diff large... My apologies. It will be much easier to view here after we merge #1333 and I rebase again.

dginev:
Thanks for merging #1333 , I have rebased here and switched the base back to master :+1:

dginev:
Alright, major headway can be reported here. All tests pass, including the newly added `expl_generate_arguments` test, which required this massive refactor. Sadly expl3 on texlive2019 hasn't passed fully _just yet_, but it should be close.

I won't jump into a long-winded explanation of the changes just yet, will try to get expl3 fully loading first. But a brief summary is that the handling of the parameter token pairs `##` and `#1-#9` has been moved up to Expandable construction time, also adding a new `CC_ARG` catcode+token type, for the constructed arguments. All other changes propagate from there.

lib/LaTeXML/Core/Definition/Expandable.pm
```diff
@@ -19,20 +19,25 @@ use LaTeXML::Common::Error;
 use LaTeXML::Core::Token;
 use LaTeXML::Core::Tokens;
 use LaTeXML::Core::Parameters;
+use LaTeXML::Package qw(TokenizeInternal);
 use base qw(LaTeXML::Core::Definition);
 
 sub new {
   my ($class, $cs, $parameters, $expansion, %traits) = @_;
+  if (!ref $expansion) {    # Tokenization DEFERRED till actually used (shaves > 5%)
```
dginev, line 27: I lost the 5% optimization actually, maybe I can re-gain it once everything is finished and one can rethink optimizing...

dginev:
With the help of Travis, I can now summarize that the regression I was studying has been resolved, and we've moved further down the expl3 file. 2018 and 2019 are regressing in the same way, so the issue may still be related to the specific changes in the PR not being complete (since we had 2018 passing prior). Let's see...

dginev:
@brucemiller to my delight I now have parity with texlives 2015,2016 and 2018 which all pass in Travis! :tada: 

I think this is a clear mark of a stopping point for this PR, where we can discuss and improve it to be mergeable into master - for what it succeeds in doing.

The next missing bit in the 2019 expl3 is at line 11,351 and seems related to a truth table in `Appendix D: Dirty Tricks` in the TeXbook, using the trick `\iffalse{\fi` to influence the TeX mouth counters (which latexml may not yet be emulating "faithfully"). I will make a new branch for that line of digging.

dginev:
Indeed slight slowdown can be spotted from this PR, though not too major. I've excluded the `83` tests since they don't yet pass, here are the parallel and serial comparisons:

```
PR parallel, 32 jobs

real	0m42.633s
user	5m17.223s
sys	0m6.380s

master parallel, 32 jobs

real	0m40.953s
user	4m56.327s
sys	0m6.254s

---

PR serial, 1 job

245.65 user
4.32 system
4:09.96 elapsed

master serial, 1 job

227.09 user
4.05 system
3:51.14 elapsed
```

lib/LaTeXML/Core/Token.pm
```diff
@@ -290,7 +322,7 @@ sub revert {
 
 sub toString {
   my ($self) = @_;
-  return $$self[0]; }
+  return $$self[1] == CC_ARG ? ("#" . $$self[0]) : $$self[0]; }
```
dginev, line 325: I hijacked the `toString` method to create the TeX-near serialization, but maybe I should have consistently used `revert` instead? Most of the cases where I needed this were using `getString`, but I thought I should let that method reliably return the actual raw value in `$$self[0]`. Generally having this new arg token makes matters tricky, since its string serialization is different from its actual integer value (extra `#`).

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -99,7 +99,19 @@ RequireResource('LaTeXML.css');
...
+    if ($inner) {
+      ($value) = $inner->reparseArgument($gullet, $value); }
+    $value; },
+  reversion => sub {
+    my ($arg, $inner) = @_;
+    (T_BEGIN,
+      ($inner ? $inner->revertArguments($arg) : Revert($arg)),
+      T_END); });
+
+DefParameterType('DefPlain', sub {
```
dginev, line 112: I could technically add the masks treatment to `Plain` and `Expanded` directly and still have all tests work, but I thought this is a hair more efficient in performance. Since the masks are calculated and carried around only when requested, not requesting them in `Plain` and `Expanded` speeds things along a small amount.

dginev:
I've successfully reintroduced the optimization with tokenizing Expandable "string definitions" only when used. Performance is slightly better, and I think looks reasonable considering I added some extra tests in this PR, which I am also running here:
```
PR latest, parallel, 32 jobs

real	0m41.809s
user	5m0.505s
sys	0m5.987s

PR latest, serial, 1 job

235.29 user
4.22 system
3:59.50 elapsed
```

dginev:
I am currently doing cartwheels of joy. After I read into the braces trick, it turned out latexml was already doing it correctly (nice!!), and instead it was a very trivial case of a missing eTeX register declaration. So trivial that I added it to this PR rather than making a new branch.

It has been a long while since I've seen this happy message:
image

The expl3 tests are now fully passing in texlive 2019! Finally!! I am actually _done_ here, as far as the 2019 release is concerned. I should probably dig up a 2020 release now and check if that also passes...

dginev:
I installed the latest Ubuntu groovy 20.10 release on my laptop tonight, and verified that the `master` branch passes all tests and still fails the expl3 tests early/badly. This PR branch sadly _also_ fails the expl3 tests, but it manages to interpret the expl3-code.tex file upto line 20379, so it gets pretty far along.

The texlive 2020 error at least seems related to the interplay of `\newlinechar` and `\endlinechar`, and I'll study it more tomorrow, this time really in a separate branch :+1:

lib/LaTeXML/Core/Token.pm
brucemiller: revert certainly sound more correct.  We possibly should scan all these variants and pin down the exact proper usecases and then use them consistently?

lib/LaTeXML/Core/Parameters.pm
```diff
@@ -61,7 +62,8 @@ sub readArguments {
   my @args = ();
   foreach my $parameter (@$self) {
     my $value = $parameter->read($gullet, $fordefn);
-    push(@args, $value) unless $$parameter{novalue}; }
+    if (!$$parameter{novalue}) {
+      push(@args, $value); } }
```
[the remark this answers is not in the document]
dginev: Haha, there used to be a call to `PrepArgTokens` in that new if body, while i was investigating what the optimal locations would be to do the param processing - it became obsolete once I settled on using `Expandable::new`. Can be reverted back to unless, no problem!

lib/LaTeXML/Core/Tokens.pm
```diff
@@ -37,16 +37,23 @@ sub Tokens {
...
+  push(@tokens, $marks) if $marks;
+  return bless [@tokens], $class; }
 
 # Return a list of the tokens making up this Tokens
 sub unlist {
   my ($self) = @_;
-  return @$self; }
+  my @unlisted = @$self;
+  if (ref $unlisted[-1] eq 'ARRAY') {
+    @unlisted = @unlisted[0 .. scalar(@unlisted) - 2]; }
```
brucemiller, line 55: So, `->unlist` *removes* the no-expand-once mask?

lib/LaTeXML/Core/Tokens.pm
```diff
@@ -84,13 +100,24 @@ sub beDigested {
 
 sub neutralize {
   my ($self, @extraspecials) = @_;
-  return Tokens(map { $_->neutralize(@extraspecials) } $self->unlist); }
+  my @neutralized = map { $_->without_dont_expand } $self->unlist;
+  if (ref @$self[-1] eq 'ARRAY') {    # preserve marks
+    push(@neutralized, @$self[-1]); }
```
brucemiller, line 105: The middle is dead code, here.  What is the expectation when both a no-expand-once (from `\the`) and dont-expand (from `\noexpand`) are present?  Or can that even happen?

lib/LaTeXML/Core/Tokens.pm
```diff
@@ -84,13 +100,24 @@ sub beDigested {
...
+  return Tokens(map { $_->neutralize(@extraspecials) } @$self); }
+
+sub without_dont_expand {
+  my ($self) = @_;
+  my @cleaned = map { $_->without_dont_expand } $self->unlist;
+  if (ref @$self[-1] eq 'ARRAY') {    # preserve marks
+    push(@cleaned, @$self[-1]); }
+  return Tokens(@cleaned); }
```
brucemiller, line 113: So, here, you remove the `\noexpand` flag, but keep the `\the` mask/mark?

lib/LaTeXML/Core/Tokens.pm
dginev: Ignores rather. I use a slice to copy the array without the last element. The tokens object still retains the masks and they can be accessed after the unlist.

lib/LaTeXML/Core/Tokens.pm
dginev: The current code does not allow for an arrayref mask made via `\the` to enter `Tokens` as a regular `Token`. It always slots that information as a mask in this magical last position of the array. Which is also why the middle isn't dead code - the call to `->unlist` ignores the masks, and then the check on `[-1]` reintroduces them in the new `@neutralized` array. 

Meanwhile, the `\noexpand` neutralization is still done in the same way, via a special `Token` that has `$_[2]` set.

lib/LaTeXML/Core/Tokens.pm
brucemiller: Oooooh, so you may have *both* array-wrapped tokens *and* a mask (both of which inhibit expansion?

lib/LaTeXML/Core/Tokens.pm
dginev: Exactly, the masks are kept, because the two mechanisms are conceptually independent from each other. I don't know if I actually need this extra care in practice, but it made sense to me to add it.

lib/LaTeXML/Core/Tokens.pm
dginev: This code may get a lot cleaner if I create a new object inheriting from `Tokens`, say `MaskedTokens`, which does all of these acrobatics internally. It will be marginally slower, since objects aren't free, but will likely clean up the logic a lot.

brucemiller: Sure, but you never use `@neutralized`, which is why it's dead.

lib/LaTeXML/Core/Tokens.pm
dginev: The array-wrapped tokens would cause a Fatal error when `Tokens()` is called, they're not allowed in as legal members. Those need to be unwrapped immediately after they are returned by `readToken` or `readXToken`, which are the only two producers of them.

brucemiller: OK, but my original point was that any caller of `->unlist` gets the no-expand-once flags/mask/marks removed, right?
(which implies that callers are expected not to expect it, and any caller that expects those masks preserved has to make its own arrangements;  that is to say that `unlist` is no longer "merely" an accessor)

dginev: I am unsure why you are searching for new behaviors of unlist? The code change here is exactly to keep it _the same_ it has always been, guarding against the new special case where there is something different at the last slot of the array. It continues to return the array of tokens, as it always has.

If someone needs the masks explicitly, there is a new accessor:
```perl
my @marks = $tokens->get_marks;
```

Though I should really decide if it should be marks or masks... In general I am abusing the array for behavior that is really meant to be in a separate slot. Refactoring to a newly added `MaskedTokens` could improve on that front.

lib/LaTeXML/Core/Tokens.pm
dginev: Ah oops! This is a bug actually, I think I mangled that unintentionally... Might not be well-tested in the testing suite too.

dginev: I added a patch, should be accurate to my intentions now.

lib/LaTeXML/Core/Tokens.pm
brucemiller: I'm testing an alternative approach to how no-expand-once (or whatever it should be called) due to `\the` is noted. Thus I'm trying to track down the places where such a flag/mark/mask are preserved and where they are dropped.

I'm inclined to think that a MaskedToken (or rather a new CC) is a better approach than two distinct mechanisms (whether smuggled or renamed to MaskedTokens), but first I have to understand and replicate the current behaviour.  Hence the questions.

dginev: Cool! The main place of import is:
 - `readToken`, `readXToken` and `readBalanced` in the `DefExpanded` and `DefPlain` parameter types for the defining primitives need to preserve and pass on these marks. The only real place they are used is in the main logic of creating a new Expandable through `do_def`, where this masks-based neutralization is crucial.

For all non-definitional uses everything is handled internally by `readXToken`, since a loop that keeps calling it will stop short of expanding an array-wrapped token, and will return it as-is, unwrapped.

I think if those two aspects work, an alternative approach should work too.

brucemiller:
On 10/13/20 2:45 PM, Deyan Ginev wrote:
> Cool! 
The main mod went fast, easy, and all tests passed --- except 2.
So, you can see why I'm fishing into the fine details, trying to
figure out why it has different behavior.

dginev:
And we're back here! For bookkeeping's sake, the two "side PRs" we got merged were at [deyan fork #8](https://github.com/dginev/LaTeXML/pull/8) and  [deyan fork #9](https://github.com/dginev/LaTeXML/pull/9).

lib/LaTeXML/Package/TeX.pool.ltxml
dginev: Upon detailed review and reflection, this is indeed a great place for calling `->without_dont_expand`.

brucemiller:
Now trying to find a more suggestive name for `KEEP_THE`, `X_THE`, `NOEXPAND1` that suggests they're related.  Thinking of `DEFERRING_THE`, `DEFERRED_THE_COMMANDS`, `(CC|T)_DEFERRED_THE` ? Got anything better?

dginev:
I've used `smuggle` a lot for the verb since that's what the book does... The piece that keeps them together is `the` in the way I have been naming them, and all documentation, be it TeX books, or eTeX/pdfTeX manuals refer to the various behavior by mentioning `\the`. Not something we can undo with great naming I am afraid. My main point has always been to use that bad name as the _last_ suffix word in a longer name, to avoid the English determiner reading. Your `DEFFERRED_THE_COMMANDS` reads as a sentence more than an array name. "Hi, I deferred the commands you asked for." kind of reading.

So maybe a trio such as:
 - `KEEP_CC_FROM_THE`
 - `PRIMITIVES_AKIN_TO_THE`
 - `CC_RETURNED_BY_THE`

Is closer to my intuition... Hard to get a great name from an unfortunate start though.

brucemiller:
I'm about to commit the smuggle variant, which sounded more fun once I remembered you using it.

dginev:
You shouldn't have commented that, if any commit should be smuggled in, that's the one :>

dginev:
`SMUGGLE_THE_COMMANDS` is absolutely hysterical, let's keep it :>

brucemiller:
Ah, yes, but....

PR #1339, opened 2020-09-14
Adding even more consistent names in physics.sty

dginev:
Follow-up to commit 640cfc7a664f541f1e66ce4f3a805e8b51af1048

I was testing the a11y generation for all tests in `t/complex/physics.tex` and noticed sine of the names looked suboptimal (`asine`, `hypsine` etc). So I quickly edited them and regenerated the tests, feel free to review/adopt/adapt.

### DIFF
--- a/lib/LaTeXML/Package/physics.sty.ltxml
+++ b/lib/LaTeXML/Package/physics.sty.ltxml
@@ -300,10 +300,10 @@ DefMacro('\lx@physics@operatorP{}{}{}', sub {
 
 my @operators = (qw(
     sin sine cos cosine tan tangent csc cosecant sec secand cot cotangent
-    hsin hypsine cosh hypcosine tanh hyptangent csch hypcosecant sech hypsecand coth hypcotangent
+    hsin hyperbolic-sine cosh hyperbolic-cosine tanh hyperbolic-tangent csch hyperbolic-cosecant sech hyperbolic-secand coth hyperbolic-cotangent
     arcsin arcsine arccos arccosine arctan arctangent arccsc arccosecant arcsec arcsecand arccot arccotangent
-    asin asine acos acosine atan atangent acsc acosecant asec asecand acot acotangent
-    exp exponential log logarithm ln naturallogarithm det determinant Pr Probability
+    asin arcsine acos arccosine atan arctangent acsc arccosecant asec arcsecand acot arccotangent
+    exp exponential log logarithm ln natural-logarithm det determinant Pr probability
     ));
 
 while (@operators) {
--- a/lib/LaTeXML/Package/physics.sty.ltxml
+++ b/lib/LaTeXML/Package/physics.sty.ltxml
@@ -87,7 +87,7 @@ sub phys_close {
 sub phys_readArg {
   my ($gullet, $required, %delimiters) = @_;
   my ($arg, $open, $close);
-  my $next = $gullet->readToken();    # or XToken ?
+  my $next = $gullet->readToken();          # or XToken ?
   if ($next && $next->equals(T_BEGIN)) {    # Read regular arg, use default delimiters
     $gullet->unread($next);
     $arg = $gullet->readArg(); }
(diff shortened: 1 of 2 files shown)


brucemiller:
Actually, this doesn't work. What you overlooked was that `\longname` is also a macro alias to `\shortname`.  So, `@operators` really needs to be triples: "shortname,longname,meaning".  But I do prefer the phrases you have for meanings.

dginev:
Ah, of course - sorry I just renamed in a quick pass. I have switched them to the triples you suggest now, please take another look

brucemiller:
Sheesh, looking over your diff, I spotted a typo --- from me!  While you're in there, could you replace all 4 occurances of `secand` with `secant`?  Thanks!!

dginev:
Ah. Actually 6 in the .ltxml file and 8 in the .xml test! All fixed :+1:

brucemiller:
Woohoo! Thanks!

dginev:
Pretty cool! Leaving behind a motivating example in this comment, for any curious eyes that may want to see how `physics.sty` can be leveraged for accessibility:

image

0-andy-0:
Cool!  but much depends on context.... 
for many cos[-1](x) does not mean the "power" 1/(cos(x)) but acos(x)
and of course cos' means -sin ;)

dginev:
We have been having this discussion with @brucemiller , and this may be a good case where we want to draw the distinction between a `power`, and a `functional-power` (aka `applicative-power`), to resolve away the ambiguity.

brucemiller:
On 9/15/20 11:19 AM, Andrew Tonks wrote:
> Cool! but much depends on context....
> for many cos-1 <x> does not mean the "power" 1/(cos(x)) but acos(x)
> and of course cos' means -sin ;)

Indeed! And one of my favorite examples :>
(which I then proceeded to forget while working on physics.sty)

Now that I've got an unbiased ear: In my experience, this idiom
is only used on a certain category of functions (trigs for example,
occasionally a random $f$, but not log...) and *only* used when the
superscript is a literal "-1".
(ie. *never* with a variable as in \sin^n; unless, of course, you're 
dealing with functional-powers all along).

Is that consistent with your experience?

0-andy-0:
> Now that I've got an unbiased ear: In my experience, this idiom is only used on a certain category of functions (trigs for example, occasionally a random $f$, but not log...) and *only* used when the superscript is a literal "-1". (ie. *never* with a variable as in \sin^n; unless, of course, you're dealing with functional-powers all along). Is that consistent with your experience?

I don't know, -1 for the inverse, not the reciprocal, of a function is common in my work, so I *am* biased!

And if I read \exp^2(x) or \ln^2(x) I would have to read the context to be sure of what was meant.
Without the brackets I would probably assume  \exp^2 x = (\exp x)^2  and \ln^2 x = (\ln x)^2,
but I would recommend the author change the notation!

Personally, *functional* powers are the default for me if the superscript is an integer

 $f^0(x)=x$, the identity function, and $f^{n+1}(x)=f^n(f(x))$, and $f^{-n}$ exists if $f$ is a bijection (1-1 & onto).

I think I have seen (and even written) $\ln^{-1}=\exp$

And if I define a function Frob,  then Frob^2(x)=Frob(Frob(x)) 
[That's a real example: the Frobenius automorphism on a finite field with p^n elements has inverse Frob^{-1}=Frob^{n-1}.
Another example you know well is the Laplacian Del^2(v)=Del(Del(v)) - this is not the square of Del v.  
In more abstract settings you define the Dirac operator on a manifold to be the functional-square-root of the Laplacian, but I have not seen the power 1/2  used.   Neither have I seen integration written as Del^{-1} ]

As you say the convention for sine and cosine is different: multiplication-powers.  
My _guess_ is that trig functions are different because:

- they have somehow "different" domain (angle) and range (length), so cos^2(x) meaning cos of cos of x fails to type check

- we know the range is the real numbers, which has a multiplication that you do without thinking.

Thus people get lazy and write sin cos for sin(x)cos(x), I mean, what else could they mean :)

In this convention cos^{7/3} makes sense, as that power is applied to the reals.
 
Now what about sinh^2(x) ?  I guess it means the square, sinh(x) * sinh(x) by analogy with trig,  but that annoys me ... what about exp^2(x), is it exp(exp(x)) or just exp(2x) ?  And ln^2(x) is (ln x)^2 or ln(ln(x)) ?

brucemiller:
I suspect you're right about the field: in more physcally oriented fields (eg. special functions), regular power is most commonly desired (except for operators, especially differential).  With the occasional special case like inverse trigs (but I think we avoided that in DLMF).  I suspect that bias probably applies to the author of physics.sty (but the users?), and likewise an explicit superscript of "-1" is likely to mean inverse (where such a thing even makes sense; what's the inverse of the determinant? :> )  I'll experiment with detecting that.

0-andy-0:
and some superscripts are not powers, in any sense.  `C^{14}` for Carbon-14 , `H^n` for n-th cohomology (or `h^n, E^n` and even `\lim^n`). What fun! How easy would it be to take a census of the uses of `#0^{#1}(#2)` in the arxiv?

brucemiller:
Indeed, in general, we treat sub & superscripts as no more
than sub & superscripts. (We imagine a hypothetical analysis
phase which could resolve based on context, types", etc)
The objective with packages like physics.sty, where there is
(or seems to be) implied semantics, is to leverage, or at least 
preserve, it.

PR #1341, opened 2020-09-23
Cleanup semantics

brucemiller:
Another iteration in cleaning up the mathematical semantics.  This one focusing on the confusion of "matrix" as a type vs a constructor.  First, we enforce (a bit more) the idea that if a math subtree has a meaning attribute, that is the meaning of the whole subtree.  Then, we take the view that when we are constructing an XMArray that we believe is matrix, rather than misguidedly giving the array a meaning attribute, we apply "meaning" to the array as a constructor. 

[This still leaves open any eventual type or property attribute where we may wish to assert that a particular symbol (say X) is a matrix]


brucemiller:
OK, I'll pretend that was reviewed :>

dginev:
Probably best, I was mentally oscillating between giving a thorough set of questions or just a simple OK, I think the simple OK is more reasonable here :+1:

PR #1343, opened 2020-09-24
Separate comments toggle from processing instruction toggle

### LINKED ISSUE #1342: no searchpaths PI when using option --nocomments
I'd like to convert the file `a.tex`:
```
\documentclass{article}
\usepackage{nada}
\begin{document}
test formula: $x+y=z$ %useless comment
\end{document}
```
using
```
latexml --includestyles --nocomments --destination=a.xml a.tex
```
and
```
latexmlpost --mathimages --destination=final.xml a.xml
```
But latexmlpost terminates with the `LaTeX Error: File nada.sty not found.`
The error disappears when I leave out the `--nocomments` option.
I found out that in `Core.pm`, line 207 the insertion of the searchpaths PI depends on the value of `INCLUDE_COMMENTS`, and I wonder why this is so. Or is there another way to convert my document to XML with math images but without tex comments?

dginev:
Fixes #1342 .

Mainly detaches using `includecomments` of LaTeXML::Core to be used only for comments, introducing a new internal option `includexmlpis` to handle processing instructions.

Importantly, adapted the test cases to add this new option, and (more hackily) had the daemon API use the explicit xslt parameter to figure out this is a test environment. Simplest refactor that avoid introducing any user-facing new constructs that came to mind.

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -192,7 +192,7 @@ sub convert {
   if ($$opts{whatsout} =~ /^archive/) {
     $$opts{archive_sitedirectory} = $$opts{sitedirectory};
     $$opts{archive_destination}   = $$opts{destination};
-    my $destination_name = $$opts{destination} ? pathname_name($$opts{destination}) : 'document';
+    my $destination_name  = $$opts{destination} ? pathname_name($$opts{destination}) : 'document';
     my $sandbox_directory = File::Temp->newdir(TMPDIR => 1);
     my $extension         = $$opts{format};
     $extension =~ s/\d+$//;
@@ -300,7 +300,7 @@ sub convert {
 
     # Close and restore STDERR to original condition.
     my $log = $self->flush_log;
-    $serialized = $dom if ($$opts{format} eq 'dom');
+    $serialized = $dom           if ($$opts{format} eq 'dom');
     $serialized = $dom->toString if ($dom && (!defined $serialized));
     # Using the Core::Document::serialize_aux, so need an explicit encode into bytes
     $serialized = Encode::encode('UTF-8', $serialized) if $serialized;
@@ -462,8 +462,8 @@ sub convert_post {
       push(@procs, LaTeXML::Post::CrossRef->new(
           db        => $DB, urlstyle => $$opts{urlstyle},
           extension => $$opts{extension},
-          ($$opts{numbersections} ? (number_sections => 1) : ()),
-          ($$opts{navtoc} ? (navigation_toc => $$opts{navtoc}) : ()),
+          ($$opts{numbersections} ? (number_sections => 1)              : ()),
+          ($$opts{navtoc}         ? (navigation_toc  => $$opts{navtoc}) : ()),
           %PostOPS)); }
     if ($$opts{picimages}) {
       require LaTeXML::Post::PictureImages;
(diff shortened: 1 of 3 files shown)


brucemiller:
Looks reasonable; Thanks!

brucemiller:
Whoops. I merged it before I realized that there's no accommodation for this in `tools/maketest`, which means my test maintenance tool no longer works right.  Alas, a quick check is that it is running purely through command-line interface, and your clever autodetect doesn't.

dginev:
Maybe a sign we need a couple of new regression tests to keep the tools running? It didn't even occur to me those could break!

brucemiller:
So, you want a test case that tests whether the test case maker works?  Oh, my brain!

dginev:
Although, btw, the tool that regenerates the tests ought to be running the _exact same code_ we run for the real test runs. Or we risk other subtle weirdness down the line.

brucemiller:
weird; my edit ended up on your comment!  Asynchronicity.

brucemiller:
Well, the tool that generates the tests runs a more command-line based version, which is what users would use, so it's more likely to compare to that expectation. The tests run by `make test` are using a somewhat ad-hoc-ed version so they're able to paper over things like searchpath PI's which a regular user can't do.  So, if one of them is "wrong"...

brucemiller:
Incidentally, a clarification: your `includexmlpis` is misnamed, since it only excludes searchpath PI's, not all PI's.

dginev:
How should we proceed? Would you like me to make a new PR with those changes?

brucemiller:
I'm torn. I hate to add a new command-line switch. OTOH, it should be possible to create xml files compatible with the test cases from the command line (very useful while doing experimental development). A sneaky switch through latexml.sty would work, but it seems like a bad idea to involve that binding in *all* test cases.  And all that leads me back to a command-line switch.

But *IF* there's to be a command-line switch it should probably be `--for-test-cases` (or some more palatable version of that), that sets all the relevant internals.  The bit that only blocks the searchpaths PI is kinda perverse, in hindsight.

dginev:
But _why_? This PR actually gets the main job done - if someone is going to create files that are test-case compatible, they might as well go ahead and use `Test::Util` directly, or even more directly pass in the option to `LaTeXML::Core` -- they are already assumed to use perl. The command-line switches are supposed to be user-facing, rather than adding dev conveniences in my mind...

brucemiller:
Well, I tried that and got
```
=> Util::Test t/complex/si
bash: Util::Test: command not found...
```
so you must mean something else.

So, if you want to update tools/maketest to use
it, that's fine so long as it works with both the
standard and daemon style tests, and works to
compare & regenerate both the xml (or html) and pdf.

dginev:
Alright, noted, I'll make a new PR updating those. And indeed, I was trying to save myself some typing, I meant `LaTeXML::Util::Test` at `./lib/LaTeXML/Util/Test.pm`

brucemiller:
Still got `command not found` :>

brucemiller:
Honestly, I would (slightly) prefer that test files could be compatible with command-line, but the big issue is that `make test` and `tools/maketests` be compatible.  Whichever one is the easiest to tweak is fine with me.

brucemiller:
random other pseudo-suggestions: have tests ignore searchpath PI's; do something about relativizing search paths; ...

PR #1344, opened 2020-09-28
semantics preserving siunitx

brucemiller:
Rewrite siunitx to preserve the intended semantics as well as desired presentation.

Note that there are still quite a few formatting options (especially regarding rounding and such) that are not yet supported.


lib/LaTeXML/Package/siunitx.sty.ltxml
```diff
@@ -105,14 +132,40 @@ AtBeginDocument('\lx@six@initialize');
...
 DefMathI('\SIUnitSymbolAngstrom',  undef, UTF(0xC5));
 DefMathI('\SIUnitSymbolMicro',     undef, UTF(0xB5));
 
+# These need less left space.
+# Should these be multiplied "ID"s, or postfix?
+DefConstructorI('\SIUnitSymbolDegree', undef,
+  "<ltx:XMTok meaning='arcminute' role='ID' lpadding='0pt'>" . UTF(0xB0) . "</ltx:XMTok>");
+DefConstructorI('\SIUnitSymbolArcminute', undef,
+  "<ltx:XMTok meaning='arcminute' role='ID' lpadding='0pt'>\x{2032}</ltx:XMTok>");
+DefConstructorI('\SIUnitSymbolArcsecond', undef,
```
dginev, line 147: Should these have both `DefMathI` and `DefConstructorI`?

brucemiller: Oh, not really; I didn't finish figuring out if `DefMath` should accept `lspace` (and friends) as options.

t/complex/si.xml
```diff
@@ -162,57 +208,63 @@
...
-          </Math><break/><Math mode="inline" tex="2\pi" text="2 * pi" xml:id="S1.SS1.p1.m6">
+          </Math><break/><Math mode="inline" tex="2\pi" text="2\pi" xml:id="S1.SS1.p1.m6">
             <XMath>
-              <XMApp>
-                <XMTok meaning="times" role="MULOP">⁢</XMTok>
-                <XMTok meaning="2" role="NUMBER">2</XMTok>
-                <XMTok font="italic" name="pi" role="UNKNOWN">π</XMTok>
-              </XMApp>
+              <XMDual>
+                <XMTok meaning="2\pi" role="NUMBER"/>
```
dginev, line 254: A latex macro inside a meaning attribute? That... feels off somehow?

dginev: I'm tempted to say `2\pi` is always an expression, rather than a number, unless it is some sort of strange embellished symbolic constant (which I don't think is the case here).

lib/LaTeXML/Package/siunitx.sty.ltxml
brucemiller: And, on top of that, I had misdefined them when I redefined them.  I added lpadding, rpadding to `DefMath`.

dginev: Nice!

lib/LaTeXML/Package/siunitx.sty.ltxml
```diff
@@ -402,156 +516,235 @@ sub six_groupdigits {
...
+    else { push(@tokens, six_format_compoundnumber($number, $bracket)); } }
+  else {
+    push(@tokens, six_format_simplenumber($number, $bracket)); }
+  return Tokens(@tokens); }
+
+# Return the plain-text string for a number, for use in meaning attribute
+# Note that the format is pretty ad-hoc, except for simple-numbers
+# and scientific-notation, where the mantissa is a simple number.
+# You'll generally not want to use this for "meaning", except in those cases
+sub six_number_string {
```
dginev, line 663: I wasn't going to say anything for a while, but `six_number` really got to me - same point as in physics.sty applies, we could always make a `Package::SIUnitx` namespace to host these methods. But if you'd really prefer a prefix, maybe something less confusing than `six`, such as `sx`?

brucemiller: Yeah, this one's much worse than physics. I'd thought about experimenting with package namespaces, but put it off till I got the basic rewrite done.

dginev:
Not to start any international trouble, but I went out of my way and double-checked the SI units as prescribed by NIST [NIST SP 330-2019](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.330-2019.pdf). The Foreword has the following clarification:

> Like its 2008 predecessor, the 2019 edition of NIST SP 330 conforms with the English text in the BIPM SI Brochure but contains a few minor differences to reflect the most recent interpretation of the SI for the United States by the Secretary of Commerce, as published in the Federal Register. These differences include the following:  
> * The spelling of English words is in accordance with the United StatesGovernment Printing Office Style Manual, which follows Webster's Third NewInternational Dictionary rather than the Oxford Dictionary. Thus the spellings“meter,” “liter,” “deka,” and “cesium” are used rather than “metre,” “litre,”“deca,” and “caesium” as in the original BIPM English text.
> * The name of the unit with symbol t and defined according to 1 t = 103 kg is called“metric ton” rather than “tonne.”•Since the preferred unit symbol for the liter in the United States is L, only L isgiven as the symbol for the liter.
> * A number of “Editors’ notes” are added in order to indicate such differenceswhere significant (except spelling differences) and to clarify the text; and
> * A few very minor editorial changes are made in order to “Americanize” somephrases

Since siunitx.sty seems to provide both a `\metre` and a `\meter` macro, I am unsure if we want to support both spellings as a `meaning` attribute, but - give the affiliation of latexml itself - I would be tempted to suggest adopting the NIST-endorsed spellings?

dginev:
Not to start any international trouble, but I went out of my way and double-checked the SI units as prescribed by NIST [NIST SP 330-2019](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.330-2019.pdf). The Foreword has the following clarification:

> Like its 2008 predecessor, the 2019 edition of NIST SP 330 conforms with the English text in the BIPM SI Brochure but contains a few minor differences to reflect the most recent interpretation of the SI for the United States by the Secretary of Commerce, as published in the Federal Register. These differences include the following:  
> * The spelling of English words is in accordance with the United States Government Printing Office Style Manual, which follows Webster's Third New International Dictionary rather than the Oxford Dictionary. Thus the spellings“meter,” “liter,” “deka,” and “cesium” are used rather than “metre,” “litre,”“deca,” and “caesium” as in the original BIPM English text.
> * The name of the unit with symbol t and defined according to 1 t = 103 kg is called“metric ton” rather than “tonne.”•Since the preferred unit symbol for the liter in the United States is L, only L isgiven as the symbol for the liter.
> * A number of “Editors’ notes” are added in order to indicate such differences where significant (except spelling differences) and to clarify the text; and
> * A few very minor editorial changes are made in order to “Americanize” somephrases

Since siunitx.sty seems to provide both a `\metre` and a `\meter` macro, I am unsure if we want to support both spellings as a `meaning` attribute, but - give the affiliation of latexml itself - I would be tempted to suggest adopting the NIST-endorsed spellings?

t/complex/si.xml
brucemiller: Yeah, it's perverse. By default, siunitx treats both pi and ldots *essentially* as digits (and you can add more fake digits!).  Probably the right solution would be to detect this as (yet another) special case.

brucemiller:
Yeah, I was bothered by those.  It's either good or bad that I can take those words directly from siunitx.sty (well, a copied bit).  Perhaps we'll want to either edit that block, or provide overrides.

lib/LaTeXML/Package/siunitx.sty.ltxml
brucemiller: You can put si into its own package, but can't really import TeX (pool or package) without a lot of exports, right?  That sounds like a good idea, but probably best when factoring the pools, which should be deferred till next release.

brucemiller:
You make a good, tempting, point; thanks for the link. But as far as the spellings are concerned, it's primarily a writing style guide for US authors (maybe even only Feds). It's derived from the UK version and defers to the French text as the "only official text".

So, arguably, keeping the UK spellings (or even better the French?) would be in keeping with the general approach to semantic annotation we're aiming for in LaTeXML. Namely, you write the text & math in whatever style you like, possibly conventions of your field or whatever, but the semantics (ideally) is as close to a universal one as we can get. The semantics of documents in different languages, fields, cultures can then be reasonably compared.

dginev:
I'm a bit lost on the "respecting different languages, fields and cultures" point when it comes to SI units. The way I see it, the moment we have an international standard as an intended "CD" for symbols produced by latexml, we have a certain obligation to ensure latexml will always emit the same markup for the same SI unit. Which spelling to choose is likely something that warrants a discussion, but I would feel a lot better if we emit a single "SI meter" csymbol, whatever its exact spelling, at any point latexml is aware it is emitting an "SI meter", be it in a document in English, French, Russian, Chinese etc.

dginev:
(on second reading) sounds like we're mostly in agreement. I don't know if Content MathML ever came up with an aliasing mechanism, but if it did, "metre" vs "meter" would be just the thing to use aliases for. Assuming no aliasing infrastructure exists, we'll have to make a (arbitrary) choice, and I don't think anyone would blame us for using NIST's guidelines for the spellings in a NIST piece of software.

brucemiller:
On 10/4/20 10:51 AM, Deyan Ginev wrote:
> I'm a bit lost on the "respecting different languages, fields and 
> cultures" point when it comes to SI units. The way I see it, the moment 
> we have an international standard as an intended "CD" for symbols 
> produced by latexml, 

There already is an "international standard", we don't have to
create a new one. There already is a CD for such symbols,
although it is experimental.  I don't think you're suggesting
we create a new one specific to latexml?

I haven't yet compared our symbols to OpenMath's, but the former
presumably could be directly mapped to the latter.


 >we have a certain obligation to ensure latexml will
> always emit the same markup for the same SI unit. Which spelling to 
> choose is likely something that warrants a discussion, but I would feel 
> a lot better if we emit a single "SI meter" csymbol, whatever its exact 
> spelling, at any point latexml is aware it is emitting an "SI meter", be 
> it in a document in English, French, Russian, Chinese etc.

Right. And all the better if it's compatible even with documents *not* 
created with latexml.

dginev:
Btw the [w3 experimental note on units](https://www.w3.org/TR/mathml-units/) just goes axiomatic on the issue:

> With the myriad of spellings available for each unit name, even within SI, it is recommended that the American English spelling with standard translations to lower case ASCII characters be used wherever possible.

brucemiller:
Oh, my. Well, that's another plausible argument for changing the 
spelling.  But it certainly confuses the issue if/when we would want to 
attach to OM cd's.

brucemiller:
So, is it only meter vs metre that bugs you?  One could imagine determining and assigning Official names to every recognized unit & macro in siunitx. Given that we were't able to read the sty file as-is (expli3), we at least were able to crib the unit definitions.  Having to maintain a parallel list could be painful, over time. OTOH, since siunitx defines `\meter` in terms of `\metre`, we could plausibly swap them (with comments).   That is assuming leaving it metre isn't pleasant.

So; several options. Which do you prefer?

dginev:
I'm definitely up for whichever gets the PR merged before the release, since I quite like adding the functionality! The real test of the names will be when they start getting used in content-consuming applications.

I've somehow gotten very fond of the simplicity of deferring to NIST's conventions, going with the American naming consistently, and calling it a day. It's not as if the names mean that much without a proper Content Dictionary :> 

I am happy to also go on the record here that I have some notable skepticism about the adequacy of relying on the existing OpenMath CDs, which we've discussed briefly in a couple of calls. To name them, [SI_BaseUnits1](https://www.openmath.org/cd/SI_BaseUnits1.html) and [units_metric1](https://www.openmath.org/cd/units_metric1.html). All mentions in them, and in general in the [OM symbols list](https://www.openmath.org/symbols/) point to `metre` instead of `meter`. My skepticism is rooted in my newly evolving understanding of anchoring CDs into primary textual resources, so I would like a CD that is very tightly following the natural text of [NIST's SI units](https://www.nist.gov/pml/weights-and-measures/metric-si/si-units) directives, with some primary references. So the idea of a "dimensionless one" as part of the `SI_BaseUnits1` CD is a bit frivolous, if the CD is to be well-encapsulated.

From this perspective, it's even safe to say that until we settle on a final CD for units, my questions can be pushed back to 0.8.6 or later. Happy each way.

brucemiller:
Just to see how it looked, I tried out the change, and it felt too nationalistic.  Google "meter vs metre", and it's clear that US is the odd man out in this particular case, including such trivia as the difference between a "micrometer" and a "micrometre". So metre is probably the more normal choice, irrespective of any particular CD's.

dginev:
In the end of the day it is a phonetic detail either way, my main concern is consistency with primary sources, as mentioned above. Different agencies have different conventions, it just seemed obvious to follow NIST in cases where it has one.

As to the US being the odd one out, it appears more complicated, to [quote from wiki](https://en.wikipedia.org/wiki/Metre#Spelling):
> Metre is the standard spelling of the metric unit for length in nearly all English-speaking nations except the United States[2][3][4][5] and the Philippines,[6] which use meter. Other Germanic languages, such as German, Dutch, and the Scandinavian languages[7] likewise spell the word meter.
> Measuring devices (such as ammeter, speedometer) are spelled "-meter" in all variants of English.[8] The suffix "-meter" has the same Greek origin as the unit of length.

PR #1345, opened 2020-09-29
use LaTeXML::Util::Test in maketests, better pathpi option name

dginev:
Further touchups on #1343 
 * renamed option to `includepathpis`
 * switched to using `LaTeXML::Util::Test` in `tools/maketests` for core test generation
 * separated out the conversion and serialization subroutines in Test, to reuse the exact same code paths for generation and testing
 * dropped the ability to control `--verbosity` in tools/maketests -- can reintroduce if considered important (I've never used it)

### DIFF
--- a/tools/maketests
+++ b/tools/maketests
@@ -237,6 +237,11 @@ sub stringifyDom {
 
 sub gen_xml {
   my ($dir, $name, $is_test) = @_;
+  # Ensure a sandbox for the LaTeXML conversion by forking and working in the child.
+  my $pid = fork();
+  if ($pid) {
+    waitpid($pid, 0);
+    return; }
   my $source = pathname_make(dir => $dir, name => $name, type => 'tex');
   my $xml    = pathname_make(dir => $dir, name => $name, type => $is_test ? 'test.xml' : 'xml');
   if (-f "$xml") {
@@ -253,7 +258,7 @@ sub gen_xml {
     close $test_fh; }
   else {
     die "Failed to convert $source: $!"; }
-  return; }
+  exit; }
 
 sub gen_pdf {
   my ($dir, $name) = @_;
--- a/tools/maketests
+++ b/tools/maketests
@@ -18,6 +18,7 @@ use Getopt::Long qw(:config no_ignore_case);
 use Pod::Usage;
 use LaTeXML;
 use LaTeXML::Util::Pathname;
+use LaTeXML::Util::Test;
 use File::Copy;
 use File::Which;
 use File::Spec::Functions;
@@ -82,6 +83,7 @@ sub maketest_latexml {
   my $src_time = pathname_timestamp($source);
   my $xml_time = pathname_timestamp($xml_file);
   my $pdf_time = pathname_timestamp($pdf_file);
+
   if ($docompare) {
     (compare_xml($dir, $name) ? push(@FAIL, $name) : push(@PASS, $name)); }
   else {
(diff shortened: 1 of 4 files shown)


brucemiller:
Hmm... proof of concept? (doesn't actually work)

There's confusion about which xml files are being written/read (with compare, it overwrites the test case and then can't find the xml to compare!)  Fixable, but...

There's also the problem with verbosity, which is set to -2 for `make test`, to keep it very quiet, but you'd want at least 0 for maketests, since the whole point is "Why is this test failing".   And when fiddling that, you realize that we're now running the same latexml object over several files (which `make test` does, but `maketests` also can (useful sometimes)), so most are causing a multitude of "subroutine foo redefined" warnings.

dginev:
I haven't used `--compare` before, and as I mentioned neither have I used `--verbosity`, so I didn't test either. The only bit I tested is whether the basic regeneration works (it does), and removed any `system()` calls to the latexml executable.

Maybe you would be willing to provide a list of invocations you regularly use, then I can both test for them and make some new unit tests that ensure they keep running in the future?

dginev:
To be explicit: I have never used `maketests` to answer "Why is this test failing", I have only ever used it to regenerate a `.xml` test file. So it may be quicker to fix the issues you're already aware of in your use case if you are fine with the refactor direction, otherwise we'll need to do a few more iterations.

dginev:
Alright, updated with the fixes you identified, thanks! Commands I tried:
```
 make; rm t/expansion/noexpand.xml ; perl tools/maketests --quiet t/expansion/noexpand.tex 
 make; rm t/expansion/noexpand.xml ; perl tools/maketests --verbose --verbose t/expansion/noexpand.tex 
 make ; perl tools/maketests --verbose --verbose --compare t/expansion/noexpand.tex
```

brucemiller:
If you're adding a new test, then presumably you're convinced that your bindings are working correctly, so `tools/maketests --xml --pdf` is handy.

But in the case where tests fail: Since `make test` intentionally only shows the 1st diff, `tools/maketests --comp` gets used the most. It shows the normal latexml output and all the diffs, so you can better establish what's going wrong (or right).  Sometimes portions of the diff are more correct than previous, but others less so. Thus there's often iteration between coding and running the comparison before finally regenerating the xml (or even extending the testcase itself).

Occasionally, there are several tests failing for related reasons in which case it is handy to run multiple comparisons at once.  In those cases the new behaviour revealing all the redefined warnings is pretty annoying (if you add some `--verb`s; but possibly they're not so necessary in that scenario)

dginev:
Thanks for the explanation, though the concrete invocations are also quite a big help. I reproduced the redefined subroutine warning you mentioned, and added a sandboxing via `fork()` that ensures each conversion is standalone, in the spirit of the original system calls.

Command I tested with:
```
make; tools/maketests --comp t/expansion/*
```

tools/maketests
```diff
@@ -237,18 +236,29 @@ sub stringifyDom {
   return [grep { /\S/ } split("\n", $string)]; }
 
 sub gen_xml {
-  my ($dir, $name) = @_;
+  my ($dir, $name, $is_test) = @_;
```
brucemiller, line 239: `$is_test` doesn't really distinguish the cases, perhaps `$is_compare` or  `$for_comparison` or something like that?

tools/maketests
```diff
@@ -237,18 +236,29 @@ sub stringifyDom {
...
     rename("$xml", "$xml.bak"); }
   #========================================
   # Generate the XML from the current latexml
-  system("$FindBin::RealBin/../blib/script/latexml", $source,
-    "--destination=" . $xml,
-    "--nocomments", (map { "--verbose" } 1 .. $verbosity)) == 0
-    or die "Failed to convert $source: $!";
-  return; }
+  my $gen_xml_core_options = {%LaTeXML::Util::Test::CORE_OPTIONS_FOR_TESTS};
+  $$gen_xml_core_options{verbosity} = $verbosity || 0;
```
brucemiller, line 252: Ah, thanks for that! (I would have asked for it :> )

tools/maketests
dginev: Sure thing, renamed :+1:

tkw1536:
+1 for using `LaTeXML::Util::Test` in `maketest`. I had some minor issues with not using that some time ago.

PR #1353, opened 2020-10-09
Fixes #1352

### LINKED ISSUE #1352: Use of \DeclareCaptionOption{parskip} leaves closing bracket behind
The `\DeclaraCaptionOption` command below produces an unexpected closing bracket (`]`) at the beginning of the document.
```
\documentclass{article}
\RequirePackage{caption}
\DeclareCaptionOption{parskip}[]{}% Workaround for conflict between captions, subfig, and parskip.
\begin{document}
Document begins here.
\end{document}
```

(Since we are currently avoiding _subfig.sty_ due to #1292, I'll try to work around this for now by just removing the problematic `\DeclareCaptionOption` call.)

dginev:
Just adding the optional argument, verified that pdflatex accepts it just fine.


brucemiller:
Cool! Thanks!

PR #1354, opened 2020-10-12
adding lexeme attribute to ltx:Math schema

dginev:
Today I noticed the (still experimental, per PR #947 ) lexemes attribute is naturally considered illegal in validation, so I thought I might as well add it to the schema. Feedback welcome on what would be the best way to document it.


brucemiller:
Not sure how to document it, either. Actually, a little unsure whether we've nailed the definition :>
But still, we should allow it!

PR #1355, opened 2020-10-14
Use meaning over role in MathML::lookupContent

dginev:
Consider the first token in the following XMath (generated by a semantic macro `\fnderive{f}{n}`):
```xml
  <XMath>
    <XMApp>
      <XMTok role="SUPERSCRIPTOP" scriptpos="post1" meaning="derivative-implicit-variable" name="fnderive@build"/>
      <XMTok role="UNKNOWN" xml:id="p1.m1.1" font="italic" fragid="p1.m1.1">f</XMTok>
      <XMDual xml:id="p1.m1.2" fragid="p1.m1.2">
        <XMRef idref="p1.m1.2.1"/>
        <XMWrap>
          <XMTok role="OPEN" stretchy="false" fontsize="70%">(</XMTok>
          <XMTok role="UNKNOWN" xml:id="p1.m1.2.1" font="italic" fontsize="70%" fragid="p1.m1.2.1">n</XMTok>
          <XMTok role="CLOSE" stretchy="false" fontsize="70%">)</XMTok>
        </XMWrap>
      </XMDual>
    </XMApp>
  </XMath>
```

The first token is the main operator, with role `SUPERSCRIPTOP` and a macro-provided meaning `derivative-implicit-variable`. I claim the expected behavior when generating Content MathML is currently wrong in master for such a case, as the handler called is the one for `Token:SUPERSCRIPTOP:?` rather than the generic `Token:?:?`.

I propose we correct to always use a provided `meaning` over `role`, also when doing the `DefMathML` table lookup. 

You can see the code diff of my reorganization to achieve that, the main improvement being _not_ defaulting to the `?` literal when a `$name` or `$role` haven't been provided, but using them undefined in some conditional checks, while adding comments about the intended fallback order.

Edit: here is a whitespace-neutral diff link:
https://github.com/brucemiller/LaTeXML/pull/1355/files?w=1

The example improves from master's:
```xml
    <m:apply>
      <m:csymbol cd="ambiguous">superscript</m:csymbol>
      <m:ci>𝑓</m:ci>
      <m:ci>𝑛</m:ci>
    </m:apply>
```

to the intended cmml:
```xml
    <m:apply>
      <m:csymbol cd="latexml">derivative-implicit-variable</m:csymbol>
      <m:ci>𝑓</m:ci>
      <m:ci>𝑛</m:ci>
    </m:apply>
```

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -24,7 +24,7 @@ our @EXPORT = (
     &pmml_mi &pmml_mo &pmml_mn &pmml_bigop
     &pmml_punctuate &pmml_parenthesize
     &pmml_infix &pmml_script &pmml_summation),
-  qw( &cmml &cmml_share &cmml_shared &cmml_ci
+  qw( &cmml &cmml_share &cmml_shared &cmml_leaf
     &cmml_or_compose &cmml_synth_not &cmml_synth_complement),
 );
 require LaTeXML::Post::MathML::Presentation;
@@ -216,7 +216,7 @@ sub lookupContent {
       # 3. If no special code, but we have a name, use a generic handler for this element
       $$MMLTable_C{"$mode:?:?"}) : (
     # 4. If we do not have a name, check for a role-based handler
-    $$MMLTable_C{"$mode:$role:?"} ||
+    ($role && $$MMLTable_C{"$mode:$role:?"}) ||
       # 5. Always use a default handler if nothing is known
       $$MMLTable_C{"$mode:?:?"}); }
 
@@ -1244,11 +1244,16 @@ sub cmml_unparsed {
     @results]; }
 
 # Or csymbol if there's some kind of "defining" attribute?
-sub cmml_ci {
+sub cmml_leaf {
   my ($item) = @_;
   if (my $meaning = (ref $item) && $item->getAttribute('meaning')) {
-    my $cd = $item->getAttribute('omcd') || 'latexml';
-    return ['m:csymbol', { cd => $cd }, $meaning]; }
+    if (my $cd = $item->getAttribute('omcd')) {
+      return ['m:csymbol', { cd => $cd }, $meaning]; }
+    elsif (($item->getAttribute('role') || '') eq 'NUMBER') {
+      # special case, numbers with a meaning attribute
+      return ['m:cn', { type => ($meaning =~ /^[+-]?\d+$/ ? 'integer' : 'float') }, $meaning]; }
+    else {
+      return ['m:csymbol', { cd => 'latexml' }, $meaning]; } }
   else {
     my ($content, %mmlattr) = stylizeContent($item, 'm:ci');
     if (my $mv = $mmlattr{mathvariant}) {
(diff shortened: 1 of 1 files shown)


dginev:
You may have guessed the further implication that it was actually impossible until now to provide a semantic annotation for any sub/super-script XMTok and have that annotation transported over to Content MathML. Your only option was a dual.

dginev:
Interestingly, the `Token:NUMBER:?` handler was written to expect a number _without_ a meaning attribute (it uses its textContent directly), but since numbers currently do indeed have a meaning attribute I had to handle them in my generic rerouting. 

That actually alerted me to `cmml_ci` being a bit clunkily named. It creates both `ci` and `csymbol`, and my last commit also added a simple `cn` case, so I just renamed it to `cmml_leaf`, which seems straightforward enough. 

Tests should pass now, though we likely need to discuss either way.

brucemiller:
Had to read your new lookup clause several times before I realized that it avoided looking up `$mode:$role:?` entirely. Which is why you had to handle numbers in `cmml_ci`, and felt even more compelled to rename it. Also, the number handler didn't "expect" a number without meaning, it just didn't require it.

I think when I originally wrote this module, I must have expected a lot more code sharing between pmml & cmml. Instead, pmml is driven almost entirely by `role` and cmml by `meaning`. In fact, the only cmml handlers affected by your change (I think) are the rather silly sub/superscripts cases: If we want to call it `ambiguous:superscript`, fine, but this is almost certainly the wrong place to do it.

Given that structure, and that we'd like to soon extend to do pmml, pragmatic & strict cmml's, we'll likely want to separate out the `DefMathML` into 2 (or 3) separate definers.  So, I'd rather not make the logic even more complex at this stage.

So clearly there's a bug, which you're fixing, but maybe the best, simplest, way to do it is have the sub/superscript handlers check for a meaning?

dginev:
> So clearly there's a bug, which you're fixing, but maybe the best, simplest, way to do it is have the sub/superscript handlers check for a meaning?

I feared you'd suggest that... I didn't anticipate your preferred way forward is to abandon the unified handling in `DefMathML` entirely, and to refactor all of the definitions into multiple calls - that would be another Large refactor in the refactor queue.

As to the logic being more complicated, that's a bit suspicious a claim, it's just the same conditional with its arms reordered for the content case, with case-by-case comments to clarify the dispatch. It's also intuitive to a very large degree, "it had meaning X, so it was mapped to a csymbol X" is a very obvious default fallback.

But sure, fixing the individual problem in the script handlers is very direct, will make a new PR.

brucemiller:
initially not all that major a refactor: just split `DefMathML` into `DefPMathML` & `DefCMathML`, and loose the dummy undef arg 99% of the time. What may (or may not) be more major is figuring out how to handle strict. May not need handlers at all, since it's just apply all-the-way-down. The main issue is `ci` vs `csymbol`. Well, except for binding. Ugh.

brucemiller:
On further reflection, it seems this approach has a more "strict cmml" pov, whereas the other sticks closer to a "pragmatic" one.  Meanwhile, it exposes some limitations/mistakes in the pragmatic code, which may or may not be worth tackling at this time.  In fact, it probably further motivates a move to make strict the primary cmml target, or perhaps even the only cmml target in a future release.  So, let's go with this approach and sort out the rest later.

dginev:
Very nice! Thanks for merging!

PR #1356, opened 2020-10-26
Stop using appveyor for Windows CI

dginev:
Removing this file, and will disable the flags in the appveyor site. Their new 10 minute limit is just too problematic for reliable CI, and I don't feel confident I can work around it.

I am attempting a circleci replacement, but we should remove the broken one already I think, no point keeping it around.


brucemiller:
sounds reasonable. Thanks!

dginev:
Quick update on the Windows front: Deyan CI can confirm the latest master tests successfully on Windows 10 with miktex on my local desktop machine.

I can in general confirm that circleci will be usable, on at least a weekly basis, to provide free Windows builds. But I am again "out of credits" before I managed to silence all miktex warnings with the "administrator rights" debacle one finds themselves in when executing things from the command line. Will get a windows CI back eventually, just not yet :>

PR #1358, opened 2020-10-26
add \meaning handling for \font-defined primitives

dginev:
Fixes the texlive 2020 regression with expl3, while also enhancing further our `\meaning` emulation.

I double-coded [the previous PR](https://github.com/dginev/LaTeXML/pull/6) to avoid rebasing, since the diff was so small it was simpler to copy it directly over the new master.


dginev:
I reran tests on a local texlive 2020, happy to report they are still passing. We should consider adding a Travis target for texlive 2020 too, maybe in place of the 2015 or 2016 target.

brucemiller:
Looks plausible. Thanks!

PR #1359, opened 2020-10-27
Prioritize meaning on script tokens for Content MathML

dginev:
Alternative PR to #1355

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -63,8 +63,8 @@ sub preprocess {
   my ($self, $doc, @nodes) = @_;
   # Set up rational, modern, defaults.
   $$self{hackplane1} = 0 unless $$self{hackplane1};
-  $$self{plane1} = 1 if $$self{hackplane1} || !defined $$self{plane1};
-  $$self{nestmath} = 0 unless $$self{nestmath};
+  $$self{plane1}     = 1 if $$self{hackplane1} || !defined $$self{plane1};
+  $$self{nestmath}   = 0 unless $$self{nestmath};
   $doc->adjust_latexml_doctype('MathML');    # Add MathML if LaTeXML dtd.
   $doc->addNamespace($mmlURI, 'm');
   return; }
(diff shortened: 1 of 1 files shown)


brucemiller:
In hindsight, I'm not as sure which approach makes future enhancements easiest; I suspect this one.

However, looking at the motivation for the other PR, and recalling that we're mostly claiming "pragmatic" cmml, I should point out that the example actually should be encoded as
```
<apply><diff/><whateverfunction/></apply>
```
even when not knowing the w.r.t variable (see https://www.w3.org/Math/draft-spec/chapter4.html#id.4.4.4).
This would require a `Apply:?:derivative-main` rule for cmml.

dginev:
Happy to repeat that I like my first PR better, as far as my preference goes :+1: As to having a special rule for the pragmatic diff, that's reasonable. But for now I am just capturing some custom symbols that come from the "accessibility CD", so the motivation for the PR was not really pragmatic CMML, but custom csymbols.

brucemiller:
Then I'm confused. The concept you used as an example should
generate `<diff/>` in our current implementation (whether by an
Apply or Token rule, either way, I guess), rather than a csymbol.
Other examples may need a csymbol.

Of course, the current example should generate a csymbol
if and when we're generating strict cmml.

dginev:
I'm not disputing adding a rule for generating the pragmatic diff element is a good idea, indeed that's nice to have. But when I opened the pr I thought of the example from the semantics-mini spec as an experimental symbol for accessibility, with a custom name. We've since also discussed other implied meanings of scripts, such as "index", "functional-power", "dimension" and others, so my initial motivation was to have the generic support available.

brucemiller:
But isn't that the PR topic? That when a script token
has a meaning, the incorrect output is generated?
And that the correct output would vary with the meaning:
sometimes diff, or some other special pragmatic element,
sometimes csymbol with a specific cd:name, sometimes default ambiguous.

So, the point is, whichever approach that can handle all these
cases cleanly is fine. But particularly should be considered,
if you're proposing to change the dispatch.

dginev:
I looked at MathML.pm with fresh eyes today and saw it has:
```perl
DefMathML("Token:?:differential",         undef, sub { return ['m:diff']; });
```

So, is there anything this PR should add? If anyone wanted to use the diff symbol itself, they should make sure their script macro (or lxDeclare) uses the `differential` meaning.

brucemiller:
On 10/28/20 9:20 AM, Deyan Ginev wrote:
> I looked at MathML.pm with fresh eyes today and saw it has:
> 
> DefMathML("Token:?:differential",undef,sub  {return  ['m:diff']; });

Interesting. That appears wrong. There doesn't seem to be
an element for differential.

> So, is there anything this PR should add? If anyone wanted to use the 
> diff symbol itself, they should make sure their script macro (or 
> lxDeclare) uses the |differential| meaning.

We're talking about the m:diff element, not the diff "symbol".
As I understood it, you wanted to mark up "f'". See 4.4.4.2.

dginev:
If you'd like to add some commits that adjusts the diff handling to your taste, be my guest. This PR (also #1355 ) was scoped to resolve a generic problem with using the meaning attribute on scripts, so in that regard it is completed and can be immediately used for csymbols, I am sticking to my story. The initial example was `\fnderive{f}{n}`, which could have been `\setdimen{R}{3}`, or `vectorindex{v}{i}`, or ...

brucemiller:
Let's go with your original pr #1355

PR #1360, opened 2020-10-27
More precise binding for pgfNumber and pgfmath@smuggleone

### LINKED ISSUE #1168: [texlive 2020] TikZ "left color" leads to Fatal error
I'm having some trouble with TikZ's shadedraw.  I think I've narrowed it down to the option `left color` causing a fatal division by zero.
```TeX
\documentclass{article}
\usepackage{tikz}
\begin{document}
\tikz{\shadedraw[color=brown](0,0)--++(0,20pt)--++(20pt,0)--++(0,-20pt)--cycle;} % recursion error and 3 undefined macros
\tikz{\shadedraw[{left color}=brown](0,0)--++(0,20pt)--++(20pt,0)--++(0,-20pt)--cycle;} % causes Fatal Illegal division by zero
\end{document}
```

dginev:
Fixes #1168 .

May not be the ideal fix, but I feel confident I debugged the root issues:
 1. The original latex `\pgfmath@smuggleone` simply assumes it has a locally defined expandable macro on input, and re-defines it locally a level up after it closes a group with `\egroup`. 
    - However, the latexml binding handles certain internal constructs by creating Primitives with subroutines as their definition, making the technique of smuggle one equivalent to `\def#1{#1}` i.e. an infinite loop.
    - Solution: added a binding that checks for the non-expandable case, and only closes a group if in it (assuming that definition was already global).

 2. Separately, our parameter type for `pgfNumber` didn't strike me as accurate. Here too I debugged with `\tracingmacros=1` and reading attentively the differences in the numbers. Very majorly, a `90` angle got passed in as a `9`, since the first `readXToken` eagerly expanded the macro that contained that value too far. In the latex sources, these parameters are just regular plain parameters to a `\def` definition, but get expanded with trickery inside their bodies. So I reworked to reflect that a little closer.

 3. Lastly, one of the division by zero fatals could be made more lax as a warning, as we have done for all other divisions with one of my previous PRs. So I also added that guard.

Feel free to improve to a more disciplined fix, if one comes to mind. I can vouch this PR creates good SVG with no_problem status for the tikz from #1168 on texlive 2020.

### DIFF
--- a/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
+++ b/lib/LaTeXML/Package/pgfmath.code.tex.ltxml
@@ -93,29 +93,13 @@ DefMacro('\@@@show@pgfmatharg@{}', sub {
     print STDERR "MATH ARG $arg\n";
     return; });
 
-# A variant on Expanded that omits the outer {}
+# A plain argument that is expanded in the parameter type definition,
+# then used for pgf computations.
 DefParameterType('pgfNumber', sub {
     my ($gullet) = @_;
-    my $token;
-    do { $token = $gullet->readXToken(0);
-    } while (defined $token && $$token[1] == CC_SPACE);    # Inline ->getCatcode!
-
-    my $result = '';
-    if ($token->getCatcode == CC_BEGIN) {
-      my @tokens = ();
-      my $level  = 1;
-      while ($token = $gullet->readXToken(0)) {
-        my $cc = $$token[1];
-        if ($cc == CC_END) {
-          $level--;
-          last unless $level; }
-        elsif ($cc == CC_BEGIN) {
-          $level++; }
-        $result .= $$token[0]; } }
-    else {
-      $result = $token->getString; }
-    $result = "0" if $result eq '.';    # Apparently "." is a valid number!
-    return $result; });
+    my $pgf_number = ToString(Expand($gullet->readArg()));
+    $pgf_number = "0" if $pgf_number eq '.';    # Apparently "." is a valid number!
+    return $pgf_number; });
 
 # This one expects {{number}{number}....} and returns an array of them
 DefParameterType('pgfNumbers', sub {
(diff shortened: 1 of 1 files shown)


dginev:
Curious detail: I double-checked the other tikz issues we have open, and this PR doesn't resolve any of them. So #1168 had a localized problem, which gets a localized solution. Makes sense given the size of tikz, and the size of our bindings for it. I have jumped into debugging the next tikz issue...

lib/LaTeXML/Package/pgfmath.code.tex.ltxml
```diff
@@ -93,29 +93,13 @@ DefMacro('\@@@show@pgfmatharg@{}', sub {
...
-          $level--;
-          last unless $level; }
-        elsif ($cc == CC_BEGIN) {
-          $level++; }
-        $result .= $$token[0]; } }
-    else {
-      $result = $token->getString; }
-    $result = "0" if $result eq '.';    # Apparently "." is a valid number!
-    return $result; });
+    my $pgf_number = ToString(Expand($gullet->readArg()));
```
brucemiller, line 100: Shouldn't this be a `readXArg`?  (which we don't have, but probably should, essentially equivalent to the `Expanded` parameter type)

dginev: Maybe? I am not quite sure. The uses I have seen are akin to:
```tex
\pgfmathcos@\pgfshade@angle
```

And the crucial aspect is that a single command sequence can be first read, then expanded, and its entire expansion is treated as the argument. I am not sure what `readXArg` would be specified to do, but it feels a bit more like `readArgX`, i.e. "first read in a plain, then expand it".

The `Expanded` parameter type is instead doing the TeX thing prescription of expanding from the get go, and taking the first non-expandable token as its argument, irrespective of how many expansion calls it had to do to get there. That's a difference to the pgf style of passing a plain argument around e.g. as in:
```tex
\def\pgfmathfpcos@#1{%
    \begingroup%
        \FPmul\pgfmathresult{#1}{0.017453292519943295}%
        \FPcos\pgfmathresult{\pgfmathresult}%
        \pgfmath@smuggleone\pgfmathresult%
    \endgroup%
}
```
```

brucemiller:
Every little bit helps; Thanks!

PR #1361, opened 2020-10-28
Allow double-loading .ldf files for babel

### LINKED ISSUE #1357: LaTeXML selects wrong language -- missing re-input of TeX sources
I have a document that employs US- and UK-english. In minimal form:
```tex
\documentclass[USenglish,UKenglish]{article}
\usepackage{babel}
\begin{document}
\languagename
\end{document}
```
While LaTeX outputs "UKenglish", LaTeXML writes "USenglish". As far as I can see, this happens because LaTeX inputs english.ldf once for USenglish and once for UKenglish, but LaTeXML omits the second input. I suppose this is because in Package.pm, `loadTeXDefinitions` says: "Don't load if we've already loaded it before." From LaTeX's perspective this is indeed expected for \usepackage and friends but not for \input. Can you put LaTeXML in a position to "re-\input" files?

dginev:
Fixes #1357, as prescribed by the issue reporter.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1875,10 +1875,6 @@ my %definition_name = (    # [CONSTANT]
   'cnf' => 'configuration',        'cfg' => 'configuration',
   'ldf' => 'language definitions', 'def' => 'definitions', 'dfu' => 'definitions');
 
-sub pathname_is_raw {
-  my ($pathname) = @_;
-  return ($pathname =~ /\.(tex|pool|sty|cls|clo|cnf|cfg|ldf|def|dfu)$/); }
-
 my $findfile_options = {    # [CONSTANT]
   type => 1, notex => 1, noltxml => 1, searchpaths_only => 1 };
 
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Package.pm
```diff
@@ -2095,7 +2095,11 @@ sub loadTeXDefinitions {
     # since someone's presumably asking _explicitly_ for the raw TeX version.
     # It's probably even the ltxml version is asking for it!!
     # Of course, now it will be marked and wont get reloaded!
-    return if LookupValue($request . '_loaded');
+    #
+    # babel.sty exception:
+    # we know the same .ldf file may be reloaded with a different option,
+    # to load an adjacently defined language, so allow that.
+    return if $type ne 'ldf' and LookupValue($request . '_loaded');
```
brucemiller, line 2102: such specific special cases can cause lots of pain later on.  Perhaps would be best to keep track from the calls which can be loaded multiple times and which not?  Not sure of the logic yet, though.

brucemiller: or at worst, a new `pathname_is_reloadable`, analogous to `pathname_is_raw`?

dginev: Alright, I did the `at worst` suggestion, since it seemed easy and reasonable. I even moved `pathname_is_raw` and improved the docs a little, for extra cleanliness.

brucemiller:
Thanks for the report & patch!

PR #1364, opened 2020-11-02
Add basic support for newest aastex63

dginev:
Quick & dirty PR aimed at content-preserving, but presentationally imperfect, aastex63.cls arXiv documents.

Took me way too long to wrestle down an error-free version of the `D` column - but I managed! Still, I am making a PR with some remaining visual issues, hoping to borrow wisdom from @brucemiller as to what should be improved before the release and what after. I'd like to get the additions in for the latest arXiv run, since I have seen new papers using aastex63.cls.

As to what I managed to achieve, my main focus was getting the `sample63.tex` official template converting without issues to HTML, [online PDF here](http://tug.ctan.org/tex-archive/macros/latex/contrib/aastex/sample63.pdf).

Here are two side-by-side tables stress-testing the tricky new `h` and `D` column specifiers. HTML on the left, with known visual deficiencies (but all content is present, and conversion is problem-free):
aastex63

I am also attaching the full HTML converted via the PR for reference of the remaining flaws:
[sample63_html.zip](https://github.com/brucemiller/LaTeXML/files/5477514/sample63_html.zip)

### DIFF
--- a/lib/LaTeXML/Package/aas_support.sty.ltxml
+++ b/lib/LaTeXML/Package/aas_support.sty.ltxml
@@ -320,35 +320,28 @@ Let(T_CS('\decimalcolnumbers'), T_CS('\deluxedecimalcolnumbers'));
 
 # http://tug.ctan.org/tex-archive/macros/latex/contrib/aastex/sample63.pdf
 # decimal alignment, 3.1.2
-DefColumnType('D', sub {
-    $LaTeXML::BUILD_TEMPLATE->addColumn(
-      before => Tokens(T_CS('\hfill')),
-      after  => Tokens(T_CS('\end@D@col@one')));
-    $LaTeXML::BUILD_TEMPLATE->addColumn(
-      after => Tokens(T_CS('\hfill')));
-    return; });
-
-DefColumnType('d', sub {    #TODO: these should be "eaten away" I think?
-    $LaTeXML::BUILD_TEMPLATE->addColumn(
-      before => Tokens(T_CS('\hfill')),
-      after  => Tokens(T_CS('\end@D@col@one')));
-    $LaTeXML::BUILD_TEMPLATE->addColumn(
-      after => Tokens(T_CS('\hfill')));
-    return; });
-
-DefMacro('\end@D@col@one', sub {
-    my ($gullet)         = @_;
-    my $smuggle_new      = 1;
-    my $current_location = $gullet->getLocator->toString;
-    if (my $smuggled_at = LookupValue('aas_smuggle_D_column')) {
-      if ($smuggled_at eq $current_location) {
-        $smuggle_new = 0;
-    } }
-    if ($smuggle_new) {
-      AssignValue('aas_smuggle_D_column', $current_location, 'local');
-      return T_CS('\@alignment@align'); }
-    else {
-      return (); } });
+sub build_d_columns {
+  $LaTeXML::BUILD_TEMPLATE->addColumn(
+    before => Tokens(T_CS('\hfill'), T_CS('\aas@start@D@column')),
+    after  => Tokens(T_CS('\aas@end@D@column')));
+  $LaTeXML::BUILD_TEMPLATE->addBetweenColumn();
+  $LaTeXML::BUILD_TEMPLATE->addColumn(
+    after => Tokens(T_CS('\hfill')));
+  return; }
+DefColumnType('D', \&build_d_columns);
+DefColumnType('d', \&build_d_columns);
+DefMacro('\aas@start@D@column XUntil:\aas@end@D@column', sub {
+    my ($gullet, $n) = @_;
+    my ($m,      $f) = SplitTokens($n, T_OTHER('.'));
+    return ($f ? (@$m,
+        T_CS('\@ADDCLASS'), T_BEGIN, T_OTHER('ltx_norightpad'), T_END,
+        T_CS('\@alignment@align'),
+        T_CS('\@ADDCLASS'), T_BEGIN, T_OTHER('ltx_noleftpad'), T_END,
+        T_OTHER('.'), @$f)
+      : ($n, T_CS('\@alignment@align')));
+##  return ($n,T_CS('\@alignment@align'));
+});
+DefPrimitive('\aas@end@D@column', '');
 
 # hidden column, 3.1.4
 Let(T_CS('\savedollar'), T_MATH);
--- a/lib/LaTeXML/Package/aas_support.sty.ltxml
+++ b/lib/LaTeXML/Package/aas_support.sty.ltxml
@@ -324,7 +324,6 @@ sub build_d_columns {
   $LaTeXML::BUILD_TEMPLATE->addColumn(
     before => Tokens(T_CS('\hfill'), T_CS('\aas@start@D@column')),
     after  => Tokens(T_CS('\aas@end@D@column')));
-  $LaTeXML::BUILD_TEMPLATE->addBetweenColumn();
   $LaTeXML::BUILD_TEMPLATE->addColumn(
     after => Tokens(T_CS('\hfill')));
   return; }
@@ -338,8 +337,9 @@ DefMacro('\aas@start@D@column XUntil:\aas@end@D@column', sub {
         T_CS('\@alignment@align'),
         T_CS('\@ADDCLASS'), T_BEGIN, T_OTHER('ltx_noleftpad'), T_END,
         T_OTHER('.'), @$f)
-      : ($n, T_CS('\@alignment@align')));
-##  return ($n,T_CS('\@alignment@align'));
+      : ($n,
+        T_CS('\@ADDCLASS'), T_BEGIN, T_OTHER('ltx_norightpad'), T_END,
+        T_CS('\@alignment@align')));
 });
 DefPrimitive('\aas@end@D@column', '');
 
--- a/lib/LaTeXML/Package/aas_support.sty.ltxml
+++ b/lib/LaTeXML/Package/aas_support.sty.ltxml
@@ -129,9 +129,11 @@ DefConstructor('\altaffiltext{}{}',
 
 DefMacro('\software{}',      '\@add@frontmatter{ltx:note}[role=software]{#1}');
 DefMacro('\submitjournal{}', '\@add@frontmatter{ltx:note}[role=journal]{#1}');
-DefMacro('\doi{}',           '\@add@frontmatter{ltx:classification}[scheme=doi]{#1}');
-DefMacro('\collaboration',   '');
-DefMacro('\nocollaboration', '');
+# Alas \doi is not frontmatter.
+DefConstructor('\doi{}',             '<ltx:ref href="https:/doi.org/#1">#1</ltx:ref>');
+DefConstructor('\@@@collaborator{}', "<ltx:note role='collaborator'>#1</ltx:note>");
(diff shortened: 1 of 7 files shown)


lib/LaTeXML/Package/aas_support.sty.ltxml
```diff
@@ -296,6 +309,68 @@ RequirePackage('deluxetable');
...
+    my ($gullet)         = @_;
+    my $smuggle_new      = 1;
+    my $current_location = $gullet->getLocator->toString;
+    if (my $smuggled_at = LookupValue('aas_smuggle_D_column')) {
+      if ($smuggled_at eq $current_location) {
+        $smuggle_new = 0;
+    } }
+    if ($smuggle_new) {
+      AssignValue('aas_smuggle_D_column', $current_location, 'local');
+      return T_CS('\@alignment@align'); }
```
dginev, line 349: I only arrived at this locator-based solution for adding an extra alignment in the table body "by the skin of my teeth".

It's a very ugly problem, because the `D` spec has a split personality. 
 - Face 1: In the table header, authors are explicitly requested to use a `\multicolumn2c` specifier for the `D` entries, to fill two entries.
 - Face 2: In the table body, authors are expected to just casually deposit their decimal numbers in a single column and the magic of `\lookfordecimal` creates a second alignment behind the scenes.

I did not emulate the decimal alignment yet, but I managed to get the "second align behind the scenes" behavior, using this very awkward locator trick. What happens if this guard isn't added? An infinite loop apparently, since the newly deposited alignment causes a new `D` first column to be created, which drops another new alignment, and so on into infinity.

So the locator assignment makes sure the expansion doesn't create a universe of "align tokens all the way down".

dginev: It would of course be easier if, as in `numprint.sty.ltxml`, the `D` column type only ever created one column. But then I would have to "detect" an extra column in the header and remove its alignment. I guess that's also doable, and may be an alternative avenue... Feedback most welcome.

lib/LaTeXML/Package/aas_support.sty.ltxml
```diff
@@ -296,6 +309,68 @@ RequirePackage('deluxetable');
...
+
+# http://tug.ctan.org/tex-archive/macros/latex/contrib/aastex/sample63.pdf
+# decimal alignment, 3.1.2
+DefColumnType('D', sub {
+    $LaTeXML::BUILD_TEMPLATE->addColumn(
+      before => Tokens(T_CS('\hfill')),
+      after  => Tokens(T_CS('\end@D@col@one')));
+    $LaTeXML::BUILD_TEMPLATE->addColumn(
+      after => Tokens(T_CS('\hfill')));
+    return; });
```
brucemiller, line 329: Nice! This is along the lines of what I was thinking of trying

lib/LaTeXML/Package/aas_support.sty.ltxml
brucemiller: And, yeah; A very scary version of what I knew would be required with "D" creating two columns.  I'll have to ponder...

However,is "split personality" a perverse pun?  It isn't so much itself split: It creates two columns, but has a before/after that causes it to read only up to one "&" but split it!  The `\multicolumn` is just there to cause the column heading to span 2 columns.  Of course the over cleverness of `Alignment` causes (in this example) the extra, empty, column (and the span) to disappear --- this threw me off while testing!

brucemiller: Incidentally: what you've written is in the ballpark of correct. It's only a question of do you *want* it to create 2 columns split on the decimal (and thus hide the real number for posterity), or just ignore the alignment (until html catches up)?

dginev: "ballpark of correct" was hard enough!

Maybe I should've gone with "Two-face" instead of split personality, if you prefer Batman references :> The backend is certainly unified (oh, it's just two columns, snooze, boring), but the *frontend* is dual - the author is expected to *write* two columns in the header, and a single column in the body, even if the single column in the body will eventually "become" two.

And calling `->addColumn` twice in a single `DefColumnType` was something I questioned the entire time, since it feels like a very suspicious technique...

I think I prefer having "two columns" everywhere for now, to stay close to the TeX implementation, and also since I feel it is more predictable to add an alignment token in the body than to remove one in the header (from a \multicolumn command.. yikes.)

brucemiller: By deciding on 2 columns, I meant whether the number would be split into two columns, or just add an empty column.  Actually, a more TeX-like approach is to provde a 'reader' for the begin of the first column, and a marker at it's end (You've got a thing about smuggling, perhaps?)  The following code seems to work, and aligns as desired:
```
DefColumnType('D', sub {
    $LaTeXML::BUILD_TEMPLATE->addColumn(
      before => Tokens(T_CS('\hfill'),T_CS('\aas@start@D@column')),
      after  => Tokens(T_CS('\aas@end@D@column')));
    $LaTeXML::BUILD_TEMPLATE->addBetweenColumn();
    $LaTeXML::BUILD_TEMPLATE->addColumn(
      after => Tokens(T_CS('\hfill')));
    return; });

DefMacro('\aas@start@D@column XUntil:\aas@end@D@column', sub {
  my($gullet,$n)=@_;
  my($m,$f)=SplitTokens($n,T_OTHER('.'));
  return ($f ? (@$m,
                T_CS('\@ADDCLASS'),T_BEGIN,T_OTHER('ltx_norightpad'),T_END,
                T_CS('\@alignment@align'),
                T_CS('\@ADDCLASS'),T_BEGIN,T_OTHER('ltx_noleftpad'),T_END,
                T_OTHER('.'),@$f)
          : ($n,T_CS('\@alignment@align')));
##  return ($n,T_CS('\@alignment@align'));
  });
DefPrimitive('\aas@end@D@column','');
```
The between column bit was an experiment that didn't work, yet. The idea is to suppress intercolumn spacing.  The temporary hack is to add the CSS class, which needs the following added to LaTeXML.css:
```
.ltx_norightpad { padding-right:0!important; }
.ltx_noleftpad { padding-left:0!important; }
```
Maybe that's what the between columns bit should be doing?

dginev: Very nice! I am unsure what the magic of the `addBetweenColumn` does there, maybe understanding that would have saved me some pain :> Definitely not overly fond of smuggling in general, just that it keeps working reliably as a low-level technique in weird TeX circumstances...

dginev: I've commited your changes to the branch since they're a clear upgrade btw.

I am now slowly pondering your code and trying to figure out why it doesn't fall into an infinite loop when it introduces the new `\@alignment@align`, while mine did, hm.

dginev: It's likely the combination of your use of `XUntil` and the timing of `after=>` that avoids the loop. I can't quite wrap my mind around tracing the individual steps, but I'll keep in mind this type of setup results in cleaner code. Thanks a lot for the upgrade!!

dginev:
Latest Table 1 screenshot, looking much better:
image

dginev:
I should also report some part of the postprocessing (the XSLT?) appears to be wrongly handling multiple dates with special roles. Here is a good-looking XML from the core processing that shows the issue:
```xml
  <resource src="ltx-ulem.css" type="text/css"/>
  <title>Dates</title>
  <date role="received">June 1, 2019</date>
  <date role="revised">January 10, 2019</date>
  <date role="accepted">November 3, 2020</date>
  <abstract name="Abstract">
    <p>abstract</p>
  </abstract>
  </document>
```

A simple:
```
latexmlpost dates.xml --dest=dates.html
```

leads to this mangled issue:
image

I don't want to over-optimize for this specific class file right now, but the dates issue felt generic enough that it may also apply to a number of other arXiv documents? Curious what you think.

brucemiller:
On 11/3/20 9:51 AM, Deyan Ginev wrote:
> Very nice! I am unsure what the magic of the |addBetweenColumn| does 

Actually, it does nothing right now.  I'm thinking it (or something like 
it) should be able to remove the "intercolumn spacing" (possibly by 
css), rather than the clumsy explicit css trick.   For example
|@{}| should supress the spacing, but array.sty's |!{}| should not.

brucemiller:
On 11/3/20 10:12 AM, Deyan Ginev wrote:
> I should also report some part of the postprocessing (the XSLT?) appears 
> to be wrongly handling multiple dates with special roles. Here is a 
> good-looking XML from the core processing that shows the issue:

Yeah, something's wrong there. Also you should be able to add a 
`name={Received}` to the keywords to `\@add@frontmatter` to get the 
label, but that's not working either.  Have to look...

brucemiller:
I have a fix for dates, part of which is aas specific, part is generic; the latter probably ought to be a separate patch, but easier to test here.  Also, there's something messed up w/enumeration ids.

brucemiller:
PR #1369 should deal with the dates, I think.

And although you understandably are itching to run arXiv, and intended this as a quick-n-dirty, the results for the sample are looking pretty good! It may be worth polishing it a bit more while we're at it, since it is hard to focus on good examples later on.

In particular, the author block(s) are kind of a mess; either we want each author to be a block, or deal with most metadata as footnotes (which is kinda the way aas treats it; not sure why it isn't happening automatically).

dginev:
Nice! And sure, we can fix things properly, it's not that _dramatic_ a rush to run arXiv, as long as we don't overextend into say something outrageous like December :>

dginev:
I took a small stab at improving the author-block, and we were indeed missing a default for the affiliation display style. It seems the current CSS always needs an extra class to lay them out in a standard fashion - either `ltx_authors_1line` or `.ltx_authors_multiline`. As it happens the `1line` style looks quite cramped for this author list (the affiliation footnotes are way too tiny and boxed in), so I went with a default of multiline. Screenshot:

image

I also handled the orcid href, and added a couple of extra touchups for the date display. I added the `name` attributes by hand there for now, since I am still unsure where they are supposed to appear from... Also patched an unrelated typo in acmart.cls while I was studying other binding examples.

There's more work to make the keywords and doi lines pretty, some tomorrow...

brucemiller:
I just pushed a couple of minor frontmatter tweaks that I still had on my side.  The only remaining aas specific thing is the treatment of the orcid. I had defined 
```
DefMacro('\author[]{}',
  '\@add@frontmatter{ltx:creator}[role=author]{\@personname{#2}}'
    . '\ifx.#1.\else\@add@to@frontmatter{ltx:creator}{\@@@orcid{#1}}\fi');
DefConstructor('\@@@orcid{}', "^ <ltx:contact role='orcid'>#1</ltx:contact>");
```
which makes it slightly more meta, and avoids committing to the author being a link to orcid.  But I'll leave that as your call.

The only 2 remaining issues are fixing the nested itemize/enumerate id's, and suppressing intercolumn space.  Neither of these are really specific to aas, and they'll require changes in other places as well.  So perhaps this can be finished, so long as we remember to fix those 2.

dginev:
Cool! Could we also decide if we're sticking with `ltx_authors_multiline` so that I update the failing test with the class?

As to the orcid link, I was trying to stay close to the original template, and they even have the orcid icon in there (which I thought is a bit too much), but definitely a link. So I think I'll keep it :>

brucemiller:
Oh,  yeah! I think the multiline thing is exactly the trick I was trying to remember!  Yeah, go with that!

PR #1365, opened 2020-11-02
Multidoc

### LINKED ISSUE #1167: Table of contents for interconnected website
I'm staring down a long (8+ hour) latexml run that I would like to speed up if possible.

Since I'm going to be calling `latexmlpost --split` on the result, one option would be if I could split the tex going into latexml.  Then I could have latexml compile one file at a time, instead of everything at once.  I tried to mimic https://dlmf.nist.gov/LaTeXML/manual/usage/ with the test document

```LaTeX
\documentclass[openany]{book}
\usepackage{latexml}
\lxDocumentID{chx1}
\begin{document}
\tableofcontents
\chapter*{Preface}
Chapter \ref{ch1}.  Chapter \ref{ch2}.
\end{document}
```
(and similar documents for Ch1.tex and Ch2.tex), and commands
```
latexml --dest=Chx1.xml Chx1
latexml --dest=Ch1.xml Ch1
latexml --dest=Ch2.xml Ch2
latexmlpost --prescan --db=my.db --dest=Chx1.html Chx1
latexmlpost --prescan --db=my.db --dest=Ch1.html Ch1
latexmlpost --prescan --db=my.db --dest=Ch2.html Ch2
latexmlpost --noscan --db=my.db --dest=Chx1.html Chx1
latexmlpost --noscan --db=my.db --dest=Ch1.html Ch1
latexmlpost --noscan --db=my.db --dest=Ch2.html Ch2
```
Everything looks to be working correctly, except that the table of contents doesn't show up.  Am I missing something that I need to do for my table of contents?  Or do I need to manually process the TOC file for my book?  Or did you have some trickery to get the table of contents of the DLMF?

PS As long as we're looking at the usage page, am I correct that
`latexmlpost --prescan --db=my.db --dest=1.html 1`
should be
`latexmlpost --prescan --db=my.db --dest=1.html ch1`
etc?

brucemiller:
This PR generalizes & adopts the code from DLMF which is used for sites built from multiple top-level documents, to establish appropriate parent/child relations between the documents.  The first document scanned is assumed to be the root of the site, all other documents will have a parent determined either by id (the document whose id is the longest prefix of the id is assumed to be the parent) or location (a document index.xml in a containing directory will be assumed to be the parent), or the site root by default.
This allows various features such as tables of contents and previous/next links to work.

Fixes #1167

I'll have to add this clarification to the documentation (and probably also fix some typos there).


lib/LaTeXML/Post/Manifest/Epub.pm
```diff
@@ -73,7 +73,8 @@ sub initialize {
   $package->setAttribute('version',           '3.0');
 
   # Metadata
-  my $document_metadata = $$self{db}->lookup("ID:" . $$self{db}{document_id});
+  my $rootentry         = $$self{db}->lookup('SITE_ROOT');
+  my $document_metadata = $$self{db}->lookup("ID:" . $rootentry->getValue('id'));
```
brucemiller, line 77: I'm assuming that this `SITE_ROOT` mechanism achieves the same thing as your `document_id` (which doesn't get stored in the permanent db, btw).  But I didn't have any separate tests (or is there an epub test?)

dginev: `931_epub.t` should have stopped you if you broke the epub generation badly, it at least checks that the correct files are present in the final generated `.epub` files. So likely it is ok? You can always make a DLMF epub to test :>

dginev: Here's a cool one-liner test with a neat result:
```
latexmlc t/ams/mathtools.tex --splitat=section --split --dest=sample.epub
```

I just checked both master and your new branch and both seemed to create a successful epub with a nice ToC and working links to the specific pages. Here is the toc screenshot as a motivational demo:
image

So the epub side should be good :+1:

brucemiller: Cool! But  now (hopefully) it should *also* work if you create a single epub from several separate documents (if you'd ever want to do that?)

dginev:
I suppose for now the only real test is DLMF? If that still works, I'd say LGTM on my part.

brucemiller:
Oh, it works there; that's where I stole the code from :>

I'm just updating the manual to explain the new behaviour.

doc/manual/manual.tex
```diff
@@ -205,27 +205,30 @@ \chapter{Using \LaTeXML}\label{usage}
 \end{description}
 \noindent The usage of these commands can be as simple as
 \begin{lstlisting}[style=shell]
-latexml doc.tex | latexmlpost --dest=doc.html -
+latexmlc doc.tex --dest=doc.html
```
dginev, line 208: I certainly like the command, but is this a good idea to change when latexmlc itself isn't officially documented in the [executables section](https://dlmf.nist.gov/LaTeXML/manual/commands/)? Or is that part of the plan?

brucemiller: too late :>

PR #1366, opened 2020-11-02
ensure intended extensions only in FindFile wrt documentstyle

dginev:
Quick follow-up to #1300 , I did some related testing with:
```tex
\documentstyle[xcolor]{article}
\begin{document}
test \color{red} test.
\end{document}
```

and realized my kpsewhich found `article.sty` in texlive and tried to load it even though I did not provide `--includestyles`. So, patched!

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1955,8 +1955,11 @@ sub FindFile_fallback {
   # Supported:
   # Numeric suffixes (version nums, dates) with optional separators
   my $fallback_file = $file;
+  my $type          = $options{type};
   if ($fallback_file =~ s/\.(sty|cls)$//) {
-    my $ltxtype = $1;
+    # if we provide the type, that remains primary.
+    $type = $1 unless $type; }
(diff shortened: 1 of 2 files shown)


dginev:
Actually, just one second to test another bit...

dginev:
Ok, my example above, the original arXiv article for aipproc.sty, and now also the following snippet convert nicely:

```tex
\documentstyle[xcolor]{foo}
\begin{document}
test \color{red} test.
\end{document}
```

Where my latest commit made sure the options are handled by OmniBus in cases where the package is "unknown" to latexml for one reason or another. This way the XML produced for the mini doc above is correct, and there isn't even an error - just a single warning that `foo` can not be found.

I also discovered I (once again) got too paranoid with using `withoptions=>1`, it wasn't actually needed when the options hash was passed in at the correct place. So this could be good to merge!

brucemiller:
Travis signed off; quick before dginev changes his mind!

dginev:
Lovely! Getting curious for a partial arXiv run on the latest month this Thursday...

PR #1367, opened 2020-11-03
Fixes #1351

### LINKED ISSUE #1351: Adding \label inside amsthm theoremstyle used to work
This use of `\label` used to work in version 0.8.4, but doesn't work on _master_:
```tex
\documentclass{article}
\usepackage{amsthm}
\newtheoremstyle{mystyle}
  {\topsep}   % ABOVESPACE
  {\topsep}   % BELOWSPACE
  {}          % BODYFONT
  {}          % INDENT (empty value is the same as 0pt)
  {\bfseries} % HEADFONT
  {\\}        % HEADPUNCT
  {.75em}     % HEADSPACE
  {#1~#2 \textit{#3}\label{mystyle:#3}} % CUSTOM-HEAD-SPEC
\theoremstyle{mystyle}
\newtheorem{functiondefinition}{Function}[section]

\begin{document}
See function~\ref{mystyle:abs}.

\begin{functiondefinition}[abs]
An absolutely fabulous function.
\end{functiondefinition}

\end{document}
```

dginev:
Should be a mostly correct fix, getting a little closer to the TeX behavior for the head formatter of a new theorem environment. Here is a screenshot showing HTML-to-PDF parity with the example of the reporting issue:

image

If tests pass, it ought to be a safe chance, or at least no more dangerous than the code we have already :>

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -2697,18 +2697,17 @@ DefPrimitive('\renewenvironment OptionalMatch:* {}[Number][]{}{}', sub {
 # The core non-customizable part is defined here.
 # For customizable theorems, see amsthm.
 AssignValue('thm@swap' => 0);
-DefRegister('\thm@style'        => Tokenize('plain'));
-DefRegister('\thm@headfont'     => Tokens(T_CS('\bfseries')));
-DefRegister('\thm@notefont'     => Tokens(T_CS('\the'), T_CS('\thm@headfont')));
-DefRegister('\thm@bodyfont'     => Tokens(T_CS('\itshape')));
-DefRegister('\thm@headpunct'    => Tokens());
-DefRegister('\thm@styling'      => Tokens());
-DefRegister('\thm@headstyling'  => Tokens());
-DefRegister('\thm@prework'      => Tokens());
-DefRegister('\thm@postwork'     => Tokens());
-DefRegister('\thm@symbol'       => Tokens());
-DefRegister('\thm@numbering'    => Tokens(T_CS('\arabic')));
-DefRegister('\thm@headformater' => undef);
+DefRegister('\thm@style'       => Tokenize('plain'));
+DefRegister('\thm@headfont'    => Tokens(T_CS('\bfseries')));
+DefRegister('\thm@notefont'    => Tokens(T_CS('\the'), T_CS('\thm@headfont')));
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -2697,18 +2697,17 @@ DefPrimitive('\renewenvironment OptionalMatch:* {}[Number][]{}{}', sub {
...
-DefRegister('\thm@notefont'     => Tokens(T_CS('\the'), T_CS('\thm@headfont')));
-DefRegister('\thm@bodyfont'     => Tokens(T_CS('\itshape')));
-DefRegister('\thm@headpunct'    => Tokens());
-DefRegister('\thm@styling'      => Tokens());
-DefRegister('\thm@headstyling'  => Tokens());
-DefRegister('\thm@prework'      => Tokens());
-DefRegister('\thm@postwork'     => Tokens());
-DefRegister('\thm@symbol'       => Tokens());
-DefRegister('\thm@numbering'    => Tokens(T_CS('\arabic')));
-DefRegister('\thm@headformater' => undef);
```
brucemiller, line 2711: Why is this definition removed?

dginev: There is a long answer and a short answer. The short answer is that it didn't seem to be used anywhere.

The long answer is that I am quite confused by the naming and kinds of constructs for these theorem customizations. In particular, it appears that the machinery which _does_ get use is a series of assignments and lookups into state, where Tokens() get hidden behind some unfortunately named keys  -- named after command-sequences in particular. So you see both assignments and lookups to a key called "\thm@headformatter", rather than using the registers here.

I was particularly alerted by the fact that this register name had a typo (single `t` for `formater`), which was suggesting removing them may be safe. And indeed the tests passed without them.

On the other hand, I did see constructs such as  `T_CS('\the'), T_CS('\thm@headpunct')`, which were suggestive that the registers may be in use... Anyhow, it helped me with development to be sure these registers weren't playing a part in the mechanism I was changing. It would be great if I could grasp the difference...

brucemiller: Oh! Didn't see the typo.  But still, because of the sneaky way registers are stored, that could have the effect of initializing it (to nothing), which apparently didn't matter. But conceivably could.  So the best solution would seem to be initializing it with the correct spelling, just in case.  :>

brucemiller: Hmm, but not trace of it in texmf-dist, so maybe ok after all.

PR #1369, opened 2020-11-04
Better formatting of multiple dates

brucemiller:
This should handle the scenario in aastex.

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
@@ -390,9 +390,10 @@
       <xsl:call-template name="authors">
         <xsl:with-param name="context" select="$context"/>
       </xsl:call-template>
-      <xsl:apply-templates select="../ltx:date" mode="intitle">
+      <xsl:call-template name="dates">
         <xsl:with-param name="context" select="$context"/>
-      </xsl:apply-templates>
+        <xsl:with-param name="dates" select="../ltx:date"/>
+      </xsl:call-template>
     </xsl:if>
     <xsl:apply-templates select="." mode="end">
       <xsl:with-param name="context" select="$context"/>
@@ -416,9 +417,10 @@
             </xsl:apply-templates>
           </xsl:element>
         </xsl:if>
-      <xsl:apply-templates select="ltx:date" mode="intitle">
+      <xsl:call-template name="dates">
         <xsl:with-param name="context" select="$context"/>
-      </xsl:apply-templates>
+        <xsl:with-param name="dates" select="ltx:date"/>
+      </xsl:call-template>
     </xsl:element>
   </xsl:template>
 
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
```diff
@@ -644,23 +646,34 @@
...
+    <xsl:param name="dates" select="ltx:date"/>
+    <xsl:if test="$dates">
+      <xsl:text>&#x0A;</xsl:text>
+      <!-- Originally, html5 seemed to suggest we might use h2 here, but that is retracted-->
+      <xsl:element name="div" namespace="{$html_ns}">
+        <xsl:attribute name="class">ltx_dates</xsl:attribute>
+        <xsl:apply-templates select="." mode="begin">
+          <xsl:with-param name="context" select="$context"/>
+        </xsl:apply-templates>
+        <xsl:apply-templates select="$dates" mode="inintitle">
```
dginev, line 660: typo? `inintitle` -> `intitle`

lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
```diff
@@ -644,23 +646,34 @@
...
+          <xsl:with-param name="context" select="$context"/>
+        </xsl:apply-templates>
+        <xsl:apply-templates select="." mode="end">
+          <xsl:with-param name="context" select="$context"/>
+        </xsl:apply-templates>
+      </xsl:element>
+    </xsl:if>
+  </xsl:template>
+
+  <xsl:template match="ltx:date" mode="inintitle">
```
dginev, line 670: same typo? `inintitle` -> `intitle`

dginev (changes requested):

lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
```diff
@@ -644,23 +646,34 @@
...
+    <xsl:param name="context"/>
+    <xsl:if test="@name"><xsl:value-of select="@name"/><xsl:text>: </xsl:text></xsl:if>
```
dginev, line 672: This is a cool approach. I tested locally and noticed that `@name` actually has no attribute to select against in the aastex63 example. The attribute for `ltx:date` that is currently in use is role, and switching this test to `@role` got me the visuals I expected locally. Was there a reason for `@name`?

brucemiller:
@Name was more intentional (most metadata has a role, but you don't 
always want that textual prefix), and name is already used in a few
cases.  A patch coming shortly after you rebase adds name= to aas_support :>

lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
brucemiller: Ah, yeah. Actually was intentional during debugging, but is probably just confusing now.  I've fixed it.

dginev:
LGTM! Didn't quite catch which branch should get the `name` patch, and who's adding it, but this PR is definitely a nice improvement already :+1:

dginev (approved):

PR #1370, opened 2020-11-05
Make textcomp.sty part of LaTeX.pool

### LINKED ISSUE #1368: Add textcomp as a dependency of LaTeX.pool (for newer texlives)
I was a bit confused what I was seeing debugging a recent arXiv preprint with a single error (at [2011.00039](https://arxiv.org/abs/2011.00039) ):
```
Error:undefined:\textonehalf The token T_CS[\textonehalf] is not defined.
```

Luckily, David Carlisle has an authoritative answer explaining that as of texlive 2020, the symbols of textcomp are now available under the native latex.ltx distribution:
https://tex.stackexchange.com/questions/80270/am-i-using-textcomp-correctly#comment1385870_80275

arXiv has just enabled texlive 2020 and already papers are using this! I suppose it's better for us to catch it early rather than late.

The big question is, should I make a patch that _always_ loads textcomp.sty when loading LaTeX.pool, or should I have a more restrained approach? As David C says, there is no harm in loading that package anyway even in texlive 2020, so might as well just flip the switch and have it on?

Some of the obvious macros I fished out of latex.ltx on my local texlive that weren't yet in LaTeX.pool were the following. Most are already defined in textcomp, and some are completely new to latexml (e.g. `\textascendercompwordmark`).

P.S. My texlive version 2019 already has them actually, so likely not that new a change?

```tex
\textperthousand
\textbardbl
\textbrokenbar
\textcelsius
\textcent
\textdegree
\textdiv
\textlnot
\textonehalf
\textonequarter
\textonesuperior
\textpm
\textquotesingle
\textquotestraightbase
\textquotestraightdblbase
\textthreequartersemdash
\textthreequarters
\textthreesuperior
\texttimes
\texttwelveudash
\texttwosuperior
\textyen
\textcapitalcompwordmark
\textascendercompwordmark
```

dginev:
Fixes #1368 . One way to approach it, also added a couple of related macros that were in latex.ltx.

The only conflict with the tests is resolving the exact unicode character for `\textperiodcentered`, which I resolved in favour of the less semantic [Middle Dot](https://www.compart.com/en/unicode/U+00B7), rather than the more mathy [dot operator](https://www.fileformat.info/info/unicode/char/22c5/index.htm).


lib/LaTeXML/Package/textcomp.sty.ltxml
```diff
@@ -15,187 +15,8 @@ use strict;
...
-  textrecipe               => "\x{211E}",
-  textthreequartersemdash  => "\x{2014}",     # whole em dash, actually
-  texttildelow             => "~",            # ??
-  texttwelveudash          => "\x{2014}");    # whole em dash, actually.
-map { DefPrimitiveI('\\' . $_, undef, $textcomp_various{$_},
-    bounded => 1, font => { encoding => 'TS1' }) } keys %textcomp_various;
-
-#======================================================================
-RawTeX('\DeclareFontEncoding{TS1}{}{}');
-#======================================================================
```
dginev, line 200: The software developer take on this is that we actually need a command that is almost exactly like `RequirePackage` but does not report loading the binding file, just treats it as a continuation of the caller. That way we could've kept the modularity, rather than move LaTeX.pool closer to 6000 lines of code.

brucemiller: Why not simply add `RequirePackage('textcomp');` to `LaTeX.pool`?  So what if it reports? Nobody reads that anyway :>

dginev: That was my first instinct! But then I saw what was done for `fixltx2e.sty` right above the code I added, and decided to "stay consistent with the current approach".

It appears that heuristic shouldn't be my default :> Happy to go either way, as long as we get the macros in.

brucemiller:
Thanks, but let's go the other way.

PR #1371, opened 2020-11-06
Always require textcomp in LaTeX.pool

dginev:
Alternative approach to #1370

### DIFF
--- a/t/structure/filelist.xml
+++ b/t/structure/filelist.xml
@@ -5,6 +5,6 @@
   <resource src="LaTeXML.css" type="text/css"/>
   <resource src="ltx-article.css" type="text/css"/>
   <para xml:id="p1">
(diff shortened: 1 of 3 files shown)


t/structure/filelist.xml
```diff
@@ -5,6 +5,6 @@
   <resource src="LaTeXML.css" type="text/css"/>
   <resource src="ltx-article.css" type="text/css"/>
   <para xml:id="p1">
-    <p>The file list inside of the cls file was ‘filelistclass.cls’.</p>
+    <p>The file list inside of the cls file was ‘textcomp.sty,filelistclass.cls’.</p>
```
dginev, line 8: unfortunate extra change to one of the test files...

brucemiller:
I think I like this approach better (tho there's still lots of dups between latex & textcomp)

PR #1372, opened 2020-11-07
Read single token if not balanced in hyperref \url

dginev:
Took a break from Hard Debugging with some easy debugging.

We have a simple Fatal:misdefined  in arXiv from the `__ANON__` subroutines, likely rare, in [this paper](https://arxiv.org/abs/astro-ph/0001345), related to hyperref's url. I isolated the tex and added a second variation to be sure I am seeing things right:

```tex
\documentclass{article}
\usepackage{hyperref}
\begin{document}
T1: {\url http://cossc.gsfc.nasa.gov/cossc/batse/}.

T2: {\url|http://cossc.gsfc.nasa.gov/cossc/batse/|}.
\end{document}
```

And noticed the PDF only takes in the first token in this aberrant situation:
url_wdelim_hyperref

while latexml is currently trying to use the leading `h` as an open delim. So this PR switches to the simpler pdflatex behavior. I also checked that this is indeed specific to hyperref itself, because if you instead do `\usepackage{url}`, pdflatex will produce what you would expect if you were using the h as a delimiter:

url_wdelim_urlsty

Now, the paper in question uses neither of those directly. 
Instead, it relies on a genericly named `aastex.cls` dated at year 2000. Some digital archeology led me to what I believe may be [that cls file](http://www.ucolick.org/~bolte/LFPAPER2/aastex.cls), and lo and behold, it defines its own local `\url`, rather than relying on hyperref. Using that (via `\documentclass{aastex}` in my snippet above), I see with pdflatex:
image

So it *did* work once! And, reading the \url definition there, it seems that it just took in the arguments and returned them, doing nothing of any interest at all with them...? 
```tex
\newcommand\anchor[2]{#2}% 
\newcommand\url{\@dblarg\@url}% 
\def\@url[#1]{\anchor{#1}}% 
```

Tempted to just ignore that patchy definition, and keep using hyperref with the harmless (i.e. no Fatal or Error) behavior with subtly broken markup. Submitted for your consideration :> 

Edit: I said possibly rare, but I am seeing some tens of these cases in the [\_\_ANON\_\_ report](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/misdefined/LaTeXML%3A%3APackage%3A%3APool%3A%3A%5F%5FANON%5F%5F), so maybe it gets a good handful of docs improved. One also hits `url.sty` directly, will add a second minor fix there...

### DIFF
--- a/lib/LaTeXML/Package/hyperref.sty.ltxml
+++ b/lib/LaTeXML/Package/hyperref.sty.ltxml
@@ -64,19 +64,19 @@ foreach my $option (    # 3.1 General Options
 # \hypersetup{keyvals} configures various parameters,
 # for each pdf keyword, provide [property,(content|resource),datatype]
 our %pdfkey_property = (
-  baseurl     => '',                                # xmp:BaseURL ??
-  pdfauthor   => ['dcterms:creator', 'content'],
-  pdfkeywords => ['dcterms:subject', 'content'],    # & pdf:Keywords
+  baseurl     => '',                                 # xmp:BaseURL ??
+  pdfauthor   => ['dcterms:creator',  'content'],
+  pdfkeywords => ['dcterms:subject',  'content'],    # & pdf:Keywords
   pdflang     => ['dcterms:language', 'content'],
-  pdfproducer => '',                                # pdf:Producer & xmp:CreatorTool
+  pdfproducer => '',                                 # pdf:Producer & xmp:CreatorTool
   pdfsubject  => ['dcterms:subject', 'content'],
-  pdftitle    => ['dcterms:title', 'content'],
+  pdftitle    => ['dcterms:title',   'content'],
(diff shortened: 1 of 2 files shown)


dginev:
Alright, second patch is in. It allowed the following to work smoothly, and got closer to the url.sty native tex definition of `\urldef`. Trick was to read the arg in first, so that any arg braces are unwrapped correctly, then unread it back and proceed to expand further.

The paper which transitioned from a Fatal to a Warning thanks to this was [math-ph/0204024](https://arxiv.org/abs/math-ph/0204024). 

```
\newcommand\email{\begingroup \urlstyle{tt}\Url}
\urldef{\danieldf}{\email}{someone@somewhere.edu}
```

brucemiller:
tricky!

PR #1373, opened 2020-11-07
Simplified lexeme scheme

dginev:
With the benefit of hindsight, here is a PR with some improvements to the lexeme serialization for NLP, focused on making things cleaner and simpler.

- For `XMTok` elements, always use the textual content as the lexeme. The attribute information is a fallback when no content is inked on the page.
- Entirely remove the `$role:` prefix in lexemes, _except_ when the role is the only available information (e.g. the `FRACOP` token)
- Add an explicit font shape for cases that overlap the text-mode chars, e.g. generated via `\mathrm`.
   -  I settled on `roman-` to alert readers it has a technical meaning (`normal` is too vague).
- Rearranged all modifiers to be prefix and the last lexeme piece to be the content. So now we have `start-ARG` and `end-POSTSUPERSCRIPT`, rather than the other way around, to be consistent with the font prefixes.

Two working examples, with a comparison between master/PR are:

1. Unicode operator and simple letters:
     ```tex
     $ a \oplus \mathrm{a} < b $
     ```
   - master:
     ``` 
     italic-a ADDOP:direct-sum a RELOP:less-than italic-b
     ```
   - new PR:
     ````
     italic-a ⊕ roman-a &lt; italic-b
     ````
2. Binomial and parentheses:
     ```tex
     $ x \neq 2\binom{m}{n}\times (x-y) $
    ```
   - master:
     ``` 
     italic-x RELOP:not-equals NUMBER:2 OPEN:( FRACOP:italic-FRACOP ARG:start italic-m ARG:end ARG:start italic-n ARG:end CLOSE:) MULOP:times OPEN:( italic-x ADDOP:minus italic-y CLOSE:)
     ```
   - new PR:
     ````
      italic-x ≠ 2 ( FRACOP start-ARG italic-m end-ARG start-ARG italic-n end-ARG ) × ( italic-x - italic-y )
     ````


dginev:
A big part of the change here is embracing Unicode inside the lexeme strings, and saving a lot of space in serialization for the operator cases. `⊕` is a lot more compact than the `directed-sum` XMTok meaning or the unicode name `circled plus`. It is also a more immediate representation of what was inked in the latex document, which is the level of information we are trying to capture with the lexemes.

Follow-up toolkits can normalize this back down to ascii with a standard de-unicode tool, as most programming languages have one available. 

Meanwhile, tokenization remains as simple as before - you can simply split by space the `lexemes` XMath attribute / or the `application/x-llamapun` annotation in MathML respectively. Lexemes never contain spaces internally.

We have shaved off more of the information early, since Bruce aptly remarked that our first experiment may be overly constraining due to latexml's limited dictionary of mathematical roles, and limited accuracy in assigning them on unknown expressions (as those in the larger arXiv). That is even truer when considering using the `meaning` attribute. This PR ends up more honest about the level of uncertainty behind math tokens, and aims for a "zero-knowledge" outcome.

I will also bookkeep that we haven't given up on serializing the full breadth of information latexml assigns to its tokens, but it appears that will have more utility as a separate serialization, which Bruce has dubbed "smart lexemes". Going for both extremes (zero-knowledge vs all-knowledge) makes a lot of sense to me, since it allows different applications to pick-and-choose what is needed. Of course, the "smart lexemes" are not in this PR, but are also not on anyone's short-term timeline, so they can be added as they become necessary. Alright, that about wraps it up!

brucemiller:
First impression: I like it.

dginev:
Cool! I also tested the indicator function:
```tex
\mathbf{1}_{A}(x)
```

master:
```
NUMBER:bold-1 POSTSUBSCRIPT:start italic-A POSTSUBSCRIPT:end OPEN:( italic-x CLOSE:)
```

new PR:
```
bold-1 start-POSTSUBSCRIPT italic-A end-POSTSUBSCRIPT ( italic-x )
```

dginev:
Also bold operators:
```tex
a \boldsymbol{+} b
```

master:
```
italic-a ADDOP:bold-plus italic-b
```

new PR:
```
italic-a bold-+ italic-b
```

The good part is that I didn't need any changes to enable these, already operational. The bad news is that the dash is starting to feel a bit awkward when the following content is not alphanumeric. For a bold minus the output is `bold--` for example... Hm.

dginev:
Alright, I have ensured `-` is treated as `minus` to avoid the confusion with the separator dash. I think I am content with having `bold-+` or `bold-×`, so this may be close to final for this pass.

brucemiller:
Maybe it's a question of which "weird" you prefer, but "-" turning into "minus" but "+" not into plus is also weird.  The treatment of "-" has the feel of being driven by the (wrong) tools, rather than logic. I do think that a "bold minus" should be distinct from non-bold, and do agree that "bold--" will trip up (say) perl's split.  But I think we'd be better off using a rarer "_" for the punctuation.

The fact that screen readers "automatically" treat "-" as a space doesn't seem too compelling, as these are lexemes, not accessibility attributes. There will be a transformation before you get accessibility info anyway, (such as "+" :> ) so having to deal with "_" doesn't seem to be a cost, even if you did want to start with lexemes (which isn't clear).

dginev:
Modern tokenizers can certainly deal with underscore just as adequately (just tested the default huggingface tokenizers for Bert and Electra for example), so `bold-minus`, `bold_minus` and `bold_-` will all get split into the expected 3 components with the default vocab.txt for Bert (or my adapted one for arXiv).

So certainly a doable change.

Using the dash *or* underscore for naive splits is a bad idea in general in any case, since even without math in the picture there are some interesting cases of plain-text dashes in the wild. And modern tokenizers can handle that. So I will switch everything to underscore, commit incoming...

brucemiller:
Or even ":" if you special case ":" and turn a bold colon into "bold:colon", at least you haven't imparted any semantics :>

dginev:
Colon I like less because it feels like a departure from mainstream NLP delimiters. Generally not used to glue word pieces together, in contrast to dash and underscore.

brucemiller:
Maybe it's my ml ignorance about "subtokens", but it seems that at the top-level, the delimiter shouldn't be seen or split on at all. Whatever lexeme you create for "bold minus sigh" should be atomic.

dginev:
Whether it is atomic or not depends on your tokenization approach, which in turn depends on your choice of a model. If you rewind back to 2016 what you are saying was canon - you would do a GloVe embedding with a huge vocabulary of atomic tokens, such as `bold_minus`, and I had one with just over a million distinct tokens for arXiv.

Fast-forward to 2019 with BERT being mainstream, and the optimal approach placed a hard limit on the vocabulary (30,000 tokens to be precise), where any token outside of that gets chunked via "subword tokenization". The two popular algorithms for doing that are BPE (byte-pair encoding) and the more linguistically motivated (and used by e.g. ELECTRA) WordPiece. They make a difference between standalone word such as `bold` and the same used as a subword, usually written `bold##`. So for `bold_-` you'd see the token array `[bold##, _##, -]`, though ideally you wouldn't be looking at that representation at all yourself. Then the information that is unique to this token triple put together will be encoded in the many levels of layers, with the help of the attention mechanism - some attention nodes will give the indexes of this triples heavy weights, and will remember important latent features for them.

Long story short, both ideas are viable and have been done.

dginev:
Commit is in. Final versions of the examples from above:
 * `italic_a ⊕ roman_a &lt; italic_b`
 * `italic_x ≠ 2 ( FRACOP start_ARG italic_m end_ARG start_ARG italic_n end_ARG ) × ( italic_x - italic_y )`
 * `bold_1 start_POSTSUBSCRIPT italic_A end_POSTSUBSCRIPT ( italic_x )`
 * `italic_a bold_+ italic_b`

dginev:
One more edge case to patch...

dginev:
Caught an edge case, my conditionals weren't letting zeros through. Also got a little ambitious and attempted to handle font information in `\text`, for shallow textual content. After fix:
```tex
$ \int_{0}^{\infty}x\,dx=\text{zero, \it yes 0} $
```

lexemes in PR:
```
∫ start_POSTSUBSCRIPT 0 end_POSTSUBSCRIPT start_POSTSUPERSCRIPT ∞ end_POSTSUPERSCRIPT italic_x italic_d italic_x = zero, italic_yes italic_0
```

It's a bit too complex to handle for arbitrary contents of `\text`, but this is still a nice upgrade over the previous state. Here's what master is currently producing for this example (not the textual bits in italic are missing):

```
INTOP:integral POSTSUBSCRIPT:start NUMBER:0 POSTSUBSCRIPT:end POSTSUPERSCRIPT:start ID:infinity POSTSUPERSCRIPT:end italic-x italic-d italic-x RELOP:equals zero,
```

brucemiller:
If you were serializing full text, with embedded math, you'd presumably do "nothing special" to the text, other than random start_math, end_math, start_text, end_text.   Does that suggest a more neutral way of dealing with `\text`? Of course, here you're only lexematizing the math alone. Moreover, with "yes" you've got text italic, not math italic.

dginev:
With the premise that my main lexeme focus is currently arXiv:

My experience is that while `\text` was _intended to_ interleave textual content within a formula, the way it is _actually used_ in arXiv is very diverse. A quick grep, finds ~207,000 uses of `\text` in just the no_problem set from the 2019 dataset. There are a lot of eyebrow-raising uses in that list, though I have no claim on providing meaningful statistics, it's too big!

Copy-pasting some expected, some surprising in a list here (and these are from `alttext` attributes, containing tex for entire math expressions):
 - `{\mathbb{Z}}/\text{image}(\pi_{1}S^{1})`
 - `H_{0},\Omega_{\text{m},0}`
 - `a-b\sqrt{2},\;\text{ where $a$ and $b$ are positive integers}.`
 - `\Delta Q_{\text{irr}},\theta`
 - `{}^{\text{\tiny{{(TM)}}}}`
 - `Q_{\text{match}}`
 - `Q_{\text{softest}}^{\text{ME}}`
 - `m_{\chi}=1\ \text{MeV}\ c^{-2}`
 - `r(F)&lt;r_{\text{gen}}(n,d)`
 - `k^{\prime}(x)\geq\sqrt{\alpha}\quad\text{for all }x\in\mathbb{R}`
 - `\textbf{K}_{L}=\text{diag}[k^{L}_{\alpha}]`
 - `B=n-\frac{A}{2}\text{.}`
 - `\text{SrTiO}_{3}`

So you see a smattering of all uses, both good and bad 
  1. used on arbitrary math symbols - variables, functions, others, for switching to the text font,
  2. used for units, again assuming for the font change,
  3. used for actual words in scripts, i.e. "embellishing language",
  4. used for the expected natural language connectives, " and ", " for all " etc., but also for punctuation
  5. used for entire dependent sentences that contain restrictions -- also with nested math in them!

If scrolling through 4-5 screens of these counts for a random sample, most of the uses I see are arguably changing the font of a single symbol, rather than adding regular words.

So maybe I should aim to make this list lexematize in an acceptable way and call it a day? 

Bookkeeping my ripgrep command that generates the list from the HTML data, for future reference:
```
rg -o -N -r '$1' 'alttext="([^"]*\\text\{[^"]*)"' no_problem/  > text_use.log
```

dginev:
Ok, ran a check, in all examples above the current PR preserves the textual argument exactly as-is when it is plain-text (as there are no font modifications). But say `\text{image}` gets mapped into a lexeme `image`, as contrasted to `\mathrm{image}` which gets mapped to `roman_image`. Which is a difference, but at the same time `\text{for all}` becomes the more textual "for all", rather than `roman_for roman_all`, which I assume we want to avoid.

Happily, the nested math works out with the current PR as well, the third example gets the lexemes:
```
italic_a - italic_b square-root start_ARG 2 end_ARG , where italic_a and italic_b are positive integers .
```

So maybe already quite reasonable? This depends a little on the application, I could imagine how a grammar would be able to leverage the `start_text`, `end_text` markers you mentioned, but for a language model they don't really add much (may end up approximating noise, especially in cases like `\text{GeV}`, `\text{km}`, `\text{m}`... ). I'm pretty content with the current state, though open to tweak it further.

dginev:
Since I had the data, and am hunting for clarity, I computed the stats for using `\text` in arXMLiv 08.2019 this morning. I am attaching the full stats as a ZIP to this comment: [arxiv_stats_macro_\text.zip](https://github.com/brucemiller/LaTeXML/files/5530743/arxiv_stats_macro_.text.zip)

Executive summary:
 - 4.5 million recorded uses of `\text` in latexml's `alttext` attribute.
 - ~218,000 of their arguments are distinct
 - i trimmed the outer spaces (so " for all " is just "for all" in the report), to reduce the number of unique hits. Assume a bunch of these have been used with added spacing, especially the actual words and phrases.

Here is a ranking of the top 100 most frequent entries:

| tex       | arXMLiv frequency |
| :-------- | ----------------: |
| \text{max} | 87684 |
| \text{and} | 79181 |
| \text{eff} | 63951 |
| \text{d} | 53319 |
| \text{s} | 53262 |
| \text{-} | 52951 |
| \text{min} | 52428 |
| \text{c} | 50488 |
| \text{th} | 50347 |
| \text{in} | 49212 |
| \text{GeV} | 42705 |
| \text{for} | 41162 |
| \text{m} | 40546 |
| \text{B} | 36726 |
| \text{T} | 35354 |
| "\text{ }" | 34037 |
| \text{e} | 33846 |
| "\text{for all}" | 29844 |
| \text{H} | 27720 |
| \text{R} | 25849 |
| "\text{ \| }" | 25595 |
| \text{S} | 24479 |
| \text{F} | 23579 |
| \text{.} | 23184 |
| \text{Tr} | 22967 |
| \text{L} | 22851 |
| \text{C} | 22722 |
| \text{p} | 22690 |
| \text{M} | 22191 |
| \text{Im} | 22070 |
| \text{A} | 21481 |
| \text{Re} | 21467 |
| \text{tr} | 21263 |
| \text{out} | 20319 |
| \text{D} | 20185 |
| \text{P} | 19623 |
| \text{mod} | 19603 |
| \text{i} | 19536 |
| \text{int} | 19533 |
| \text{b} | 19210 |
| \text{I} | 19130 |
| \text{E} | 19077 |
| \text{cm} | 18536 |
| \text{diag} | 18221 |
| \text{on} | 18217 |
| \text{f} | 17828 |
| \text{tot} | 16772 |
| \text{N} | 16547 |
| \text{K} | 15923 |
| \text{g} | 15343 |
| \text{o} | 14931 |
| \text{if} | 14807 |
| \text{supp} | 14330 |
| \text{r} | 13767 |
| \text{or} | 13530 |
| \text{MeV} | 13474 |
| \text{a} | 13185 |
| \text{loc} | 12949 |
| \text{as} | 12512 |
| \text{ext} | 12475 |
| \text{dim} | 12469 |
| \text{t} | 12221 |
| \text{opt} | 12157 |
| \text{O} | 11714 |
| \text{U} | 11070 |
| \text{eV} | 11018 |
| \text{TeV} | 10914 |
| \text{eq} | 10814 |
| \text{G} | 10706 |
| \text{exp} | 10582 |
| \text{} | 10546 |
| \text{h} | 10364 |
| \text{rank} | 10331 |
| \text{id} | 9737 |
| \text{2} | 9658 |
| \text{V} | 9490 |
| \text{SU} | 9167 |
| \text{det} | 9034 |
| \text{Aut} | 8638 |
| \text{Hom} | 8557 |
| \text{n} | 8241 |
| \text{obs} | 8130 |
| \text{II} | 8012 |
| \text{x} | 7973 |
| \text{with} | 7960 |
| \text{nm} | 7849 |
| \text{--} | 7837 |
| \text{deg} | 7829 |
| \text{ad} | 7799 |
| \text{dist} | 7727 |
| \text{SO} | 7701 |
| \text{span} | 7696 |
| \text{st} | 7659 |
| \text{cut} | 7605 |
| \text{crit} | 7581 |
| \text{ref} | 7542 |
| \text{v} | 7469 |
| \text{ex} | 7368 |
| \text{w} | 7353 |
| \text{Var} | 7288 |

lib/LaTeXML/MathParser.pm
```diff
@@ -731,44 +735,43 @@ sub node_to_lexeme_full {
...
-          $lexemes .= $child->textContent() . ' '; }
-        elsif (my $child_lexeme = $self->node_to_lexeme_full($child)) {
+    # can be either a single node XMText, we're looking at a leaf text node
+    # or \text{}-like construct, with multiple math formulas and interleaved text
+    foreach my $child ($node->childNodes) {
+      if (ref($child) eq 'XML::LibXML::Text') {
+        $lexemes .= $child->textContent() . ' '; }
+      else {
+        my $child_lexeme = $self->node_to_lexeme_full($child);
+        print STDERR "child lexeme: $child_lexeme\n";
```
brucemiller, line 757: debugging leftover?

brucemiller:
other than that stray line, ready to merge; then I can make another cpan test....

dginev:
Thanks for catching that and sorry! Should be good to go :+1:

dginev:
A note for future eyes, as I'm preprocessing the arXiv lexemes and deliberating markup in the wild...

A new claim has been made by the chemistry community that is now experimenting with transformer models, in particularly in the recent introduction of the [SELFIES](https://aspuru.substack.com/p/molecular-graph-representations-and) representation.

Problem summary:
> The problems stem from the fact that SMILES strings have a complex grammar. When used in machine learning models, most of the results are entirely invalid. Either they do not correspond to correct syntactical molecules or violate fundamental constraints from physics and chemistry. 

Solution:
> Even entirely random SELFIES strings represent correct molecular graphs.

So, this is fundamentally a very simple insight, which may give a great performance boost. The idea is to design the encoding in such a way that a model can never output a completely ungrammatical serialization, thus maximizing the learning signal, and adding a bit of a QA upgrade.

So, what does this imply for our math lexemes? A fragile example in our current lexemes would be:
```
start_POSTSUPERSCRIPT 2 end_POSTSUPERSCRIPT
```

because the model may produce invalid output by forgetting the left or right delimiting start/end lexeme. Instead, the analogous approach to what the chemical community did is what we already do for font information -- a single prefix morpheme, such as `italic_` or `bold_`.

If the `SELFIES` conclusion transfers over to our math lexemes, that would suggest we can upgrade by doing:
```
psup_2
```

and for `x^{2^3}` we could do:
```
italic_x psup_2 psup_psup_3
```

With the benefit that a model would never be able to create an invalid superscript missing a start or end -- it will just return the content to the baseline. This also localizes the errors to the actual argument of the script - instead forgetting to close a script may wrongly imply the entire RHS of the expression is inside the script.

A bit late to redo this now on the latexml side, but I am seriously considering whether I should already change it in the llamapun side... @brucemiller I wonder if the SMILES blog post sounds reasonable to you, given that we are also about to experiment with the same class of transformer architectures.

dginev:
A bigger example to make the scheme even clearer:
```
x^{{2+3}^{4-5}} = 0
```

maps to:
```
italic_x
   psup_2 psup_+ psup_3 psup_psup_4 psup_psup_- psup_psup_5
= 0
```

compared to the current:
```
italic_x
start_POSTSUPERSCRIPT
  2 + 3 
  start_POSTSUPERSCRIPT
    4 - 5
  end_POSTSUPERSCRIPT
end_POSTSUPERSCRIPT
= 0
```
```

**Comment by dginev** [comment]
Hm, this would actually also imply tables would be better grasped by marking them by delimiting infix rather than circumfix.

As in:
```
start_ARRAY start_ROW 
  start_CELL Line end_CELL
  start_CELL Model end_CELL
  start_CELL italic_F sub_roman_line end_CELL
  start_CELL italic_r sub_roman_inner sup_† end_CELL
  start_CELL italic_r sub_roman_outer end_CELL
  start_CELL italic_χ sup_2 / roman_dof end_CELL
  end_ROW
end_ARRAY
```

may be better modeled if written as:
```
start_ARRAY
  Line T_ALIGN
  Model T_ALIGN
  italic_F sub_roman_line T_ALIGN
  italic_r sub_roman_inner sup_† T_ALIGN
  italic_r sub_roman_outer T_ALIGN
  italic_χ sup_2 / roman_dof T_ALIGN
  T_CR
end_ARRAY
```

I think the array start/end wrapper is still needed, whatever it be called. But one is indeed reducing the token load by half when using an infix align mark and an infix end-of-row mark. Which means the model can spend more attention on the content, and use less attention on the boilerplate -- it also fits the SELFIES mantra of making it impossible to create unbalanced grammatical spans, with the exception of still needing an `end_ARRAY`. Localizing annotations to each token and using separators instead of wrappers appears to be easier to model for transformers (repeating myself, maybe also to internalize it on my end, since I only recently found the argument convincing).

PR #1374, opened 2020-11-07
[arXiv patches] \endthebibliography, \@definecounter, low-level font size macros

dginev:
Looking at the [error:unexpected](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/unexpected?all=false) report, I was surprised to see 1.18% of tasks sharing an error about an unexpected `\endlx@list` macro.

Debugging two articles:

1. In the first, [astro-ph/0001014](https://arxiv.org/abs/astro-ph/0001014), I noticed that the issue comes from redefining the bibliography macros in the custom local `.sty` file. We are successfully locking those when the new definition is done via `\def`, but for the ending macro the switch is done via `\let`:
    ```tex
    \let\endthebibliography=\endlist
    ```

   And locking macros against `\let`s isn't a viable strategy. However, I noticed we always execute a sub `beforeDigestBibliography` when `\thebibliography` is invoked, and it appears quite safe to simply redefine the constructor for `\endthebibliography` right then and there. Which I did, and reached a successful fix.

   A secondary issue was adding the internal macro behind `\newcounter`, namely `\@definecounter`, so that it is available for use in the local `.sty` files. Which is easy since the former is just the latter with an added check that the CS is definable.

2. In the second paper I checked, [cond-mat/0001346](https://arxiv.org/abs/cond-mat/0001346), there is the exact same issue with the bibliography ending, so there is hope this fix can help a large number of articles. 

   Here I also noticed the seemingly harmless use of the LaTeX low-level macros for storing pt sizes for fonts. These ought to be a recent introduction to the arXiv reports, as they are encountered in the local `.sty` files we are now interpreting raw from the paper's root directory. So I added them in, and the article converted nicely. I don't believe there should be any major side-effects, and hopefully it helps other articles avoid errors.

   So, these could be easy wins in arXiv, and easy merges to latexml!

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -5410,6 +5410,31 @@ DefMacro('\fontsize{}{}', Tokens());
 # https://tex.stackexchange.com/questions/112492/setfontsize-vs-fontsize#112501
 DefMacro('\@setfontsize{}{}{}', '\let\@currsize#1');
 
+DefMacroI('\@vpt',    undef, T_OTHER('5'));
+DefMacroI('\@vipt',   undef, T_OTHER('6'));
+DefMacroI('\@viipt',  undef, T_OTHER('7'));
+DefMacroI('\@viiipt', undef, T_OTHER('8'));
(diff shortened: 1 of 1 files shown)


brucemiller:
Cool; thanks!

PR #1375, opened 2020-11-07
[arXiv patches] Allow semiverbatim in \figcaption filenames

dginev:
A number of documents in the 1.77% [error:unexpected:_](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/unexpected/%5F?all=false) report seem to be due to underscores in captions, in particular at the filename optional argument, e.g. in [astro-ph/0001049](https://arxiv.org/abs/astro-ph/0001049):

```tex
\figcaption[dev_join.eps]{The equivalent number of Gaussian standard...}
```

It turns out this is handled correctly as semiverbatim in the aas_support binding with:
```perl
DefMacro('\@figcaption OptionalSemiverbatim {}', ...
```

but in the cases where the macro is used inside a `{figure}`, it gets redirected to the main LaTeX.pool `\caption` macro. And there the expected optional argument was plain, rather than semiverbatim. Changing that gets the article in question to succeed with `no problems`, so another one in the books.

### DIFF
--- a/lib/LaTeXML/Package/aas_support.sty.ltxml
+++ b/lib/LaTeXML/Package/aas_support.sty.ltxml
@@ -304,8 +304,9 @@ DefMacro('\plotfiddle Semiverbatim {}{}{}{}{}{}',
 # For figures added externally; Used at end of file.
 # \figcaption[filename]{text\label{key}}
 # But sometimes used just like \caption!
-DefMacro('\@figcaption OptionalSemiverbatim {}', '\begin{figure}[#1]#2\end{figure}');
-DefMacro('\figcaption', sub {
+DefMacro('\@figcaption {}', '\begin{figure}#1\end{figure}');
+# Note that the optional [filename] seems unused, so we just read and drop it.
+DefMacro('\figcaption OptionalSemiverbatim', sub {
     (((LookupValue('current_environment') || '') =~ 'figure')
       ? T_CS('\caption') : T_CS('\@figcaption')); });
 


brucemiller:
But the optional arg to standard LaTeX's `\caption` is for the table of figures, so this isn't quite right.   I think that `\figcaption`, when used inside a figure, should redirect to a different aas-specific caption macro.  The 2nd argument being a file name, not TOF entry. Not actually sure what we should do with it; Drop it?  Eventually, you'd call `\caption` w/o the optional arg.

For that matter, the defn of `\@figcaption` isn't right either. The optional argument to `\begin{figure}` is the placement code (h,t,b). Should we presume great optimism and wrap it in an `\includegraphics`? :>

brucemiller:
checking aastex63.cls, it seems that they just drop the filename argument. Why is it there?

dginev:
It is quite tempting to just drop it to be honest...

brucemiller:
If so, simplest solution would be to add `OptionalSemiverbatim` arg to `\figcaption`, (removing it from `\@figcaption`) and then let it redirect as before.

dginev:
Various versions of aastex vaguely refer to "identify"-ing the file via the optional argument, but I don't think any of them "acted" on that verb. E.g:
http://web.mit.edu/newtex/doc/latex/aastex/RelNotes.v40
> \figcaption[file_name]{TEXT}
  This command can be used for generating figure captions on a figure caption
  page: this command is not intended to be used in a figure environment. The
  captions are simply printed in paragraph mode and the figure numbers are
  incremented.  A \label command can be included in the {TEXT}.  The optional
  [file_name] parameter is intended to be used to identify the figure's EPS 
  file, if it exists - use the actual file name as it appears on disk.

I'll switch this around as per your latest suggestion, thanks!

PR #1376, opened 2020-11-08
[arXiv patches] only need to expand in \newenvironment

dginev:
Another underscore error, [in this paper](https://arxiv.org/abs/cs/0208043), was wrongfully illegal in:
```tex
\newenvironment{theorem_cite}[1]
```

Turns out this is perfectly workable in the native TeX, due to `\csname` and `\endcsname` wrapping the `theorem_cite`. LaTeXML was doing a bit too much here, all the way to digesting, rather than simply expanding. So I switched to that and the paper seems to succeed well with the theorems looking good. latexml tests pass in a breeze. Should be an easy merge.


brucemiller:
Hmm. I wonder if digest was used to reduce such cases to unicode (tho it didn't help here).  But a key thing is that however the environment name is handled, it should match `\begin`. And in fact, the latter only use expand.  So, let's see what breaks :>

PR #1378, opened 2020-11-09
Itemization ids

brucemiller:
This PR makes the ids of all nested lists (itemize, enumerate, description, ...) relative to the item's id (irrespective of what kind of list). Alas it makes the ids longer and the code is a bit more obscure, but avoids id clashes.

### DIFF
--- a/t/daemon/formats/jats.xml
+++ b/t/daemon/formats/jats.xml
@@ -118,12 +118,12 @@ Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor 
           <list-item id="S1.I1.i2">
             <p id="S1.I1.i2.p1">two</p>
             <p>
-              <list list-type="bullet" id="S1.I1.I1">
-                <list-item id="S1.I1.i2.i1">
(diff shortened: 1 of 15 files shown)


t/daemon/formats/jats.xml
```diff
@@ -138,12 +138,12 @@ Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor
           <list-item id="S1.I2.i2">
             <p id="S1.I2.i2.p1">two</p>
             <p>
-              <list list-type="order" id="S1.I2.I1">
-                <list-item id="S1.I2.i2.i1">
-                  <p id="S1.I2.i2.i1.p1">one</p>
+              <list list-type="order" id="S1.I2.i2.I2">
+                <list-item id="S1.I2.i2.I2.i1">
+                  <p id="S1.I2.i2.I2.i1.p1">one</p>
```
dginev, line 143: `S1.I2.i2.I2.i1.p1` is a fascinating construction, but certainly quite usable. As long as we don't have to present it to any ID-skeptical committees, seems good to me! An verbose structural ID is definitely better than a clashing ID with a patch.

PR #1380, opened 2020-11-10
minor perl warning appeasement in main latexmlc calls

dginev:
Very minor warnings (undef vs empty string) that showed up in some of the main latexmlc calls, best to get rid of them and keep the noise levels low.


brucemiller:
looks reasonable; thanks!

PR #1383, opened 2020-11-13
[arXiv patches] Use .bbl heuristic when unpacking; handful extra latex macros

dginev:
Fixes [1905.00934](https://arxiv.org/abs/1905.00934). More importantly, it has the potential of avoiding confusion in detecting the main TeX source file from an arXiv paper directory. 

For the example of this paper, there were equally viable `main.tex` and `main.v10.tex` source files in the root directory. If following the official `arXiv::AutoTeX` logic, the first lexicographic entry would be considered primary, i.e. `main.tex`, but in this case that is the wrong choice. I didn't find official language how arXiv would have found the right file, and the timestamps seem identical.

However, one of the files has a `.bbl` of the same name, while the other doesn't. So, for arXiv, this looks like a meaningful heuristic as a final tie-breaker between otherwise equally viable files. The one with a `.bbl` is primary.

Also, the custom `.sty` included with this paper used a couple of latex internals we hadn't defined, which seemed quite easy to just include as-is, so also added those. Paper completes with a single warning after the PR, for a single math formula that failed to parse.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -981,7 +981,14 @@ DefPrimitive('\markboth{}{}',    undef);
 DefPrimitiveI('\leftmark',  undef, undef);
 DefPrimitiveI('\rightmark', undef, undef);
 DefPrimitive('\pagenumbering{}', undef);
-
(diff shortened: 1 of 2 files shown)


brucemiller:
Sounds plausible, too!

PR #1385, opened 2020-11-13
[arXiv patches] avoid loading a missing .bbl file

dginev:
Certain arXiv articles seem to have an alternatively named .bbl file (because why not), and we're currently filing a hard Error for missing the wrongly expected bbl. For instance in this example, in file `main.tex`:

```tex
\input{OOD.bbl}
\bibliographystyle{ACM-Reference-Format}
\bibliography{OOD}
```

Even after emitting the Info message that `main.bbl` isn't found, we also emit an Error from trying to `\input` it anyway. This PR simply avoids trying to load a file we can't find. No other fix is needed since the alternative file `OOD.bbl` gets loaded just fine.

Also, made `canContain` a little quieter when called with an empty `$tag`, since this doc also did that a bunch.

Article in question was [1905.01726](https://arxiv.org/abs/1905.01726)

### DIFF
--- a/lib/LaTeXML/Common/Model.pm
+++ b/lib/LaTeXML/Common/Model.pm
@@ -238,7 +238,7 @@ sub decodeQName {
   my ($self, $codetag) = @_;
   if ($codetag =~ /^([^:]+):(.+)$/) {
     my ($prefix, $localname) = ($1, $2);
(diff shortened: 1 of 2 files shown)


brucemiller:
Looks plausible! Thanks!

PR #1386, opened 2020-11-14
Further handling of \the tokens, cleveref.sty regression

dginev:
Work in progress, for now my fix can handle a very specific new example that we had regressed on after my expl3 refactor:
```tex
\documentclass{article}
\makeatletter
\def\@tempc{fuu}
\toksdef\@toksa=0
\@toksa={\expandafter\def\csname\@tempc @foo\endcsname}

\the\@toksa{[Foo]}

\begin{document}
 Test:\fuu@foo.
\end{document}
```

More to come.


lib/LaTeXML/Core/Stomach.pm
```diff
@@ -147,6 +151,14 @@ sub invokeToken {
   my ($self, $token) = @_;
   no warnings 'recursion';
 INVOKE:
+  # If we are invoking a new token, we'll do a new read of arguments,
+  # and that should void any remaining CC_SMUGGLE_THE noexpands
+  if ($$token[1] == CC_SMUGGLE_THE) {
```
dginev, line 156: Actually, if I just avoid this guard on `$token` but run the removal of the smuggle cc *every time* invokeToken is called, cleveref.sty loads correctly, _and_ all tests pass. So this may be the only code path that needs attention - in one sense.

However, it is a terrible idea from a pragmatic standpoint - this is a very hot path in the code, and gets called **11,572** times just for fully running an empty `\usepackage{cleveref}` doc. So one would prefer solving the smuggle issue either earlier, or only when obviously needed, so that we can stay cheap in runtime...

dginev: With the guard on, as you see it in the diff here, it actually doesn't get called at all during a `\usepackage{cleveref}` but fixes our synthetic example from the PR description. So at least it's fast :>

dginev:
I may have done this better, and dare I say _right_ ?

A very important performance trick is to set a flag in the gullet whenever we add the "cc_smuggle_the" tokens, and only do whichever mitigation tactic we choose when  that flag is set. Turns out this can save a _lot_ of time, as the hash lookup needed is cheap, but going through the pushback (often) is expensive.

So my parallel make test completes with just +1 second of runtime added, with the latest approach. Neat!

I also think I managed to fix with more conceptual clarity (i.e. without touching Stomach at all). In particular, in the spirit of TeX's `scan_toks`, I switched to "setting up a clean slate" whenever a new call to `->readArguments` is made (and the flag mentioned is set). As long as the smuggle tokens are gone whenever a new scan starts, we have some guaranteed parity with the TeX book.

The latest commit is fast, passes tests, and loads cleveref. I feel good about the morning.

lib/LaTeXML/Core/Parameters.pm
```diff
@@ -59,6 +60,7 @@ sub revertArguments {
 sub readArguments {
   my ($self, $gullet, $fordefn) = @_;
   my @args = ();
+  $gullet->setup_scan() if $$gullet{pushback_has_smuggled_the};
```
dginev, line 63: This is the crux of the fix. Quite happy with it for now!

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -101,6 +101,15 @@ sub flush {
   $$self{mouthstack} = [];
   return; }
 
+sub setup_scan {
+  my ($self) = @_;
+  $$self{pushback_has_smuggled_the} = 0;
+  # setup new scan by removing any smuggle CCs
+  for my $token (@{ $$self{pushback} }) {
+    if ($$token[1] == CC_SMUGGLE_THE) {
+      $token = $$token[2]; } }
```
dginev, line 110: btw this for loop ought to be the fastest way to sneak out the catcode without needless overhead. Running a `map` that replaces the pushback, which is what I tried in my first exploratory commit, is noticeably slower.

brucemiller: That actually *replaces* the tokens? darm perl!

lib/LaTeXML/Core/Parameters.pm
```diff
@@ -17,6 +17,7 @@ use LaTeXML::Common::Object;
 use LaTeXML::Common::Error;
 use LaTeXML::Core::Parameter;
 use LaTeXML::Core::Tokens;
+use LaTeXML::Core::Token qw(CC_SMUGGLE_THE);
```
brucemiller, line 20: leftover?

lib/LaTeXML/Core/Parameters.pm
brucemiller, line 63: how expensive is moving the `if` to Gullet's `setup_scan`?  I'd think not so much, and would avoid the secret knowledge & `$$`.

lib/LaTeXML/Core/Parameters.pm
```diff
@@ -68,6 +70,7 @@ sub readArgumentsAndDigest {
   my ($self, $stomach, $fordefn) = @_;
   my @args   = ();
   my $gullet = $stomach->getGullet;
+  $gullet->setup_scan() if $$gullet{pushback_has_smuggled_the};
```
brucemiller, line 73: ditto

lib/LaTeXML/Core/Gullet.pm
dginev: yeah, I had to double-check the docs to make sure. Rule of thumb is that using an iterator is both fastest+safest (in any lang), and apparently within one of these iterator-style loops the variable is an _alias_ for the array element.

lib/LaTeXML/Core/Parameters.pm
dginev: Well, it's north of 4,000 pointless function calls... and it's a lot cheaper to check the hash value than to call *and* check. `readArguments` is called quite often as you can imagine.

Let me get a number from the parallel make test, comparing the two... +2 seconds in the parallel make test. Not the end of the world, my first naive stab *doubled* the time of make test :> But it's such an easy optimization that it's kind of a shame not to do it...

dginev: ... but 2 seconds aren't that big a deal, committed!

dginev:
Thanks for the feedback! I now also found a small doc as a cleveref test, and found a tiny glitch in the binding, so making another small patch, adding the test and this should be good to go.

PR #1387, opened 2020-11-15
[arXiv patches] Stubs for a few extra pdftex primitives, bound by expl3

dginev:
I remembered engrafo has some [latexml-running tests](https://github.com/arxiv-vanity/engrafo/tree/master/tests/integration) for arxiv-vanity and tried to run them as sources of potential regressions yesterday. Still going through the last few, mostly OK.

A paper that was a bit "too loud" with pointless errors was [1803.07574](https://github.com/arxiv-vanity/engrafo/tree/master/tests/integration/papers/1803.07574), mostly due to expl3 wanting to bind some extra pdftex primitives which we hadn't added stubs for.

So I added stubs, trying to be at least a little precise with some of the argument specs (but not with all, just enough to get the paper+expl happy). I also added a missing `{acknowledgements}` definition to the rare class file of the article. Moving to the other tests...

### DIFF
--- a/lib/LaTeXML/Package/pdfTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/pdfTeX.pool.ltxml
@@ -134,8 +134,12 @@ DefRegister('\pdfretval'               => Number(0));
 # \pdfximage [ image attr spec ] general text (h, v, m)
 # \pdfrefximage object number (h, v, m)
 # \pdfannot annot type spec (h, v, m)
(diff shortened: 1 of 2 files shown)


dginev:
(aside: just finished examining the other suspicious files, not all engrafo tests pass under pdflatex, and latexml mirrors the errors well. So this should be the first and last PR from that side of things)

brucemiller:
Looks plausible :>

PR #1388, opened 2020-11-15
Less restrictive of recursion (usually redundant svg:g)

brucemiller:
Block recursion warnings in a few places where we traverse the XML tree from top-to-bottom, which can be deep, but without cycles.  Also increase the XSLT recursion limit.  These are mostly due to redundant `svg:g` elements, which we'll want to fix, but the results are often acceptable anyway.

Probably will be following up with a couple of semi-related patches.

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -162,9 +162,9 @@ sub canContain {
 sub canContainIndirect {
   my ($self, $tag, $child) = @_;
   my $model = $$self{model};
-  $tag = $model->getNodeQName($tag) if ref $tag;          # In case tag is a node.
+  $tag   = $model->getNodeQName($tag)   if ref $tag;      # In case tag is a node.
   $child = $model->getNodeQName($child) if ref $child;    # In case child is a node.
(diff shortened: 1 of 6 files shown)


dginev:
Looks plausible :>

brucemiller:
Running into some nasty issues processing the bibliography of 2010.15968.  Firstly, there's belt&suspenders handling of the log, so it's really hard to see what's happening within the processing of the bib.  Having savagely hacked that, it appears to be another belt&suspenders with both `includestyles` set, and since the packages declared in PI's are loaded using `preload`, they're getting processed raw, even when the main document didn't ... and in fact, couldn't have.  Just FYI; Gotta take a break from debugging further.

brucemiller:
It makes sense for `preload` from the commandline to be assertive and use `includestyles`, but internally, from `MakeBibliography`, this is pretty severe. Not sure how to distinguish these cases.  Why does `MakeBibliography` (also) enable `includestyles` in the recursive call?

This only affects the one arXiv paper (that I know of), but makes it fatal.

brucemiller:
This should recover a percent; we can deal with the log/includestyles issue later.

PR #1390, opened 2020-11-16
Windows CI via CircleCI

dginev:
Took me an ungodly amount of time, but I now have CircleCI passing the basic latexml tests under MikTeX on Windows.

Without caching this takes 19 minutes, and seemingly "740 credits". Since my weekly quota is 2500 credits, this would just be 3-4 builds per week, which is ... underwhelming? But we mostly care about windows regressions around release time, and having it run once a week is fine too.

Caching saves about 2 minutes of setup, but that leaves us with 17 mins of runtime... In any case, it works, with the main remaining downside of being unable to silence the ignorable warnings of miktex through kpsewhich.

Here is a link to the last build:
https://app.circleci.com/pipelines/github/dginev/LaTeXML/121/workflows/49c9f338-a001-4f4f-9ff0-8b0e61de6a9c/jobs/124

Not the most pressing of upgrades, but it's good to be done setting it up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,9 +40,9 @@ jobs:
           command: $env:Path="C:\Program Files\MiKTeX\miktex\bin\x64\;C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;$env:Path";
               gmake
       - run:
-          name: gmake test 9
+          name: gmake test 8
           command: $env:Path="C:\Program Files\MiKTeX\miktex\bin\x64\;C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;$env:Path";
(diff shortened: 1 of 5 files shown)


tkw1536:
FYI, GitHub Actions could do this as well (they offer Mac + Windows VMs). Although I am not sure how difficult it would be to set up.

dginev:
Thanks, could be quite quick to double-check that actually... The reason this PR was very painful was having to debug MikTeX without access to the machine. Rewriting this file in the GH actions style should be something I can try quickly...

brucemiller:
I guess I ignored this too long; If it's still of interest? I'll accept...

dginev:
The CI idea is, but not Circle-CI. I should reroute here and try the github actions implementation of the same idea, since the infinite quotas are worth the trouble.

dginev:
Will reopen with the Github actions approach soon, this can be made obsolete for now.

Issue #1150, opened 2019-05-30
revtex4 empty citations regression


dginev:
Would love us to re-patch this before the next major rerun.

Example:
https://corpora.mathweb.org/preview/arxiv_testbench/tex_to_html/0704.3193?style=simple

snippet - notice the empty parentheses :
```
between the first and third atoms CHARMM22 ().
```

Also, the previews conveniently include the conversion logs at the end of the document, so feel free to scroll to the bottom for the bit that loads revtex4. I recall the last time Michael reported this same issue for revtex articles, but I can't immediately remember where the patch landed.

dginev:
For the record, when I generate the PDF of that article locally The citation comes out as numeric style `[20]`.

![image](https://user-images.githubusercontent.com/348975/58636652-cb6fa700-82be-11e9-8e10-dc835e111531.png)

dginev:
In turn typeset in the bibliography as:

![image](https://user-images.githubusercontent.com/348975/58636735-f659fb00-82be-11e9-9e90-2c94f1e149ce.png)

dginev:
A minimal example waiting to become a test can be seen at my working branch here:
https://github.com/dginev/LaTeXML/commit/714b4a68a067c363bc7c4b2f9604efc1c8de2823

brucemiller:
Probably the most correct solution would be to defer setting the style till the end, but it's going to be pretty messy to unravel the current setup.  However, it's easy to recognize the case after the fact, and avoid adding the parens (d1fd81271156588356672fd092f42df949f047d6); this should work for now.  Thanks!

PR #1392, opened 2020-11-17
[arXiv patches] Maintain the empty () hack for revtex4-1 bibs with no year

dginev:
I looked at the warning documents quickly and spotted a visual regression that we had addressed, before in #1150 , the empty `()` in citations without a year, mostly in the arXiv case of bbl-only information.

At some point the delimiter changed from ` year phrase2` to `{}year{}phrase2` and the guard wasn't migrated, so it didn't take effect. I've updated that regex and the following minimal-ish example converts more acceptably again:

```tex
\documentclass{revtex4-1}
\begin{document}

Abdul-Rahman et al. in arXiv:1901.09297~\cite{Abdul-Rahman2019} provided an elegant approach and proved analytically ...

\begin{thebibliography}
\bibitem{Abdul-Rahman2019}
 H. Abdul-Rahman, M. Lemm, A. Lucia, B. Nachtergaele, and A. Young,  A Class of Two-Dimensional AKLT Models with a Gap, arXiv:1901.09297.
\end{thebibliography}
\end{document}
```

### DIFF
--- a/lib/LaTeXML/Post/CrossRef.pm
+++ b/lib/LaTeXML/Post/CrossRef.pm
@@ -559,10 +559,10 @@ sub make_bibcite {
           # HACK! Avoid empty () from situations where we've set the show (CITE_STYLE) too early
           # and don't actually have author-year information!
           my $n = $1;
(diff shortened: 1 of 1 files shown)


dginev:
Ah, that wasn't the PR, it was #1150 , updating reference.

PR #1394, opened 2020-11-18
include bugtracker link in the META file so MetaCPAN will link to it

szabgab:


dginev:
Thanks for suggesting this @szabgab ! Looks reasonable to me.

Do you happen to know why metacpan isn't just defaulting any github-based repository to the `/issues` tracker link, when no tracker is provided? Seems like a more powerful PR, you add that once to metacpan and can enable all github-based packages at once.

brucemiller:
Thanks for the patch! Just a moment too late for 0.8.5 :>

@dginev Not everybody actually uses the github Issues for
their bugtracking (though probably most do)

szabgab:
I seem to recall a discussion ages ago in which people were against this idea. I thin the same people were also against the the idea to be able to opt-out from RT. So each distribution will have one.

The PR you suggest would be probably easier, convincing the people probably not.

dginev:
As long as the discussion has been had, that's fine. It felt like an obvious point to raise (now that github is as ubiquitous as it is, and its issues are standard in projects in all sorts of programming languages). Thanks for the PR!

PR #1396, opened 2020-11-22
Authorea fork parity reminder, enumitem.sty question

tervenar:
Add the following package [https://github.com/jbezos/enumitem/blob/master/enumitem.sty](enumitem).

### DIFF
--- a/lib/LaTeXML/Package/SIunits.sty.ltxml
+++ b/lib/LaTeXML/Package/SIunits.sty.ltxml
@@ -1,0 +1,29 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# |  SIunits                                                              | #
+# | Implementation for LaTeXML                                          | #
+# |=====================================================================| #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
+# |  United States Government & not subject to copyright in the US.     | #
+# |---------------------------------------------------------------------| #
+# | Thanks to the arXMLiv group for initial implementation              | #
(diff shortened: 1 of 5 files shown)


dginev:
Hey @tervenar thanks for the PR, but it is quite confusing what exactly is going on here and what the intention of your request was?

First, I see the PR contains four of my old commits from 2016, indeed from the mainline `Authorea` fork of latexml and its master branch. I recall adding them there because we wanted to have the features up-and-running on a short notice for Au users even before they got ready for latexml inclusion. We should have a chat with @brucemiller about what may be needed to develop extra to get them into master.

As to `enumitem.sty` which is the title of the PR, that package isn't in the commits diff, so I am assuming this is a feature request? As it happens latexml already has a binding available for that at:
> ./lib/LaTeXML/Package/enumitem.sty.ltxml

So you may want to elaborate a bit on what exactly your motivation for the PR today was and the direction in which you'd like us to improve. Thanks for opening it in any case, it's good to remind me we had unfinished improvements at the Authorea fork that we should polish and merge into master, thanks!! Had completely slipped my mind.

brucemiller:
We already have bindings for `csquotes` and `siunitx`.  Is `SIunits` compatible with `siunitx` (the latter is apparently an enhanced version of the former)?  If so, we probably can define a binding just by loading `siunitx`.  `hepunits` is apparently based on `siunitx`, and a binding could probably leverage that binding (so you'd get good semantics, to boot)

brucemiller:
We certainly can't accept this PR, but it's a good reminder for
the two packages I mentioned (and although they aren't all that
high on arXiv).

PR #1404, opened 2020-12-09
Update to support noted deprecations from latest IEEEtran.cls

dginev:
One case of undefined `{keywords}` in arXiv was due to IEEEtran deprecating its older aliases, but older arXiv sources still using them (naturally).

An example article is cs0010013, downloadable [here](https://corpora.mathweb.org/entry/import/1824861).

I added the entire deprecation block from the end of [IEEEtran.cls](http://mirrors.ctan.org/macros/latex/contrib/IEEEtran/IEEEtran.cls), hopefully catching some of the other spurious errors.

### DIFF
--- a/lib/LaTeXML/Package/IEEEtran.cls.ltxml
+++ b/lib/LaTeXML/Package/IEEEtran.cls.ltxml
@@ -377,8 +377,10 @@ Let('\iedlabeljustifyr', '\IEEEiedlabeljustifyr');
 Let('\QED',       '\IEEEQED');
 Let('\QEDclosed', '\IEEEQEDclosed');
 Let('\QEDopen',   '\IEEEQEDopen');
-Let('\proof',     '\IEEEproof');
-Let('\endproof',  '\endIEEEproof');
+DefConstructor('\qed',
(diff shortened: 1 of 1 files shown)


dginev:
The changes also fix a number of undefined cases of `{proof}` in arXiv, and I also just added a patch to an old use of `\qed`, e.g. for [1102.3390](https://arxiv.org/abs/1102.3390).

dginev:
Of course why should it be easy when things can be complicated. Different versions of IEEEtran end up used with different forms of keywords - both `\keywords{}` macro and `{keywords}` environment, analogously to the problem we once had with `\abstract`. So I adapted that treatment and tested with the two invocations, which seems to be working fine. Now also in the additional (newer) `1804.01512` article.

dginev:
Done debugging here. I ended up touring the `undefined:{proof}` report through the ages of arXiv, and tested sources all the way upto October 2020. Which ended quite rewarding - it both caught the issue with the dual form of keywords, as well as a couple of additional reasons for an undefined proof that are in adjacent PRs not related to IEEEtran.

So all in all, I feel quite good about rerunning that report after these fixes.

PR #1407, opened 2020-12-09
add proof support in OmniBus

dginev:
Oversight on my part last time I extended OmniBus, I missed {proof}, which has 0.53% reported at arXiv. Still checking if other packages need it, but the two papers I double-checked were falling back exactly on OmniBus.


dginev:
Examining the undefined `{proof}` errors further, at:

And `1004.1430` in particular, I've realized that arXiv just has most variations, and OmniBus should just go ahead and be ready to load amsmath whenever any of the common ones get triggered. So I have extended with another handful of envs, and got another couple of articles upto Warning status.

brucemiller:
Just a random tangent (poking at a couple of fatals); Is there likely any harm to require `aas_macros` in `OmniBus` to partly cover various unknown astronyomy styles?

dginev:
Looks very useful and very harmless. Let's throw it in!

brucemiller:
Cool!

PR #1408, opened 2020-12-09
Ignore spaces in args for package and class loader macros

dginev:
This is fundamentally very simple, but still kind of icky, since I am not sure where the perfect patch would be located.

The minimal example (found in arXiv) is the following valid use of `\usepackage`:
```tex
\documentclass{article}
\usepackage{ams thm}
\begin{document}
test
\end{document}
```

Turns out the arguments ignore/discard the spaces. Since we are handling the argument parsing for package and class loading macros in the `afterDigest` perl subroutines, I went ahead and changed a bunch of them to actively discard spaces before passing the names on into the Package.pm loader machinery. Could fix a number of weird missing style files? And ideally is a no-op for anything else.

I also used a tight imperative for loop instead of the multiple grep/map calls, since I keep remembering that `map` is much slower than `for` in Perl.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -51,7 +51,9 @@ DefConstructor('\documentclass OptionalSemiverbatim SkipSpaces Semiverbatim []',
   afterDigest  => sub {
     my ($stomach, $whatsit) = @_;
     my $options = $whatsit->getArg(1);
-    LoadClass(ToString($whatsit->getArg(2)),
+    my $class   = ToString($whatsit->getArg(2));
+    $class =~ s/\s+//g;
+    LoadClass($class,
       options => [($options ? split(/\s*,\s*/, ToString($options)) : ())],
       after   => Tokens(T_CS('\AtBeginDocument'), T_CS('\warn@unusedclassoptions')));
     return; });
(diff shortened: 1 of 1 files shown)


dginev:
Also, because amsthm couldn't load, this translated into errors for undefined `{proof}`, which is how I stumbled onto this issue. Bowling pin action...

brucemiller:
Hopefully there're no valid spaces in names! :>

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -792,9 +795,11 @@ DefConstructor('\usepackage OptionalSemiverbatim Semiverbatim []',
...
     my $options  = $whatsit->getArg(1);
     my $packages = $whatsit->getArg(2);
-    my @pkgs     = grep { $_ } grep { !/^\s*%/ } split(/\s*,\s*/, ToString($packages));
     $options = [($options ? split(/\s*,\s*/, (ToString($options))) : ())];
-    map { RequirePackage($_, options => $options) } @pkgs;
+    for my $pkg (split(',', ToString($packages))) {
+      $pkg =~ s/\s+//g;
+      next if !$pkg || $pkg =~ /^%/;
+      RequirePackage($pkg, options => $options); }
     return });
```
brucemiller, line 803: Very strange.  With https://corpora.mathweb.org/entry/import/2447837 I'm having suddenly most packages not load (even though they end up in PI's), so that the `\usetikzlibrary` line fails undefined.  It's so nonsensical, I can't pin down what's happening. Seems like a Perl bug, or even a memory error (I even rebooted).  Do you see anything odd?

dginev: I think I just found a fix for this, but it is so sinister! Found it independently too. PR incoming...

dginev: Linking to PR to keep the issue connected: #1411

PR #1411, opened 2020-12-11
espcrc2.sty and Safe use of subs in bindings

dginev:
The `espcrc` (1 and 2) part of the PR is easy, just a very old style used by arXiv that adds `\address`. 

The second part of the PR hit a phenomenally sinister bug, which hopefully was freshly introduced by "better" use of `RequirePackage`. Though I am _still_ not sure if it is fresh or old, and the exact mechanics behind the last steps of it.

A very good testing file is [cs/0309037](https://corpora.mathweb.org/entry/import/2272595) ran with includestyles, e.g.
```
latexmlc test.zip --dest=test.html --includestyles
```

If you observe the log and the raw loaded `aadebug.cls`, you will see a report of only **the first** of the following calls to `\RequirePackage`:
```tex
\RequirePackage[latin1]{inputenc}
\RequirePackage[T1]{fontenc}
\RequirePackage{palatino}
\RequirePackage[dvips]{graphicx}
\RequirePackage[usenames,dvipsnames]{color}
%\RequirePackage[pdftex]{hyperref}
\RequirePackage{hyperref}
```

Jumping in the code and debugging for a while with prints, I can also report the other macros are indeed correctly processed, called, their `beforeDigest` is called, but crucially their `afterDigest` subroutine _isn't_.

In particular the two lines you will see changed in the diff of this PR related to
```perl
my $post = $$self{afterDigest};
```

Printing a `Dumper` of the post variable (in master) revealed that while the first call to `\RequirePackage`'s afterDigest was perfect, in the second call (and all following), the post variable contained `[ undef ]` as its content.

I tried tinkering around some more and tried to isolate a minimal example, but the code is actually robust in isolation. My best theory right now is that because in the case of this paper we have a nested call to InputDefinitions (inputenc.sty -> latin1.def), the nested interplay _somehow_ leads to that afterDigest value getting erased from the Primitive definition. This could be completely off the mark and wrong.

However, by serendipity and chance, fixing this is actually much easier than understanding the bug. If one avoids using the `$$self` field directly but instead moves the subroutine items into a new array, as in:
```perl
my @post = @{ $$self{afterDigest} || [] };
```

This error never occurs. Which makes sense because replacing the sub with `undef` in `@post` will **not** propagate to the Primitive definition itself, it will just locally modify that array.

Bizarre and complicated. I will try hunting down the exact line that erases the field tomorrow, but the patch seems very solid when testing.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -445,6 +445,7 @@ lib/LaTeXML/Package/epsf.tex.ltxml
 lib/LaTeXML/Package/epsfig.sty.ltxml
 lib/LaTeXML/Package/epstopdf.sty.ltxml
 lib/LaTeXML/Package/esint.sty.ltxml
+lib/LaTeXML/Package/espcrc.sty.ltxml
 lib/LaTeXML/Package/etex.sty.ltxml
 lib/LaTeXML/Package/etoolbox.sty.ltxml
 lib/LaTeXML/Package/eucal.sty.ltxml
--- a/lib/LaTeXML/Core/Definition/Constructor.pm
+++ b/lib/LaTeXML/Core/Definition/Constructor.pm
@@ -102,10 +102,10 @@ sub invoke {
     my $value = $props{$key};
     if (ref $value eq 'CODE') {
       $props{$key} = &$value($stomach, @args); } }
-  $props{font}    = $font                           unless defined $props{font};
-  $props{locator} = $stomach->getGullet->getLocator unless defined $props{locator};
-  $props{isMath}  = $ismath                         unless defined $props{isMath};
-  $props{level}   = $stomach->getBoxingLevel;
+  $props{font}        = $font                           unless defined $props{font};
+  $props{locator}     = $stomach->getGullet->getLocator unless defined $props{locator};
+  $props{isMath}      = $ismath                         unless defined $props{isMath};
+  $props{level}       = $stomach->getBoxingLevel;
   $props{scriptlevel} = $stomach->getScriptLevel if $ismath;
   # Now create the Whatsit, itself.
   my $whatsit = LaTeXML::Core::Whatsit->new($self, [@args], %props);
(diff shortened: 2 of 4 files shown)


dginev:
arXiv link to source since the cortex links are a bit fragile for downloading from github at the moment: https://arxiv.org/e-print/cs/0309037 (you need to be at the web page with a valid captcha proof in your browser...)

brucemiller:
The corrupting of the `afterDigest` array kinda correlates with the suspicion I had, although I was probably mostly on the wrong specific track. I was concerned that using `for` vs `map` (or even `foreach`) was exposing that wonderful Perl feature where if you modify the right (or wrong) variable within a loop (or call or...) that things outside the (apparent) block get modified as well.

dginev:
Ah, good connection - indeed that change may have been the trigger. It's a relief the debugging pieces are falling in place.

To end on an upbeat note, `for` may be leading to corrupting the data, but it does it _faster_! :>

brucemiller:
Thanks for that! :>

Although I had also been thinking that we'd need the accumulated speedup from a full arXiv run to make up the time we spent debugging.  I think I like your observation better.

brucemiller:
I'll merge, since it looks plausible.  Is there an implied reason to generally prefer for, foreach, map?  Certainly the speed of map isn't a concern when it iterates over loading files, but other times it might.

dginev:
Thanks for merging!

I am mostly working on memory from the last Big Benchmarking expedition I went onto years ago, and maybe I should recheck if the newest perls are still suffering the difference I had spotted then.

The basic rule of thumb was that functional syntax is expensive in perl because it does *all* intermediate allocations you may imagine from the written code. Meanwhile, compiled functional languages (haskell, scala, even functional idioms in things like Rust and C++), manage to compress the execution code massively from what was written by the programmer, often achieving "zero cost abstraction".

Since in perl you pay for everything, you end up wanting to do less. So whenever the purpose of a map is to iterate over a list (rather than to explicitly build a new list), it can be more efficient to just do the for/while loop directly. And if memory serves me well, it gets noticeable exactly when you stack multiple functional blocks together `grep {} map {} grep {} map {} ...`, since each block will do its own array-related allocations. Meanwhile having a single `for` loop doing multiple actions on the same item can end up cheaper.

I won't go as far as to say we _need_ a refactor, and this shouldn't be the biggest of issues... Sorry for the new bugs!

brucemiller:
Right; so prefering for(each) over map if you're not accumulating the 
result makes sense.

Googling around suggests that for is the same as foreach, except when it 
isn't? Sometimes folks are suggesting for has more of the "danger" than 
foreach, but I don't know if they're informed.

Stylistically, I tend to prefer foreach when iterating over a list, and
for(init,test,step) when indexing is involved.  Just seems clearer 
intent, for my tastes.

Issue #1244, opened 2020-03-09
TeX drawing inside equations


dginev:
We should ideally have a decent support for visual snippets inside equations (see StackOverflow link below for a mathematically reasonable example). This requires non-obvious interplay between MathML and SVG, which may in turn require the specifications to provide additional guidance.

Thanks to Peter K Sweby for requesting the feature on the mailing list.

Example 1:
```tex
\[
w(x,y)=\mbox{\tikz{\draw[line width=1pt] (0,0) rectangle(30pt,40pt);}}.
\]
```
Example 2:
![image](https://user-images.githubusercontent.com/348975/76262542-0251f080-6233-11ea-92d4-af3330327aae.png)

https://tex.stackexchange.com/questions/179035/tikz-figure-inside-equation
```tex
\documentclass{article}
\usepackage{tikz}

\def\straight{\tikz[baseline=.1ex]{
\fill (0,0) circle (1pt) coordinate (A);
\fill (0,1.5ex) circle (1pt) coordinate (B);
\draw (A)--(B);}
}

\def\curved{\tikz[baseline=.1ex]{
\fill (0,0) circle (1pt) coordinate (A);
\fill (0,1.5ex) circle (1pt) coordinate (B);
\useasboundingbox (0,0) rectangle (.7ex, 1.5ex);
\draw (A) .. controls +(1ex,-1ex) and +(1ex, 1ex) .. (B);}%
}

\begin{document}

\[ P(\straight) + P(\curved) = p \]

\end{document}
```

brucemiller:
Not sure when this was last tried, but these 2 examples work for me, at least to the extent of embedding tikz svg into math (the 2nd examples are missing the connecting lines, but they do that whether embedded in math or not --- ie. a separate tikz bug).

dginev:
Interesting, the examples are indeed "modestly successful" with the latest master. The first question was to even ask what is the best way to stack MathML and SVG together, but it sounds like you've thought about this before? The markup generated is an `<mtext>` element that hosts the `<svg>` drawing, which is probably the best MathML 3 has to offer.

If that is the case, this may be closeable already - I had forwarded the issues as a question from the mailing list, to which I didn't personally have an answer to at the time.

brucemiller:
A quick experiment in the inspector, it looks like the connecting paths only missing a `stroke` attribute (or got nested wrong in the `svg:g`).  Possibly an easy fix?

dginev:
[...]

PR #1413, opened 2020-12-15
Pgf cleanup

brucemiller:
This PR runs a cleanup of `svg:g` elements after they close to check for redundancies, or when they are completely masked by a child, or if a child establishes a complete set of the same attributes, and so can be pulled out.  This vastly reduces both the size and depth of tikz/pgf documents, particularly when run with post-2017 texlives.

In the process, a couple of other tikz regressions were suddenly discoverable and were fixed.  More can probably be found and fixed?  Also, since this fixes some tikz regressions that didn't cause outright failure, but very poor rendering, a rerun of arXiv tikz might be in order (when this is complete).

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -1815,7 +1815,6 @@ sub wrapNodes {
     if ($self->isOpen($n)) {
       $leave_open = 1;
       last; } }
-  print STDERR "WRAPPING with $qname, but keeping upen!\n" if $leave_open;
   my $model  = $$self{model};
   my $parent = $nodes[0]->parentNode;
   my ($ns, $tag) = $model->decodeQName($qname);
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -851,12 +851,12 @@ sub closeToNode {
     $n        = $n->parentNode; }
   if ($t == XML_DOCUMENT_NODE) {    # Didn't find $node at all!!
     Error('malformed', $model->getNodeQName($node), $self,
-      "Attempt to close " . Stringify($node) . ", which isn't open",
+      "Attempt to close to " . Stringify($node) . ", which isn't open",
       "Currently in " . $self->getInsertionContext()) unless $ifopen;
     return; }
   else {                            # Found node.
     Error('malformed', $model->getNodeQName($node), $self,
-      "Closing " . Stringify($node) . " whose open descendents do not auto-close",
+      "Closing to " . Stringify($node) . " whose open descendents do not auto-close",
       "Descendents are " . join(', ', map { Stringify($_) } @cant_close))
       if @cant_close;               # But found has intervening non-auto-closeable nodes!!
     $self->closeNode_internal($lastopen) if $lastopen; }
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -722,7 +722,7 @@ sub openText {
     (($t == XML_DOCUMENT_NODE)    # Ignore initial whitespace
     || (($t == XML_ELEMENT_NODE) && !$self->canContain($node, '#PCDATA')));
   return if $font->getFamily eq 'nullfont';
-  print STDERR "Insert text \"$text\" /" . Stringify($font) . " at " . Stringify($node) . "\n"
+  print STDERR "openText \"$text\" /" . Stringify($font) . " at " . Stringify($node) . "\n"
     if $LaTeXML::Core::Document::DEBUG;
 
   # Get the desired font attributes, particularly the desired element
@@ -746,7 +746,9 @@ sub openText {
       last if ($$self{model}->getNodeQName($n) ne $elementname) || $n->getAttribute('_noautoclose');
       $n = $n->parentNode; }
     $self->closeToNode($closeto) if $closeto ne $node;    # Move to best starting point for this text.
-    $self->openElement($elementname, font => $font, _fontswitch => 1) if $bestdiff > 0; # Open if needed.
+    $self->openElement($elementname, font => $font,
+      _fontswitch => 1, _autoopened => 1)
+      if $bestdiff > 0;                                   # Open if needed.
   }
   # Finally, insert the darned text.
   my $tnode = $self->openText_internal($text);
@@ -760,7 +762,7 @@ sub openText {
 sub openElement {
   my ($self, $qname, %attributes) = @_;
   NoteProgress('.') if ($$self{progress}++ % 25) == 0;
-  print STDERR "Open element $qname at " . Stringify($$self{node}) . "\n" if $LaTeXML::Core::Document::DEBUG;
+  print STDERR "openElement $qname at " . Stringify($$self{node}) . "\n" if $LaTeXML::Core::Document::DEBUG;
   my $point = $self->find_insertion_point($qname);
   $attributes{_box} = $LaTeXML::BOX unless $attributes{_box};
   my $newnode = $self->openElementAt($point, $qname,
@@ -776,7 +778,7 @@ sub openElement {
 # This is kinda risky! Maybe we should try to request closing of specific nodes.
 sub closeElement {
   my ($self, $qname) = @_;
-  print STDERR "Close element $qname at " . Stringify($$self{node}) . "\n" if $LaTeXML::Core::Document::DEBUG;
+  print STDERR "closeElement $qname at " . Stringify($$self{node}) . "\n" if $LaTeXML::Core::Document::DEBUG;
   $self->closeText_internal();
   my ($node, @cant_close) = ($$self{node});
   while ($node->nodeType != XML_DOCUMENT_NODE) {
@@ -833,6 +835,7 @@ sub isCloseable {
 # Close $qname, if it is closeable.
 sub maybeCloseElement {
   my ($self, $qname) = @_;
+  print STDERR "maybeCloseNode(int) $qname\n" if $LaTeXML::Core::Document::DEBUG;
   if (my $node = $self->isCloseable($qname)) {
     $self->closeNode_internal($node);
     return $node; } }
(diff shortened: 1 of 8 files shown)


brucemiller:
Probably this is as good as it gets for now (until some alignment work).  It appears to recover all my offline TeXamples cases, and in fact improves the output of several of them.

dginev:
Impressive, all three of my [showcase](https://latexml.mathweb.org/editor) examples convert accurately with the latest state of the PR, fixing even the wrongly rotated triangle. Very cool! I've tested under texlive 2020 just now, while I'm waiting for a new cooler pump for my PC tower which has the texlive 2019 installed.

But definitely an upgraded outcome from master as far as my examples are concerned - neat!

dginev:
I have also verified that the math-with-svg examples from #1244 convert correctly with the PR, and we can close that issue after a merge here, quite nice. Still some minor sizing discrepancies, but those can only be solved with font metrics - the core tikz interpretation appears spot on!

lib/LaTeXML/Core/Document.pm
```diff
@@ -1082,17 +1087,39 @@ sub openText_internal {
...
     print STDERR "Inserting text node for \"$text\" into " . Stringify($point) . "\n"
       if $LaTeXML::Core::Document::DEBUG;
     $point->appendChild($node);
     $self->setNode($node); }
   return $$self{node}; }                                # return the text node (current)
 
+# Since xml text nodes don't have attributes to record the origining box,
+# we need to manage the accumulation of autoOpen'ed boxes
+# Indeed, propogate it to ancestors if they were autoOpened for same cause (box)
+sub appendTextBox {
```
dginev, line 1109: Very interesting technique, clever too. Properly tracking the boxes for auto-opened nodes is super useful, wish I had realized it was missing.

dginev: Any thoughts about tracking this better on the generic level in latexml? Or is the need to bookkeep specific to the pgf binding? I'm still absorbing the fine details.

lib/LaTeXML/Core/Document.pm
```diff
@@ -1082,17 +1087,39 @@ sub openText_internal {
...
+    my $p = $node;
+    # AND, propogate change to autoOpen'd ancestors based on same initial box
+    while (($p = $p->parentNode) && ($p->nodeType == XML_ELEMENT_NODE)
+      && $p->getAttribute('_autoopened')
+      && (($self->getNodeBox($p) || '') eq $origbox)) {
+      $self->setNodeBox($p, $newbox); } }
+  return; }
+
 # Question: Why do I have math ligatures handled within openMathText_internal,
 # but text ligatures handled within closeText_internal ???
```
dginev, line 1124: The above old question about ligatures still sounds curious.

lib/LaTeXML/Core/Document.pm
```diff
@@ -1821,9 +1873,21 @@ sub wrapNodes {
     $self->setNodeBox($new, $box); }
   foreach my $node (@nodes) {
     $new->appendChild($node); }
-  $self->afterClose($new);
+  $self->afterClose($new) unless $leave_open;
   return $new; }
 
+# Check if $node, or any of it's children, are the current node, and thus still "open"
+# Maybe a better way, such as explicitly marking _open ?
```
dginev, line 1880: this is interesting for XML reasons. The recursion could get expensive if someone wraps a big subtree with a new parent, while some of its terminal leaf nodes is still open.

I would even be tempted to suggest avoiding the checks altogether and requiring a separate method for firing the `afterClose`. If we think about unsuspecting web developers using a `wrapNodes` method, we would only expect a re-parenting of a set of nodes, without extra side-effects. The method is a little tricky to think about (for me) as it is. Maybe even changing it to:
```perl
wrapNodes($self, $parent_node, @nodes_to_reparent)
```

would be safer, while adding a higher-level method which could work with a qname, e.g.
```perl
# fires ->afterClose
wrapClosedNodes($self, $qname, @nodes)

# does not fire ->afterClose
wrapOpenNodes($self, $qname, @nodes)
```

Recursing to check if open every time may get expensive, looking at where it is currently used:
```
lib/LaTeXML/Core/Rewrite.pm
187:    $node = $document->wrapNodes('ltx:XMWrap', @nodes); }
204:    my $wrapper = $document->wrapNodes('ltx:XMWrap', @nodes);
206:    $node = $document->wrapNodes('ltx:XMDual', $wrapper);

lib/LaTeXML/Core/Alignment.pm
590:  $document->wrapNodes('ltx:thead', @heads) if @heads;
591:  $document->wrapNodes('ltx:tbody', @rows)  if @rows;
592:  $document->wrapNodes('ltx:tfoot', @foots) if @foots;

lib/LaTeXML/Core/Document.pm
431:        my $text = $self->wrapNodes($elementname, $child);
2406:=item C<< $node = $document->wrapNodes($qname,@nodes); >>

lib/LaTeXML/Package/TeX.pool.ltxml
2018:      push(@newnodes, $document->wrapNodes('ltx:p', @n)); }
3388:        $text = $document->wrapNodes('ltx:text', $text); }
3477:            $document->wrapNodes('ltx:XMText', $m); }
```

To my intuition it may even be better to require from to the binding developer to call `->afterClose` explicitly, removing it from the current method, compared to deeply recursing in the children for every call...

lib/LaTeXML/Package/pgfmath.code.tex.ltxml
```diff
@@ -196,39 +197,41 @@ DefMacro('\pgfmathmod@ pgfNumber pgfNumber', sub {
...
-    pgfmathresult(rad2deg($_[1])); return });
+    return pgfmathresult(rad2deg($_[1])); });
 DefMacro('\pgfmathatan@ pgfNumber', sub {
-    pgfmathresult(atan($_[1])); return });
-DefMacro('\pgfmathatantwo@ pgfNumber', sub {
-    pgfmathresult(atan2($_[1], $_[2])); return });
+    return pgfmathresult(rad2deg(atan($_[1]))); });
+DefMacro('\pgfmathatantwo@ pgfNumber pgfNumber', sub {
+    return pgfmathresult(rad2deg(atan2($_[1], $_[2]))); });
+# Reverse of tikz, but...
```
dginev, line 207: ...but? Would support a full comment :>

lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
```diff
@@ -448,60 +438,112 @@ DefConstructor('\lxSVG@discardpath@clipped{}',
...
+        $mergeable = 0; } }
+    if ($mergeable) {
+      foreach my $key (sort keys %av) {
+        $nodeattr{$key} = $av{$key};
+        $node->setAttribute($key => $av{$key}); }
+      $nmerged++;
+      $document->unwrapNodes($c); } }
+###  Debug("COLLAPSE empty=$nempty, redundant=$nredundant, merged=$nmerged, popped=$npopped, pushed=$npushed")
+###    if ($nempty || $nredundant || $nmerged || $npopped || $npushed);
+  return; }
```
dginev, line 546: Cool cleaning subroutine! Can't trace all the logic just by reading through, since lots of markup weed-whacking is going on. Would have been a candidate for a unit test in a different world... But I guess adding one tikz `.tex` test which uses it would do too!

lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
```diff
@@ -1031,7 +1095,6 @@ DefConstructor('\lxSVG@halign BoxSpecification',
     my $template = LaTeXML::Core::Alignment::Template->new((@nonreps ?
           (columns => [@nonreps], repeated => [@cols])
         : (columns => [@cols])));
-    #  print STDERR "Template = ".Stringify(Tokens(@template))."\n => ".$template->show."\n";
     # Now read & digest the body.
     # Note that the body MUST end with a \cr, and that we've made Special Arrangments
     # with \alignment@cr to recognize the end of the \halign
```
dginev, line 1100: The `pgfsys-latexml.def.ltxml` binding file has gotten a lot simpler and cleaner, very very nice.

dginev (approved):
Looks like a pretty great refactor+upgrade effort!

Thanks for doing all the horrid debugging work to arrive at the PR. I've left some minor code comments, and ran some good-looking tests, so very uplifting to review. The one bit that worries me is the recursive check for `isOpen`, but that's about it.

lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
brucemiller: There was a lot of "Two wrongs makes a right" going on there. Which is challenging to debug, if you "fix" an "obvious" mistake and it gets worse!

lib/LaTeXML/Core/Document.pm
brucemiller: It eventually did get more generic, as I had to add a similar box patch-up when adding xml elements to auto-opened nodes.  It is also generic in that it applies to all xml, but it was driven by pgf being the application that depends most on having the right boxes attached for sizing estimates.

lib/LaTeXML/Core/Document.pm
brucemiller: but still a good question :>

lib/LaTeXML/Core/Document.pm
brucemiller: Concern noted, but one motivation for the "fully automatic" was *exactly* the poor ole unsuspecting developer wondering why the close-node daemon was sometimes running and sometimes not. While it may be expensive to sniff it out from within `Document`, it's hard to figure out at the binding level whether it's needed or not, and then to get it done at the right time.

I'd rather improve how the check is done: The suggested alternative was off-hand; probably marking `_closed` would be safer than `_open`, since it would be hard to keep track of the latter, but with the former, once closed it's done.  Still, it would require more testing & exploration than fits in this PR; might be better to come back to this topic separately with some appropriate benchmarks.

Another aside: I do like the built-in automation here. And with the box management, I was hoping that a further evolution could possibly clean up a lot of unreadable code used to manage MathForks and such. But that's for another time, as well.

lib/LaTeXML/Package/pgfmath.code.tex.ltxml
brucemiller: no comment

PR #1414, opened 2020-12-16
Stricter get_converter which never overrides a set cache_key

dginev:
I'm briefly helping the MathIR folks from [ARQmath](https://www.cs.rit.edu/~dprl/ARQMath/), and preparing a turnkey solution for daemonized processing of tens of millions of formulas with latexml, notably from StackExchange.

On big part of that is stripping down all overhead one can find in the hand-over from the conversion harness to the LaTeXML.pm `->convert` call. Luckily, that is already mostly taken care of quite nicely in LaTeXML.pm, and some of the trickier bits I can ship separately as `latexmls` enhancements. The one bit that I would like to fine-tune here is allowing for reusing the same `cache_key` without specifying any other option. In latexmlc that would look as, for example:

Initial call with a range of options:
```
latexmlc --cache_key=example --expire=60 --whatsout=math --whatsin=math --pmml --cmml 'literal:\frac{x}{y}'
```

Multiple follow-up calls that only pass in the cache key and reuse a daemonized converter (via latexmls):
```
latexmlc --cache_key=example 'literal:\sqrt{x}'
latexmlc --cache_key=example 'literal:\sqrt{y}'
...
```

Of course I won't be using latexmlc for [the production system](https://github.com/dginev/latexml-rust), since - naturally - that adds a lot of extra overhead. But you can imagine the equivalent HTTP POST request as well:
```
curl -d 'cache_key=example&source=literal:\sqrt{x}' localhost:3334
```
etc.

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -45,7 +45,6 @@ our $LOG_STACK = 0;
 
 # Switching to white-listing options that are important for new_latexml:
 our @COMPARABLE = qw(preload paths verbosity strict comments inputencoding includestyles documentid mathparse);
-our %DAEMON_DB = ();
 
 sub new {
   my ($class, $config) = @_;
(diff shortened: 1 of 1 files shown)


dginev:
Would appreciate a recent-ish merge here (whenever time permits), now that I have my holiday "ducks in a row" for [latexmls](https://github.com/dginev/LaTeXML-Plugin-latexmls). I shipped a v1.5 on that front which allows very direct control over the server pids by a harness process, and if I get the `cache_key` upgrade from the LaTeXML PR here I get to "all systems go" for converting the math of StackExchange with optimal performance.

brucemiller:
I guess this looks reasonable :> Sorry for the delay.

dginev:
No worries at all, thanks for merging! Off to prepare some dockerfiles...

PR #1423, opened 2021-01-02
[arXiv patches] iopart support tweaks

dginev:
Upgrades [cond-mat/0002162](https://arxiv.org/abs/cond-mat/0002162) from Error to Warning status. Small oversights in iopart binding.


dginev:
Some other small tweaks are likely possible to fish out from the [misdefined reports](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/misdefined?all=false), this was from the `#1` entry.

PR #1426, opened 2021-01-05
closer to native latex.ltx, add \abstracts to OmniBus

dginev:
So, um, I hope you don't take this the wrong way...

But people are doing weird stuff with the latex.ltx primitives in arXiv! In particular, [gr-qc/0003053](https://arxiv.org/abs/gr-qc/0003053) uses `\savebox1{arg...}` and `\usebox1`, rather than the officially documented backslash-starting box name.

And.. well, it actually works, because the native latex.ltx code allows it to work!

I'm a little hestient to suggest merging here, but I am wondering what other tests one would need to be convinced the native TeX definition works well. It seems to be improving the arXiv article in question at least, but... lilkely worth discussing :>

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4390,49 +4390,36 @@ DefPrimitive('\newsavebox DefToken', sub {
     DefRegisterI($_[1], undef, Number($n));
     AssignValue('box' . $n, List()); });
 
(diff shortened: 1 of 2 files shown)


dginev:
I threw in a couple of minor latex.ltx internals that I spotted in arXiv, though they're not even significant enough to fix a full article (the one I checked had more serious errors besides these easy bits). But might as well throw it in ...

brucemiller:
Seems worth a try :>

PR #1427, opened 2021-01-06
allow multiple midscripts as nested mover/munder in pmml

### LINKED ISSUE #1425: Upgrade mid-level-scripts logic
0.32% of arXiv has an `unexpected` error that looks rather approachable/resolvable on the surface. But I am unsure what the best PR tactic would be. It comes down to upgrading our script treatments to recognize some workable cases that are currently, well, unexpected.

Minimal math snippet:
```tex
{\sum_a}'
```

As used in arXiv:
```tex
The Hamiltonian is then written as
\begin{equation}
H  =  - J {\sum_{\langle ij \rangle}}' \cos \, (\phi_{i} - \phi_{j})
+ \lambda J \, \cos \, (\phi_0 - \phi_1)
\end{equation}
```
that is equation 7 from [cond-mat/0001095](https://arxiv.org/abs/cond-mat/0001095), and in case you're wondering what the prime means, here's a quote:
> where the prime indicates the omission of the AFM bond from the summation

it is _not_ a derivative, but a decoration - in this one example.

A very short term question is - should we relax the Error to Warning status? Failed math parsing is a Warning, and this seems connected. Or alternatively, can we adapt the trees of these cases to where the post-processor can use them well? The (recently upgraded by me) message in question is `Error:unexpected:multiple-mid-level-scripts`:

image

dginev:
Fixes #1425 , especially the nested stackrel example.

It also avoids the error for the `{\sum_a}'` example, but adds the prime as a middle position script, rather than a post script. That certainly requires a separate patch.

But compared to simply dropping the extra middle position script, I think this PR is a straightforward improvement.

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -874,7 +874,7 @@ sub stylizeContent {
 # These are the strings that should be known as fences in a normal operator dictionary.
 my %fences = (                                                    # CONSTANT
   '('        => 1, ')' => 1, '[' => 1, ']' => 1, '{' => 1, '}' => 1, "\x{201C}" => 1, "\x{201D}" => 1,
-  "\`"       => 1, "'" => 1, "<" => 1, ">" => 1,
+  '`'        => 1, "'" => 1, "<" => 1, ">" => 1,
   "\x{2329}" => 1, "\x{232A}" => 1, # angle brackets; NOT mathematical, but balance in case they show up.
   "\x{27E8}" => 1, "\x{27E9}" => 1,    # angle brackets (preferred)
   "\x{230A}" => 1, "\x{230B}" => 1, "\x{2308}" => 1, "\x{2309}" => 1);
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Post/MathML.pm
```diff
@@ -874,7 +874,7 @@ sub stylizeContent {
 # These are the strings that should be known as fences in a normal operator dictionary.
 my %fences = (                                                    # CONSTANT
   '('        => 1, ')' => 1, '[' => 1, ']' => 1, '{' => 1, '}' => 1, "\x{201C}" => 1, "\x{201D}" => 1,
-  "\`"       => 1, "'" => 1, "<" => 1, ">" => 1,
+  '`'        => 1, "'" => 1, "<" => 1, ">" => 1,
```
dginev, line 877: latexmllint told me to change this, so I did

dginev:
With the PR, the first example produces a reasonable presentation, though not maximally compact:
```
latexmlmath "\displaystyle {\sum_a}'"
```

```xml
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="\displaystyle{\sum_{a}}^{\prime}" display="block">
  <mover>
    <munder>
      <mo largeop="true" movablelimits="false" symmetric="true">∑</mo>
      <mi>a</mi>
    </munder>
    <mo movablelimits="false">′</mo>
  </mover>
</math>
```

I could try to detect this case and pack together...

lib/LaTeXML/Post/MathML.pm
```diff
@@ -984,26 +984,23 @@ sub pmml_script_mid_layout {
...
-        "Multiple mid-level (limit) scripts; extras are DROPPED!", "base: " . $base->toString(),
-        map { @$_ } @$midscripts); }
-    { local $LaTeXML::MathML::NOMOVABLELIMITS = 1;
-      ##### TRY this to block an extra mstyle
-      local $LaTeXML::MathML::STYLE = $base->getAttribute('mathstyle') || $LaTeXML::MathML::STYLE;
-      $base = pmml($base); }
+    local $LaTeXML::MathML::NOMOVABLELIMITS = 1;
+    ##### TRY this to block an extra mstyle
+    local $LaTeXML::MathML::STYLE = $base->getAttribute('mathstyle') || $LaTeXML::MathML::STYLE;
     # Get the (possibly padded) over & under scripts (if any)
```
brucemiller, line 990: You don't really want these 2 locals scoped to the entire routine, but just to formatting the base.

brucemiller:
Except for the comment above, this is actually exactly the needed patch. At least, once the script positions & levels have been correctly determined;  the existing `MathML.pm` code correctly sorts the scripts into pre, mid and post and passes them to the correct handlers; the only issue was the mid handler didn't handle multiple (nested) mid scripts, which you fixed.

However, since I needed to test in the context of the other scriptpos tuning, I simply stole your patch for that PR. Thanks!

dginev:
Excellent, very happy I got some part of it right :>

PR #1429, opened 2021-01-07
More precise \read

dginev:
Got a report from @Jazzpirate that he's seen some subtle differences between the content read by latexml's `\read` and pdflatex's.

So I tried to sync them better, and added a new test. There may be a cleaner approach, but I'm semi-worried it may cascade into too many changes, so I tried to get a PR out to discuss before I get too far into the weeds.

The io.xml has another discrepancy from the PDF after I regenerated it (had more in master), in particular getting the empty middle line in:
```
Line:  “another line spaces ”
Line:  “”
Line:  “Foo!”
```

which was entirely [missing](https://github.com/brucemiller/LaTeXML/blob/master/t/digestion/io.xml#L36) in the master io.xml, but my PR adds an extra spurious `\par` that should be avoided eventually - not sure how yet. The rest ought to be aligned.


lib/LaTeXML/Core/Mouth.pm
```diff
@@ -212,7 +212,7 @@ sub handle_EOL {
   # but it makes nicer XML with occasional \n. Hopefully, this is harmless?
   my $token = ($$self{colno} == 1
     ? T_CS('\par')
-    : ($STATE->lookupValue('PRESERVE_NEWLINES') ? Token("\n", CC_SPACE) : T_SPACE));
+    : (($STATE->lookupValue('PRESERVE_NEWLINES') || 0) == 1 ? Token("\n", CC_SPACE) : T_SPACE));
```
brucemiller, line 215: This doesn't seem to be needed (or wanted?) as far as the fidelity of `\read` is concerned. You always get catcode `CC_SPACE`, but whether it contains a space or cr. It does affect the XML output, though.

lib/LaTeXML/Core/Mouth.pm
```diff
@@ -230,7 +230,12 @@ sub handle_comment {
   $$self{colno} = $$self{nchars};
   my $comment = join('', @{ $$self{chars} }[$n .. $$self{nchars} - 1]);
   $comment =~ s/^\s+//; $comment =~ s/\s+$//;
-  return ($comment && $STATE->lookupValue('INCLUDE_COMMENTS') ? T_COMMENT($comment) : undef); }
+  my $include_comments = $STATE->lookupValue('INCLUDE_COMMENTS');
+  return unless $include_comments;
+  if ($include_comments == 1) {
```
brucemiller, line 235: Subtle! Probably warrants a comment about what 2 signifies?

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -2294,12 +2294,12 @@ DefPrimitive('\read Number SkipKeyword:to SkipSpaces Token', sub {
...
       $stomach->bgroup;
       AssignValue(PRESERVE_NEWLINES => 2);
-      AssignValue(INCLUDE_COMMENTS  => 0);
+      AssignValue(INCLUDE_COMMENTS  => 2);
       my @tokens = ();
       my ($t, $level) = (undef, 0);
       while ($t = $mouth->readToken) {
         my $cc = $t->getCatcode;
-        push(@tokens, $t);
+        push(@tokens, $t) unless $cc == CC_COMMENT;
```
brucemiller, line 2302: The spurious `\par` is due to `@tokens` being empty, even though we read "something".  A solution would be to add an `$ntokens` to the my, initialized to 0; increment it inside the `while`, and then a couple of lines below this code, change push's if test from `@tokens` to `$ntokens` (or code to the same effect).

brucemiller:
I'm missing what the rebase did; was it intentional?

dginev:
@brucemiller intentional yes, and ideally it did nothing material. I got a little worried to be working from behind a diff, so rebased the branch just in case. I am yet to address the rest of the comments, doing so in a bit.

lib/LaTeXML/Core/Mouth.pm
dginev: Indeed, this exact difference aims to reduce the number of line-breaks that reach the XML in a way that is closer to TeX.

So in `io.pdf` you will see:
> Line: “This file exists! ”

and

> Line: “This file exists!fl”

while in the `io.xml` you will see them [with line breaks](https://github.com/brucemiller/LaTeXML/pull/1429/files#diff-22a41af1c2e122c91da0fc36220d5200299d7dda29b9058430bdd29d10d58db9), without this change.

lib/LaTeXML/Core/Mouth.pm
dginev: Indeed, done.

lib/LaTeXML/Package/TeX.pool.ltxml
dginev: Once you pointed this out, it became completely obvious. Funny how that works. Fix is in - thanks!

brucemiller:
This is mergeable, modulo the cr business; Most of the testcase diff is due to that.  I'd prefer to keep the cr's, unless there's a good reason not to.

dginev:
Well the good reason was that it's closer to what pdftex does. I don't know if Dennis' tests will be affected by the subtle difference, but maybe it's fine either way?

brucemiller:
In what way is it closer? Unless I'm missing something, both forms
are essentially the same (the cr & space aren't distinguishable in the pdf)

lib/LaTeXML/Core/Mouth.pm
dginev: @brucemiller are we discussing the comment above that I exemplified with io.pdf? Just making sure I'm not misunderstanding the point.

brucemiller:
Right. The html for both versions looks, to my eyes, identical to each 
other and the pdf.  The XML, well.... which one is more "correct"?

dginev:
Ok, I backed off from changing the cr behavior, double-checked the tests, should be better!

brucemiller:
Cool! Looks good! Thanks!

PR #1432, opened 2021-01-09
Basic support for standalone.sty

### LINKED ISSUE #678: implement standalone.cls/sty
It would be very nice to have `standalone.cls.ltxml` and `standalone.sty.ltxml`, I use it in my tikzinput package, which I can only get out of "image-mode" if I have LaTeXML bindings for `standalone`. 

The former would probably just be a very thin wrapper around `OmniBus.cls.ltxml` and the latter needs to read everything between `\begin{document}` and `\end{document}` (excluding the environment), and nothing else. That should be relatively simple I hope.

dginev:
Fixes #678

At least a basic use of standalone.sty should be supported now.

Example [from the docs](https://ctan.math.illinois.edu/macros/latex/contrib/standalone/standalone.pdf), if you have two files:

- main.tex
```tex
% Main file
% Real document class:
\documentclass{article}
% Use the ’standalone’ package:
\usepackage{standalone}
% Load all packages needed for all sub−files:
\usepackage{tikz}
% Inside the real ’document’ environment
% read the sub−file with ’\input’
\begin{document}
% ...
\begin{figure}
  \input{fig}
  \caption{A subfile}
\end{figure}
% ...
\end{document}
```

- fig.tex
```tex
\documentclass[tikz]{standalone}
\usetikzlibrary{calc}
\begin{document}
\begin{tikzpicture}
  \draw (0,0) rectangle (2,1) node [midway] {Example};
\end{tikzpicture}
% Further ’tikzpicture’ environments are possible which will create further pages.
\end{document}
```

For the basic use a simple binding ended up more manageable than the convoluted TeX internals that we can't support fully yet. I added a few minor pieces that get us a little closer to a raw interpretation of `standalone.sty`, but to get everything correctly I'd need to enable shellescapes, and emulate them accurately - which is more work than the feature itself is worth. Relaxing the usual latex macros in a local scope from a perl binding works just fine for this basic use.

### DIFF
--- a/lib/LaTeXML/Package/standalone.sty.ltxml
+++ b/lib/LaTeXML/Package/standalone.sty.ltxml
@@ -32,7 +32,8 @@ DefPrimitive('\@standalone@documentclass[]{}', sub {
     Let(T_CS('\end{document}'),   T_CS('\@standalone@end@input'));
     return; });
 
-Let('\documentclass', '\@standalone@documentclass');
+AtBeginDocument(sub {
+    Let('\documentclass', '\@standalone@documentclass'); });
 
 #**********************************************************************
 
--- a/lib/LaTeXML/Package/standalone.sty.ltxml
+++ b/lib/LaTeXML/Package/standalone.sty.ltxml
@@ -1,0 +1,39 @@
+# -*- CPERL -*-
+# /=====================================================================\ #
+# |  standalone.sty                                                           | #
+# | Implementation for LaTeXML                                          | #
+# |=====================================================================| #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
+# |  United States Government & not subject to copyright in the US.     | #
+# |---------------------------------------------------------------------| #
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Package/standalone.sty.ltxml
```diff
@@ -0,0 +1,39 @@
...
+    $stomach->bgroup;
+    AssignValue(inPreamble => 1);
+    for my $package (split(",", ToString($packages))) {
+      RequirePackage($package);
+    }
+    Let(T_CS('\begin{document}'), T_CS('\@standalone@start@input'));
+    Let(T_CS('\end{document}'),   T_CS('\@standalone@end@input'));
+    return; });
+
+Let('\documentclass', '\@standalone@documentclass');
```
brucemiller, line 35: Wouldn't this be better deferred, say in a `AtBeginDocument`, in case `\usepackage` comes before the real `\documentclass` (say, preloaded)?

dginev: It would! I'll keep that in mind for future bindings too, thanks for the tip. Now also pushed

PR #1433, opened 2021-01-09
Fix scriptpos

### LINKED ISSUE #1425: Upgrade mid-level-scripts logic

brucemiller:
This PR fixes some corner cases with multiple sub/super/over/under scripts, including when `\limits`, and `\nolimits` are used.  It includes (steals) the patch from #1425 to 

 closes #1427 (supersedes)
fixes #1425


lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -5479,22 +5479,27 @@ DefMathI('\biguplus', undef, "\x{228E}",
...
+  afterDigest => sub { mergeLimits('post'); },
+  properties  => { isSpace => 1 });
+DefConstructorI('\displaylimits', undef, '',
+  afterDigest => sub {
+    mergeLimits((($_[1]->getProperty('mathstyle') || '') eq 'display' ? 'mid' : 'post')); },
+  properties => { isSpace => 1 });
+
+sub mergeLimits {
+  my ($pos) = @_;
+  for (my $i = scalar(@LaTeXML::LIST) - 1 ; $i >= 0 ; $i--) {
```
dginev, line 5495: it's been a while since I saw an old-school for loop! Hope there are no off-by-one errors in there :> Looks like it could also work with:

```perl
for my $box(reverse(@LaTeXML::LIST)) {
  ... }
```

but your version is certainly the faster code.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -5479,22 +5479,27 @@ DefMathI('\biguplus', undef, "\x{228E}",
...
+    my $box   = $LaTeXML::LIST[$i];
+    my $prev  = $box->getProperty('scriptpos') || '';
+    my $level = ($prev =~ /^\w*(\d+)$/ ? $1 : $STATE->getStomach->getScriptLevel || '');
```
dginev, line 5498: minor, but `$STATE->getStomach->getScriptLevel` could probably be called once outside of the loop and reused.

dginev (approved):
Looks pretty great to me, LGTM.

Question: isn't this the type of PR that may have some cascading effects into DLMF? Or is it mergable without rerrunning there?

I have no real complaints, and the long list of erased `scriptpos="mid"` in the test suite look quite appealing.

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: the loop should typically be short anyway

brucemiller:
Yep, already checked dlmf; no consequential diffs, just a lot of the pointless ones you noted.

PR #1434, opened 2021-01-09
Ensure empty dates dont deposit any content in HTML

dginev:
I finally felt brave enough to open a `no_problem` article in arXiv, namely [the HTML of 1504.05855](https://corpora.mathweb.org/preview/arxmliv/tex_to_html/1504.05855?style=simple), and spotted a cosmetic oversight.

The current XSLT allows a `\date{}` to propagate into the HTML as `<div class="ltx_dates">()</div>`, due to a test which checks only if a node exists, but not whether it is empty.

Not my most notable PR ...


brucemiller:
Sure, why not :>

PR #1442, opened 2021-01-17
Modernize HTML5: add viewport meta tag

xworld21:
This is an attempt at modernising the generic HTML5 output.

I have added a meta viewport tag, strictly for the HTML5 output only, with default value `width=device-width, initial-scale=1, shrink-to-fit=no` (taken from [arXiv Vanity](https://www.arxiv-vanity.com/)). This should make the output mobile friendly in most cases. I think it's reasonable to turn it on by default, as the current output is generally bad on mobile, and the tag has no effect on desktop anyway.

I'd like to make a few more changes like this, so I thought of adding XSLT parameters to configure these updates. As I mentioned on the mailing list, inlining SVG images and using `` instead of `<object>` would be next, and for that a parameter like `INLINE_SVG` is probably warranted to maintain compatibility with old browsers (and possibly the use cases mentioned in [`LaTeXML-xhtml-misc.xsl`](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/resources/XSLT/LaTeXML-misc-xhtml.xsl#L188)?).

Now for the viewport tag, one can override the `mode="head-viewport"` template, which is easy enough to do, but to get a sense of how things would look like I have added a parameter `META_VIEWPORT` to change the viewport value or remove the tag altogether.

PS: I also took the liberty of moving the content-type meta tag at the top of the head, since it is supposed to appear as early as possible [1].

[1] https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Introduction_to_HTML5#declaring_the_character_set_with_the_%3Cmeta_charset%3E

### DIFF
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -59,7 +59,8 @@
 
 /* .ltx_chapter_title, etc should be in ltx-article.css etc.
  */
-.ltx_page_main { margin:0px; padding:1em 3em 1em 2em; }
+.ltx_page_main { margin:0px; padding:1em 3em 1em 2em;
+    padding:min(1em,1.5%) min(3em,4.5%) min(1em,1.5%), min(2em, 3%); }
 .ltx_tocentry  { list-style-type:none; }
 
 /* support for common author block layouts.*/
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -59,7 +59,8 @@
 
 /* .ltx_chapter_title, etc should be in ltx-article.css etc.
  */
-.ltx_page_main { margin:0px; padding:1em 3em 1em 2em;
+.ltx_page_main { margin:0px auto; max-width: 36em; max-width: 72ch;
+    padding:1em 3em 1em 2em;
     padding:min(1em,1.5%) min(3em,4.5%) min(1em,1.5%), min(2em, 3%); }
 .ltx_tocentry  { list-style-type:none; }
 
--- a/lib/LaTeXML/resources/CSS/LaTeXML-navbar-left.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML-navbar-left.css
@@ -9,4 +9,5 @@
 .ltx_page_main {
     position:absolute; left:190px; top:0px; right:2px; 
     margin:0px; padding:1em 3em 1em 2em;
+    padding:min(1em,1.5%) min(3em,4.5%) min(1em,1.5%), min(2em, 3%);
     width:70%; }
--- a/lib/LaTeXML/resources/CSS/LaTeXML-navbar-right.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML-navbar-right.css
@@ -7,4 +7,5 @@
     margin-left:-2em; }
 .ltx_page_main {
     margin:0px; padding:1em 3em 1em 2em;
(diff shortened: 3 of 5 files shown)


dginev:
I quite like both the direction to modernize the HTML, as well as the style of PR - brief and super easy to review. Schema decisions tend to be in the court of @brucemiller so I'll let him navigate, but definitely appreciate the PR!

xworld21:
I have reduced this PR to the essential: it now adds a `META_VIEWPORT` parameter which is initially empty, with default value `width=device-width, initial-scale=1, shrink-to-fit=no` only in the HTML5 output. This is a much simpler and better approach, since `USE_HTML5` is also true in EPUB3, where the viewport tag is used for different reasons (fixed layout documents).

It is probably a good idea to tweak the default CSS as well: the default margins on mobile are pretty large.

I moved the `content-type` changes to #1446.

brucemiller:
I've gotten pretty rusty on the details of viewports, mostly because when I last experimented with them they didn't do much! :>  But this looks pretty useful, and is at least a good start.  Thanks much!!

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -59,7 +59,9 @@
 
 /* .ltx_chapter_title, etc should be in ltx-article.css etc.
  */
-.ltx_page_main { margin:0px; padding:1em 3em 1em 2em; }
+.ltx_page_main { margin:0px auto; max-width: 36em; max-width: 72ch;
```
brucemiller, line 62: The effects of this snuck up on me: Is there a reason to specify max-width in by default (especially so small)?  To my tastes, it's unpleasant on the desktop with a large browser window, and doesn't seem to have any constructive effect in my limited testing on mobile.

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -59,7 +59,9 @@
...
+    padding:1em 3em 1em 2em;
+    padding:min(1em,1.5%) min(3em,4.5%) min(1em,1.5%), min(2em, 3%); }
```
brucemiller, line 64: Also: Isn't the "," after the 3rd min clause a typo? (same in the previous two .css)

lib/LaTeXML/resources/CSS/LaTeXML.css
xworld21: Sorry, I must have brought this over from my own stylesheet! Not sure how it ended up there. My bad. I agree that it shouldn't be the default, it all depends on your use case.

lib/LaTeXML/resources/CSS/LaTeXML.css
xworld21: Must have been a bad day when I submitted this... yes, the last comma is a typo. The line is supposed to replicate the previous one with the extra min(..., ...%).

And now that I see this, I should have documented the magic "1.5%" number. It's chosen so that the old 1em wins when the width is around 1100px.

lib/LaTeXML/resources/CSS/LaTeXML.css
xworld21: Ok, I just checked what I was doing in January. I think I pushed this commit by semi-accident. I was preparing the margin change but I hadn't really finished. At least I made separate commits, so you can just revert aebd080119f4baaf9a89307069fdae556c0949b2 to limit the change to the min(...) tweak.

I'll review the padding change as well, because now I do not feel sure that was fully baked either. I think the sidebar is still broken on mobile. I should at least double check that it's not any worse than before.

Issue #1440, opened 2021-01-17
Generated XHTML fails (a few) validity checks


nxg:
The XHTML generated by LaTeXML passes a large fraction of the checks provided by the (very thorough) W3C [EPUB validator](https://github.com/w3c/epubcheck/), but not all of them. The attached stylesheet (see internal comments for notes and rationale) normalises the XHTML so that it passes these checks.

The stylesheet also reworks footnotes into a form which is EPUB-friendly, but more generally is the format recommended by the [DAISY accessibility consortium](http://kb.daisy.org/publishing/docs/html/notes.html), and I think also implicitly recommended by the XHTML specification (though it's hard to pin down a specific location within the (X)HTML(5) sprawl).

The failures are:

  * `<object>` elements with an `@alt` attribute (the fallback content should be element-content).
  * both `<tbody>` and `<tr>` as children of `<table>`.
  * A `<meta>` content-type indicating XHTML rather than HTML.

The third one is a specifically EPUB issue; the first two are XHTML validity issues. Though this is using an EPUB validator – because in my particular case I'm generating XHTML en route to EPUB – I'm reporting XHTML validity issues which will very probably be relevant for the xhtml output as well. The validity errors are also reported by Emacs nxml-mode, so, again, this isn't an EPUB-specific issue.

Stylesheet: [sanitise-xhtml.xslt.gz](https://github.com/brucemiller/LaTeXML/files/5826192/sanitise-xhtml.xslt.gz)

xworld21:
For reference, I looked at the `<meta>` issue, I think it comes from the living HTML5 standard [1]. If I understand correctly what it says, `content-type` is not allowed when using the XML syntax of HTML5, which is what the EPUB spec calls 'XHTML' [2].

[...]

PR #1444, opened 2021-01-24
Improve SVG accessibility

xworld21:
The current way to include external SVG images (via the `<object>` tag) has accessibility issues, see e.g. [1].

This PR replaces the `alt` attribute with `aria-label`, and uses the normal `` tag as fallback content (without duplicating the `id`, of course!), which seems to be better for screen readers, although still not very good. This should also fix part of #1440.

To reduce the amount of duplication, I have moved the code generating the `description` and common attributes (`height`, `width`, ...) into new modes of templates matching `ltx:graphics`.

[1] https://www.powermapper.com/tests/screen-readers/elements/#object

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-misc-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-misc-xhtml.xsl
@@ -144,38 +144,10 @@
     <xsl:element name="img" namespace="{$html_ns}">
       <xsl:attribute name="src"><xsl:value-of select="f:url(@imagesrc)"/></xsl:attribute>
       <xsl:call-template name="add_id"/>
-      <xsl:call-template name="add_attributes">
-        <xsl:with-param name="extra_style">
-          <xsl:if test="@imagedepth">
-            <xsl:value-of select="concat('vertical-align:-',@imagedepth,'px')"/>
-          </xsl:if>
-        </xsl:with-param>
-      </xsl:call-template>
-      <xsl:if test="@imagewidth">
-        <xsl:attribute name='width'>
-          <xsl:value-of select="@imagewidth"/>
-        </xsl:attribute>
-      </xsl:if>
-      <xsl:if test="@imageheight">
-        <xsl:attribute name='height'>
-          <xsl:value-of select="@imageheight"/>
-        </xsl:attribute>
-      </xsl:if>
-      <xsl:choose>
-        <xsl:when test="@description">
-          <xsl:attribute name='alt'>
-            <xsl:value-of select="@description"/>
-          </xsl:attribute>
-        </xsl:when>
-        <xsl:when test="ancestor::ltx:figure/ltx:caption">
-          <xsl:attribute name='alt'>
-            <xsl:value-of select="ancestor::ltx:figure/ltx:caption/text()"/>
-          </xsl:attribute>
-        </xsl:when>
-        <xsl:otherwise>
-          <xsl:attribute name='alt'></xsl:attribute> <!--required; what else? -->
-        </xsl:otherwise>
-      </xsl:choose>
+      <xsl:apply-templates select="." mode="graphics-attributes"/>
+      <xsl:attribute name='alt'>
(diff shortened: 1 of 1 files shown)


xworld21:
I refactored the `ltx:graphics` code using modes and thrown in an `USE_IMG_FOR_EXTERNAL_SVG` parameter, although I admit it is just a proof-of-concept at this stage. With modes, it should be easy to customise the output in case one wants more accessible SVGs, but the occasional animation (for example, use the `as-dynamic-svg` mode only for files ending with `-animated.svg`). But there may be smarter ways of approaching this problem.

And `as-dynamic-svg` is probably a misnomer, because the `` allows for SVG+SMIL animations.

dginev:
Thanks for all the great PR work @xworld21 , much appreciated! We'll be getting back to you with reviews and thoughts next week. The helpful links you're sprinkling in are quite appreciated, since the biggest piece of work is for us to understand what's the current best practice and then hopefully just press the "merge" button 👍🏻

brucemiller:
Interesting.... have to think about it!

xworld21:
@brucemiller @dginev I think it's better if I split the PR into two parts: the first two commits, which are needed to be standards-compliant no matter what, and the rest of the changes, which require some thinking. I should first figure out the possible use cases for `` vs. `<object>`. E.g. most LaTeX(ML)-produced images are static and better served via ``, but the `animate` package, which is currently not supported, produces *interactive* SVGs via `dvisvgm`; there could be other things like this.

brucemiller:
That might be a good idea; Exactly the implications of img and object changes hasn't quite sunk in, for me.

brucemiller:
I'm kinda unclear whether "using object" has accessibility issues, or whether a (mostly obsolete) browser has accessibility issues.  And all this is conflated with (1) the lack of clarity about aria, (2) the lack of an appropriate description for images (but separately solveable, perhaps) and (3) that half the information about the choice between img vs object tends to be about what particular versions of browsers are doing or not.

And it seems kinda unfortunate to cast the solution in the form that the converter chooses either accessibility or scripting on all svg.  It would seem better to stick with object for svg, since they *might* have scripting, webfonts, etc, and figure out how best to add the correct description, whether with aria or otherwise.

xworld21:
> I'm kinda unclear whether "using object" has accessibility issues, or whether a (mostly obsolete) browser has accessibility issues.

Unfortunately, `<object>` has accessibility issues even now in 2022. I don't know the technical reasons, but it just trips up screen readers, and worst of all, it messes up keyboard navigation pretty badly.

I would counter this with: when would an SVG file contain scripts, fonts, etc? In the standard usage, latexml produces external SVGs only from mathimages, and those can definitely go in `` (whereas other SVG content is inlined). The only bad situation would be a user providing their own SVGs with fonts/interactivity (in fact, I do this in BookML for the animate package, and I inspect each SVG from the XSL stylesheet to decide whether to use `` vs. `<object>`), but I bet it's pretty rare.

dginev:
> in fact, I do this in BookML for the animate package, and I inspect each SVG from the XSL stylesheet to decide whether to use  vs. <object>

This could be a nice technical solution to the conundrum. Default to the simpler img, and try to detect object-worthy cases. Maybe it can be a point of consensus?

It is hard to strike a balance between catering to the "tools of the day" and over-catering to them, but if `<object>` tags are known to be badly supported and offer little added value for us in *most* cases...

xworld21:
For reference, crude detection of scripted content in XSLT at https://github.com/vlmantova/bookml/blob/f3ae748328bcb20604954e025bba81be3fc10e12/XSLT/xhtml5.xsl#L313

Not that I'd suggest doing this in XSLT: scripted content is easy, but checking for external resources would be a lot easier and more robust from perl. Postprocessing already inspects images to find the size, so it should be easy to add a check for scripts and resources, and flag the outcome via an attribute.

xworld21:
Small update: one can also embed images via `<svg><use href="..."></svg>`. However, the link *must* contain a fragment, so it only works with SVGs that have an id on the root. SVG 2 is supposed to change that, but I understand it's not even implemented yet.

PR #1446, opened 2021-01-31
Improve content-type compliance

xworld21:
This PR adjusts the XHTML5 output (HTML5 in XML syntax) to behave correctly according to the HTML5 spec: the `content-type` declaration must not appear in the XML serialisation [1] and `charset` can be omitted as it is only allowed to facilitate migration [2]. It also moves the `content-type` declaration before the title so that the encoding appears within the first 1024 characters of the document [3] (unless a user overrides the `head-begin` template, in which case they will hopefully notice the warning).

I tweaked the stylesheet with the idea that `USE_NAMESPACES` means we are dealing with an XML serialisation, while `USE_HTML5` distinguishes between HTML5 and HTML4/XHTML1.1.

In turn, the EPUB3 stylesheet does not need to override the `head-content-type` template because it is already handled correctly in the generic xhtml stylesheet.

This fixes part of #1440, at least for when one generates EPUBs manually without using `latexmlc`. The PR adds an XHTML5 stylesheet to facilitate the process.

[1] https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-content-type
[2] https://html.spec.whatwg.org/multipage/semantics.html#the-meta-element
[3] https://html.spec.whatwg.org/multipage/semantics.html#charset


brucemiller:
This looks pretty sensible. Thanks!

Issue #1450, opened 2021-02-07
download button in listings needs download attribute


xworld21:
The download button for listings has stopped working for a while, since browsers stopped allowing navigation to data URLs [1]. I figure the appropriate solution is to add the empty `download` attribute to tell the browser to download the content, rather than navigating to it [2].

I suppose it would be even better to use the actual file name when it is available (i.e. when using `\lstinputlisting`).

[1] [Data URLs - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs)
[2] [HTML Standard](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-download)

dginev:
was fixed by #1451 but remained open, closing

PR #1451, opened 2021-02-07
add download attribute to listings data

xworld21:
Add download attribute to listings as per #1450

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-block-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-block-xhtml.xsl
@@ -154,6 +154,9 @@
         <xsl:call-template name="add_data_attribute">
           <xsl:with-param name="name" select="'href'"/>
         </xsl:call-template>
+        <xsl:if test="$USE_HTML5='true'">
+          <xsl:attribute name="download" namespace="{$html_ns}"/>
+        </xsl:if>
         <xsl:text>&#x2B07;</xsl:text>
       </xsl:element>
     </xsl:element>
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/resources/XSLT/LaTeXML-block-xhtml.xsl
```diff
@@ -154,6 +154,9 @@
         <xsl:call-template name="add_data_attribute">
           <xsl:with-param name="name" select="'href'"/>
         </xsl:call-template>
+        <xsl:if test="$USE_HTML5='true'">
```
brucemiller, line 157: But data links work in xhtml generally, just (apparently) not in epub; do we need a separate switch for epub?

xworld21: The reasoning behind the `USE_HTML5` test is because the `download` attribute was not valid pre-HTML5 – it does not appear in the XHTML1.0/1.1 DTDs. An abundance of precaution, maybe?

For EPUB, I was going to submit a PR overriding `match="ltx:listing[@data]"` with an empty template in the EPUB3 stylesheet, rather than adding a flag in the XHTML template. Does that sound reasonable?

xworld21: @brucemiller the EPUB3 fix was actually a one-liner #1464, I could have sent it right away. With the two PRs, the end result is: all plain HTML outputs have the old download button; HTML5 and "XHTML5" get the additional `download` attribute; EPUB3 has no button at all (at least until there is a way of including the code as an attachment).

brucemiller:
OK, this seems worthwhile; Thanks!

PR #1459, opened 2021-02-09
[WIP] utf-8 binmode in latexmlc STDOUT

dginev:
Long overdue correction, which was concealing subtleties in encoding when debugging the `latexmls` server executable.

Unicode issues are very subtle for me, since I haven't done the work of mastering the low-level code points, but the general rule that seems robust was:
 - Use `binmode(HANDLE, ":encoding(UTF-8)");` on each handle we are printing to, and add those calls in each executable file that will be printing (this may change with the logging refactor, but is very much a path to sanity in the main branch).
 - Avoid at all costs double-encoding, i.e. avoid early calls to `encode()` inside the `.pm` application logic. I believe the mainline latexml codebase is largely complaint with that, but today I realized `latexmls` wasn't, due to a seemingly innocent call to `encode_json`.

Small change for the LaTeXML repo itself, but a good reveal for an issue in `LaTeXML::Plugin::latexmls`

### DIFF
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -56,15 +56,15 @@ my $keyvals = $opts->scan_to_keyvals(\@ARGV);
 # Client-side options:
 my ($port, $address, $expire, $local) = map { $opts->get($_) } qw(port address expire);
 $address = '127.0.0.1' if !$address || ($address eq 'localhost');
-$address =~ s/^(\w+)\:\/\///;    # strip away any protocol
+$address =~ s/^(\w+)\:\/\///;     # strip away any protocol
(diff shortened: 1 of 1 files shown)


dginev:
As usual, it's never that simple with unicode.

I think I won't touch the latexml repository at all for this round, maybe I can see things more clearly as part of the logging PR refactor. LaTeXML.pm is standardized on returning encoded utf-8, in the perl sense, and my need for raw bytes in the decoded sense is very specific to my Rust pipeline.

I'll add an option in latexmls that explicitly requests decoded bytes, and leave it at that for now.

PR #1462, opened 2021-02-10
Implement \DeclareUnicodeCharacter

### LINKED ISSUE #1461: [add missing macro] add \DeclareUnicodeCharacter to inputenc.sty
`inputenc.sty.ltxml` does not implement
```LaTeX
\DeclareUnicodeCharacter{codepoint}{expansion}
```
which takes a hexadecimal `codepoint` and makes the corresponding character expand to `expansion`.

xworld21:
Fixes #1461.


xworld21 (commented):
I have answered my own question and switched prototype. I should have thought this over a bit more before submitting!

lib/LaTeXML/Package/utf8.def.ltxml
```diff
@@ -29,6 +29,17 @@ AssignValue(INPUT_ENCODING => undef);
 # And set the Mouth to simply pass-thru the utf8.
 AssignValue(PERL_INPUT_ENCODING => 'UTF8');
 
+DefPrimitive('\DeclareUnicodeCharacter Digested {}', sub {
```
brucemiller, line 32: Looks quite plausible, but wouldn't Expanded be more appropriate than Digested?  A simple example of use would be helpful (if not an actual test case).

lib/LaTeXML/Package/utf8.def.ltxml
```diff
@@ -29,6 +29,17 @@ AssignValue(INPUT_ENCODING => undef);
...
+    my ($stomach, $hexcode, $expansion) = @_;
+    my $char = $hexcode->toString();
+    if ($char =~ /^[0-9a-fA-F]+$/) {
+      $char = chr(hex($char));
```
dginev, line 36: I just ran into [a mistake I made](https://github.com/brucemiller/LaTeXML/pull/1471/commits/09c2751436a4eb63b44ab67fb03e66079f56b426) that is also visible here - we should be using `UTF()` instead of `chr()` for these open-ended unicode codes.

I noticed it when passing resulting strings out of latexml (and out of perl) and landing on `error : string is not in UTF-8
` breakage.

xworld21: I would have never noticed this, thanks!

lib/LaTeXML/Package/utf8.def.ltxml
xworld21: Very good point, I don't think there is any use case for Digested. It's just me picking the wrong option (still learning!). After reading the `inputenc` source, I believe Expanded is the good answer: the code responsible for reading the codepoint is `\count@"#1\relax`, which I think expands `#1` without executing.

But you made notice something else: the current code mangles spaces incorrectly. `inputenc` wraps the second argument in the macro `\IeC{}`, to prevent active characters from gobbling spaces. Plus `\IeC` acts as `\protect`, and that is also important to reproduce.

It's likely that `\IeC` is needed in `\DeclareInputText` too. I'll come up with some test cases and think about a sensible implementation.

xworld21: I think I understand now. The argument should definitely be Expanded. If you pass unexpandable tokens, they will get executed in the middle of `utf8.def`, and that is definitely not the intended use of `inputenc`.

Regarding the `\IeC{}` trick in `inputenc`, it is used to "avoid any space after the command being gobbled up when the text is written out to an auxiliary file" (from `utf8ienc.dtx`). LaTeXML does not write auxiliary files, and I gather it does not emulate writing to them, so the `\IeC{}` trick is not needed. Likewise for the `\protect` hidden inside `\IeC{}`. So if I am correct on this, the current implementation is fine.

Do you need a `t/encoding/utf8.tex`? I can write some examples (e.g. mapping € to `\euro`, or √ to `\sqrt`), even though I can't think of any edge case that needs to be tested.

brucemiller:
A test case demo'ing `\DeclareUnicodeCharacter`? Yeah, that would be nice, if it isn't too much trouble. And would help avoid regressions.

xworld21:
Ok, so the remaining outstanding issue (for me) is `\protect`. I am trying to understand: are there any situations in which LaTeXML *needs* commands to be protected? If so, there might be situations where the `inputenc` binding behaves differently because of the missing `\IeC` protection. I can't figure this one out from code, I am afraid.

brucemiller:
Cool! Thanks!!!

Issue #1457, opened 2021-02-08
download button in listings data broken in EPUB3


xworld21:
When clicking on the download button of a `\begin{listings}` environment, some EPUB readers (Apple Books!) navigate to the file but do not offer a way back. In the case of Apple Books, one needs to trigger a reflow by resizing the window to resume reading. This is arguably a bug in Apple Books. This happens even if I add the `download` attribute using #1451.

Still, I checked the EPUB3 spec, and if I understand correctly, using data URLs in `<a>` tags breaks the standard:
- resources of non-core media types are foreign resources [1], which can be attached, but must be treated according to certain rules, and have a reasonable fallback; note that even `text/plain` is *not* a core media-type;
- I also guess that all navigation links must correspond to entries in the manifest, which is probably the reason for Apple Books getting lost.

Given that the button is broken, I suggest to remove it (PR to follow soon). The note in [1] suggests an alternative implementation, but it seems like the users are supposed to unzip the EPUB package by themselves.

I should add that `epubcheck` does not complain if a listing, and thus its download button, is present, so I may be getting this very wrong.

[1] https://www.w3.org/publishing/epub32/epub-spec.html#sec-foreign-restrictions

PS: minimal example
```LaTeX
\documentclass{article}
\usepackage{listings}
\begin{document}
Click the download button to navigate away... and never come back.
\begin{lstlisting}
  # code
\end{lstlisting}
\end{document}
```

brucemiller:
I recently discovered that data urls had stopped working, being apparently a security risk. Managing an additional generated resource (like images) in seperate files would be a big mess, but I wonder if there is some other mechanism that would provide access to data embedded within the page?  Would you be able to access the listing data if it were provided as a normal separate file (within the epub), using a normal relative link?

xworld21:
@brucemiller 

> Would you be able to access the listing data if it were provided as a normal separate file (within the epub), using a normal relative link?

I believe this is the correct approach. An EPUB3 reader may or may not offer functionality to extract the file, but that's up to the reading software, not LaTeXML; furthermore, most readers will probably show `text/plain` content even if it's not a core media type. At least that's how I understand the foreign resources section of the spec. Let me run the experiment with a couple of readers, and I'll let you know!

I suppose we could also file an issue against [epubcheck](https://github.com/w3c/epubcheck), since it does not seem to notice data URLs. It might be a way to get a clarification on the spec.

> I recently discovered that data urls had stopped working, being apparently a security risk.

[...]

PR #1464, opened 2021-02-11
remove download button from listings in EPUB3

xworld21:
Quick fix for #1457 until a better solution is available.


brucemiller:
This is simple, and only affects epub, so it's probably a good temporary solution even though it isn't really satisfying. But Thanks!!!

PR #1466, opened 2021-02-16
Updated texvc.sty

dginev:
Encountered while "lexematizing" StackExchange, which loads the texvc macro subset (via mathjax).

It turns out that the catcode switch for `$` and `%` are quite rocky over SE expressions, and were not playing along with `\text` well. So I opened the latest texvc.sty I could find, which seems to be:

http://www.tug.org/texlive///Contents/live/texmf-dist/tex/latex/texvc/texvc.sty

and authored by @physikerwelt some years back?

Since that file only introduces macros, I reworked our binding to do that as well. Updated packages+macro set as well. Would be nice to have merged soon, but I can also keep a local SE conversion running on this branch, fine with me either way.

### DIFF
--- a/lib/LaTeXML/Package/texvc.sty.ltxml
+++ b/lib/LaTeXML/Package/texvc.sty.ltxml
@@ -29,10 +29,13 @@ use LaTeXML::Package;
 # recognizes --- there is no regular LaTeX texvc.sty, that I am aware of.
 
 RequirePackage('amsmath');
+RequirePackage('amsfonts');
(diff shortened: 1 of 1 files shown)


dginev:
If we get this PR merged soon, I can base the Continuous Integration on the dependent repository on the latexml master, rather than on an offshoot branch... And it's an easy one too I think 🤞🏻

brucemiller:
Looks safe; Thanks!

PR #1469, opened 2021-02-20
Improve line splitting and trailing spaces removal

### LINKED ISSUE #1467: Web2C-compliant line-breaks in Mouth?
This bug is... irrelevant, because it is triggered only by files with spurious carriage returns next to single CRLF's, and that's hardly going to happen in real life. I only stumbled upon it by reading `Mouth.pm`.

If you run `latexml` on [this attachment](https://github.com/brucemiller/LaTeXML/files/5991393/MWE-wrong-newline-split.tex.zip), you will get
```xml
  <para xml:id="p1">
    <p>Line 1.
Line 2.</p>
  </para>
```
instead of the 'correct'
```xml
  <para xml:id="p1">
    <p>Line 1.</p>
  </para>
  <para xml:id="p2">
    <p>Line 2.</p>
  </para>
```

The cause is the use of both Perl's own `readline` as well as `split`, which has a peculiar behaviour with trailing empty records. I am already working on a fix as a Perl & LaTeXML internals learning exercise - I think I can simplify some bits of the mouth, which is probably worth doing anyway.

Edit: more [realistic example](https://github.com/brucemiller/LaTeXML/files/5998362/MWE.zip) with classic mac newlines from the included file getting discarded.

xworld21:
This PR splits lines and removes trailing spaces in a way that is closer (and maybe identical?) to Web2C. It fixes #1467 but also a small number of other discrepancies.

The changes, barring mistakes on my part, should sum up to:
- only literal spaces (ASCII 32) are stripped, while other `\s` characters such as tabs are preserved;
- *all* literal spaces are stripped, regardless of whether they are preceded by a `\`;
- stripping is only done to lines coming from external sources, not to internal strings (e.g. the ones passed to `DefMacro`) which we may assume to be already stripped as intended (e.g. some end with control spaces);
- carriage returns are only inserted back into the string as part of the `\endoflinechar` process, and nowhere else (as far as I could tell, the carriage return would be stripped by `/\s*$//g` in all codepaths; I may well be wrong here).

Edit: of course I made a final change without testing it... I misunderstood how `readRawLine` worked. It is fine now.


dginev:
Hey @xworld21 thanks for another nice PR! Would you feel upto adding the small example you had in the issue as a test file as well? You'd need the `.tex`, a generated `.pdf`, and an `.xml`. They'll probably fit best under `t/tokenize`.

 You'll also have to add the three new entries to the MANIFEST file to get the CI tests passing.

xworld21:
Done! There are two new tests, one for handling different newline characters, one for handling trailing spaces that are not ASCII 32.

brucemiller:
This is an impressive bit of deciphering! And a subtle patch; still manages to get trailing control space right (or even more right). Thanks!!

PR #1471, opened 2021-02-23
add \unicode extension from MathJax to texvc.sty

dginev:
Thanks for merging #1466 ! It unlocked another debugging round for me, which led to my seeing a construct that had previously dodged my attention - the MathJax-postulated `\unicode` macro. Thankfully that is incredibly simple, docs here:

http://docs.mathjax.org/en/latest/input/tex/extensions/unicode.html

It's used in StackExchange, which relies heavily on texvc.sty, and I think it is a rather good location to unify all not-quite-tex extensions. But if you'd like a stranger `mathjax.sty.ltxml` contribution, I could also refactor to that...

### DIFF
--- a/lib/LaTeXML/Package/texvc.sty.ltxml
+++ b/lib/LaTeXML/Package/texvc.sty.ltxml
@@ -56,7 +56,7 @@ DefPrimitive('\unicode[][]{}', sub {
       $code = hex $code }
     else {
       $code = int($code); }
-    my $code_char = chr($code);
+    my $code_char = UTF($code);
     return Box($code_char, undef, undef,
       T_OTHER($code_char)); });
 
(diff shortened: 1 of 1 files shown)


brucemiller:
Should this be done as a macro, as opposed to a primitive creating a `Box`? Is it *only* used for creating a unicode character in output, or could you use, say, `\unicode{24}` to enter math mode :>

dginev:
A primitive indeed fits better, since this is really intended as a crutch in mathjax territory, rather than as any meaningful participant in a TeX expansion flow.

I've changed to that (though writing the `undef, undef` hurt my eyes again). Also enabled the decimal use, I spotted in the docs - should be good to go.

brucemiller:
Looks reasonable! Thanks!!

PR #1472, opened 2021-02-28
Replace object alt attribute with aria-label and text fallback

xworld21:
These are the uncontroversial parts of #1444, needed to fix standards-compliance as per #1440 and to achieve some low effort accessibility improvement.


brucemiller:
This looks reasonable. I suspect the use of `@alt` on `object` was probably a left-over from code for `img`.
Thanks very much!

PR #1475, opened 2021-03-02
Binding with proofwiki's mini macro language for MathJax

dginev:
After texvc.sty, I've now stumbled on proofwiki.org and its mini macro language for MathJax. The binding is a direct copy-over from the mathjax macro definitions on each proofwiki page, and seems quite straightforward.


dginev:
I'm just rebasing here, no changes to the PR itself.

brucemiller:
ok, looks good; Thanks!

PR #1480, opened 2021-03-05
Improve TeXlive detection during testing

brucemiller:
Check LATEXML_KPSEWHICH variable, when detecting TeXLive version for testcase exclusions


dginev:
Looks good!

PR #1485, opened 2021-03-10
fix pgf shading

### LINKED ISSUE #1479: TikZ shading stuck on first color
This is a followup to #1168, where I've been trying to use TikZ's shading in my drawings.  It seems that we no longer have the fatal error thanks to #1360, but LaTeXML can't change the shading color.  With the following example:

``` tex
\documentclass{article}
\usepackage{tikz}
\begin{document}
\tikz{\shadedraw[{left color}=blue](0,0)--(0,20pt)--(20pt,20pt)--(20pt,0)--cycle;}
\tikz{\shadedraw[{left color}=red](0,0)--(0,20pt)--(20pt,20pt)--(20pt,0)--cycle;}
\tikz{\draw[fill=red](0,0)--(0,20pt)--(20pt,20pt)--(20pt,0)--cycle;}
\tikz{\shadedraw[color=red](0,0)--(0,20pt)--(20pt,20pt)--(20pt,0)--cycle;} 
\tikz{\draw[{right color}=red](0,0)--(0,20pt)--(20pt,20pt)--(20pt,0)--cycle;}
\end{document}
```

I get the pdf:
texOutput

But the html:
LaTeXMLoutput

It looks like `\draw[fill]` is ok, but `\shadedraw` and `\draw[{left color}]` use whatever the first shading color happened to be.

brucemiller:
Recorded stops must be cleared globally; deconvolute the code while we're at it

Fixes #1479

### DIFF
--- a/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
+++ b/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
@@ -782,27 +782,23 @@ DefMacro('\lxSVG@sh@create', sub {
       T_BEGIN, T_CS('\pgf@sys@shading@end@rgb'), T_END,
       T_END,   T_BEGIN, T_END); });
 
-DefPrimitive('\lxSVG@sh@color{Float}{Float}{Float}', sub {
-    AssignValue(pgf_sh_color => Color('rgb', map { $_->valueOf } @_[1 .. $#_]));
-    return; });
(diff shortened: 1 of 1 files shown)


dginev:
Very impressive, it actually works!

HTML above (with this PR), PDF under:

image

The only visible discrepancy is the shading ratio in the last square, but I have no idea if that's just a difference between SVG and PDF in general. Impressively accurate fix!

dginev (approved):

PR #1488, opened 2021-03-14
Change test utflettercase.tex to use \MakeLowercase und \MakeUppercase

stamer:
- utflettercase.tex only compiles with error into pdf with texlive >= 2015 (previous versions not tested)
- (iso8859-1 letters are created for \lowercase)
- instead use the Latex commands \MakeLowercase and \MakeUppercase
- pdf (visual) and xml output is not changed.

### DIFF
--- a/t/expansion/utflettercase.tex
+++ b/t/expansion/utflettercase.tex
@@ -1,7 +1,7 @@
 \documentclass{article}
 \usepackage[utf8]{inputenc}
 \begin{document}
-\uppercase{(aou) (äöü) (ÄÖÜ)}
+\MakeUppercase{(aou) (äöü) (ÄÖÜ)}
 
-\lowercase{(aou) (ÄÖÜ) (äöü)}
+\MakeLowercase{(aou) (ÄÖÜ) (äöü)}
 \end{document}


dginev:
Hi @stamer I am not sure I follow the reasoning for changing the test case here. On a quick look I see we're testing the two macros you changed to in `t/expansion/lettercase.tex`, so they have their own monitoring.

Are you saying latexml shouldn't support `\uppercase` and `\lowercase` over unicode at all?

dginev:
Ah, indeed, I am reading:
> The TeX primitive commands \uppercase and \lowercase are set up by default to work only with the 26 letters a–z and A–Z. 

https://latexref.xyz/Upper-and-lower-case.html

dginev:
Ok, now that I've recovered my memory, I still think we should keep this in as a feature. Indeed if you run:
```
$ pdflatex t/expansion/utflettercase.tex
```

It will come back with errors, alerting you not to use these macros on unicode. However, if you instead run
```
$ xelatex t/expansion/utflettercase.tex
```

You will see the macros succeed. I think it is a reasonable expectation of latexml to be able to deliver the xelatex behavior here, so we should keep the test and fix it where it appears problematic.

stamer:
Hi @dginev, no I am not saying it should not support \uppercase and \lowercase. But if the tex to pdf is not working, it is not necessary to test it. I thought it was pdflatex as it included inputenc. Is there any way to find out which to use when?
 
`perl tools/maketests --pdf t/expansion/utflettercase.tex` fails, as it tries to use pdflatex.

dginev:
I think the general rule of thumb is that we try to be closer to pdflatex, _except_ in cases where unicode becomes important - for those we try to be as unicode-native as possible, and if xetex/xelatex can create a PDF, so should latexml. As to which PDFs were made with xelatex, this one may have been the only one, especially since I was the one who added it.

brucemiller:
It certainly is an insufficiency in our test harness
that we haven't formalized which engine each test
is intended for (or even provide hints).

When the first tests were set up, we didn't have a mechanism
handy, but in the meantime we've got properties for which
styles & TeXlive versions are required.  Should be easy now to
add a record of which engine is needed, as well.

stamer:
Ok, then maketests would need to parse the .t file as well. Would it be sufficient to add a simple `% tex= "xelatex"` (as used in llmk) to the test document? maketests already parses the tex file for `documentclass` in order to determine to use pdflatex instead of pdftex.

brucemiller:
Argh! I didn't think that through. The .t spec currently is only used 
for make test, which does't care one bit about the pdf. Hmmm, your 
suggestion would be one way to do it.... gotta think

dginev:
Closing the PR to avoid accidentally merging the wrong fix.

A little xelatex annotation would indeed be a sensible technique. Or we could rename some tests as `.xe.tex` and do extension-based checks in test generation. The current `maketests` script will have an easier time doing extension logic than grepping the files, though it can naturally evolve to do either - but I do like complying with some existing build system. And `llmk` is a fun one - I've been in touch with its creator Takuto Asakura, who's also doing some math linguistics work.

PR #1489, opened 2021-03-14
Minor XSLT linting

xworld21:
Just a few typographical changes in the XSLT files. Each commit explains the type of change, e.g. replace tabs with spaces.

This is in preparation for an actual XSLT linter in `latexmllint` (not ready yet!) and probable stylesheet patches down the line.


dginev:
So you'd like us to merge in a linting style we can't ourselves reproduce? That's actually a difficult proposition in principle. 

On the other hand, having a consistent linter for the XSLT side of things sound lovely to me, especially if you coordinate it with Bruce's stylistic choices before you submit the PR 👍🏻

dginev:
You've also picked a bit of a silly minor test failure along the way:

https://github.com/brucemiller/LaTeXML/pull/1489/checks?check_run_id=2107502198#step:7:40

xworld21:
@dginev ah, this is not the output of the linter, it's just the minimum to get reasonable diffs. Mainly, the XML parser normalizes the white space inside the attributes, so the information about tabs vs newlines vs spaces is lost, which means that the spaces need to be used consistently before the linter is applied.

I'll submit the issue proposing the linter, with the link to my branch, since it was actually quite easy to integrate in `latexmllint`.

xworld21:
> You've also picked a bit of a silly minor test failure along the way:
> 
> https://github.com/brucemiller/LaTeXML/pull/1489/checks?check_run_id=2107502198#step:7:39

That's... interesting, the other tests are passing, right? I was suspicious of the TEI stylesheets, some things are not wrapped in `<xsl:text>` and they probably should.

brucemiller:
We acquired some conflicts with the patch from #1501, but I think I've resolved them correctly. The overall style & indendation of the JATS stylesheet is still a bit inconsistent. But overall, this is a nice cleanup. Thanks!

PR #1491, opened 2021-03-16
Remove month from date before generating suffixes

### LINKED ISSUE #1486: Issues with labelling of bibliographic references
This was a problem with LaTeXML 0.8.2 and still affects LaTeXML 0.8.5. 
I am expecting references to the works (a mix of articles and books) by the same author published during a year to be labelled with lables consiting of the year of publication and a subscript, e.g. 2002a, 2002b, etc for publications published by an author during 2002. 
However, if a month of publication is also given (as for journal articles), publications published during different months are labelled independently with label subscripts being restarted for each month separately. Publications for which no publication month is given are assigned subscripts independently of the monthly-labelled ones, i.e. they are incremented in a space of a year.

I have noticed that in the XML document to which bib file is converted `<bib-date role="publication">` elements contain years for the publications for which only years of publication are given, and strings consisting of a year, hyphen and numerical representation of the month, where publication dates include months, i.e. "2020-12" for an article published in Decemeber 2020. It would appear that the contents of `<bib-date role="publication">`  are used as keys when the labels are constructed, which results in incorrect labels.

I am using bibliography in a separate bib file, which I am converting to XML independently of the main document. I am using the aa (Astronomy  & Astrophysics) bibliography style.

brucemiller:
As it sez...

Fixes #1486

### DIFF
--- a/lib/LaTeXML/Post/MakeBibliography.pm
+++ b/lib/LaTeXML/Post/MakeBibliography.pm
@@ -326,7 +326,8 @@ sub getBibEntries {
         $sortnames = $names = $t->textContent; }
       my $date  = $doc->findnode('ltx:bib-date[@role="publication"] | ltx:bib-type', $bibentry);
       my $title = $doc->findnode('ltx:bib-title',                                    $bibentry);
-      $date            = ($date  ? $date->textContent  : '');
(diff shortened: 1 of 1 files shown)


dginev:
Nice one-liner PR! If tests pass, LGTM

PR #1492, opened 2021-03-16
Suggestion for hinting for maketests

stamer:
Here is one suggestion to provide a hint to compile a testcase. It solves the problem that t/expansion/uflettercase.tex needs xelatex to successfully compile. But maketests tries to either pdftex or pdflatex if called with --pdf option. Use a simple hinting (compatible to llmk) to find out which compiler should be used. 

t/expansion/utflettercase contains `% tex = xelatex`. This is detected and xelatex is used to successfully compile the testcase to pdf. If no hint is given, previous detection for pdflatex is used.

### DIFF
--- a/t/expansion/utflettercase.tex
+++ b/t/expansion/utflettercase.tex
@@ -1,3 +1,6 @@
+%+++
+% tex = xelatex
+%+++
 \documentclass{article}
 \usepackage[utf8]{inputenc}
 \begin{document}
--- a/t/expansion/utflettercase.tex
+++ b/t/expansion/utflettercase.tex
@@ -1,6 +1,6 @@
-%+++
-% tex = xelatex
-%+++
+%%%
+%%% TeX-engine: xelatex
+%%%
 \documentclass{article}
 \usepackage[utf8]{inputenc}
 \begin{document}
--- a/tools/maketests
+++ b/tools/maketests
@@ -277,9 +277,9 @@ sub gen_pdf {
     if (/documentclass/) {
         $program = 'pdflatex';
         last; }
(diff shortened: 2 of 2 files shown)


dginev:
Looks good to me! Thanks!!

dginev:
As an alternative example (not a recommendation, I currently hold no preference myself), I see that the `latexmk` build system uses a variable named `$pdf_mode`, though that is a config parameter, rather than an inline annotation:
https://tex.stackexchange.com/a/501503/78008

stamer:
llmk uses the term latex. Then I thought, well it is not just latex, but tex or pdftex might be candidates as well. Therefore I used `tex = xyz`, but `latex = xyz` is also recognized. I have not a real preference, `pdf_mode` is good. Another option would be to use the term `pdf_command`.

brucemiller:
Exactly, it's not necessarily TeX or LaTeX, but by the same token,
it's not necessarily pdf (although it currently always is). It's about
which TeX "engine" to generate a reality check on the XML.  We want
to emulate xetex or xelatex; someday we may even want to emulate luatex.
That's why I'm thinking in terms of "engine".

Google....
Hmm, it appears that AucTeX recognizes
```
%%% Local Variables:
%%% TeX-engine: pdflatex
%%% End:
```
although it seems that it accepts only specific engines.
(I've never used it)

OTOH, `latexmk` has a variable `$pdf_mode` (set to a number choosing
from a list of commands)

I'd suggest using the `TeX-engine` type of line (w/o bothering
with the Local Variables wrapper.

stamer:
Ok, sounds good, I will take care of that tonight.

brucemiller:
Looks good! This should help enhance tests in the future. Thanks!

PR #1493, opened 2021-03-17
fix halign hack

brucemiller:
Fix the badfix (begin/endgroup should NOT look like braces for alignments)


brucemiller:
Well, why didn't you say so before! :>

PR #1494, opened 2021-03-17
add minimal minimal.cls.ltxml binding

brucemiller:
Another easy one :>


dginev:
That's a little strange to see, given that I added it in #1475 some 16 days ago :> Why not merge there and close here? May have compounding returns!

brucemiller:
Well, mine was better.  But oh well...

PR #1507, opened 2021-03-31
Don't loose preambles of \lstnewenvironment.

brucemiller:
Don't loose preambles of \lstnewenvironment; Thanks Tim Prescott


dginev:
Seems reasonable!

PR #1509, opened 2021-04-06
Enable fontenc-aware lower/uppercase for cyrillic

### LINKED ISSUE #1508: fontenc is missing cyrillic letters
Compiling the following:
``` LaTeX
\documentclass{article}
\usepackage[T1]{fontenc}
\begin{document}
\MakeLowercase{A}
\end{document}
```
dies because many cyrillic letter commands are undefined, so it passes the 100 error limit.

Looking at fontenc.sty.ltxml, it seems that 33 pairs are defined in lines 36--56 (which doesn't have a parallel in TeX's fontenc.sty).  On the other hand, lines 63--87 (which matches TeX's fontenc.sty) adds 98 pairs to the \@uclclist, so that when \MakeLowercase is called, 65 pairs of undefined macros are called.

Unfortunately, I can't tell where TeX defines the letters, and my knowledge of cyrillic (none) isn't enough to know which macros should become which letters.  It almost seems like TeX isn't defining the letters in fontenc.sty, but only defining the capitalization rules.

dginev:
Fixes #1508 

A bit of a "request for comments" PR, since I am unsure what the best general strategy would be for latexml, given how the unicode situation is dramatically different between pdflatex and perl.

First, some good news:
 1. The native definitions of `\MakeUppercase` and `\MakeLowercase` seem to work like a charm. I refactored to them to play around, leaving them in for effect.
 2. A restricted attempt to use perl's `uc` and `lc` commands for these macros achieves the desired effect for cyrillic -- seemingly without currently known side effects?

Then the tricky bits:
 - fontenc itself is a bit of a mixed bag... We are defining only a core set of the cyrillic macros, and indeed another way to get rid of the errors mentioned in the issue is to define the advanced/rarer macros directly. In latex, those are actually loaded in through the definition files, namely any of `x2enc.dfu`, `t2cenc.dfu` or `utf8enc.dfu`, from what I can see.
 - I did not address the macros directly since it's also safe to say no one is realistically using them directly - people type in cyrillic characters, and pdftex maps into these macros internally. But latexml doesn't, as unicode letters haven't been mapped into active characters - which has its upsides.

So for now I just validated that the issue is composed of these pieces, and validated the simple fix of running perl's `uc` or `lc` on tokens with the `other` catcode that pass through these macros solves the general problem as far as latexml is concerned. So a new minimal example that works with this PR is:

```tex
\documentclass{article}
\usepackage[T2A]{fontenc}
\usepackage[utf8]{inputenc}
\begin{document}
\MakeLowercase{Б}
\MakeUppercase{я}
\end{document}
```

Feedback welcome.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -5096,20 +5096,47 @@ Let('\@sptoken', T_SPACE);
 
 DefMacroI('\@uclclist', undef, '\oe\OE\o\O\ae\AE\dh\DH\dj\DJ\l\L\ng\NG\ss\SS\th\TH');
 
-DefMacro('\MakeUppercase{}', sub {
-    my @t = LookupDefinition(T_CS('\@uclclist'))->getExpansion->unlist;
-    my @x = (T_CS('\def'), T_CS('\i'), T_BEGIN, T_LETTER('I'), T_END,
-      T_CS('\def'), T_CS('\j'), T_BEGIN, T_LETTER('J'), T_END);
-    while (@t) { push(@x, T_CS('\let'), shift(@t), shift(@t)); }
-    my $arg = Expand(Tokens(T_BEGIN, @x, $_[1]->unlist, T_END));
-    (T_CS('\uppercase'), T_BEGIN, $arg->unlist, T_END); });
+sub ltx_adjust_case {
+  my ($arg, $flag_lower) = @_;
+  my @adjusted = ();
+  for my $token ($arg->unlist) {
+    if ($$token[1] == CC_OTHER) {
+      my $letter = $$token[0];
+      my $cased  = $flag_lower ? lc($letter) : uc($letter);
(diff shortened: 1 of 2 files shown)


dginev:
Ok, traced this in latex.ltx: the main discrepancy was that the `\DeclareFontEncoding` macro loads a respective `...enc.dfu` file, when available. 

Adding that to the binding in turn makes my minimal example succeed on lower/uppercasing cyrillic without any extra help.

One additional niggle was moving the declaration of `\DeclareUnicodeCharacter` to LaTeX.pool, since a related change is that textcomp.sty.ltxml will now load ts1enc.dfu, and we are loading that binding whenever LaTeX.pool loads. This is a slight bit different from what latex.ltx does, which is to define and use `\DeclareEncodingSubset`, but I think we can defer emulating that 100% accurately until we may need it.

May be good to merge - review welcome @brucemiller

brucemiller:
Looks like a quite reasonable approach. Cool!

brucemiller (approved):

PR #1510, opened 2021-04-06
CSS tweaks

brucemiller:
This makes two column layout use CSS flexbox, and adds media query to have the 2 columns unwrap when the screen is too narrow.

Also fixes a couple of minor typos from #1442.

@xworld21 Could you take a look that I've not undone anything important to your work?

### DIFF
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -57,11 +57,10 @@
 
 .ltx_outdent { margin-left: -2em; }
 
-/* .ltx_chapter_title, etc should be in ltx-article.css etc.
- */
-.ltx_page_main { margin:0px auto; max-width: 36em; max-width: 72ch;
+/* .ltx_chapter_title, etc should be in ltx-article.css etc. */
+.ltx_page_main { margin:0px auto;
     padding:1em 3em 1em 2em;
-    padding:min(1em,1.5%) min(3em,4.5%) min(1em,1.5%), min(2em, 3%); }
+    padding:min(1em,1.5%) min(3em,4.5%) min(1em,1.5%) min(2em, 3%); }
 .ltx_tocentry  { list-style-type:none; }
 
 /* support for common author block layouts.*/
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -244,16 +243,19 @@ span.ltx_rowspan { position:absolute; top:0; bottom:0; }
...
-.ltx_page_columns:after {
-    content:"."; display:block; height:0; clear:both; visibility:hidden; }
+    flex:100%; }
+@media screen and (max-width: 80em){
+  .ltx_page_columns {
+     flex-direction: column; }
+  .ltx_page_column1 > ul {
+     margin-bottom: 0; }
+  .ltx_page_column2 > ul {
+     margin-top: 0; }
```
dginev, line 257: Why the special treatment for `ul` ? What about `ol`, other blocky bits? Haven't tested, just looked oddly specific.

brucemiller: Yeah, probably ol, dl should be there. Basically the idea is that 2 columns is often used to split a table of contents/index/other list. And these often have top/bottom margins. But when you're putting 2 of those lists back into (the illusion of) one list, you want to remove the margin between them.

dginev: Love the concept, and thanks for extending. I guess we can always extend more on a need basis, as we read more on our phones.

This is also the type of general issue that I feel very tempted to loop back with the arxiv-vanity folks, who have been doing extensive CSS "reimagining" on top of latexml, but I am not at all certain they have the time/desire to spend time reviewing latexml's CSS. I'll tag @bfirsh just so that he is aware, but no assumptions he has the time.

dginev (approved):
Looks good to merge, as far as I'm concerned - neat!

lib/LaTeXML/resources/CSS/LaTeXML.css
xworld21: @brucemiller are there strong reasons for not using the native CSS multicolumn here?

brucemiller:
Strong reason? Conservative paranoia? Is it well enough supported for 
lists now? I should do some experimentation.

xworld21:
Reference about multicolumn https://caniuse.com/multicolumn. If seems better than flex, actually (looking at IE10-11). I haven't looked at it seriously – it may not play well with the actual tags you are using.

dginev:
> Reference about multicolumn https://caniuse.com/multicolumn. If seems better than flex, actually (looking at IE10-11). I haven't looked at it seriously – it may not play well with the actual tags you are using.

Last time I had checked, flexbox had the advantage of being very popular and I seen as a big part of "the future" for responsive design. And usually it can deliver what multicol can, while playing nicely with other flex-ible content. So while I can imagine the latexml CSS relying heavily on flex for eventually doing *all* responsive design, I am not sure I want to start thinking about how multicol and flex will interplay when they start stacking together.

xworld21:
I think multicolumn does the columns for you, i.e. you don't have to guesstimate where to split the content in half, whereas with flexbox you have to create two blocks (which LaTeXML already does anyway), and it supposedly works better in IE10-11. OTOH, fewer future headaches sounds like a good reason to stick to flex.

Issue #1511, opened 2021-04-09
packParameters in wrong places


brucemiller:
@dginev I get that `packParameters` needs to be done early, and in many places that a token list has been read in. But it seems either too many places, or the logic is inverted (can only disable w/secret option), or the timing just isn't quite right.  The attached example indicates that `\read` and `AddToMacro` (called by `\g@addto@macro`, eg.) probably should *not* call pack parameters.

Simple-minded tests such as having `$gullet->readArg` always do it, or having it never done, of course cause explosions of errors. Perhaps the answer is simply to use the `DefPlain` parameter type more frequently? (whenever the tokens will end up as a macro defn?)

Demo: [hashtest.zip](https://github.com/brucemiller/LaTeXML/files/6287808/hashtest.zip)

brucemiller:
Fixed by #1514

PR #1514, opened 2021-04-15
[WiP] Pack parameters

brucemiller:
Apparently, we need to process the parameter token `#` relatively early upon reading a sequence of tokens (to recognize doubled hash, or `#1` parameter specs). Unfortunately, it isn't completely clear exactly when that needs to occur. For example `\def` does it by processing the body while reading, rather than when the definition takes place.  But it turns out that `\read` and `\g@addto@macro` don't carry out this processing, although you do get errors *using* the line that was read in, if you aren't careful. Thus #1511. 

Currently, there's a secret `noprep` option to Expandable that inhibits "parameter packing" (the pre-processing of the parameter tokens `#`).  This PR renames and exposes the `noprep` option to be visible to `DefMacro` in general. This is then used by `\read`, and `AddToMacro` (eg. by `\g@addto@macro`). 

Unfortunately, this PR only fixes those cases that were accidentally discovered to give errors at the wrong time. It is likely that there are other macros/functions that need to have `nopackParameter`. But more fundamentally, I suspect there is a deeper timing error, and this PR simply masks some instances.  At least, it proves a test case covering them.


brucemiller:
Aside: The `\detokenize` in the file read section, may be a bit of red herring. It's not critical to the issue, it's only there to *see* the tokens in the definitions/lines.

dginev:
In the B Book this "parameter packing" is referred to as "Scan and build the parameter part of the macro definition", at item `474.` and is exclusively done in the `scan_toks` routine, itself responsible for building token lists in particular procedural time points. A good mention about where `scan_toks` is needed can be seen in `464.`:

image

That is indeed suggestive the "parameter packing" isn't done in `\read`, since we have a special note that it is used in `\write`.

The `\g@addto@macro` definition in latex.ltx uses `\the` which should - if we've implemented it correctly - inhibit the "packing" behavior as well, as the param tokens are neutralized when returned from `\the`:
```tex
\long\def\g@addto@macro#1#2{%
  \begingroup
    \toks@\expandafter{#1#2}%
    \xdef#1{\the\toks@}%
  \endgroup}
```

A lot of this confusion comes from us not having an easy way to directly implement an exact equivalent of `scan_toks` in latexml -- I tried for the expl3 work, but it wasn't really feasible as we have many different avenues in which tokens are entered and removed from the Gullet. The one solid handle available were the `readToken` methods, which is why we ended up complicating those further. But that has indeed left us in a position where we have to emulate indirectly the rule of "pack parameters only when scan_toks is called", and we have to find all code locations where that applies.

For what it's worth, if it turns out hard to spot where the packing needs to be disabled, we can always invert the flag and only mark the locations where it should be enabled, disabling it by default. I wish I understood it well enough as to have a clear intuition which is more intuitive - but I can't make that claim for now.

lib/LaTeXML/Package.pm
```diff
@@ -2285,7 +2286,7 @@ sub AddToMacro {
   else {
     DefMacroI($cs, undef, Tokens(map { $_->unlist }
           map { (blessed $_ ? $_ : TokenizeInternal($_)) } ($defn->getExpansion, @tokens)),
-      scope => 'global'); }
+      nopackParameters => 1, scope => 'global'); }
```
dginev, line 2289: Right, so the perl `AddToMacro` tries to emulate a latex technique which leverages `\the`. It is completely natural that it would need this guard added to it, and we (= I) may want to grep over latex.ltx to find other cases using `\the` that we have reimplemented in perl. They'll need that flag as well.

t/tokenize/hashes.xml
```diff
@@ -21,4 +21,38 @@
...
+      <p>Reading file badchars.tex:<break/>
+Line:“Bad chars: ^b$@##˙x ! ”<break/>
+
+Line:“More Bad chars: ^b$@##˙x ! ”<break/>
+
+Line:“And More Bad chars: ^b$@##˙x ! ”<break/>
+
+Line:““par ”<break/>
+
+Full Contents:“Bad chars: ^b$@##˙x ! More Bad chars: ^b$@##˙x ! And More Bad chars: ^b$@##˙x ! “par ”<break/></p>
```
dginev, line 55: The test matches the PDF nicely. I'm only doubting my eyes a little with the `Full Contents` spacing at the ends of the lines - the PDF seems to have an extra space after the exclamation chars. But that may be fixed by the other PR?

Looks good!

dginev (approved):

t/tokenize/hashes.xml
brucemiller: Hmm, is that two spaces, or just a wide space (due to punctuation?). Funny, I can't see that at all!!

PR #1516, opened 2021-04-17
For JATS, include ltx:title/ltx:tag

### LINKED ISSUE #1503: Theorem title in Jats-XML
When i convert a theorem to JATS-XML the title is not appearing in the XML.

For a theorem environment defined like

**test.tex**
```tex
\documentclass{article}
\newtheorem{theorem}{Theorem}[section]

\begin{document}
\begin{theorem}
Let f be a function.
\end{theorem}
\end{document}
```
I get with `latexml test.tex`:
```xml
<?latexml searchpaths="/home/robert/Work/ems/tex-json"?>
<?latexml class="article"?>
  <theorem class="ltx_theorem_theorem" inlist="thm theorem:theorem" xml:id="S0.Thmtheorem1">
    <tags>
      <tag>Theorem 0.1</tag>
      <tag role="refnum">0.1</tag>
      <tag role="typerefnum">Theorem 0.1</tag>
    </tags>
    <title class="ltx_runin"><tag><text font="bold">Theorem 0.1</text></tag></title>
    <para xml:id="S0.Thmtheorem1.p1">
      <p><text font="italic">Let f be a function.</text></p>
    </para>
  </theorem>
</document>
```

When i convert to JATS-XML with `latexmlc test.tex --dest=test.jats.xml --pmml --stylesheet=LaTeXML-jats.xsl`:
```xml
<?xml version="1.0"?>
<article>
  <front>
    <article-meta>
      <contrib-group/>
      <!-- The element theorem with attributes
	class=ltx_theorem_theoreminlist=thm theorem:theoremxml:id=S0.Thmtheorem1fragid=S0.Thmtheorem1 
      is currently not supported for the front matter.
    -->
    </article-meta>
  </front>
  <body>
    <statement id="S0.Thmtheorem1">
      <title/>
      <p id="S0.Thmtheorem1.p1">
        <italic>Let f be a function.</italic>
      </p>
    </statement>
  </body>
  <back>
    <!-- The element theorem with attributes
	class=ltx_theorem_theoreminlist=thm theorem:theoremxml:id=S0.Thmtheorem1fragid=S0.Thmtheorem1 
      is currently not supported for the back matter
    -->
    <app-group/>
  </back>
</article>
```

It seems the conversion should happen here, but is not picking up the title:
```xml
<xsl:template match="ltx:theorem/ltx:title">
    <title>
      <xsl:apply-templates select="@*|node()"/>
    </title>
  </xsl:template>
```

brucemiller:
Include ltx:title/ltx:tag when converting to JATS; update test case

fixes #1503

### DIFF
--- a/t/daemon/formats/jats.xml
+++ b/t/daemon/formats/jats.xml
@@ -15,7 +15,7 @@
             <addr-line>Somewhere, USA</addr-line>
           </address>
           <!-- The element contact with attributes
-        role=current_address 
+        role=current_address
       is currently not supported for the front matter.
     -->
           <email>joe@blow</email>
@@ -54,7 +54,7 @@
   </front>
   <body>
     <sec id="S1">
-      <title>Introduction</title>
+      <title>1. Introduction</title>
       <p id="S1.p1">As discussed in <xref ref-type="bibr" rid="bib.bib1">[AS64]</xref>,
 Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
       <p>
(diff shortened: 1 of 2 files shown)


dginev (approved):

t/daemon/formats/jats.xml
```diff
@@ -54,7 +54,7 @@
   </front>
   <body>
     <sec id="S1">
-      <title>Introduction</title>
+      <title>1. Introduction</title>
```
dginev, line 57: The one question I'd have here is whether there is a way to wrap some ephemeral element around the numbers, also in JATS. I have found that very useful in the HTML output as for certain NLP tasks being able to reliably discard the numbering makes life very convenient.

It's not the hardest regex to write by hand if you were using this JATS form, so just something to remark in passing.

dginev: Ah, I opened the PRs in inverse order, just saw that you've raised the same point in the other one. Apologies, feel free to disregard.

PR #1517, opened 2021-04-19
Fix scantokens

### LINKED ISSUE #1515: Discrepancy with behaviour of tex primitives (related to tokenization, \newlinechar and \scantokens)
LaTeXML seems to tokenize characters differently than TeX does; specifically, it seems to tokenize *beyond* the `\newlinechar` in `\scantokens`.

Here's a "minimal" example:
If the main file looks like this:
```tex
\documentclass{article}
\begin{document}
	\def\foo{foo!}
	\def\bar{bar!}
	\makeatletter\def\foo@bar{foobar!}\makeatother
	\newread\myread
	\def\readloop{
		\ifeof\myread\else
			\read\myread to \filecontenttemp
			\edef\filecontent{\filecontent\filecontenttemp^^J}
			\readloop
		\fi
	}
	\def\readfile{
		\catcode`\\=12
		\immediate\openin\myread=test2.tex\relax
		\def\filecontent{}
		\readloop
		\catcode`\\=0
		\immediate\closein\myread
	}
	\readfile
	\message{\meaning\filecontent}
	\expandafter\scantokens\expandafter{\filecontent^^J}
\end{document}
```

and `test2.tex` looks like this:
```tex
\foo
\bar
\makeatletter
\foo@bar
\makeatother
\foo\bar
```

...then TeX will produce as output `foo!bar!foobar!foo!bar!`, but LaTeXML will tokenize the `\foo@bar` *before* the `\makeatletter` is executed, yielding as output `foo!bar!foo!@bar foo!bar!`

brucemiller:
This is a more accurate implementation of `\scantokens`, which sets up a mouth for the imaginary file, but defers reading from it, so that expansion/digestion interleaving works as expected.

To make things interesting, it seems to introduce an extra space (as a newline in LaTeXML's case) in the `multi-escaped-param` test. But if you examine the pdf carefully, I think you'll find that there are 3, not 2, spaces before the result.

And even more interesting, `tools/maketests` formerly discarded blank lines, which gave different test results between `make test` and `maketests`.

What fun! Should be all better now.

Fixes #1515

### DIFF
--- a/lib/LaTeXML/Package/eTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/eTeX.pool.ltxml
@@ -16,6 +16,8 @@ use warnings;
 use LaTeXML::Package;
 
 # See http://tex.loria.fr/moteurs/etex_ref.html
+# Or better yet, see the full manual
+# http://texdoc.net/texmf-dist/doc/etex/base/etex_man.pdf
 # Section 3. The new features
 
 #======================================================================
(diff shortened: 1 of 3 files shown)


dginev (approved):
No objections here 👍🏻

PR #1519, opened 2021-04-20
Fix bigop sizes

### LINKED ISSUE #1502: \big(vee|wedge|otimes|...) are rendered with a non-optimal character
Now that I have noticed, I can't unsee it: operators like `\bigvee`, `\bigwedge`, `\bigotimes` are rendered as `\vee`, `\wedge`, `\otimes` with an extra `mathsize="160%"`, but they have their own native Unicode 1.1 counterparts, and those render better.

I suspect this affects many/all of the "n-ary" characters (e.g. in [this list](https://www.fileformat.info/search/google.htm?q=n-ary&domains=www.fileformat.info&sitesearch=www.fileformat.info&client=pub-6975096118196151&forid=1&channel=1657057343&ie=UTF-8&oe=UTF-8&cof=GALT%3A%23008000%3BGL%3A1%3BDIV%3A%23336699%3BVLC%3A663399%3BAH%3Acenter%3BBGC%3AFFFFFF%3BLBGC%3A336699%3BALC%3A0000FF%3BLC%3A0000FF%3BT%3A000000%3BGFNT%3A0000FF%3BGIMP%3A0000FF%3BFORID%3A11&hl=en), although I am sure there are better references). I assume all the good fonts (MathJax TeX, STIX2, etc.) have 100% coverage of those.

Unless there is a strong backward compatibility reason, I'll start removing all the `mathsize`'s which can be done directly in Unicode. I hope it's not that many!

brucemiller:
This PR uses the correct (I think) Unicode for various `\bigXXX` symbols in TeX and also a few in stmaryrd.  It also treats those in the latter style as having movable scripts.

I couldn't find large size Unicode for `\bigbox, \bigcurlywedge,\bigcurlyvee,\bigtriangledown,\bigtriangleup`, so I kept the Big sizing.  If you find the right codes, let me know! (Also the rendering for `\bigparallel` is dubious, but the codepoint seems to be the correct one).

Fixes #1502

Thanks, @xworld21

### DIFF
--- a/t/fonts/stmaryrd.xml
+++ b/t/fonts/stmaryrd.xml
@@ -750,29 +750,29 @@
               <XMRef idref="S0.Ex11.m1.9"/>
             </XMApp>
             <XMWrap>
-              <XMTok fontsize="160%" name="bigbox" role="BIGOP" xml:id="S0.Ex11.m1.1">□</XMTok>
+              <XMTok fontsize="160%" mathstyle="display" name="bigbox" role="SUMOP" scriptpos="mid" xml:id="S0.Ex11.m1.1">□</XMTok>
               <XMTok font="italic" name="quad" role="PUNCT"> </XMTok>
-              <XMTok fontsize="160%" name="bigcurlywedge" role="BIGOP" xml:id="S0.Ex11.m1.2">⋏</XMTok>
+              <XMTok fontsize="160%" mathstyle="display" name="bigcurlywedge" role="SUMOP" scriptpos="mid" xml:id="S0.Ex11.m1.2">⋏</XMTok>
               <XMTok font="italic" name="quad" role="PUNCT"> </XMTok>
-              <XMTok fontsize="160%" name="bigcurlyvee" role="BIGOP" xml:id="S0.Ex11.m1.3">⋎</XMTok>
+              <XMTok fontsize="160%" mathstyle="display" name="bigcurlyvee" role="SUMOP" scriptpos="mid" xml:id="S0.Ex11.m1.3">⋎</XMTok>
               <XMTok font="italic" name="quad" role="PUNCT"> </XMTok>
-              <XMTok fontsize="160%" name="biginterleave" role="BIGOP" xml:id="S0.Ex11.m1.4">⫼</XMTok>
+              <XMTok mathstyle="display" name="biginterleave" role="SUMOP" scriptpos="mid" xml:id="S0.Ex11.m1.4">⫼</XMTok>
               <XMTok font="italic" name="quad" role="PUNCT"> </XMTok>
               <XMDual role="ADDOP" xml:id="S0.Ex11.m1.5">
                 <XMTok mathstyle="display" meaning="intersection-plus" name="bignplus" role="ADDOP"/>
                 <XMWrap rule="kludge">
-                  <XMTok fontsize="160%" mathstyle="display" meaning="intersection" name="bigcap" role="SUMOP" scriptpos="mid">⋂</XMTok>
+                  <XMTok mathstyle="display" meaning="intersection" name="bigcap" role="SUMOP" scriptpos="mid">⋂</XMTok>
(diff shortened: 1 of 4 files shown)


t/fonts/stmaryrd.xml
```diff
@@ -750,29 +750,29 @@
               <XMRef idref="S0.Ex11.m1.9"/>
             </XMApp>
             <XMWrap>
-              <XMTok fontsize="160%" name="bigbox" role="BIGOP" xml:id="S0.Ex11.m1.1">□</XMTok>
+              <XMTok fontsize="160%" mathstyle="display" name="bigbox" role="SUMOP" scriptpos="mid" xml:id="S0.Ex11.m1.1">□</XMTok>
```
dginev, line 753: Was the math role change from `BIGOP` to `SUMOP` intentional here?

xworld21:
> I couldn't find large size Unicode for `\bigbox, \bigcurlywedge,\bigcurlyvee,\bigtriangledown,\bigtriangleup`, so I kept the Big sizing. If you find the right codes, let me know! (Also the rendering for `\bigparallel` is dubious, but the codepoint seems to be the correct one).

MathJax and unicode-math use 25B3 and 25BD for `\bigtriangle{up,down}`, and they claim U+2016 is `\Vert` (in html `&amp;Vert;', actually), so those seem ok. I have found a "white large square", but it's not at all a large "white square"! I also couldn't find the curly ones anywhere.

brucemiller:
> Was the math role change from |BIGOP| to |SUMOP| intentional here?

Semi-intentional, I copied the code from TeX's `\bigXXX`. Probably 
there's no real distinction; the important thing was to do the variable 
sizing and script positioning (which does happen with the stmaryrd symbols).

brucemiller:
> MathJax and unicode-math use 25B3 and 25BD for |\bigtriangle{up,down}|, 
> and they claim U+2016 is |\Vert| (in html `&Vert;', actually), so those 
> seem ok. 

How do they distinguish the sizes between the `\big` and non-`\big`?

dginev:
> Was the math role change from |BIGOP| to |SUMOP| intentional here?
> Semi-intentional, I copied the code from TeX's `\bigXXX`. Probably there's no real distinction; the important thing was to do the variable sizing and script positioning (which does happen with the stmaryrd symbols).

Ok, as long as it's considered kosher, all good on my end. But I'm starting to get an inkling we need a grammar refactor related to `BIGOP` and `SUMOP`, which may be just a strategic rename at the minimal, to make clear that not everything marked `SUMOP` is a sum, and not everything big is marked `BIGOP`.

xworld21:
> How do they distinguish the sizes between the `\big` and non-`\big`?

Ah, I didn't notice it looks the same as `\triangledown`!

My understanding after a quick look:
- unicode-math flags U+25BF (small triangle) as the non-big version (see [page 31](http://mirror.ox.ac.uk/sites/ctan.org/macros/unicodetex/latex/unicode-math/unimath-symbols.pdf)), even though `\smalltriangledown` is a thing;
- MathJax renders U+25BF the same as U+25BD when presented as unicode; from LaTeX `\triangledown`, it adds the class attribute `TEX-A` to use a font which renders U+25BD with a smaller glyph.

Oh well. So font size it is, I guess.

brucemiller:
> Oh well. So font size it is, I guess.

No problem; looking forward to getting a new Issue raised after a few 
more Unicode updates :>

brucemiller:
>  But I'm 
> starting to get an inkling we need a grammar refactor related to |BIGOP| 
> and |SUMOP|,

Indeed. Keeping in mind that the roles are primarily syntactic, rather 
than semantic, and that the current status reflects all the flaws of 
evolution relative to design... The idea behind BIGOP was primarily 
operators which act upon a following expression, typically w/o 
parentheses and likely without scripts; whereas SUMOP adds scripts that 
can be over/under or following. The latter is like summation, but not 
necessarily linear sum-like.  There are likely few, if any, places in 
the software where they're distinguished.

So, yeah, a refactoring and renaming should happen some day :>

PR #1520, opened 2021-04-21
fixup tei stylesheet for tabular with and without thead/tbody

brucemiller:
As it says; this should make the TEI stylesheet work correctly on tabulars with and with `ltx:thead` and `ltx:tbody` (which are produced either by extra markup, or `GUESS_TABULAR_HEADERS`)

This was discovered on a tangent to #1518

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-tei.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-tei.xsl
@@ -781,11 +781,11 @@
         </table>
     </xsl:template>
 
-    <xsl:template match="ltx:tabular/*">
+    <xsl:template match="ltx:tabular">
             <xsl:apply-templates select="@*|node()"/>
     </xsl:template>
 
-    <xsl:template match="ltx:tbody">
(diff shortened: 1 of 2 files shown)


brucemiller:
Might as well piggy-back that fix onto this PR.

dginev (approved):
👍🏻

lib/LaTeXML/resources/XSLT/LaTeXML-tei.xsl
```diff
@@ -781,11 +781,11 @@
         </table>
     </xsl:template>
 
-    <xsl:template match="ltx:tabular/*">
+    <xsl:template match="ltx:tabular">
             <xsl:apply-templates select="@*|node()"/>
     </xsl:template>
 
-    <xsl:template match="ltx:tbody">
+    <xsl:template match="ltx:thead | ltx:tbody">
```
xworld21, line 788: Also `tfoot`, I think?

brucemiller: Oh, yeah. Avoided PR dance and pushed directly in 58e6b1e2

PR #1524, opened 2021-05-04
Expl3 optimizatoin

brucemiller:
Yep, misspelled. A series of smallish, but cumulative, optimizations. The main strategy is avoiding layers of subroutine calls, using operations that are faster in Perl, and open-coding accessors where not too offensive. Generally focussed on speeding up the expl3 test case, which appears to be 20% faster (tikz samples about 10%, make test around 7-8%)

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -268,7 +268,7 @@ sub isColumnEnd {
   return; }
 
 sub readToken {
-  my ($self, $toexpand) = @_;
+  my ($self) = @_;
   #  my $token = shift(@{$$self{pushback}});
   my ($token, $cc, $atoken, $atype, $ahidden);
   while (1) {
@@ -319,17 +319,14 @@ sub readXToken {
   return shift(@{ $$self{pending_comments} }) if $commentsok && @{ $$self{pending_comments} };
   my ($token, $cc, $defn, $atoken, $atype, $ahidden);
   while (1) {
-    while ($token = shift(@{ $$self{pushback} })) {    # Check in pushback
-      if (($cc = $$token[1]) == CC_SMUGGLE_THE) {      # ONLY in pushback!
-        return $LaTeXML::SMUGGLE_THE ? $token : $$token[2]; }    # Expands to smuggled token
-      elsif ($cc == CC_COMMENT) {
+    # NOTE: CC_SMUGGLE_THE should ONLY appear in pushback!
+    while (($token = shift(@{ $$self{pushback} })) && $CATCODE_HOLD[$cc = $$token[1]]) {
+      if ($cc == CC_COMMENT) {
         return $token if $commentsok;
         push(@{ $$self{pending_comments} }, $token); }
       elsif ($cc == CC_MARKER) {
-        $self->handleMarker($token); }
-      else {
-        last; } }
-    if (!defined $token) {                                       # Else read from current mouth
+        $self->handleMarker($token); } }
+    if (!defined $token) {    # Else read from current mouth
       while (($token = $$self{mouth}->readToken()) && $CATCODE_HOLD[$cc = $$token[1]]) {
         if ($cc == CC_COMMENT) {
           return $token if $commentsok;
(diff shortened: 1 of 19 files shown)


dginev:
Quick confirmation on the speedup claim, on my desktop PC I am seeing:
```
$ time make test TEST_FILES=t/83*
```

master 
```
real	1m12.626s
user	1m12.130s
sys	0m0.417s
```

PR optimization branch

```
real	0m56.832s
user	0m56.400s
sys	0m0.402s
```

which is indeed a 25% speedup. Nice!

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -319,17 +319,14 @@ sub readXToken {
   return shift(@{ $$self{pending_comments} }) if $commentsok && @{ $$self{pending_comments} };
   my ($token, $cc, $defn, $atoken, $atype, $ahidden);
   while (1) {
-    while ($token = shift(@{ $$self{pushback} })) {    # Check in pushback
-      if (($cc = $$token[1]) == CC_SMUGGLE_THE) {      # ONLY in pushback!
-        return $LaTeXML::SMUGGLE_THE ? $token : $$token[2]; }    # Expands to smuggled token
-      elsif ($cc == CC_COMMENT) {
+    # NOTE: CC_SMUGGLE_THE should ONLY appear in pushback!
```
dginev, line 322: good note, and indeed.

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -338,47 +335,47 @@ sub readXToken {
...
+      return ($cc != CC_SMUGGLE_THE) || $LaTeXML::SMUGGLE_THE ? $token : $unexpanded; }
+    ## Wow!!!!! See TeX the Program \S 309
     elsif (!$LaTeXML::ALIGN_STATE    # SHOULD count nesting of { }!!! when SCANNED (not digested)
       && $LaTeXML::READING_ALIGNMENT
       && (($atoken, $atype, $ahidden) = $self->isColumnEnd($token))) {
       $self->handleTemplate($LaTeXML::READING_ALIGNMENT, $token, $atype, $ahidden); }
-    ## Note: special-purpose lookup in State, for efficiency
-    elsif (defined($defn = LaTeXML::Core::State::lookupExpandable($STATE, $token, $toplevel))) {
+    ## Note: use general-purpose lookup, since we may reexamine $defn below
+    elsif ($LaTeXML::Core::State::CATCODE_ACTIVE_OR_CS[$$token[1]]
```
dginev, line 348: minor but can't you use `$cc` instead of another lookup to `$$token[1]` in here?

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -338,47 +335,47 @@ sub readXToken {
...
+      && defined($defn = LaTeXML::Core::State::lookupMeaning($STATE, $token))
```
dginev, line 349: also, unsure if there are performance implications, but it's a bit briefer to write `$STATE->lookupMeaning` than the explicit Perl 4 namespaced subroutine call. But I wouldn't be surprised if you reply that you've measured and the arrow syntax is somehow slower :>

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -338,47 +335,47 @@ sub readXToken {
...
-          push @expansion, $exp_t; }
-        elsif ($r eq 'LaTeXML::Core::Tokens') {
-          push @expansion, @$exp_t; }
-        else {
-          Fatal('misdefined', $r, undef, "Expected a Token, got " . Stringify($_)); } }
+      my $r;
+      my @expansion = map { (($r = ref $_) eq 'LaTeXML::Core::Token' ? $_
+          : ($r eq 'LaTeXML::Core::Tokens' ? @$_
+            : Fatal('misdefined', $r, undef, "Expected a Token, got " . Stringify($_),
+              "in " . ToString($defn)))) }
```
dginev, line 358: I'll take your word that the map is faster than the for loop, I guess you never know until you benchmark... I recall rewriting the map to the for here, so my bad if I made it slower.

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -535,30 +531,51 @@ sub readKeyword {
   return; }
 
 # Return a (balanced) sequence tokens until a match against one of the Tokens in @delims.
-# In list context, also returns the found delimiter.
+# Note that Braces on input hides the contents from matching,
+# so this assumes there wont be braces in $delim!
+# But, see readUntilBrace for that case.
+# NOTE: Add error handling, $token undef detection, etc!!!!!
```
dginev, line 537: Nice `NOTE` :>

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -535,30 +531,51 @@ sub readKeyword {
...
+          @ring = (); }                                                # and retry
+        else {
+          push(@ring, $token); } }
+      my $i;
+      for ($i = 0 ; ($i < $ntomatch) && ($ring[$i]->equals($want[$i])) ; $i++) { }    # Test match
+      last if $i >= $ntomatch;                                                        # Matched all!
+      push(@tokens, shift(@ring)); } }
+  if (!defined $token) {                                                              # Ran out!
+    $self->unread(@tokens);    # Not more correct, but maybe less confusing?
+    return; }
```
dginev, line 573: A bit scary that `readUntil` got a refactor of this size. Is the new code closer to TeX, or ideally functional identical to the old?

lib/LaTeXML/Core/Parameters.pm
```diff
@@ -60,10 +60,8 @@ sub readArguments {
   my ($self, $gullet, $fordefn) = @_;
   my @args = ();
   $gullet->setup_scan();
-  foreach my $parameter (@$self) {
-    my $value = $parameter->read($gullet, $fordefn);
-    push(@args, $value) unless $$parameter{novalue}; }
-  return @args; }
+  my ($p, $v);
+  return map { $p = $_; $v = $p->read($gullet, $fordefn); ($$p{novalue} ? () : $v); } @$self; }
```
dginev, line 64: is this map _also_ faster than foreach? Should I entirely revise my rule of thumb and start treating map as faster? Shocking if so.

dginev (approved):
Impressive that tests pass with such a variety of refactors, nice! Big fan of the simplification diffs, mildly worried about hot-coding some of the methods, but generally got a "makes sense, should work" impression from the PR.

Especially nice find with avoiding to create identical objects and returning `$self` when a method doesn't perform any modifications.

lib/LaTeXML/Core/Gullet.pm
brucemiller: indeed; too much cut'n'paste rearranging!

lib/LaTeXML/Core/Gullet.pm
brucemiller: Can't say that it's true in general, but seemed so in these cases.

lib/LaTeXML/Core/Gullet.pm
brucemiller: yeah, except I actually did add the error handling, but forgot to remove the comment!

lib/LaTeXML/Core/Gullet.pm
brucemiller: Yeah, scary perhaps, 'cause it is a change to api to conform with the way it's actually used (which then allowed for some nice optimization).

lib/LaTeXML/Core/Parameters.pm
brucemiller: Seems that it was. But then I was benchmarking on a windy day, so who knows?

PR #1525, opened 2021-05-06
softer Fatal errors

dginev:
Work-in-progress, focused on usability, and especially the user story of:
> I tried using latexml, but it produced nothing on my first try, so I gave up without investigating.

The basic idea is avoiding the hard `die()` calls in (core processing) Fatals, and instead resolving them by strategically resetting all available runtime state, so that the only next step in the usual processing flow is to end the Core phase and serialize to output whatever was processed so far.

Because this approach tries to remain inside the main processing flow, it needs a number of sanity tests. Fatals occurring at each stage - during reading (Mouth), expansion (Gullet), digestion (Stomach), construction (Document), math parsing, rewrite, etc. may need to be guarded differently. 

Some perl-based fatals also may need special attention. Timeouts are a curious case, since the `die()` call was a bit of a solid guarantee that we will escape infinite loops without negotiations. With the simple `return;`, it becomes crucial to vet the possible causes of infinite loops and preemptively remediate them in the Fatal call. That is somewhat impossible in general - you'd need to vet each loop in latexml. But the common cases can be vetted - in particular infinite TeX expansion. Probably I should add a detection of cases where the Fatal just can't escape out of the trap it is in, and have the `die()` as a final fallback.

But for that I need more tests. So, WIP status until I prep those. Feedback already welcome!


dginev:
Tricky CI situation, the new timeout fatal test seems to be failing based on the exact moment of execution when the `alarm()` fires. Sadly I can't reproduce this locally at all (always passes here), but will take a closer look ...

dginev:
Ok, lesson learned - CI is a bit unpredictable when redirecting a stream in a system() call a-la unix. In my case it was:
```
2>/dev/null
```

Instead, I now moved to using `IPC::Open3`, which keeps the testing harness nice and quiet while behaving very cleanly both locally and in CI. So now I have the Fatals `timeout`s consistently producing partial output. 

PR is good for review, I only need to figure out which other tests need adding, to avoid nasty surprises in arXiv.

dginev:
I just ran through `astro-ph/0002507` which is in the [undexpected:<endgroup>](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/unexpected/%3Cendgroup%3E?all=false) Fatals report, and was pleased to see most - virtually all - of the content successfully transition to HTML successfully. The fatal in that specific example was a follow-up to:
```
Error:unexpected:\end{multicols} Can't close environment multicols;
```

itself due to reading a custom local `emulate_apj.sty` file. So definitely a rare long-tail kind of Fatal. It's encouraging to see that we will be able to recover some reasonable partial HTML even in these hard cases, and this may also improve arXiv's perception of the latexml conversion runs.

dginev:
Impressively, I managed to recover [1107.4709](https://arxiv.org/abs/1107.4709) today - which was converting mostly successfully but hit a very minor technicality in the `.bbl` file which lead to a perl:die and no output.

It was very helpful to trace the exact recovery point Perl returns to when a die is triggered, which I hadn't tested prior - apparently it's only captured at the eval {} block in LaTeXML.pm that called `->digestFile`. So that took a bit of separation and care to be able to re-enter `->finishDigestion` in cases where some boxes got rescued during the Fatal.

And very impressively I saw the entire thesis, even though still with 5 warnings, 33 errors and 1 fatal error, produce a large and readable HTML output.

For bonus points I also patched the specific Fatal that caused the issue, but it was a rare edge case, so the general treatment is really what's of value (in 7cd4426  )

dginev:
Should also cover the later stage `->convertDocument` Fatal recovery now, e.g. for `physics0505012` in CorTeX. I also patched the perl:die I debugged with as a bonus, related to `ltx:ERROR` elements sneaking into math parser on broken docs. Surprisingly good output got generated with the parsing fatal, encouraging!

dginev:
PR is looking very solid against the arXiv `Fatal:misdefined` subreport. I patched a number of needless Fatals after also testing that the documents produce meaningful partial outputs when encountering them, a win-win all around. Almost done with my vetting steps...

dginev:
Done testing (with arXiv) here. Seems robust, and happily a lot of the later tests I tried had the Fatal already resolved either by the latest master, or by the small patches I sprinkled in.

May take a stylistic re-read in the morning, but I will be rerouting to the logging PR next, and this one can follow it.

t/96_fatal.t
```diff
@@ -1,40 +1,47 @@
...
+  # DWYM - if this runs for the first time, add the .test.html as the reference file.
+  if (! -f "$broken_base.html") {
+    pathname_copy("$broken_base.test.html","$broken_base.html")
+  }
+  if (my $testhtml = process_htmlfile("$broken_base.test.html", $broken_base)) {
+    if (my $html = process_htmlfile("$broken_base.html", $broken_base)) {
+      is_strings($testhtml, $html, $broken_base); } }
+  # cleanup the test file
+  unlink("$broken_base.test.html");
 }
```
brucemiller, line 44: (without having followed all the details) Is it still the case that `tools/maketests` can cope with (in particular, create them!) the Fatals tests?

dginev: The fatals tests had no output, so until now nothing generated them. I haven't followed the `tools/maketests` to make these examples, would need to recast them as daemon `.spec` tests if you'd like that to be possible.

brucemiller: Oh, I thought they were daemon tests. I think it would be good if they fit under `maketests` umbrella, since that makes them much easier to update (if the inner XML changes), or add new tests. Otherwise, there's a lot of digging for exactly how they're supposed to be run.  I guess the fact they (eventually) die may add a new wrinkle, though.

brucemiller:
pretty overwhelming, but perhaps less confusing w/o conflicts? :>

dginev:
Rebased and refactored some of the bits I changed to the new logging. Happy I added tests so that I don't have to guess if the various bits still work - they should!

But I was quite brave with some of the changes, so an extra couple of tests may be needed. Especially with the recursive Fatals, I can't quite tell if I got all the possible "die" combos.

MANIFEST
```diff
@@ -880,13 +880,24 @@ t/complex/xii.pdf
...
 t/daemon/api/port.xml
 t/daemon/citations.tex
 t/daemon/complex/exhaustive.spec
 t/daemon/complex/exhaustive.xml
 t/daemon/document.tex
+t/daemon/fatal_100.tex
+t/daemon/fatal_egroup.tex
+t/daemon/fatal_misdefined.tex
+t/daemon/fatals/fatal_100.spec
+t/daemon/fatals/fatal_100.status
```
brucemiller, line 893: Hadn't we removed `*.status` files?

dginev: Dangers of time-traveling... removed again.

lib/LaTeXML/Core/State.pm
```diff
@@ -709,15 +709,15 @@ sub getStatusMessage {
   my ($self)         = @_;
   my $status         = $$self{status};
   my @report         = ();
-  my $warning_status = $$status{warning} && "$$status{warning} warning" . ($$status{warning} > 1 ? 's' : '');
-  my $error_status = $$status{error} && "$$status{error} error" . ($$status{error} > 1 ? 's' : '');
-  my $fatal_status = $$status{fatal} && "$$status{fatal} fatal error" . ($$status{fatal} > 1 ? 's' : '');
+  my $warning_status = $$status{warning} && colorizeString("$$status{warning} warning" . ($$status{warning} > 1 ? 's' : ''),'warning');
```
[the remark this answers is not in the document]
dginev: Error.pm sounds like a great place for this indeed! Should I move it in this PR?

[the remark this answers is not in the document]
brucemiller: Perhaps it's more related to the eventual LaTeXMLpm refactor. Clearly there needs to be some kind of job-specific status "object" (even if it's just a hash); there's some necessary communication between LaTeXMLpm, Error.pm, possibly State....  So, maybe not quite yet :>

dginev:
I also noticed some messiness when running with a terminal spinner and encountering a fatal. So I added explicit calls for `UseSTDERR(undef)` to latexml and latexmlc, with an extra `_spinnerclear()` when STDERR is getting released. So now Fatals on the command line clean up after themeselves for those two executables.

Maybe I should also add this to latexmlmath and latexmlpost?

dginev:
Running some more tests here and monitoring the logging. I spotted I made a mistake that broke the locators in Fatals, as I reset the gullet stack *before* I serialized the message out, thus always getting a default locator string. Added a commit that defers the "stack reset" piece of logic in Fatal to last stage of the subroutine, which corrects my mistake.

These were especially visible if you ctrl+c and trigger an interrupt Fatal, which is one way to escape an infinite loop.

dginev:
Three more upgrades in my last commit:
 - double-check my newly introduced calls to `$state` in Fatal actually have a defined state present
 - use the return value of `alarm()` to have proper tight timeout bounds - thanks to Bruce for pointing that out!
 - added some fake mistakes in post-processing that created "die", "timedout" and "interrupted" Fatals during the ProcessChain call. LaTeXML.pm was a bit sloppy with that, should be much cleaner now - no doubly wrapped Fatals (die+timedout) or sloppy messages.

LaTeXML.pm is definitely on a path where a clean refactor will become obvious - some subroutines can be extracted out (e.g. the alarm-eval guards), and some of the initialization/session logic can be made less redundant and more compartmentalized. Likely as a follow-up PR some time soon after.

lib/LaTeXML/Core/Document.pm
```diff
@@ -562,15 +562,17 @@ sub serialize_attr {
 sub absorb {
   my ($self, $object, %props) = @_;
   no warnings 'recursion';
-  # Nothing? Skip it
-  my @boxes   = ($object);
+  # Record the boxes to absorb in the Document object, so that we can recover
+  # during Fatal errors by dropping the remaining ones.
+  $$STATE{boxes_to_absorb} = [$object];
+  my $boxes   = $$STATE{boxes_to_absorb};
```
brucemiller, line 568: Isn't this whole `boxes_to_absorb` thing a no-op?  Although Error clears it, it's never really used.

dginev: It is currently a no-op in the available tests yes. I would need to get extra crafty to be able to design a test for what I was trying to rescue, namely a case where a Fatal is encountered while looping over the `@$boxes` in `Document::absorb` and the flow is returned back to that loop after the Fatal is handled.

If those `$boxes` don't get reset to `[]` the loop would continue, rather than terminate right away. But none of my tests have this type of behavior captured yet.

brucemiller: Nothing about tests, it's just a no-op as it is: assigning $$STATE{boxes_to_absorb}=[]` doesn't affect the local array reference `$boxes`. It would if you modified the referenced array itself, it would, but in that direction dragons lie.

We're back to the observation that we really do not want to try to blindly resume. And in the case of `absorb`, it's relatively easy: Fatal should throw (die) to the enclosing catch (eval), and you get whatever document was built up to that point.  Which, I'm guessing, is what you're trying to achieve by clearing the pending boxes (except that you don't).

dginev: Ah, you're right, my code swaps the reference, rather than empty the existing reference. I should've written:

```perl
my $boxes_to_absorb = $$STATE{boxes_to_absorb};
@$boxes_to_absorb=() if ref $boxes_to_absorb eq 'ARRAY';
```

"Blindly resuming" has the potential to recover with less breakage than the hard die, which is why I've attempted it as a first stab. Of course it also needs to be tested well, and I'm missing a test here.

The problem with switching to `die` everywhere is that I'd need to refactor the Core.pm methods to also use `eval {}` so that e.g. the math parsing can take place after a fatal in say `absorb`. Currently a die in absorb will float all the way up to LaTeXML.pm, concluding the `Core::convertDocument` phase right then and there. The returns were trying to dance around that.

dginev:
We never really resolved how to proceed with the PR ideas, and I keep hoping we somehow see a best path forward. Some ideas that were floated were:
1. Merge it largely as-is with a careful review
2. Refactor most Fatals down to Errors, and only leave the really hard "process deaths" as Fatals
    - Maybe emulate the PR behavior for these error cases with a new switch on the lines of `Error(... , finalize => 1)` which will yank the remaining processing data.
3. Remain with a hard `die` for all Fatals, and lose some of the recovery potential, but remain identical to master in the runtime flow when Fatals are encountered.
4. Defer any of these changes until after LaTeXML.pm has been refactored to a point where latexml and latexmlc can be reconciled, and maybe each individual step of processing has its own `eval { } ` scope.

Also worth noting that no PR is ever final code, and a later refactor of LaTeXML.pm will undoubtedly also touch these bits, in whichever way they end get resolved. There's a lot of Fatal guarding taking place in LaTeXML.pm.

lib/LaTeXML/Core/Document.pm
dginev: > in that direction dragons lie.

I was also hoping I could enumerate and tame these particular dragons. If a soft `return` from a fatal ends up creating/stepping on a trap door, I can currently imagine three scenarios:
1. processing hits a harsher issue, triggering an unrecoverable fatal (such as a perl die), and ends up `die()`-ing up to the LaTeXML.pm eval anyway, a little bit later
2. or, a new class of infinite loops gets introduced in situations I haven't yet thought about. Then latexml either
   -  reaches a predefined timeout and ends up with a hard `die` from the alarm or
   - the user gets fed up and interrupts the process, which also should trigger a hard `die`.

Maybe that's all too clunky / too much boilerplate for anyone else to be convinced in. Just explaining where I was coming from.

brucemiller:
On 6/14/21 9:28 PM, Deyan Ginev wrote:
> "Blindly resuming" has the potential to recover with less breakage

Yes, it has that potential. But it also has the potential to cause 
extremely confusing breakage. Indeed, since it is so wildly unnatural 
that "Fatal" just plows on through, it seems an almost guaranteed potential.

brucemiller:
On 6/14/21 10:02 PM, Deyan Ginev wrote:
>     in that direction dragons lie.
> 
> I was also hoping I could enumerate and tame these particular dragons. 

The particular dragon I'm worried about here is unmaintainability.  You 
have to pay very close attention to the fact that `$boxes` was 
effectively tied to an instance variable that is being modified behind 
your back.

brucemiller:
The more I think about it, and review the code, the more firmly I'm convinced of the (3)+(2)+(4) solution (referncing your cases above). Namely, Fatal must remain, well, Fatal, but with the PR (or something like it :> ) we manage to get *something* out of the processing. Then we proceed to maximize that something according to the nature of the errors and the characteristics of the processing.  For example, a (2) pass would convert any (current) Fatals that are actually safe-ish to Errors. A pre-(4) pass would involve more Core/LaTeXML refactoring but for example would recognize that for example when absorbing into the document you don't actually need much in the way of recovery --- you've got all you're going to get. But you do have to decide on and arrange for any further processing (math parsing, xslt, etc).  That refactoring amounts to a the beginning of the reconsilliation for (4).

What I'm afraid of with (1) is on a couple of fronts. We'll be building in mysterious mechanisms that will likely make future debugging and refactoring harder. And, if by luck we manage to survive blindly plowing through fatals and getting lots of big new documents, any attempt to refactor that doesn't get every last node out will be perceived as regression, and so harder socially and psychologically, as well as programming-wise.

brucemiller:
Good to factor out the separate simpler PRs, to keep focus, but I trust that a (slightly) simpler version of the PR is in the works?

dginev:
> Good to factor out the separate simpler PRs, to keep focus, but I trust that a (slightly) simpler version of the PR is in the works?

Yes, I'm trying to make all the cake be had and eaten... should be in shortly

PR #1529, opened 2021-05-12
JATS, TEI stylesheet cleanup

### LINKED ISSUE #1499: MathML namespace
When i compile LaTeX to XML the Math tags have a prefix `m:`.

For example

test.tex
```latex
\documentclass{article}
\begin{document}
$a = b^2$
\end{document}
```
with `latexmlc test.tex --dest=test.jats.xml --pmml --stylesheet=LaTeXML-jats.xsl` to JATS-XML

test.jats.xml
```xml
<?xml version="1.0"?>
<article>
  <front>
    <article-meta>
      <contrib-group/>
    </article-meta>
  </front>
  <body>
    <p id="p1">
      <inline-formula id="p1.m1">
        <m:math xmlns:m="http://www.w3.org/1998/Math/MathML" alttext="a=b^{2}" display="inline">
          <m:mrow>
            <m:mi>a</m:mi>
            <m:mo>=</m:mo>
            <m:msup>
              <m:mi>b</m:mi>
              <m:mn>2</m:mn>
            </m:msup>
          </m:mrow>
        </m:math>
      </inline-formula>
    </p>
  </body>
  <back>
    <app-group/>
  </back>
</article>
```

According to https://accessibility.psu.edu/math/mathml/ the m namespace
> refers to an earlier MathML variant used by Microsoft. As only Internet Explorer ever recognized it, it is recommend to avoid using m namespace MathML unless there is an issue of backwards compatability with a particular student.
Note: In m namespace MathML, all tags include a prefix m: (e.g. <m:math>)

Is this the case in LaTeXML? In other places i've seen `mml:` used or no prefix, which seems to be only required in certain settings (https://stackoverflow.com/questions/26093182/use-of-namespace-in-mathml-rendering)

Is it possible to configure this for LaTeXML output?

brucemiller:
Quite by accident, I discovered some oddities in these stylesheets and that they were not reusing the base LaTeXML stylesheet modules as much as I had expected. Among other things that made it quite easy to use the `mml` namespace prefix (even if it annoys :> ). And also discovered that instances `ltx:MathFork` (used for aligned sets of equations) was getting *all* math branches copied, resulting in redundant math.  Along with other general stylesheet formatting, namespace cleanup.

Fixes #1499.

Please @rgieseke, @xworld21 and any other JATS or TEI users, please take a look that I haven't inadvertently broken anything.

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-jats.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-jats.xsl
@@ -17,14 +17,15 @@
 \=========================================================ooo==U==ooo=/
 -->
 <xsl:stylesheet
-   version     ="1.0"
-   xmlns:xsl   ="http://www.w3.org/1999/XSL/Transform"
-   xmlns:ltx   ="http://dlmf.nist.gov/LaTeXML"
-   xmlns:str   ="http://exslt.org/strings"
-   xmlns:m     ="http://www.w3.org/1998/Math/MathML"
-   xmlns:xlink ="http://www.w3.org/1999/xlink"
-   extension-element-prefixes="str"
-   exclude-result-prefixes="ltx str m xlink">
+    version     ="1.0"
+    xmlns:xsl   ="http://www.w3.org/1999/XSL/Transform"
+    xmlns:ltx   ="http://dlmf.nist.gov/LaTeXML"
+    xmlns:str   ="http://exslt.org/strings"
+    xmlns:m     ="http://www.w3.org/1998/Math/MathML"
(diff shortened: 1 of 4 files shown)


dginev:
Neat, diffs look good to me on a quick skim. 

Do we know if there are official validators for JATS and TEI that we could double-check our testing targets with? Some of the ePub issues in my queue came exactly from the official ePub validator for example.

brucemiller:
I'm going to go ahead and merge (since the namespace stuff cause glitches in other branches, whatev') but it would be good to be able to validate indeed!

PR #1532, opened 2021-05-17
Terminal logs disabled by default, enable on demand

dginev:
Since we're moving to closer reliance to LaTeXML.pm, _and_ we want backwards compatibility with the third-party applications that rely on LaTeXML.pm, it seems inevitable that the terminal logging should be switched off by default.

Especially since all terminal-facing applications that use (or may use) LaTeXML.pm that we know of are in the main latexml repository - `latexml`, `latexmlpost`, `latexmlc` and `latexmlmath`. For these I switched the terminal logging on as early as possible, already at the package loading step, so that we also retain the new fancy terminal spinners as the default on their end.

There is _no_ new command-line option to control whether an executable uses the terminal or not. That is a non-negotiable feature of the application - it is either a terminal app, or some API-oriented service. Which seems reasonable based on past experience.

I also got blocked by perl critic when trying to commit, due to using a `select` idiom which is [frowned upon by Conway](https://metacpan.org/pod/Perl::Critic::Policy::InputOutput::ProhibitOneArgSelect), so I refactored that based on the best practice - which seems harmless in newer perls?


dginev:
Submitting here for early discussion of whether this is the right abstraction to control the terminal. This change gets `ltxmojo` specifically to act as before.

Oh, and I should also mention that I saw all uses of `generateMessage` now had a sub-call to `colorizeString`, so I refactored them to their intuitive dependent form - simplifies the calling pattern.

Makefile.PL
```diff
@@ -33,7 +33,7 @@ our ($KPSEV, $TEXMF);
 my ($texmfspec) = grep { /^TEXMF/ } @ARGV;
 if ($texmfspec && $texmfspec =~ /^TEXMF\s*=(.*)$/) {
   $TEXMF = $1;
-  @ARGV  = grep { $_ ne $texmfspec } @ARGV; }    # Remove so MakeMaker doesn't fret.
+  local @ARGV = grep { $_ ne $texmfspec } @ARGV; }    # Remove so MakeMaker doesn't fret.
```
dginev, line 36: Perl::Critic insisted...

dginev:
From what I've tested today, the changes in this PR seem to preserve the expected logs for [latexmls](https://github.com/dginev/LaTeXML-Plugin-latexmls), [latexml-runner](https://github.com/dginev/latexml-runner) and [cortex](https://github.com/dginev/cortex).

For CorTeX I needed to revise the `Status:conversion:LVL` string printing to happen unconditionally - it is a reliable way to parse the exact status code of the job, without having to worry about core vs post conversions, and finer details. It was also the main source of truth used in cortex, and since the new PR failed to print it by default (verbosity was set to >1 ?), cortex was seeing all jobs as Fatals.

But things look backwards compatible after my last commit here, so I seem to be done with the baseline testing.

brucemiller:
Hmm, I'm thinking that to head off some conflicts, I should merge what you have here before embarking on the stuff we talked about today. Maybe you can make whatever patches & renaming you want, then I'll merge?

dginev:
Roger that @brucemiller . I just added the changes I felt I had some solid footing on after conversing today, so if the state of the PR looks acceptable, feel free to merge and continue in the main one.

Feel free to let me know if you think I should add more adjustments - I retested with ltxmojo, cortex and latexmls, and it appearead OK on the quick tests.

brucemiller:
Merged, but...

"disabled by default" violates imnsho, "do the obvious thing (by default)".  Can we revisit exactly what characterizes the situations where you do *not* want output to stderr and *disable* it in those situations only? (rather than having a dozen places where you enable it)

dginev:
> "do the obvious thing (by default)"

There is no "obvious thing" for LaTeXML.pm, there are "obvious things" for the individual executables that rely on it. Since LaTeXML.pm aims to expose an API layer for the various capabilities of the software suite, as a "programmable interface", the only obvious piece is that any _required_ data will flow in/out from the various method calls.

How I/O is used is in no way obvious when we magnify on LaTeXML.pm. It only becomes obvious in latexml, latexmlpost, latexmls, ltxmojo, ...

brucemiller:
Yeah, I know what an API is. Sounds like we need a better one.

dginev:
What we can certainly do is list out all places we know which will need updates in either case.

1. For "request to enable", changes are needed in latexml, latexmlpost, latexmlmath, latexmlc (in-place), possibly makesite and other DLMF executables.

2. For "request to disable", changes are needed in latexmls, ltxmojo, [latexmlstomp](https://github.com/KWARC/LaTeXML-Plugin-latexmlstomp), [cortex worker](https://github.com/dginev/LaTeXML-Plugin-Cortex/blob/master/bin/latexml_worker#L298), from the ones I've written.

The big hassle with the "request to disable" scenario is that I can only execute it after I tag a specific new version of latexml (0.8.6) and issue new versioned releases in each of the separate repositories hosting these binaries with updated Makefile.PL and the updated latexml version. Which is more admin compared to updating the in-house executables in the same PR that changes Error.pm, while keeping the LaTeXML.pm behavior with the previous default.

brucemiller:
Well, gee; sounds like a good incentive to try a little harder to characterize what distinguishes the case (1) and (2) calls. Use of ports? log files vs scalar refs, ... something?

dginev:
They use different communication channels. 

The (1) cases assume working in a classic command-line setting where I/O happens directly on the terminal, and there is a notion of a "user session". Redirects are possible, but the default is "interactive".

The (2) cases assume working in an API setting, where I/O is not done via the standard handles. Instead the information flow is self-contained in the method calls, and is finalized by serializing to some envelope format and dispatched out of the application - e.g. a JSON-based HTTP web service or a ZIP file sent via the ZeroMQ protocol to the CorTeX server. But it's open-ended really.

brucemiller:
So case (1) don't actually even use the LaTeXML.pm API for the simple 
reason that it isn't appropriate for their purposes. So, at the risk of 
making the current API even less appropriate, I'm going to suggest that 
we switch to a `DisableSTDERR()`, and call that as needed within 
LaTeXML.pm so that those case (2) programs don't have to change.

As we later develop the API that is appropriate for case (1), we'll do 
or not do whatever is required for those applications.

dginev:
Sure, I'll just keep disagreeing :> 

latexmlc has already unified the latexml+latexmlpost invocation _via_ LaTeXML.pm, and had experimental mechanisms for enabling/disabling the STDERR capability. So LaTeXML.pm has been used as the API interface for executables in both cases (1) and (2) above, and is a good candidate for a versatile interface layer.

Stating sentiments such as 
> as we later develop the API that is appropriate for case (1)

implies LaTeXML.pm is broken or insufficient for case (1), which I fail to see, especially since I've been using latexmlc on the command line for years without serious issues.

dginev:
On a technical note, the disable/enable question should be external to LaTeXML.pm.

The `->convert` interface can be used in either scenario, so it shouldn't be taking any position on I/O -- the executables are really the only pieces that know which I/O is appropriate. In the case of latexmlc the answer to whether the session is interactive is only clear when the options are processed, since it can be used as a remote client (via `--address` and `--port`, in which case it acts as a simple HTTP client), or it can be used as a local executable, in which case an interactive session takes place.

brucemiller:
On 5/19/21 3:29 PM, Deyan Ginev wrote:
> On a technical note, the stderr disable/enable question should be 
> external to LaTeXML.pm
One of the motivations for this exercise was exactly because
LaTeXML.pm's experimental attempt to manage STDERR failed,
by hiding bibtex errors and causing a lot of confusion
and wasted time. So, no, LaTeXML.pm, has not worked form me without
issues and, as it stands, is not a good candidate.
Perhaps it can be saved?

Either we can recognize when STDERR should be disabled
based on config/options that indicate this is being run as a "service" 
of some kind, or we rely on disabling it from the calling executables.

> The |->convert| interface can be used in either scenario, so it 
> shouldn't be taking any position on I/O -- the executables are really 
> the only pieces that know which I/O is appropriate. In the case of 
> latexmlc the answer to whether the session is interactive is only clear 
> when the options are processed, since it can be used as a remote client 
> (via |--address| and |--port|, in which case it acts as a simple HTTP 
> client), or it can be used as a local executable, in which case an 
> interactive session takes place.
So, then if --address and/or --port are used, then you should likely 
disable STDERR.   That seems straightforward, enough.  And reasonably 
implemented in LaTeXML.pm (or Config.pm), contrary to your opening.

dginev:
> One of the motivations for this exercise was exactly because
LaTeXML.pm's experimental attempt to manage STDERR failed,
by hiding bibtex errors and causing a lot of confusion
and wasted time.

If you had presented that concern upfront, we may have been able to constructively discuss what the underlying cause of the problems there were -- recursive calls are a _separate_ consideration to interactive sessions, as they will be as interactive as the outer call was.

I explicitly made a choice to suppress the messages when I coded the experimental feature of MakeBibliography sending a `->convert` call, so that choice needs to be improved upon when it causes problems. Rather than concluding LaTeXML.pm is completely broken and must be treated like something to be discarded, the correct conclusion is that my decision in suppressing the messages needs to be revised and the captured inner log needs to be reintegrated with the outer one.

> Perhaps it can be saved?

It certainly should be possible, given it doesn't need "saving".

> So, then if --address and/or --port are used, then you should likely
disable STDERR. That seems straightforward, enough.  And reasonably
implemented in LaTeXML.pm (or Config.pm), contrary to your opening.

No, the use case is the _inverse_ of this. When you have these options, LaTeXML.pm is not used at all, an HTTP request is sent to a remote latexml somewhere else. When you _do not_ have the options, LaTeXML.pm is used, and the session becomes interactive. So you can then enable the STDERR logging in the latexmlc executable, when that logic branch is active.

brucemiller:
> No, the use case is the /inverse/ of this.

Well, you'll figure it out, once you're willing to try.

dginev:
> Well, you'll figure it out, once you're willing to try.

I will indeed, in the respective executable. Hopefully with a latexml 0.8.6 version bump.

brucemiller:
Actually, if (if!) there were a reason to want to "enable" STDERR, a better strategy, um, I mean, Rationale, would be to hide it within a `InitializeSTDERR`. That would be responsible for, if needed and indeed even possible, to figure out the terminal width, position and I dunno what else. The purpose being to allow cooperation (meaning what?) between use of Error between caller & callee (eg. makesite & latexml, dlmfpost); or even more critical: threaded calls to the latter.  Twas a good first step to have reliable, complete log files, but we're still going to get all sorts of clashes with multiple spinners running, if they don't know where they are.

dginev:
>  `InitializeSTDERR`

Sure, if that sounds like something you want to pursue with further algorithms added for ever more advanced logging, you can indeed flesh out a full logging framework interface, with a mandatory initialization method, and a config option with various possible customization parameters. Rust has a handful of those, it's usually an `.init` or `.init_config` method called at the top of the main executable that sets the stage for things like streams and verbosity. [some examples here](https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html). I don't particularly like getting all the way to creating builder and adaptor objects, but having a set of configuration parameters and a flavor of an init method is definitely quite pleasant.

PR #1533, opened 2021-05-17
Add binding for 'ifplatform.sty'

### LINKED ISSUE #1531: Support for ifplatform.sty (and use \ifwindows tests)
Hello!

With LaTeXML version 0.8.5 and with the following input file: 

```latex
\documentclass{book}
\usepackage{ifplatform}

\begin{document}

\ifwindows
Windows
\else 
Other 
\fi

\end{document}
```

I get the following output: 

```console
/opt/local/bin/latexml --verbose --verbose --verbose --verbose --debug=Model  minimal.tex 2>&1
latexml (LaTeXML version 0.8.5)
processing started Mon May 17 12:55:17 2021

(Digesting TeX minimal...
(Loading /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/TeX.pool.ltxml...
(Loading /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/eTeX.pool.ltxml... 0.00 sec)
(Loading /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/pdfTeX.pool.ltxml... 0.00 sec) 0.10 sec)
(Processing content /Users/aivo/tmp/debug/minimal.tex...
(Loading /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/LaTeX.pool.ltxml...
(Loading /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/textcomp.sty.ltxml... 0.01 sec) 0.14 sec)
(Loading /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/book.cls.ltxml... 0.01 sec)
Warning:missing_file:ifplatform Can't find binding for package ifplatform
	at minimal.tex; line 3 col 0 - line 3 col 1
	search paths are /Users/aivo/tmp/debug
	Next token is T_CS[\par]
	In Core::Definition::Constructor[\usepac... /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/LaTeX.pool.ltxml; line 798
	 <= Core::Stomach[@0x7fbdcdf8c518] <= ...

Info:dependencies:dependencies Loading dependencies for /opt/local/share/texmf-texlive/tex/latex/ifplatform/ifplatform.sty: shellesc,pdftexcmds,catchfile,ifluatex
	at minimal.tex; line 3 col 0 - line 3 col 1

(Loading /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/ifluatex.sty.ltxml... 0.00 sec)
Error:undefined:\ifwindows The token T_CS[\ifwindows] is not defined.
	at minimal.tex; line 6 col 0 - line 6 col 10
	Defining it now as with \newif
	Next token is T_LETTER[W]
	Stack Trace:
		LaTeXML::Core::State=HASH(0x7fbdcdda7...->generateErrorStub(Core::Gullet[@0x7fbdcdf79290],T_CS[\ifwindows]) @ /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Core/Gullet.pm line 290
	Core::Gullet[@0x7fbdcdf79290]->readXToken(1,1) @ /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Core/Stomach.pm line 95
	Core::Stomach[@0x7fbdcdf8c518]->digestNextBody() @ /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Core.pm line 170
	Core[@0x7fbdca89c530]->finishDigestion() @ /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Core.pm line 159
	LaTeXML::Core::__ANON__(LaTeXML::Core::State=HASH(0x7fbdcdda7...) @ /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Core.pm line 249
	Core[@0x7fbdca89c530]->withState(CODE(0x7fbdca89c908)) @ /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Core.pm line 161
	Core[@0x7fbdca89c530]->digestFile('minimal.tex','mode','TeX','preamble',undef,'postamble',undef) @ /opt/local/bin/latexml line 103
	eval {...} @ /opt/local/bin/latexml line 98

 1.69 sec) 1.79 sec)
(Building...
(Loading compiled schema /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/resources/RelaxNG/LaTeXML.model... 0.01 sec). 0.04 sec)
(Rewriting... 0.00 sec)
(Finalizing... 0.00 sec)
Conversion complete: 1 warning; 1 error; 1 undefined macro[\ifwindows]; 1 missing file[ifplatform.sty].
processing finished Mon May 17 12:55:19 2021
<?latexml searchpaths="/Users/aivo/tmp/debug"?>
<?latexml class="book"?>
<?latexml package="ifplatform"?>
  <resource src="ltx-book.css" type="text/css"/>
  <para xml:id="p1">
    <p>Other</p>
  </para>
</document>
```

How can I debug this situation further? I have tried with various --verbose and --debug combinations, but I cannot figure out what is happening. It seems that even though the ifplatform.sty is loaded and parsed, the commands defined there do not have effect.

tkw1536:
This PR adds a binding for the 'ifplatform.sty' package, emulating the package as closely as possible.

This commit does not add any tests for the package, because the output is operating system dependent. As such it would likely fail for a large part of users of LaTeXML. 

Fixes #1531.

### DIFF
--- a/lib/LaTeXML/Package/ifplatform.sty.ltxml
+++ b/lib/LaTeXML/Package/ifplatform.sty.ltxml
@@ -1,0 +1,68 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# |  ifplatform                                                         | #
+# | Implementation for LaTeXML                                          | #
+# |=====================================================================| #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
+# |  United States Government & not subject to copyright in the US.     | #
+# |=====================================================================| #
+# | Thanks to Tom Wiesing <tom.wiesing@gmail.com>.                      | #
+# | Additional thanks to GitHub user "aivokalu" for part of the         | #
+# | original implementation.                                            | #
+# |---------------------------------------------------------------------| #
+# | Bruce Miller <bruce.miller@nist.gov>                        #_#     | #
+# | http://dlmf.nist.gov/LaTeXML/                              (o o)    | #
+# \=========================================================ooo==U==ooo=/ #
+package LaTeXML::Package::Pool;
+use strict;
+use warnings;
+use LaTeXML::Package;
+
+# Define the names for the common operating systems
+DefMacro('\windowsname',    'Windows');
+DefMacro('\notwindowsname', '*NIX');
+DefMacro('\linuxname',      'Linux');
+DefMacro('\macosxname',     'Mac\,OS\,X');
+DefMacro('\cygwinname',     'Cygwin');
+
+# internal 'if' being used to detect if shell escape is enabled.
+# The concept doesn't exist in LaTeXML the way it exists in pdflatex,
+# so we hardcode it as true.
+DefConditional('\ifshellescape', sub { 1; });
+
+# Do a check for the various operating systems
+# and store the result in appropriate variables
+my $isWindows = $^O eq "Win32";
+my $isLinux   = $^O eq "linux";
+my $isMacOSX  = $^O eq "darwin";
+my $isCygwin  = $^O eq "cygwin";
+
+# Check which platform we are on so that we can define the platformname macro.
+# Also define the '\unknownplatform' macro from the output of uname.
+my ($platformNameMacro, $unknownPlatformName) = ('', '[Unknown]');
+if ($isWindows) {
+  $platformNameMacro = '\windowsname'; }
+elsif ($isLinux) {
+  $platformNameMacro = '\linuxname'; }
+elsif ($isMacOSX) {
+  $platformNameMacro = '\macosxname'; }
+elsif ($isCygwin) {
+  $platformNameMacro = '\cygwinname'; }
+else {
+  $platformNameMacro = '\unknownplatform';
+  eval { require POSIX; } or last;
+  ($unknownPlatformName) = POSIX::uname(); }
+
+DefMacro('\unknownplatform', $unknownPlatformName);
+Let('\platformname', $platformNameMacro);
+
+# \newif s for the various operation systems.
+# In the original package these rely on 'shell escape' being enabled, and otherwise return false.
+# This is not emulated here, it is considered an implementation detail.
+DefConditional('\ifwindows', sub { $isWindows; });
+DefConditional('\iflinux',   sub { $isLinux; });
+DefConditional('\ifmacosx',  sub { $isMacOSX; });
+DefConditional('\ifcygwin',  sub { $isCygwin; });
+
+1;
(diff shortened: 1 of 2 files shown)


tkw1536:
I have tested this binding only on OS X, as that is the only system I have access to. I would appreciate if someone could test on Windows and Linux respectively. 

A reasonable test file is the following:

```tex
\documentclass{article}
\usepackage{ifplatform}

\begin{document}

Windows: \ifwindows{true}\else{false}\fi


Linux: \iflinux{true}\else{false}\fi


MacOSX: \ifmacosx{true}\else{false}\fi


Cygwin: \ifcygwin{true}\else{false}\fi


Platformname: \platformname


Unknown Platform: \unknownplatform

\end{document}
```

tkw1536:
I have also managed to test this manually on a linux server. However I don't have access to a windows / cygwin instance.

lib/LaTeXML/Package/ifplatform.sty.ltxml
```diff
@@ -0,0 +1,68 @@
...
+DefMacro('\windowsname',    'Windows');
+DefMacro('\notwindowsname', '*NIX');
+DefMacro('\linuxname',      'Linux');
+DefMacro('\macosxname',     'Mac\,OS\,X');
+DefMacro('\cygwinname',     'Cygwin');
+
+# internal 'if' being used to detect if shell escape is enabled.
+# The concept doesn't exist in LaTeXML the way it exists in pdflatex,
+# so we hardcode it as true.
+DefConditional('\ifshellescape', sub { 1; });
```
brucemiller, line 32: Not so much that  the concept doesn't exist, but that we haven't implemented it, and security is a big disincentive to do so.  Wouldn't it be better to return 0, so that code won't then *try* to execute a shell escape and have it ignored (fail)?

tkw1536: The reason i have it set to true is that the package internally uses `--shell-escape` to determine if Mac, Linux or other OSs are being used. From the manual:
> If you also wish to determine the difference between which Unix-variant you are using (i.e., also detect \iflinux, \ifmacosx, and \ifcygwin) then shell escape must be enabled. This is achieved by using the -shell-escape command line option when executing LATEX
When shell escape is not enabled, these always return false. In the binding we can of course detect those operating systems without needing to enable (or even implement) such an option. 

I have a strong suspicion (although no data to back this up) that `\ifshellescape` is also misused outside the package to check if non-windows OS detection is enabled. And since it is enabled in the binding, in my view it makes sense to set it to 1.

brucemiller: OK, let's go with that, but keep in the back of our mind that it's a bit weird :>

PR #1540, opened 2021-06-01
latexmlc refactor for new logging framework

dginev:
First stab at this refactor, following a user-centric mindset. Decisions:
* The `.status` tests were an experiment, and are at this point a distraction for refactoring, so I removed them.
    * If we want to have stable tests for conversion status, we can parse the now mandatory `.log` files and/or use the corresponding `$$response{status}`, `$$response{status_code}`  reports from LaTeXML.pm.
* I retouched LaTeXML.pm to use `Note()` calls for some messages that would also make sense for the command-line, which worked nicely.
* That allowed removing all mentions of STDERR from `latexmlc` itself, now fully delegating to LaTeXML.pm
*  In "client" mode the conversion is now extra silent, but that use of latexmlc is arguably quite rare. Especially now that `latexml-runner` exists, it is easy to imagine users who care about the performance of a daemonized latexmls would also care for the non-trivial overhead induced by the latexmlc client and use the rust toolkit instead.

So this PR may be good for a discussion/review at this stage.


MANIFEST
```diff
@@ -881,48 +881,34 @@ t/complex/xii.xml
 t/daemon/amsarticle.tex
 t/daemon/broken.tex
 t/daemon/api/port.spec
-t/daemon/api/port.status
```
brucemiller, line 884: Is there perhaps any git-related reason that we may want to *not* delete these now, but simply ignore them. And just update (possibly) the contents with a later change?

bin/latexmlc
```diff
@@ -133,56 +132,43 @@ if ((!$sock) && $local && ($expire == -1)) {
...
-      print $log_handle $log;
-      close $log_handle;
-    }
-  } else { print STDERR "\n", $log, "\n"; }    #STDERR log otherwise
-}
+if (!$is_archive && $log) {
+  my $log_path = $opts->get('log') || 'latexml.log';
+  open(my $log_handle, ">", $log_path) or die("Fatal:I/O:forbidden Couldn't open log file $log_path : $!\n");
+  print $log_handle $log;
+  close $log_handle; }
```
brucemiller, line 152: So, not using `UseLog()`? And always "latexml.log", rather than "$jobname.latexml.log"?  Is this part of what you're wanting to defer till later, or what you're trying to push as the "right" way?

MANIFEST
dginev: I think they pollute the filesystem+manifest quite a bit, and could be replaced by a generic "expect status_code 0 unless explicitly requested otherwise". Then one could change the daemon tests from e.g.
```perl
latexml_tests('t/daemon/fatal');
```

to
```perl
latexml_tests('t/daemon/fatal',
  expect => {
    status_code => 3
  });
```

So fingers crossed the files don't return. The string status is a lot more volatile (e.g. some messages have ids that will need to be maintained, or timestamps or...), and what we really want to test is the _code_ rather than the specific message.

bin/latexmlc
dginev: 1. I was just learning of UseLog(), so it slipped my mind to try to use it when flushing the full `$$response{log}` from a conversion. It would look something like:
    ```perl
    UseLog($log_path);
    print $LaTeXML::Common::LOG $log;
    UseLog(undef);
    ```
which is mildly odd... Maybe we need a DumpToLog method in Error.pm that does this?

2. The `latexml.log` considerations are a bit unclear to me. Until now latexmlc didn't use a default `latexml.log` at all, so adding it in as-is is in a way a simple upgrade. But the big challenge is that in my uses of latexmlc I don't have a jobname way too often. I test around with various `literal:` formulas and fragments, for which cases I think the default `latexml.log` would be OK, or comparably a `literal.latexml.log` ?

Getting this discussion is exactly what I was hoping for actually :>

MANIFEST
brucemiller: ok

bin/latexmlc
brucemiller: (1) Yeah, odd. But what's odd (to me) is that you're returning the log as a string at all (in general).  If the log should go in a file, get the name, and somebody call's `UseLog($logpath)`, and you're done. In a server context where you don't want the file, use a string ref for `$logpath`, and then send the string wherever it goes.

(2) Sure, `latexml.log` is quite reasonable for the cases where you *don't* have a reasonable jobname; but when you do, not so much.

dginev: (1) Well, an API call should return the payload, so a `$$response{log}` is completely natural from the LaTeXML.pm perspective. latexmlc may decide to write the value down to a file, ltxmojo/latexmls may decide to ship it back over the wire.

But it doesn't need to *always* get returned, and in some cases it comes back empty already -- whenever there is an archive being written with `--log` set, LaTeXML.pm would create that file and return the raw bytes for the archive in `{result}`, without setting `{log}` at all. I keep thinking of this as some sort of "interactive" flag, though still a bit fuzzy what's the least confusing way to make it configurable.

(2) Ok, I'll try to rework that part.

dginev: Thinking on your suggestion of:
>  In a server context where you don't want the file, use a string ref for $logpath, and then send the string wherever it goes.

That does make sense after playing with the thought for a while. Ideally we at least deal with "filehandles and string refs", which look the same to Perl, rather than "pathnames and string refs" which are very different kinds of objects. Maybe we need a `UseLogHandle()` that I can call with the string ref from LaTeXML.pm, and which `UseLog()` can call after it opens the file and gets a handle on it?

Then... latexmlc can figure out a `--log` option for Config before the conversion starts, and `->convert()` can be interactive, as are latexml and latexmlpost. 

Meanwhile, latexmls/ltxmojo won't set a `--log` for most jobs (archives being the one exception), which LaTeXML.pm will interpret -- as it already does -- as a fallback to a local string ref, to be returned serialized in the response.

Yeah that can work, let me try...

brucemiller: Sure, but the "payload" can be whatever you (or I) say it is: a pathname, eg, not necessarily the whole log in a string.

Currently, `UseLog` can already distinguish between a string (pathname) and string ref; there can be more cases if we can distinguish them and they're useful. 

For latexml/latexmlpost, if no `--log` was specified, I'd think the log should default to jobname based.  But I can see the ltxmojo cases want to be able to say no-log; either with an explicit `--log=none` or otherwise detect these cases.
(the word "interactive" seems to go too far)

dginev:
Ok, a bunch of new changes after stabilizing and testing against the web services again. Some you'll like, some we'll have to discuss :>

* Fallback to `latexml.log` when no jobname is available for latexml and latexmlc
* Added spinner messages when latexmlc does remote conversions via latexmls
* Delegated more of the logging of latexmlc to LaTeXML.pm, and the `Note` and `NoteLog` methods
  * while still handling the case where a full log returns from a remote latexmls call
* Stabilized the calls to `->bind_log` and `->flush_log` in LaTeXML.pm to follow the new Error.pm capabilities, essentially making them smaller and calling them only when needed.
* Added an escape hatch for the archive conversion `whatsout`, which treats the `--log` parameter differently than a usual conversion, due to packing.
* Made sure logs are fresh when latexmlc is called on the same jobname by removing any existing log files.
* And most controversially, added a `FinalizeLog()` method, to avoid calling a method with a meaningful `(undef)` argument, which I find to be very scary and dangerous :>

dginev:
I should mention that, crucially, the LaTeXML.pm logic is still successfully "holding it together" and has the following two possible logging behaviors:
  * If a `--log` parameter is set, LaTeXML.pm will use it directly and return an empty `$$response{log}`.
     * used by latexmlc, and maybe one future day by latexml, latexmlpost and latexmlmath
  * If a `--log` parameter is **not** set, LaTeXML.pm will use the `$$latexml{log}` string storage, accumulate it and return the entire log as a single string in `$$response{log}`.
    * used by ltxmojo, latexmls, ltxpsgi and all their downstream consumers (latexml_runner, remote calls of latexmlc, latexml_ruby, etc).

brucemiller:
Mostly positive...
In you 2nd message, is the assumption (for now) that the calling executable is responsible for setting `--log` to some default logfile, if it desires?

I never was comfortable with `bind_log` and `flush_log`, and they essentially toverlap with what`UseLog` is trying to do. It seems that they are employed at too low a level, which requires lots of extra logic about counting, closing, appending, etc.  Hopefully, we can eliminate the redundancy, but perhaps that's for the next iteration?

And what's so scary about "Use log none"? Zheesh, kids these days!

dginev:
> In you 2nd message, is the assumption (for now) that the calling executable is responsible for setting `--log` to some default logfile, if it desires?

Exactly.

LaTeXML.pm should add directives for each of STDERR and Log, but shouldn't make any decisions whether they're on/off, as it is shared between applications that have each of the possible choices.

The executables know their requirements very concretely, and they also know if they should require a default log file or not.

> I never was comfortable with `bind_log` and `flush_log`, and they essentially toverlap with what`UseLog` is trying to do. 

Another way of saying that is that "they inspired the UseLog upgrades" :>

> It seems that they are employed at too low a level, which requires lots of extra logic about counting, closing, appending, etc. Hopefully, we can eliminate the redundancy, but perhaps that's for the next iteration?

Next iteration sounds good! The main extra piece they currently add on top of Error.pm is the string storage, held by the main converter object. That storage could *also* be moved to Error.pm if one preferred. But Error.pm is very anti object orientation, with the package-level variables, so it feels like a hypothetical move in a Perl 4 direction, rather than a Perl 7 direction. (See e.g. [the latest perl newsletter](https://perlweekly.com/archive/514.html) discussing yet another OO package as "the highlight of the future Perl7").

> 
> And what's so scary about "Use log none"? Zheesh, kids these days!

Well, the proverbial bug example is that you call `UseLog($path)` but make a typo somewhere (or a programming logic edge case, either one) that fails to assign `$path` with a value. So magically your run of latexml will become *very* silent, and you'll have no idea why. Fun! :> Or to simple perl programmers like me - just scary.

brucemiller:
So you program something wrong and something bad happens? :> Unconvinced. And I dislike the asymmetry.

dginev:
> So you program something wrong and something bad happens? :> Unconvinced.

Might as well disable the Perl Critic lints if that's the overarching philosophy we're headed to... We already avoid [returning undef](https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::ProhibitExplicitReturnUndef), might as well avoid relying on it on input. 

To address the assymetry, we could switch to initialize/finalize, or use/release, or something similar.

brucemiller:
If hyperbole is what we're after, let's also forbid `$x = undef;`

dginev:
`UseLog(undef)` is back, good for re-review 👍🏻

PR #1543, opened 2021-06-04
[arXiv patch] more internal \input macros, typo and lighter font Info logs

dginev:
### Book discussion

An impressive book-sized manuscript made the rounds recently, at [arXiv:2105.10386](https://arxiv.org/abs/2105.10386), totaling 419 PDF pages.

With this PR, the conversion with latexml parses over 16,000 math expressions, makes an index of 500 entries and finishes in a total of 12 minutes, without errors. It has 415 warnings, mostly due to math parsing.

The run seems to do quite well with the custom `.sty` file of the book, OmniBus recovers well the custom `.cls` and loads the dependencies. Impressively latexml does quite well with raw interpretation of three unsual styles: `nomencl.sty`, `combelow.sty` and `letltxmacro.sty`, all of which are loaded raw.

The book can be converted into a 8.5MB-sized ePub file, which I'll attach here for a preview until we merged. Also, here's a screenshot from calibre, which really lags under the size, but manages to render in 3 columns on my wide screen:
image

[aobf.epub.gz](https://github.com/brucemiller/LaTeXML/files/6601495/aobf.epub.gz)
(of course I had to gzip it for github to accept the attachment...)

### Patches

The book was written in a way where it was almost fully supported by latexml, but for a handful of custom bits.
 * A binding typo had to be fixed (which I also fixed in #1525 , but that's ok, either PR can be rebased when the time comes).
 * Some internal input-related macros had to be added
 * And I suppressed one of the font-related info messages which was printing once on each line of the TeX source file (i.e. thousands of prints with the same message).

For some strange reason the babel greek test broke, adding an extra space char, so I have to troubleshoot that before we can merge...

### DIFF
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -35,7 +35,7 @@ use lib catdir($RealBin_safe, "..", "lib");
 use LaTeXML;
 use LaTeXML::Common::Config;
 use LaTeXML::Common::Error;
-use LaTeXML::Util::Pathname qw(pathname_is_literaldata);
+use LaTeXML::Util::Pathname;
 use URI::Escape;
 use HTTP::Response;
 use HTTP::Request;
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -993,7 +993,7 @@ DefPrimitive('\markboth{}{}',    undef);
 DefPrimitiveI('\leftmark',  undef, undef);
 DefPrimitiveI('\rightmark', undef, undef);
 DefPrimitive('\pagenumbering{}', undef);
-DefMacro('\@mkboth', '\@gobbletwo');    # default, just in case
+#DefMacro('\@mkboth', '\@gobbletwo');    # default, just in case
 DefMacro('\ps@empty',
   '\let\@mkboth\@gobbletwo\let\@oddhead\@empty\let\@oddfoot\@empty' .
     '\let\@evenhead\@empty\let\@evenfoot\@empty');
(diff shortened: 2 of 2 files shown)


dginev:
Apparently babel was upset I refactored to call `Input()` on the result of `FindFile($shortname)` rather than the base `$shortname` that was given on input `\InputIfFileExists`. 

I backed away from such bravery, stayed as close as possible to the master code, and fingers crossed everything is solid now.

brucemiller:
Looks plausible if you fix the conflict :>

dginev:
Reconverted the arXiv book in question with the new logging framework, was an interesting test (spotted a dependency oversight in my latest latexmlc updates - fixed here).

One new observation for me is that the "Building" spinner phase can be a little quiet in a large book. The spinner spins with "Building" for a good minute with no further elaboration.

The "math parsing" stage is nice and loud *if* you get parsing failures :> If all were to parse well, the "Parsing 15924 formulae" message would have been a bit too static, and that may be another easy upgrade where we can show an incrementing counter upto the total number, as the spinner is updating. `parsing 1 of 15924` , `parsing 2 of 15924` ... Already a cool experience to see it spinning around!

---

A curious glitch is that the spinner seems to be getting stuck in post-processing. I saw:
```
 / post-processing Scan OPS/aobf.xhtml processing
```
followed by:
```
 / post-processing MakeIndex OPS/aobf.xhtml processing
```

where the `/` marker didn't spin - it was just a red `/` for the entire duration of post-processing. I'll check if there's a quick way to kick it into spinning again.

Happily, 12 minutes later, I got a nice book ePub out again with a warning status conversion, so the main deliverable is operational.

brucemiller:
Meaning ready to merge or more tweaks?

I'd noticed that *sometimes* we *do* have a number of things to be processed, so that the spinner could include a % (or count or...).  I figured it'd be better to deal with that later, though (if at all)

dginev:
This one can be merged freely, i can tweak the logging anywhere :> New branch is easy so feel free...

PR #1545, opened 2021-06-09
Enable \if to expand \let bindings

dginev:
I think this was a regression during the big upgrade run for supporting expl3.

Our `\if` token reader had omitted the check whether the read token was `\let` to another one, which is expected to be covered (e.g. described in the TeXbook exercise 20.11).

I also introduced the explicit test from the book, so that we don't regress again. The current master branch would fail the new test, this PR passes it.

I am unsure if I am _fully_ accurate with the way I placed the `\let` check, but tests still pass and I recovered a Fatal arXiv document, [cs/0003055](https://arxiv.org/abs/cs/0003055), to a "no problem" status, while constructing a very elaborate custom plot drawing. The resulting drawing for that paper is quite ugly -- but the content is clearly correct, just hard to position without some help. I didn't know troubleshooting the paper would lead to this PR, but I am happy this was the *only* actual issue with the paper, even when it died with 100 errors -- they were just compounding on the single mistake.

Sadly I also introduced a lot of spacing diffs in TeX.pool, so reporting that my `perltidy -v` is `v20210402` at the moment.


dginev:
It is much easier to view the PR without the spacing diff, at:
https://github.com/brucemiller/LaTeXML/pull/1545/files?w=1

brucemiller:
Sneaky!

PR #1547, opened 2021-06-12
Patch expl3.sty regression on texlive 2020

### LINKED ISSUE #1454: [texlive 2020] expl3 regressed on @expl@push@filename@aux@@
It's really tricky to pin down where exactly the error comes from, but the ultimate source seems to be that expl3 modifies `\RequirePackage`, `\usepackage` and similar commands to use `\@pushfilename`, which causes errors of the sort
`Error:expected:{} Missing argument {} for Core::Definition::Expandable[]\@expl@push@filename@aux@@ {}{}{}]`. 

I assume this line here attempted to fix this error, but has probably been broken in a more recent version of expl3:
https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/Package/expl3.sty.ltxml#L24

The error can be easily reproduced by calling `\usepackage` after expl3 has been loaded, e.g.

```
\documentclass{article}
\usepackage{expl3}
\usepackage{float}

\begin{document}foo\end{document}
```

dginev:
Fixes #1454 .

Now that I updated my texlive and can reproduce locally, I traced this issue to one of my older patches in `Package::InputDefinitions` which hardcoded a little too closely the LaTeX 2 behavior of a standalone `Digest(T_CS('\@pushfilename'))` while loading e.g. a new .sty package with definitions.

In LaTeX 3 that internal macro gets extended to do more token interplay with the rest of the input stream, so my standalone `Digest` call creates the errors reported in the issue - and I could reproduce those clearly.

However, the expl3 metadata fishing is a bit of a hack in all honesty: It relies on the exact placement of `\@pushfilename` in the original `latex.ltx`. You move that macro e.g. a line or two down, or worse - change the definition pattern on the line after - and it will completely break its behavior. So my instinct again was that we shouldn't be emulating that directly. Which lead me to the comment we have had in `expl3.sty.ltxml`:

```
# Redefine aux macro to avoid reliance on argument ordering artefacts in latex.ltx
# The purpose here was originally to set `\g_file_curr_name_str`,
#  but I am unsure if it is used anywhere currently?
```

Which deactivated the `\@pushfilenameaux` in previous texlives.  I think the original point still holds, so I also deactivated the equally hacky `\@expl@push@filename@aux@@`, making the surface error disappear.

Happily, our expl3 tests had not regressed, and the core interpretation is still solid on texlive 2020 (for the degree we've tested it).

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -103,7 +103,7 @@ our @EXPORT = (qw(&DefAutoload &DefExpandable
     &LookupLCcode &AssignLCcode
     &LookupUCcode &AssignUCcode
     &LookupDelcode &AssignDelcode
-    ),
+  ),
 
   # Random low-level token or string operations.
   qw(&CleanID &CleanLabel &CleanIndexKey  &CleanClassName &CleanBibKey &NormalizeBibKey &CleanURL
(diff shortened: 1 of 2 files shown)


brucemiller:
Ah, ok.

PR #1550, opened 2021-06-15
Touchup latexml/latexmlc exit

dginev:
Making sure we get slightly cleaner stderr cleanup on exit and status_code check for latexmlc, especially for Fatals (split from #1525 )

### DIFF
--- a/bin/latexml
+++ b/bin/latexml
@@ -165,6 +165,7 @@ $serialized = undef;
 
 CheckDebuggable();
 
+UseSTDERR(undef);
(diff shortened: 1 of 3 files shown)


bin/latexml
```diff
@@ -165,6 +165,7 @@ $serialized = undef;
 
 CheckDebuggable();
 
+UseSTDERR(undef);
```
brucemiller, line 168: presumably latexmlpost deserves one of these too

dginev: done!

PR #1552, opened 2021-06-15
Better recovery on Fatal

dginev:
The essence of the overloaded previous PR #1525 .

- includes the more precise `alarm()` TTL
- starts the chunking into more `eval {}` blocks in LaTeXML.pm, which facilitates more chances for recovering partial documents
- keeps the Fatal logic as in master, but introduces a `hardYankProcessing` sub that will drop any remaining processing data from the Core pipeline, as well as store the work done so far into a recoverable slot in State.
- adds some missing calls to colorizeString when generating messages
- adds some relevant tests with recovery on Fatal.
  - I also broke an arXiv article by hand and double-checked the partial outputs still work, seems good for the `popFrame` fatal case.

May still be reexamined and rediscussed, should be easier to grock.


lib/LaTeXML/Common/Error.pm
```diff
@@ -285,6 +285,8 @@ sub Fatal {
 
 sub hardYankProcessing {
   my $state = $STATE;
+  # Nothing we can do if we are called without a global $STATE bound
+  return unless $state;
```
dginev, line 289: benefits of moving this code out into a subroutine - I can now stop early when the state is missing, e.g. if Fatal is called from some obscure place in initialization, post-processing, makesite ...

dginev:
Also take a look at a brainstorming branch I made derived from here, as to how the new `yank` subroutine can be used to make a new kind of `FinalError`, which is softer than a Fatal, but harsher than an Error:

https://github.com/dginev/LaTeXML/pull/11/files

PR #1553, opened 2021-06-16
Softer errors

brucemiller:
Replace some recoverable Fatal errors with Error.


lib/LaTeXML/Core/State.pm
```diff
@@ -507,7 +507,7 @@ sub pushFrame {
 sub popFrame {
   my ($self) = @_;
   if ($$self{undo}[0]{_FRAME_LOCK_}) {
-    Fatal('unexpected', '<endgroup>', $self->getStomach,
+    Error('unexpected', '<endgroup>', $self->getStomach,
```
dginev, line 510: Did you try triggering this one? This is one place where you can hit an infinite loop (or just 100 errors really) since `\end{document}` in LaTeX.pool loops over popFrame.

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -363,8 +363,9 @@ DefConstructorI(T_CS('\end{document}'), undef, sub {
       && (!($ifstack = $STATE->lookupValue('if_stack')) || !$$ifstack[0])) { }    # OK!
     else {
       my @lines = ();
-      while (!$STATE->isValueBound('current_environment', 0)
-        || ($STATE->valueInFrame('current_environment', 0) ne 'document')) {
+      while ((!$STATE->isValueBound('current_environment', 0)
+          || ($STATE->valueInFrame('current_environment', 0) ne 'document'))
+        && ($STATE->getFrameDepth > 0)) {
```
dginev, line 368: Ah, I see you've added the guard, scratch my last comment and sorry for the noise.

dginev:
@brucemiller what is your `perltidy -v` ?

Mine is `v20210402`. Worried about the spacing changes, in particular worried about them bouncing back and forth.

dginev (approved):
I like all changes actually! No need to die on any of them if we can avoid it, even if they end up causing some contained havoc in the document.

brucemiller:
On 6/16/21 1:42 PM, Deyan Ginev wrote:
> @brucemiller <https://github.com/brucemiller> what is your |perltidy -v| ?
> 
> Mine is |v20210402|. 
I'm stuck at v20201001
You don't like accordion music?

dginev:
```
# Difference at line 14 for t/daemon/fatals/fatal_egroup
#       got : '      </Math> here and then suddenly we experience a bad    mystery dropped content error.</p>'
#  expected : '      </Math> here and then suddenly we experience a bad </p>'
# Looks like you failed 2 tests of 7.
```

Amazing, you broke my fatal egroup test by no longer having that Fatal on egroup. Tsk! How am I to test that now :>

brucemiller:
Ouch! I shoulda pulled first, or something...
Well, you can have failing tests or failing programs; but not both :>

dginev:
> Ouch! I shoulda pulled first, or something...
> Well, you can have failing tests or failing programs; but not both :>

I think for now it makes sense to just delete the `fatal_egroup` files (.tex, .xml, .spec and MANIFEST entry) and merge the PR as-is. It's a great upgrade, worth losing a test! :>

PR #1554, opened 2021-06-22
Less arxiv fatals

brucemiller:
This cleans up a handfull of random fatals that were easy to isolate & fix with the generally cleaned up state of things.

Most significant may be a stubby binding for `psfrag`. It doesn't actually add the fragments, but creates footnotes to inform of what would have been added to the figures. So an imperfect, but error free, result.  Could be non-trivial move of error to warning, since psfrag was used a lot in older arXiv papers.

Still WiP, since I may sneak in a few other minor cleanups.

### DIFF
--- a/lib/LaTeXML/Post/CrossRef.pm
+++ b/lib/LaTeXML/Post/CrossRef.pm
@@ -657,12 +657,12 @@ sub generateRef {
         $pending = $$self{ref_join}; } # inside/outside this brace determines if text can START with the join.
       $id = $entry->getValue('parent'); } }
   if (!@stuff) {                       # Try first child for a title-less document?
-    my $entry = $$self{db}->lookup("ID:$reqid");
-    if (($entry->getValue('type') || '') eq 'ltx:document') {
-      foreach my $c (@{ $entry->getValue('children') }) {
-        if (my $centry = $$self{db}->lookup("ID:$c")) {
-          if (my @s = $self->generateRef_aux($doc, $centry, $reqshow)) {
-            push(@stuff, @s); last; } } } } }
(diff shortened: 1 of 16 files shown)


lib/LaTeXML/Post/CrossRef.pm
```diff
@@ -657,12 +657,12 @@ sub generateRef {
...
-      foreach my $c (@{ $entry->getValue('children') }) {
-        if (my $centry = $$self{db}->lookup("ID:$c")) {
-          if (my @s = $self->generateRef_aux($doc, $centry, $reqshow)) {
-            push(@stuff, @s); last; } } } } }
+    if (my $entry = $$self{db}->lookup("ID:$reqid")) {
+      if (($entry->getValue('type') || '') eq 'ltx:document') {
+        foreach my $c (@{ $entry->getValue('children') }) {
+          if (my $centry = $$self{db}->lookup("ID:$c")) {
+            if (my @s = $self->generateRef_aux($doc, $centry, $reqshow)) {
+              push(@stuff, @s); last; } } } } } }
```
dginev, line 665: shouldn't there be a warning if one is missing? Or are such warnings already emitted in Core processing, so it would be redundant in Post?

brucemiller:
On 6/22/21 10:35 PM, Deyan Ginev wrote:
> shouldn't there be a warning if one is missing?

It *should* be caught by `$self->note_missing` at the end, which 
acummulates all such into a single warning.

dginev (approved):
LGTM! We can keep upgrading on a rolling basis. With some luck the conversion will finish by end of today and I can test the PR right after.

PR #1556, opened 2021-06-29
Detect experimental MathML support from maybeMathjax.js

dginev:
It appears there is a nice and simple trick one can currently use to detect an "intention for" MathML support in browsers, namely having made available the interface:

```javascript
var has_mathml_interface = typeof (MathMLElement) == "function";
```

This neatly allows detecting unusual MathML browsers, such as a regular Chrome with experimental features enabled.

I tested with:
```
latexmlc --pmml --cmml --whatsin=math 'literal:Q_{m,n}^{[i+j<m+n]}(z_{i,j})' \
              --dest=test.html --javascript=LaTeXML-maybeMathjax.js
```

and the detection worked well in Firefox and the two Chrome variants. I would like to update the official script and deploy this to CorTeX, so that we feast our eyes on arXiv documents rendered with Chrome's MathML support.

### DIFF
--- a/lib/LaTeXML/resources/javascript/LaTeXML-maybeMathjax.js
+++ b/lib/LaTeXML/resources/javascript/LaTeXML-maybeMathjax.js
@@ -1,7 +1,7 @@
 //======================================================================
 // Load MathJax, IFF the current browser can't handle MathML natively.
 
-(function() {
+(function () {
     var mathjax_url =
         "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js?config=MML_HTMLorMML";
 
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/resources/javascript/LaTeXML-maybeMathjax.js
```diff
@@ -1,7 +1,7 @@
 //======================================================================
 // Load MathJax, IFF the current browser can't handle MathML natively.
 
-(function() {
+(function () {
     var mathjax_url =
         "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js?config=MML_HTMLorMML";
```
xworld21, line 6: A semi-related comment: I think MathJax 3 changed the URL conventions. E.g. I directly download `es5/mml-chtml.js` in my templates, because it is about 160k smaller. The `config=` line sounds like a leftover MathJax 2?

PR #1557, opened 2021-06-30
More arxiv patches

brucemiller:
As it says, a few random patches; some bindings...


lib/LaTeXML/Package/fancybox.sty.ltxml
```diff
@@ -0,0 +1,49 @@
...
+use warnings;
+use LaTeXML::Package;
+
+# These could be made to depend on \fboxsep, \fboxrule, \cornersize
+DefMacro('\cornersize OptionalMatch:* {}', '');
+our $fancybox_border = 'border:1px solid black;';
+our $fancybox_shadow = 'border:1px solid black; box-shadow: 5px 5px 10px black;';
+our $fancybox_double = 'border:3px double black;';
+our $fancybox_oval   = $fancybox_border . 'border-radius:5px;';
+our $fancybox_Oval   = 'border:2px solid black;' . 'border-radius:5px;';
```
dginev, line 24: Isn't it more robust to make these into class names, and attach a custom `fancybox.css` to the PR that people can customize/override?

But yeah, maybe that's long-term, maybe a comment that says "TODO: Refactor to classes..." is the best I can ask for in the short-term.

lib/LaTeXML/Package/mathabx.sty.ltxml
```diff
@@ -0,0 +1,811 @@
...
+# DefMath('\mayacntb',Tokens());
+# DefMath('\mayacntc',Tokens());
+# DefMath('\mayacnter',Tokens());
+# DefMath('\mayadelimiters',Tokens());
+# DefMath('\mayadigit',Tokens());
+# DefMath('\mayaexpansion',Tokens());
+# DefMath('\mayarecurse',Tokens());
+# DefMath('\mayawidth',Tokens());
+
+#======================================================================
```
dginev, line 810: Wow, that's a lot! Impressive amount of new math definitions.

lib/LaTeXML/Package/proof.sty.ltxml
```diff
@@ -0,0 +1,82 @@
...
+DefMath('\lx@proof@logical@and', "\x{2003}", meaning => 'and', role => 'ADDOP');
+DefMacro('\lx@proof@split@and{}', sub {
+    my ($gullet, $tokens) = @_;
+    my @clauses = SplitTokens($tokens, T_ALIGN);
+    if (!@clauses) {
+      return; }
+    elsif (scalar(@clauses) == 1) {
+      return Tokens(@{ $clauses[0] }); }
+    else {
+      return I_apply(undef, T_CS('\lx@proof@logical@and'), map { Tokens(@$_); } @clauses); } });
```
dginev, line 28: `I_apply(undef,` is something I have to go and look up, since I forgot what the undef does here.

... grepping over the code base suggests so far we've only used `I_apply({},` is there a reason why it's instead undef here? 

... oh I see, the first argument is meant to be key-value pairs, and undef would skip over that case as well at the moment. I guess I'll just post my musings as a comment.

dginev (approved):
Upgrades! Nice!!

Cool PR with lots of new bits that look nice on the surface.

brucemiller:
Probably should just wrap up this PR with what it has. 

(I'm working on adjustbox, which possibly can be read raw, but requires a few support patches; thus, a separate PR)

dginev:
That sounds like a great plan to me! Nothing more to add on my end from a cursory review, also since I am not familiar enough with the `.sty` files to spot anything subtle.

lib/LaTeXML/Package/makecell.sty.ltxml
```diff
@@ -36,5 +36,55 @@ AtBeginDocument('\let\lx@orig@theadfont\theadfont'
...
+  afterDigest => sub {
+    my ($stomach, $whatsit) = @_;
+    my ($diagH, $diagV, $space, $A, $B) = $whatsit->getArgs;
+    $diagH = ToString($diagH);
+    $diagV = ToString($diagV);
+    my $flip = ($diagH < 0) ^ ($diagV < 0);
+    my ($Aw, $Ah, $Ad) = $A->getSize; $Ah = $Ah->add($Ad);
+    my ($Bw, $Bh, $Bd) = $B->getSize; $Bh = $Bh->add($Bd);
+    my $w = $space->getWidth;                            # + tabcolsep
+    my $h = $w->multiply(Float(abs($diagV / $diagH)));
```
dginev, line 77: `$diagH` should probably be guarded to not be zero, or a perl die is to follow

PR #1558, opened 2021-06-30
update MathJax url to load mml-chtml component only

xworld21:
Something I spotted in quick PR by @dginev: MathJax version 3 is being called with the options meant for version 2 (see [Loading MathJax](http://docs.mathjax.org/en/latest/web/configuration.html#loading-mathjax) versus the old [Combined Configurations](https://docs.mathjax.org/en/v2.7-latest/config-files.html#combined-configurations)).

I believe the url below is the intented one (MathML input, CommonHTML output; there is [no "or MathML" anymore](http://docs.mathjax.org/en/latest/web/components/output.html#output-chtml)).

### DIFF
--- a/lib/LaTeXML/resources/javascript/LaTeXML-maybeMathjax.js
+++ b/lib/LaTeXML/resources/javascript/LaTeXML-maybeMathjax.js
@@ -3,7 +3,7 @@
 
 (function() {
     var mathjax_url =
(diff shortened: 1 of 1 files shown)


dginev:
FWIW, this looks good to me (I updated it on our arXMLiv preview site after I saw the suggestion by @xworld21 in the last PR), and seems to be working out just fine in the previews.

So the PR is good to merge in my eyes 👍🏻

brucemiller:
Cool!  It seems to work for me. Thanks!

PR #1561, opened 2021-07-10
update fancyhdr.sty

xworld21:
This is a ~~trivial~~ update to the prototypes in `fancyhdr.sty`: some optional arguments were missing, there are some new macros that can be ignored, and [edit] more importantly
- `\iffootnote` has changed behaviour in version 3.8, September 2016 (from conditional to command)
- `\nouppercase` is only defined when inside a header or a footer, so I have commented out the definition

I have updated the test file accordingly. [Edit] I have also rearranged the definitions to follow the package documentation.

For reference, I updated the prototypes by skimming through http://tug.ctan.org/tex-archive/macros/latex/contrib/fancyhdr/fancyhdr.pdf

I have not implemented `\fancycenter[][]{}{}{}` command (roughly like an `{lcr}` tabular, but with extra effort to have the centre at the centre), also new from version 4.0, because it can be used everywhere in the document. ~~Should I do anything about it, like add a comment, or a TODO item?~~ [Edit] there is a comment about it not being implemented.


lib/LaTeXML/Package/fancyhdr.sty.ltxml
```diff
@@ -24,19 +24,19 @@ use LaTeXML::Package;
...
 DefMacroI('\fancyhead', '[]{}', '');
 DefMacroI('\fancyfoot', '[]{}', '');
 
-DefMacroI('\fancypagestyle', '{}{}', '');
-DefMacroI('\fancyhf',        '[]',   '');
+DefMacroI('\fancypagestyle', '{}[]{}', '');
+DefMacroI('\fancyhf',        '[]{}',   '');
 
 DefMacroI('\fancyheadoffset', '[]{}', '');
 DefMacroI('\fancyfootoffset', '[]{}', '');
```
brucemiller, line 42: Probably useful (upon reading the manual) to add **```DefMacro('\fancyhfoffset[]{}','');`` as well.

There are several other macros, but seem intended to be used inside the heads/foots, which we never expand, so presumably we can ignore them.

xworld21: I reordered the macros to match the docs because I kept getting lost. `\fancyhfoffset` was actually there, but I did miss another macro. Now I am fairly sure it's all there. Also, I fixed `\iffootnote`, because it's not a conditional.

*Except*, pdflatex does not compile the test, mostly because the test uses the wrong `\iffootnote` (presumably because of an old version?). I am converting the PR to draft because I need to figure out another couple of things first.

lib/LaTeXML/Package/fancyhdr.sty.ltxml
```diff
@@ -24,46 +24,56 @@ use LaTeXML::Package;
...
-DefMacroI('\plainfootrulewidth', undef, '0pt');
-DefMacroI('\footruleskip',       undef, '.3\normalbaselineskip');
+# not implemented yet: \fancycenter[][]{}{}{}, since 4.0
+
+# always false as LaTeXML does not paginate
+DefMacroI('\iffloatpage', '{}{}', '#2');
+DefMacroI('\iftopfloat',  '{}{}', '#2');
+DefMacroI('\ifbotfloat',  '{}{}', '#2');
+DefMacroI('\iffootnote',  '{}{}', '#2');    # since 3.8
+# DefConditional('\iffootnote', undef);     # pre-3.8 implementation
```
xworld21, line 54: This could break old files! How do you deal with this kind of breaking changes?

brucemiller: Hmm, fancyhdr (currently) defines them as macros in terms of `\ifx`. It seems dubious style to define `\iffootnote{}{}` as a regular macro, rather than real conditional `\iffootnote#1\else#2\fi`, but it seems like it would be a major change to redefine it that way from one version to the next. I don't see any hint of that in the documentation. Are you sure it ever was a conditional (ie. `\newif`)?

I suppose there could be corner cases where your definition above would work differently than an expansion in terms of `\ifx`. If that was a concern, you could copy fancyhdr's defns, and add your own things like
```
DefConditional('\if@fcolmode');
DefConditional('\iff@nch@footnote');
Let('\topfloat','\empty');
Let('\botfloat','\empty');
```
(those might be good to define, anyway)

xworld21: fancyhdr really did a breaking change in 2016, `\iffootnote` was a conditional and was changed into `\iffootnote{}{}`, see https://tug.org/svn/texlive/trunk/Master/texmf-dist/tex/latex/fancyhdr/fancyhdr.sty?r1=41944&r2=42004.

I can't see a way out, other than reading `fancyhdr.sty` itself?

brucemiller: Wow. Reading `fancyhdr.sty` would probably just get you the new version anyway! :>  Luckily only `iffootnote` got this special treatment. You could just go with the current behaviour and let the few old uses of `\iffootnote` break.  If you wanted to get fancy, you might gamble that there's always an opening '{', so that something like
```
DefMacro('\iffootnote`,sub {
  return ($_[0]->ifNext(T_BEGIN) ? T_CS('\lx@iffootnote@macro') : T_CS('\lx@iffootnote@cond`)); });
```
with the 2 latter being the new & old defns, respectively.
(You might look at `OmniBus.cls.ltxml`, which is a collection of dirty tricks, trying to be a class file for all times & places)

xworld21: If you are ok with the breakage, maybe it's better to go with the current behaviour. Hopefully the old uses are all within headers and footers. I am thinking, the hack can get tricky to debug when the heuristics is wrong.

Ideally, I would like to extract the version number. Another hack could be to redefine `\ProvidesPackage` so that it saves the version number and calls `\endinput`, then reads `fancyhdr.sty` (or some alternative form of grepping). Too complicated?

brucemiller: I think I'm fine with a little potential breakage. As you note: if it's used at all, it will likely be wthin one of the header macros which will never be expanded anyway.

PR #1563, opened 2021-07-11
Add binding for 'actuarialangle.sty'

xworld21:
As per subject, this is a small binding for [actuarialangle](https://mirrors.ctan.org/macros/latex/contrib/actuarialangle/actuarialangle.pdf). It translates `\actuarialangle` to the corresponding MathML `menclose` and `\overanglebracket` to its Unicode equivalent.

The output quality is hit and miss, unfortunately. Firefox renders the angled bracket properly, but messes up the actuarial angle; MathJax does the opposite.

Edit: here is a small example
```latex
\documentclass{article}
\usepackage[thinspace]{actuarialangle}
\begin{document}
\[ \actuarialangle{n} \quad a_{\actuarialangle{n}} \]

\[ \angl{ab} \quad \angln \quad \anglr \quad \anglk \]

\[ \overanglebracket{abcde} \quad \group{xy} \quad A_{\group{xy}:\angln} \]
\end{document}
```
which `pdflatex` renders as
image

### DIFF
--- a/lib/LaTeXML/Package/actuarialangle.sty.ltxml
+++ b/lib/LaTeXML/Package/actuarialangle.sty.ltxml
@@ -1,0 +1,40 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# |  actuarialangle                                                     | #
+# | Implementation for LaTeXML                                          | #
+# |=====================================================================| #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
+# |  United States Government & not subject to copyright in the US.     | #
+# |---------------------------------------------------------------------| #
+# | Bruce Miller <bruce.miller@nist.gov>                        #_#     | #
+# | http://dlmf.nist.gov/LaTeXML/                              (o o)    | #
+# \=========================================================ooo==U==ooo=/ #
+package LaTeXML::Package::Pool;
+use strict;
+use warnings;
+use LaTeXML::Package;
+
+InputDefinitions('actuarialangle', type => 'sty', noltxml => 1);
+
+DefConstructor('\actuarialangle{}',
+  "<ltx:XMApp>"
+    . "<ltx:XMTok role='ENCLOSE' enclose='actuarial' meaning='actuarialangle' />"
+    . "<ltx:XMWrap>#1</ltx:XMWrap>"
+    . "</ltx:XMApp>");
+
+# define \overanglebracket, unless the nobracket option is passed
+if (IfCondition('\ifacta@bracket')) {
+  DefMath('\overanglebracket{}', "\x{23E0}", operator_role => 'OVERACCENT'); # TOP TORTOISE SHELL BRACKET
+  DefMath('\group{}', '\overanglebracket{#1}');
+}
+
+# preserve reversion
+Let('\lx@actuarialangle@angl', '\angl');
+DefMath('\angl{}', '\lx@actuarialangle@angl{#1}');
+
+DefMath('\angln', '\angl{n}');
+DefMath('\anglr', '\angl{r}');
+DefMath('\anglk', '\angl{k}');
+
+1;
--- a/MANIFEST
+++ b/MANIFEST
@@ -336,6 +336,7 @@ lib/LaTeXML/Package/aastex.sty.ltxml
 lib/LaTeXML/Package/accents.sty.ltxml
 lib/LaTeXML/Package/acmart.cls.ltxml
 lib/LaTeXML/Package/acronym.sty.ltxml
+lib/LaTeXML/Package/actuarialangle.sty.ltxml
 lib/LaTeXML/Package/ae.sty.ltxml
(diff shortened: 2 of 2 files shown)


dginev:
It's really great to see an external binding contribution, it's been a while since we got some of those! Small or large, your effort is very much appreciated - thank you @xworld21

dginev:
Could you also add a little TeX snippet demonstrating using the binding in the PR description? It's quite helpful to quickly test the branch.

lib/LaTeXML/Package/actuarialangle.sty.ltxml
```diff
@@ -0,0 +1,43 @@
...
+  "<ltx:XMApp>"
+    . "<ltx:XMTok role='ENCLOSE' enclose='actuarial' meaning='actuarial-angle' />"
+    . "<ltx:XMWrap>#1</ltx:XMWrap>"
+    . "</ltx:XMApp>");
+
+DefMath('\overanglebracket{}', "\x{23E0}", operator_role => 'OVERACCENT'); # TOP TORTOISE SHELL BRACKET
+
+Let('\group', '\overanglebracket');
+
+# preserve reversions
```
xworld21, line 30: by the way, if there is a better way to preserve the reversions, please let me know! I copied the idea from `\prescript` in `mathtools`

xworld21:
Of course, no PR is too small: I forgot to omit `\overanglebracket` if requested via the option `nobracket`. Now it should all be good.

lib/LaTeXML/Package/actuarialangle.sty.ltxml
brucemiller: If I'm understanding what you're trying to do, I think `DefMath` has some built-in magic for dealing with such cases.
Does the following replacement for the last two lines do what you want:
```
  DefMath('\group{}','\lx@actuarialangle@group{#1}');
```

Likewise the following could replace the loop body below:
```
  Let('\lx@actuarialangle@'.$cmd, '\\'.$cmd);
  DefMath('\\'.$cmd, '\lx@actuarialangle@'.$cmd);
```

xworld21: That works very well, I have updated the PR, thanks.

lib/LaTeXML/Package/actuarialangle.sty.ltxml
```diff
@@ -0,0 +1,36 @@
...
+
+# define \overanglebracket, unless the nobracket option is passed
+if (IfCondition('\ifacta@bracket')) {
+  DefMath('\overanglebracket{}', "\x{23E0}", operator_role => 'OVERACCENT'); # TOP TORTOISE SHELL BRACKET
+  DefMath('\group{}', '\overanglebracket{#1}');
+}
+
+# preserve reversion
+Let('\lx@actuarialangle@angl', '\angl');
+DefMath('\angl{}', '\lx@actuarialangle@angl{#1}', role => 'ENCLOSE');
```
xworld21, line 34: So I have decided that preserving the reversions of `\angln` and others is overkill. However, the TeX reversion of `\angln` comes out as `{\angl{n}}` instead of just `\angl{n}`. Also, if I use `\angln`, the `<Math>` tag loses the `text` attribute. I find this a bit odd.

brucemiller: Ah, my mistake: Omit the `role=>'ENCLOSE'` from the `DefMath`; that says that the whole angl expression is an ENCLOSE, which it isn't and messes up parsing, so you lose the `text` attribute.

The extra braces on `\angln` are because that's how the style defines it; To the extent that `\angln` is a Thing, you could define it (and its friends) simply with
```
DefMath('\angln','\angl{n}');
```
and you could even add specific properties (like meaning) to it, if you wanted.

xworld21: Uh, I don't remember why I added `role => 'ENCLOSE'` anymore – I was making too many small experiments. It was not in your original suggestion, and I removed it. Now the `text` attribute works. I guess I would have seen a warning in the log? I am still getting used to the new error reporting style.

And indeed, `\angln` is defined as `{\angl n}`. I replaced them with `DefMath`'s.

brucemiller:
Cool! Looks good, thanks so much!!

brucemiller:
On 7/17/21 9:33 AM, Vincenzo Mantova wrote:
? I guess I
> would have seen a warning in the log? I am still getting used to the new 
> error reporting style.

Yeah, so am I (although I like it, obviously :> )
The warning would have shown on STDERR also, but more subtle
as a single line. Perhaps too easy to overlook?

xworld21:
> Perhaps too easy to overlook?

Indeed, this is what actually happens
```
Warning:not_parsed:>ENCLOSE MathParser failed to match rule 'Anything' at actuarial.tex; line 8 col 2
```
Trouble is, I was running `latexml` without a destination file, so it would just disappear too quickly.

PR #1564, opened 2021-07-12
Add binding for 'prettyref.sty'

xworld21:
I saw prettyref on [this list](https://github.com/brucemiller/LaTeXML/wiki/Porting-LaTeX-packages-for-LaTeXML) and I know it works with `--includestyles` because I use it regularly, so this is an easy one.

A small example:
```latex
\documentclass{article}

\usepackage{amsthm}
\usepackage{prettyref}

\newtheorem{thm}{Theorem}
\newtheorem{prop}[thm]{Proposition}
\newrefformat{prop}{Proposition~\ref{#1}}
% the thm format is builtin

\begin{document}
\begin{thm}
  \label{thm:theorem}
  A theorem.
\end{thm}

\begin{prop}
  \label{prop:proposition}
  A proposition.
\end{prop}

References: \prettyref{thm:theorem}, \prettyref{prop:proposition}.
\end{document}
```

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -585,6 +585,7 @@ lib/LaTeXML/Package/pgfutil-common.tex.ltxml
 lib/LaTeXML/Package/physics.sty.ltxml
 lib/LaTeXML/Package/pifont.sty.ltxml
 lib/LaTeXML/Package/placeins.sty.ltxml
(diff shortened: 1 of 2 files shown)


brucemiller:
Cool! Thanks!!

Issue #1566, opened 2021-07-12
\NewDocumentCommand undefined on TeX Live 2021


xworld21:
On the topic of expl3 regressions, I get the following on master, using `expl3.sty` version 2021-02-18, `xparse.sty` version 2021-03-12 (in a TeX Live 2021 installation snapshotted on 8 April 2021).

```
$ tools/maketests --compare t/expl3/xparse

========================================
Compare Expected t/expl3/xparse to current output XML:
Added: 5a6,7
>   <ERROR class="undefined">\NewDocumentCommand</ERROR>
[rest omitted]
```
 Some log
```
On the topic of expl3 regressions, I get the following on TeX Live 2021 (snapshot from 8 April 2021).
latexml (LaTeXML version 0.8.5)
(Digesting TeX xparse...
(Loading [...]/lib/LaTeXML/Package/TeX.pool.ltxml...
(Loading [...]/lib/LaTeXML/Package/eTeX.pool.ltxml... 0.00 sec)
(Loading [...]/lib/LaTeXML/Package/pdfTeX.pool.ltxml... 0.01 sec) 0.10 sec)
(Processing content [...]/t/expl3/xparse.tex...
(Loading [...]/lib/LaTeXML/Package/LaTeX.pool.ltxml...
(Loading [...]/lib/LaTeXML/Package/textcomp.sty.ltxml...
(Processing definitions /nix/store/c4hlgygrk5zyghmkw3nw3lm8yjxf7dm7-texlive-combined-full-2021.20210408/share/texmf/tex/latex-dev/base/ts1enc.dfu... 0.02 sec) 0.52 sec) 0.66 sec)
(Loading [...]/lib/LaTeXML/Package/article.cls.ltxml... 0.01 sec)
(Loading [...]/lib/LaTeXML/Package/xparse.sty.ltxml...
(Processing definitions /nix/store/c4hlgygrk5zyghmkw3nw3lm8yjxf7dm7-texlive-combined-full-2021.20210408/share/texmf/tex/latex/l3packages/xparse/xparse.sty...
(Loading [...]/lib/LaTeXML/Package/expl3.sty.ltxml...
(Processing definitions /nix/store/c4hlgygrk5zyghmkw3nw3lm8yjxf7dm7-texlive-combined-full-2021.20210408/share/texmf/tex/latex/l3kernel/expl3.sty...
[...]

PR #1569, opened 2021-07-17
Add binding for 'xparse-2018-04-12.sty'

xworld21:
This is about fixing #1566. It looks like `xparse.sty` has been split off into a few frozen versions. From the comments:
```tex
%% This is file `xparse-2018-04-12.sty', generated from
%% xparse.dtx  (with options: `core,package', on 2021-01-20)
%% then adapted and frozen for compatibility.  Development of
%% xparse.dtx will continue in the LaTeX2e kernel as ltcmd.dtx.
```
If I understand correctly, some core commands, like `\NewDocumentCommand`, are being moved into LaTeX2e kernel, so `xparse.sty` has to detect the LaTeX version, then load the appropriate version of `xparse.sty` to fill the gaps.

At the moment, LaTeXML is detected as an old format, and `xparse.sty` tries to load the 2018-04-12 version. The binding is needed to prevent LaTeXML falling back to `xparse.sty.ltxml`.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -695,6 +695,7 @@ lib/LaTeXML/Package/xfrac.sty.ltxml
 lib/LaTeXML/Package/xkeyval.sty.ltxml
 lib/LaTeXML/Package/xkvview.sty.ltxml
 lib/LaTeXML/Package/xparse.sty.ltxml
+lib/LaTeXML/Package/xparse-2018-04-12.sty.ltxml
 lib/LaTeXML/Package/xspace.sty.ltxml
 lib/LaTeXML/Package/xunicode.sty.ltxml
 lib/LaTeXML/Package/yfonts.sty.ltxml
--- a/lib/LaTeXML/Package/xparse-2018-04-12.sty.ltxml
+++ b/lib/LaTeXML/Package/xparse-2018-04-12.sty.ltxml
@@ -1,0 +1,20 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# |  xparse-2018-04-12                                                  | #
+# | Implementation for LaTeXML                                          | #
+# |=====================================================================| #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
+# |  United States Government & not subject to copyright in the US.     | #
+# |---------------------------------------------------------------------| #
+# | Bruce Miller <bruce.miller@nist.gov>                        #_#     | #
+# | http://dlmf.nist.gov/LaTeXML/                              (o o)    | #
+# \=========================================================ooo==U==ooo=/ #
+package LaTeXML::Package::Pool;
+use strict;
+use warnings;
+use LaTeXML::Package;
+
+# Should just work?
+InputDefinitions('xparse-2018-04-12', type => 'sty', noltxml => 1);
+1;


dginev:
I'm not crazy with decisions the LaTeX 3 work is inducing but ... if it's done in CTAN and we support xparse, and it's causing regressions, we ought to include it. The open question is how many of these xparse.sty files we'll end up with, plenty of more dates in the calendar. But hopefully they use the freezing sparingly.

LGTM, thanks for debugging this!

brucemiller:
On 7/17/21 10:27 AM, Deyan Ginev wrote:
> I'm not crazy with decisions the LaTeX 3 work is inducing but ... 

I haven't looked at texlive's implementation, in particular exactly
where/how it's decided which xparse to load.
But it sure would be preferable if we could avoid adding an endless
number of binding files and rather have `xparse.sty.ltxml` figure out
magically which of the `xparse-bluemoon.sty` files to load.
I dunno if that's possible to figure out, though?

xworld21:
> On 7/17/21 10:27 AM, Deyan Ginev wrote: I'm not crazy with decisions the LaTeX 3 work is inducing but ...
> I haven't looked at texlive's implementation, in particular exactly where/how it's decided which xparse to load. But it sure would be preferable if we could avoid adding an endless number of binding files and rather have `xparse.sty.ltxml` figure out magically which of the `xparse-bluemoon.sty` files to load. I dunno if that's possible to figure out, though?

I *think* that the 2018-04-12 will stay there for the foreseeable future, based on https://github.com/latex3/latex3/blob/main/l3packages/CHANGELOG.md#2021-02-02 and the fact that the code is part of the source repository. But that's just a guess, I haven't found more explanations than the comment in the frozen file.

The version detection is done by `xparse.sty` itself by checking if `\NewDocumentCommand` exists, and the value of `\fmtversion` (a date). LaTeXML does not define either of them, so it should always end up with 2018-04-12.

brucemiller:
It's not so much that xparse-2018-04-12 will go away, but that there 
will be an arbitrary number of new ones.  It's really a question of
which tail we want to end up chasing: new versions, versus the logic
of the choice changing over time.

I think I'm inclined initially to chase the latter tail and simply
modify `xparse.sty.ltxml` to do something like
```
our $xparse_version = "xparse-2018-04-12";
if(FindFile($xparse_version, type=>'sty')){
   LoadDefinitions($xparse_version, type=>'sty',noltxml=>1); }
else {
   LoadDefinitions('xparse','type=>'sty',noltxml=>1); }
```
which gives a little wiggle room to try different versions, if needed.
We can always switch to a more explicit approach if we have to,
and avoid clutter in the meantime.

BTM: Actually, `LaTeX.pool.ltxml` does define `\fmtversion, but defines
it as "2018/12/01"; possibly wrong date and/or format?

xworld21:
> BTM: Actually, `LaTeX.pool.ltxml` does define `\fmtversion, but defines it as "2018/12/01"; possibly wrong date and/or format?

Ah, you are right, `\fmtversion` is defined. LaTeX gives me 2020-10-01, so dashes instead of slashes. But `xparse.sty` uses `\@parse@version` so hopefully the slashes do not make a difference.

dginev:
So, you may have seen a comment by me about the decisions of the LaTeX 3 team ... 😅 The slashes and dashes are both valid syntaxes. And the `\fmtversion` macro has actual intended uses for random calendar dates. See specific comment, and entire thread, discussing these bits here:

https://github.com/latex3/latex2e/issues/186#issuecomment-557274225

xworld21:
Please note that the frozen version of xparse is not the full package, you still need to load the rest of `xparse.sty`.

I am fairly confident that
```perl
our $xparse_version = "xparse-2018-04-12";
if (FindFile($xparse_version, type => 'sty')) {
    # TODO copy the package options
    InputDefinitions($xparse_version, type => 'sty', noltxml => 1); }
InputDefinitions('xparse', 'type => 'sty', noltxml => 1);
```
would replicate the current logic (modulo package options!). The file test is reasonable, and things happen in the right order, so it should be ok.

Do you prefer this approach? I haven't quite understood the advantage of duplicating the code in `xparse.sty`. It seems like the maintainance burden is the same – if a new xparse needs a new frozen version, you will have to add it anyway.

xworld21:
...ah, no, xparse calls `\file_input_stop:` after loading the frozen version. Forget what I just said then. The alternative approach sounds saner.

brucemiller:
Actually, there's one thing I'm not clear about:
Once LaTeXML is reading a raw style/tex file, it should
continue reading others raw, providing it's correctly determined
the correct file to read. So if within xparse.sty (or .tex) it says
to input xparse-2018-04-12, it should do that without needing a
LaTeXML binding.

So, would LaTeXML work correctly if it just had a "correct" value
for `\fmtversion' ? (ie. w/o change to the xparse binding)

xworld21:
> So if within xparse.sty (or .tex) it says to input xparse-2018-04-12, it should do that without needing a LaTeXML binding.

The fallback mechanism strips the version number and tries to load `xparse.sty.ltxml`, which of course does not help (in fact, I get a warning from the line below when I hit the regression), that's why I suggested the additional binding
https://github.com/brucemiller/LaTeXML/blob/4186d763cf8fc82d31a5d59701b63bced1b0bf29/lib/LaTeXML/Package.pm#L1993

The `\fmtversion` is just used for date comparison, before 2020-10-01 you get 2018-04-12, otherwise you get 2020-10-01 (or none at all if certain commands are defined/undefined), so the LaTeXML value is fine from that point of view.

dginev:
Yep, the fallback mechanism for arXiv creates this difficulty for xparse.sty. Also why my first comment was happy to accept the PR as a quick resolution. Especially since I can imagine an arXiv article from this/next year that *both* needs the fallback mechanism *and* needs this frozen xparse package.

brucemiller:
On 7/18/21 9:04 AM, Deyan Ginev wrote:
> Yep, the fallback mechanism for arXiv creates this difficulty

Oh, yeah, ouch.  But the fallback is exactly there to allow
LaTeXML to use a (more general) binding to avoid reading the
raw TeX. But here, it's intending to read the raw tex anyway,
so it shouldn't be avoiding it.

That suggests to me that the fallback logic is slightly off.
In fact, why didn't it go ahead and read the frozen xparse
from the if branch of `FindFile_aux` immediately preceding
the fallback? Shouldn't `!$options{notex} && $interpretting`
have been true?

brucemiller:
Oh! because we're cleverly avoiding kpsewhich at that point...sigh.

dginev:
Right, how is the fallback to know which cases are endorsed texlive logic, and which custom arxiv patches to standard styles? A paper could include its own xparse variant which we don't want to interpret directly for now.

brucemiller:
On 7/18/21 12:43 PM, Deyan Ginev wrote:
> Right, how is the fallback to know 

My thinking was that if it was willing to accept a non-ltxml file,
so a "versioned" one should be good enough.  Willing to accept is
subtly different than accept only (`noltxml`).

dginev:
Well, in our current arXiv setup, we are "willing to accept" raw `.sty` files from the paper's current directory and load them directly, such as `mymacros.sty`. But if these local files happen to be versioned instances of standard packages (say `amsmath2020_arxiv.sty`), we are even more willing to load an existing binding (`amsmath.sty.ltxml`) since - for now - the benefits of having the binding definitions outweigh the negatives of not having whichever bits were overridden or added.

If someone includes `xparse2020_arxiv.sty`, I would still want the `xparse.sty.ltxml` loaded instead. But if this was something we vetted and eager to support directly, a binding `xparse2020_arxiv.sty.ltxml` is a great way of transparently marking that.

I think I would still prefer to merge the PR as-is, rather than do extra guessing in the fallback logic, since the use cases are somewhat hard to predict.

My long-term solution is to drop the fallback entirely and make `--includestyles` the default behavior, but we're not there just yet.

brucemiller:
I'm getting a little confused in the discussion about which alternative is alternative to what, but --- if it works --- I'd prefer one of the `FindFile` versions, rather than adding clutter to `LaTeXML/Packages/` (there's already way too much).  Does one of those versions work correctly?

dginev:
> Does one of those versions work correctly?

Can we automatically tell which versioned files are to be loaded (as with this use of xparse) vs which are to instead fall back to an existing binding? The latter has a good example with:
```
/usr/share/texlive/texmf-dist/tex/latex/amsmath/amsmath-2018-12-01.sty
```

Which we don't want to load raw, especially in arXiv. At the moment I don't see a way to tell the difference without listing the files *somewhere*. It doesn't need to be a new file if that's an issue, we could create an `AllowRawPackage($name)` directive. Though I prefer cluttering the Package/ directory to inventing more API layers. We could do a "versioned" subdirectory? :>

brucemiller:
I'm confused about the sense of your question. My query about "Does one of those versions..." was refering to the two different proposals ("versions" of the code, *not* versions of the souce file) for adding perl code to `xparse.sty.ltxml` which used `FindFile` to look for the presence of `xparse-2018-04-12.sty`.  Here the specific file and version are explict, either in the added code, or added binding.

As to a more general strategy: (1) I've wondered about having a list of safe-to-load-raw files, rather than always cluttering up with a trivial binding. But, having to track down a separate list, and the fact that in many cases the trivial binding still needs a couple of tweaks, has kept me from pursuing that.  And (2) there's probably a strategy that notes that if we're already trying to read raw, using a versioned file is OK, but it's tricky to get right in an arXiv context (and you were reluctant as well).

dginev:
> adding perl code to xparse.sty.ltxml which used FindFile to look for the presence of xparse-2018-04-12.sty

Shouldn't we let the raw interpretation of xparse.sty decide when to load that, rather than pre-deciding it ourselves? I'm not sure the choice is as easy as "always load it when it's there".

That's why I preferred a variant where the xparse.sty.ltxml binding can instead say `AllowRawPackage("xparse-2018-04-12");` and whenver Package.pm gets to the fallback logic, that exact name will be exempted from falling back.

brucemiller:
But that would require rethinking the fallback logic.
And if we're rethinking that logic, it's tempting to
try to figure out exactly what the correct logic *is*;
I suspect that if we did, we wouldn't need the declaration at all.

xworld21:
Just to confirm that both 'option A' (trivial passthrough binding) and 'option B' (code in https://github.com/brucemiller/LaTeXML/pull/1569#issuecomment-882045278) work. Except in option B I probably need to manually pass the package options (how do I do that?).

If you give me the order, I'll push the option you decide is best.

> it's tempting to try to figure out exactly what the correct logic *is*;

What about passing the `$discarded` to the binding in some way, and let the binding be in charge? Then e.g. `xparse.sty.ltxml` could use `InputDefinitions('xparse' . $discarded, ...);` or something of that sort, no need for special logic or hardcoded versions. `amsmath` could implement some tweaks for certain values of `$discarded` etc.

brucemiller:
Sorry to steal your thunder, but your debugging was very helpful in working out an alternative! Thanks

PR #1570, opened 2021-07-17
Invoke commands by passing the arguments in a list of strings

xworld21:
Reason: most tests above 90 fail when a parent folder contains a space, or if a required Perl module has a space in its path (noticed on Windows, where it's much more likely to actually happen). ~~I have added double quotes around all paths.~~

Edit: I replaced the `$invocation` strings with arrays of strings, each representing the argument in its plain form (unquoted, unescaped). I think this is the most correct solution.

If you like this approach, I can grep for all the other `system()` and backticks and do the same, because e.g. the kpsewhich calls are also giving me problem related to spaces in the arguments.


brucemiller:
Cool! Thanks!!

PR #1573, opened 2021-07-19
Parse fenced modified factors

dginev:
Motivating examples:
* `P(x=0)`
* `\text{Pr}(a < b < c)`
* `\mbox{BR}(B_{d}\to\pi^{\mp}K^{\pm})`
* `{d\sigma\over d\hat{t}}(ab\to cd)`

I added a simple case of modified factors that is [common on arXiv](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/warning/not%5Fparsed/ATOM%3EOPEN?offset=100&page_size=100) to the MathGrammar coverage. It's possibly slightly risky for existing parses, but I hope that adding it as a last fallback rule in `factorOpen` minimizes the changes of existing parses.

I didn't do any heavy-lifting on the semantic end, but kept the `Fence()` treatment used by the various `modifier` variants. It can be revisited whenever we get the modifier markup upgraded.

Might as well have these limited cases parse, as the parentheses make them relatively easy to expect and infer.


brucemiller:
Another nice, subtle, enhancement! Thanks!!

PR #1574, opened 2021-07-20
Adjustbox

brucemiller:
Binding for adjustbox.

Initially, just the groundwork that allows `adjustbox.sty` to be processed without errors:  a few random latex defintions; processing registers correctly w.r.t `\protect`; expansion of keyvals parameters and general KeyVal cleanup and simplification.

Next up is getting the correct (visual) results from adjustbox.

Of interest to @dginev, @tkw1536


lib/LaTeXML/Package/xkeyval.sty.ltxml
```diff
@@ -20,16 +20,17 @@ use LaTeXML::Util::Pathname qw(pathname_split);
 ###
 ### Basic \setkeys
 ###
+AssignValue('keyval.sty_loaded' => 1, 'global');    # pretend keyval loaded too.
```
dginev, line 23: is it loading in the pdflatex flow? If so, why not `RequirePackage` instead?

lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -24,14 +24,17 @@ use LaTeXML::Util::Image;
 
 DefParameterType('GraphixDimension', sub {
     my ($gullet) = @_;
-    if ($gullet->ifNext(T_OTHER('!'))) {
-      $gullet->readToken();
+    $gullet->skipSpaces;
+    my $next = $gullet->readXToken;
+    # Wrong !?!?!
```
dginev, line 29: curious if there are details :>

dginev (commented):
Exciting!

Not much to comment for now on my side. The `readXToken` bits will be interesting to flesh out fully while keeping everything else happy.

lib/LaTeXML/Package/xkeyval.sty.ltxml
brucemiller: It's really just to keep `keyval` from being loaded later and overwriting `xkeyval` defs (the latter is proper superset)

lib/LaTeXML/Package/graphics.sty.ltxml
brucemiller: just wondering if interpretation is correct

brucemiller:
ok, let's call that done;; modulo reviews, CI, etc

Issue #1572, opened 2021-07-18
new terminal output shows escape characters in cmd.exe


xworld21:
As per subject, the new terminal output writes escape characters even when the terminal does not support them. At the very least, I see this on cmd.exe (Windows 10, StrawberryPerl 5.32.something). So something about the terminal capabilities goes wrong there.

If you point me at the right file, I can try to debug the problem.

brucemiller:
I guess I knew that was going to happen, but just crossed my fingers. There's no real attempt to detect the terminal capabilities (how?). FWIW, it's in `LaTeXML::Common::Error`.

xworld21:
> There's no real attempt to detect the terminal capabilities (how?)

I would guess [Term::Cap](https://perldoc.perl.org/Term::Cap), but indeed, how exactly? I know by `.bashrc` (straight from Debian) uses `tput setaf 1` to detect color support. That could be a start.

brucemiller:
On 7/18/21 1:25 PM, Vincenzo Mantova wrote:
>     There's no real attempt to detect the terminal capabilities (how?)
> 
> I would guess Term::Cap

My "how?" was a bit facetious; when I looked into it, it seemed the 
solutions were worse (heavier) than the problem.  But I guess it has to 
be looked into...

xworld21:
> My "how?" was a bit facetious; when I looked into it, it seemed the solutions were worse (heavier) than the problem

I can imagine!

By the way, I just tried emptying out `%color_theme`, which almost works (the spinner is still colored, but at least the output is readable). But it has the side effect of restoring the old standard error, which I did not expect.

xworld21:
Uh, this is a lot better:
```perl
   use Win32::Console;
   Win32::Console::OutputCP(65001);
   system('');
```
from https://incusdata.site/pan/pl/Perl-%20Getting%20Started.html

Just plugged in, and I get colored spinners in cmd.exe.

So now the question in, for which values of `^O` does this need to happen? Is it safe to do in every Windows case?

dginev:
> So now the question in, for which values of ^O does this need to happen? Is it safe to do in every Windows case?

See what happens when you have it enabled in PowerShell, which is what I expect most Windows users would be running nowadays - at least that's what I have been using in recent years. If it is unchanged, it may be safe to have enabled on Windows in general.

xworld21:
> See what happens when you have it enabled in PowerShell

[...]

PR #1575, opened 2021-07-20
Enable VT100 emulation on Windows

xworld21:
~~Tiny~~ small fix to #1572 for Windows.

~~Running `system('')` enables the escape sequences, and disables them on exit, which is the desired behaviour with virtually no work. Also it is completely harmless even if it runs in the wrong terminal.~~ Now based on `Win32::Console`.

I think I did this in the right spot, fenced behind the availability of `Win32::Console` ~~(copied from the similar test from `Util::Pathname`)~~.

It works in both cmd.exe and PowerShell, at least for the versions I have at hand.

### DIFF
--- a/lib/LaTeXML/Common/Error.pm
+++ b/lib/LaTeXML/Common/Error.pm
@@ -67,7 +67,23 @@ sub UseSTDERR {
     $IS_TERMINAL = -t STDERR;
     binmode(STDERR, ":encoding(UTF-8)");
     use IO::Handle;
-    *STDERR->autoflush(); }
+    *STDERR->autoflush();
+
+    # Win32 console handling
+    if (eval { require Win32::Console; }) {
+      # set utf-8 codepage
+      # CP_UTF8 = 65001
+      Win32::Console::OutputCP(65001);
+
+      # get standard error console
+      our $W32_STDERR = Win32::Console->new(&Win32::Console::STD_ERROR_HANDLE());
+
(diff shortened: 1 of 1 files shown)


dginev:
Given that this only executes on Windows, that's fine by me. A bit of a shame that we don't have the Windows CI enabled yet, so that the PR gets properly tested, but that time will come soon too...

xworld21:
> Wouldn't have assumed `system('')` would be ever useful for anything in perl.

I wish I could find some documentation about it... fiddling with the console mode itself is too complicated, it requires saving the state and restoring it at the end (or so it seems to me).

xworld21:
Ok, the `system('')` is far too obscure. It works because it launches `cmd.exe /d/x/c`, which happens to set utf-8 and escape sequences, because... why exactly?

I have written a proper implementation with `Win32::Console`. It falls back to `Win32::Console::ANSI` if enabling the VT emulation does not seem to work.

dginev:
Fantastic work, thank you! I think you've dramatically increased the chances of the PR getting merged without the shamanic `system('')` line 😄 Thanks again.

brucemiller:
Wow; the code's scarier, but less mysterious than `system('');`, and is procted by the `eval{ require...`, so seems like it ought to be safe.  Am sure glad we ended up with the `UseSTDERR` as a hook; it seemed unnecessary before, but is really the only sensible place for this kind of special handling.

Emphatic Thanks!

PR #1576, opened 2021-07-23
Enable Windows VT100 emulation only if in a terminal

xworld21:
Small update to #1575... I forgot the easy part of the PR, which is to only execute the console changes if `$IS_TERMINAL` is true.


brucemiller:
Ha! I thought about that earlier, and forgot about it too!! :>

PR #1577, opened 2021-07-23
Read $dtdfile into a string before parsing

### LINKED ISSUE #1571: LibXML::Dtd->new() fails when $dtdfile contains a space
Another space related bug. This line
https://github.com/brucemiller/LaTeXML/blob/c2958fbe2f9d2d2a7d616fd59ab71441d1a9fcc7/lib/LaTeXML/Common/Model/DTD.pm#L120
fails when `$dtdfile` contains a space.

Two possible fixes: use `URI::file` to convert the path to a `file:` URI, or read the entire file into a string and use `Dtd->parse_string()` instead of `Dtd->new()`. Both seem to work on my machine.

xworld21:
Fix #1571.

After reading more than should have been necessary, I am convinced that this approach (reading the file into a string, then call `parse_string`) works correctly re DTD encoding. The DTD will not know its `public_id` but it hardly seems relevant.

The only other approach is to turn `$dtdfile` into a `file://` URI, but that would be more code (or a `URI::file` dependency).

### DIFF
--- a/lib/LaTeXML/Common/Model/DTD.pm
+++ b/lib/LaTeXML/Common/Model/DTD.pm
@@ -16,6 +16,7 @@ use LaTeXML::Util::Pathname;
 use LaTeXML::Global;
 use LaTeXML::Common::Error;
 use LaTeXML::Common::XML;
+use URI::file;
 
 #**********************************************************************
 # NOTE: Arglist is DTD specific.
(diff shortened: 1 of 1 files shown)


dginev:
I was half-hoping you opened a PR that removes the DTD functionality, at which point I was to start arguing in your favor :> But sure, might as well keep it functioning if we're keeping it around.

So there is no way to use `XML::LibXML::Dtd->new` in a manner that respects the encoding?

xworld21:
> So there is no way to use `XML::LibXML::Dtd->new` in a manner that respects the encoding?

Not so much the encoding, but parsing the filename correctly. I can make `new` work by making `$dtdfile` a file URI. Actually, is there some util within LaTeXML for URI encoding? I didn't look too much into this.

xworld21:
Just to clarify, this also seems to work
```perl
use URI::file;
# ... #
$dtd = XML::LibXML::Dtd->new($$self{public_id}, URI::file->new($dtdfile));
```

dginev:
Nice! 

`URI::file->new($dtdfile))` is fine by me, looks like it could be a robust fix.

Is `URI::file` available with the current `URI` dependency? If so, feel free to go for that, though you may want @brucemiller to OK that as well.

brucemiller:
Indeed, seems to already be a dependency. Looks good; Thanks!!!

PR #1579, opened 2021-07-24
Use double quotes when compiling MathGrammar

xworld21:
Another Windows paper cut: the MathGrammar compiles properly, but the `-e "MathGrammar.pm";` test fails because the perl code is passed in single quotes, and Windows does not do single quotes.


brucemiller:
Oh, my! Thanks!!

PR #1580, opened 2021-07-24
Portable update_revision

xworld21:
Another Windows motivated PR. The `update_revision` part of the Makefile essentially expects a POSIX system, but we can do without. This is a bit delicate, I think I have tested it properly, but please double check.

The changes:
1. Only check that `.git` exists (sometimes it is a file, e.g. when using submodules - not Windows related, but relevant);
2. Replace backticks with `$(shell ...)`, and fiddle with the quotes, again because Windows does not understand them;
3. Replace `test` with a tiny perl script, because `test` is not available from Windows.


brucemiller:
Seems to work for linux, so if it does Windows too, it looks good! Thanks!!

PR #1581, opened 2021-07-25
Don't set autclose when flushing an input mouth

brucemiller:
The previous setting of autoclose when flushing seems no-longer necessary (if ever? it would seem to be tested by `t/expansion/endinput`, which still passes) and occasionally causes mouths to be closed too early; chaos ensues (esp. pgfplots).


dginev:
It can be hard to reason about the auto-closing, so I don't mind the fix at all - especially given that the tests pass and the regressions is gone. Here is a screenshot of a simple HTML doc i converted with this branch

pgfpicture

So, LGTM!

dginev (approved):

PR #1582, opened 2021-07-25
Guard for Tokens in UnTeX

dginev:
Stumbled on a curious fatal error in the [sandbox reports](https://corpora.mathweb.org/corpus/sandbox%5Farxiv%5F1905/tex%5Fto%5Fhtml/fatal/perl/die?all=false), namely:

> Can't locate object method "getCatcode" via package "LaTeXML::Core::Tokens" at /usr/local/share/perl/5.30.0/LaTeXML/Core/Token.pm line 146

A quick check reveals a number of `reversion` attributes of `DefParameterType` declarations in TeX.pool which return a `Tokens` object rather than a perl list of `Token` objects. For now I made `UnTeX` more robust, where it will actively expect `Tokens` objects and unlist them.

In a hypothetical rewrite, I see I have made the choice of standardizing DefParameterType reversions to always return a single `Tokens` object. Which is almost identical to a list of individual Token objects, but a hypothetical compiler would indeed force the choice to be made and followed. Since we don't live in that hypothetical world, here's a runtime patch for UnTeX :)

### DIFF
--- a/lib/LaTeXML/Core/Token.pm
+++ b/lib/LaTeXML/Core/Token.pm
@@ -136,7 +136,9 @@ my $UNTEX_LINELENGTH = 78;    # [CONSTANT]
 sub UnTeX {
   my ($thing, $suppress_linebreak) = @_;
   return unless defined $thing;
-  my @tokens = (ref $thing ? $thing->revert : Explode($thing));
+  my @tokens = ref $thing ?
+    map { ref $_ eq 'LaTeXML::Core::Tokens' ? $_->unlist : $_ } $thing->revert :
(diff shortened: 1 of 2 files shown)


brucemiller:
Re: standardizing. It's not just parameter types, but standardizing *all* reversions to return a Tokens would be a good thing. I think the current expectation was that it returns a list of Token(s), but apparently sometimes Tokens. `Explode` is another such. Should do a clean sweep at some point and get ALL such (except, of course `->unlist`, :>) to return Tokens. (or a single Token is also acceptable)

PR #1585, opened 2021-07-28
more portable test for urls (esp. Windows)

### LINKED ISSUE #1567: Common::Locator::getShortSource confuses 'C:\...' with a url on Windows
First report from Windows (not enjoying this!).

Plenty of tests fail on Windows on master. The failing lines are
https://github.com/brucemiller/LaTeXML/blob/13f394ec64e8a6acc3d3645bdcb569da3ab03cf8/lib/LaTeXML/Common/Locator.pm#L55-L57
This runs `url_split` on paths like `C:\Users\...`, receives two arguments instead of three because there are no slashes, then fails because `$ext` is undefined.

Changing `':'` with `':/'` in `Locator.pm` makes most tests happy (others still fail for different reasons). But that's not a good solution because it ignores potentially valid URLs.

At any rate, `url_split` returns two values, not three, which is also suspicious.

Edit: the definition of "[starts with a Windows drive letter](https://url.spec.whatwg.org/#start-with-a-windows-drive-letter)" seems relevant.

brucemiller:
Try not to treat a Windows pathname `C:\foobar` as a url.

Fixes #1567 (hopefully)

### DIFF
--- a/lib/LaTeXML/Common/Locator.pm
+++ b/lib/LaTeXML/Common/Locator.pm
@@ -52,7 +52,7 @@ sub getShortSource {
   my ($self, $stringSource) = @_;
   my $source = $$self{source};
   return ($stringSource || 'String') unless ((defined $source) && $source);
-  if (index($source, ':') != -1) {
+  if (index($source, ':') > 1) {    # assuming legit protocols are more than 1 char (NOT windows C:/)
     my ($path, $base, $ext) = url_split($source);
     return "$base.$ext"; }
   else {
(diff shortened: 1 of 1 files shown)


dginev:
I like the subtlety in this particular case, nice and simple! LGTM

xworld21:
Thanks! I was dragging my feet on this. I have had the code `index($source, ':') >= 2` on my Windows VM and it fixes this particular problem.

Just to make sure this is not forgotten: in #1567 I also mentioned that `url_split` returns two values while `getShortSource` expects three.

PR #1586, opened 2021-07-28
Timeout-related arXiv patches

dginev:
Apparently seeing an arXiv PR gets one curious to also make another small arXiv PR...

This patch upgrades `arXiv:cond-mat/0003435` from a timeout Fatal to a Warning status (due to math parsing).

* lock more natbib macros to avoid redefinitions in arXiv that lead to badly broken misinterpretation with infinite loops in `\bibitem`.
  * Also make sure we have enough locked to keep the citation filling operational.
* add some minor internal revtex macros to silence some otherwise unimportant errors.

Hopefully no unintended consequences... 

There is a general pattern I've spotted, which is that quite often we don't want to allow redefining macros which have a constructor in their expansion, especially for core features of LaTeX, but maybe in general. OR, vice-versa, only allow a redefinition if the new definition *also* has a primitive or constructor in the expansion, to ensure we don't lose the level of supported XML markup. I haven't acted on this observation here, but felt worth recording.

### DIFF
--- a/lib/LaTeXML/Package/amstext.sty.ltxml
+++ b/lib/LaTeXML/Package/amstext.sty.ltxml
@@ -20,7 +20,7 @@ use LaTeXML::Package;
 #**********************************************************************
 RequirePackage('amsgen');
 
-DefConstructor('\text{}', "<ltx:text _noautoclose='true'>#1</ltx:text>", mode => 'text');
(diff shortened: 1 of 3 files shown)


brucemiller:
Good observation about locks, but I've been hesitant to go too far; I've seen cases where things get redefined and then used with different argument patterns. Perhaps a more correct generalization is where latex has layers, we should have the same macro upper levels (which can be redefined), but have the lower layer be a constructor. Alas, that's hard to analyze & code.

PR #1587, opened 2021-07-29
fix malicious typo

brucemiller:
two steps forward, one back.

[wip, till sandbox run finished]


dginev (approved):
I have an automatic instinctive reaction: "a typo in a binding suggests we need a new/extended test using the binding".

Aside from that, good quickfixes!

brucemiller:
Your reaction is quite reasonable; the usual excuses being too many bindings, too many tests already, time consuming to create, pstricks is never modified and seldom used (except 1.5% of arxiv!?!). That said; patches welcome :>

Anyway, to my eye this should fix the only regression I noticed;  Fatal/misdefined/no_type

PR #1588, opened 2021-07-29
allow trailing punct in scripts

dginev:
I'm surprised how hard this was, since I intended it to be an easy upgrade.

I found one adaptation to the script rules, trying to only minimally modify MathGrammar, that also allows trailing punctuation in scripts. Motivated by an arXiv formula, as usual.

On the surface, one simply needs to tack on an `endPunct(?)` clause to the end of the `Subscript` and `Superscript` rules.

But for that to work, one has to also correctly propagate the information to the semantic subroutines, and upgrade `NewList` and `extract_separators` for the possible absent/present trailing punctuation content.

That said, after a bit too much trial and error, I think I have a working variant, with a few added tests. Give it a spin and let me know what you think.


lib/LaTeXML/Core/Document.pm
```diff
@@ -1595,6 +1595,7 @@ sub setNodeBox {
 
 sub getNodeBox {
   my ($self, $node) = @_;
+  return unless $node;
```
dginev, line 1598: i ended up blowing up on this while prototyping, so kept the guard in just in case.

brucemiller:
Interesting; the changes weren't quite as scary as you implied, but still a bit hairier than you'd expect.

In the end, you're just ignoring that trailing punctuation, right? The hair is propogating things through to the natural place to construct the `XMDual`. And the surprise is the magic upper layer that moves the trailing punctuation up a level (presumably that's still correct enough?).

dginev:
My first attempt had the trailing punct XMTok disappear entitely, so just having it remain in the final XML was a win.

And yes, the magic that moves it out of the list application feels "ok enough" to me. They're void of actual meaning, so as long as they get to the presentation tree, we're good.

PR #1589, opened 2021-07-29
arXiv patch: custom tabulars emit less errors

dginev:
Rescues a short undergrad thesis posted on arXiv from a Fatal to an Error. In particular, [1905.00636](https://arxiv.org/abs/1905.00636).

The main challenge was a custom tabular environment `{game}` defined in their local `sgamevar.sty` style file. I don't think we're quite there with interpreting custom aligned environments, but we ought to be good enough to not fatal for a mid-sized document.

I extended the same trick I introduced for `&`, also for `\omit`, `\noalign` and `\span`, and added a boxing group when encountering these via `->bgroup`. That reduced the document down to 33 errors, and produced an HTML with basically all content present, save for the custom tables.

A minor bit that I also spotted in the doc was that `[colorlinks]` loads `color.sty` in hyperref, so that the `\color` macro is available, etc. So I added a conditional to do that in the hacky option-handling code in the hyperref binding.

### DIFF
--- a/lib/LaTeXML/Core/Stomach.pm
+++ b/lib/LaTeXML/Core/Stomach.pm
@@ -172,7 +172,7 @@ INVOKE:
       # Locally deactivate to avoid a flurry of errors in the same table.
       # Alert the end user once per table, allowing longer documents to not
       # hit 100 errors too quickly.
-      $STATE->assignMeaning(T_ALIGN,
+      $STATE->assignMeaning($token,
         $STATE->lookupMeaning(T_CS('\relax')), 'local')
         if Equals($token, T_ALIGN);
       Error('misdefined', $token, $self,
@@ -365,7 +365,7 @@ sub setMode {
     # but inherit color and size
     $STATE->assignValue(font => $STATE->lookupValue('savedfont')->merge(
         color => $curfont->getColor, background => $curfont->getBackground,
-        size => $curfont->getSize), 'local'); }
+        size  => $curfont->getSize), 'local'); }
   return; }
 
 sub beginMode {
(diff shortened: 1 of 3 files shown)


lib/LaTeXML/Core/Stomach.pm
```diff
@@ -172,7 +172,7 @@ INVOKE:
       # Locally deactivate to avoid a flurry of errors in the same table.
       # Alert the end user once per table, allowing longer documents to not
       # hit 100 errors too quickly.
-      $STATE->assignMeaning(T_ALIGN,
+      $STATE->assignMeaning($token,
```
dginev, line 175: Ah, and this bit - in this article a custom macro was `\let` to `&`, so the guard on `T_ALIGN` wasn't catching that custom macro. By using the exact token which snuck into the Stomach error, this should quiet a few further errors.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -2520,11 +2520,41 @@ DefConstructorI('\hidden@align', undef, "",   alias => '');
...
+    $_[0]->bgroup;
+    Error('unexpected', '\noalign', $_[0], "\\noalign cannot be used here");
+    $STATE->assignMeaning(T_ALIGN,
+      $STATE->lookupMeaning(T_CS('\relax')), 'local');
+    $STATE->assignMeaning(T_CS('\noalign'),
+      $STATE->lookupMeaning(T_CS('\relax')), 'local');
+    $STATE->assignMeaning(T_CS('\omit'),
+      $STATE->lookupMeaning(T_CS('\relax')), 'local');
+    $STATE->assignMeaning(T_CS('\span'),
+      $STATE->lookupMeaning(T_CS('\relax')), 'local');
```
brucemiller, line 2532: Wouldn't `Let` be clearer (and significantly less scary?)

dginev: It would, done!

brucemiller:
OK, let's go with that.

PR #1592, opened 2021-07-31
Safe(r) backticks

xworld21:
Another fix about passing the arguments in a list so that they are escaped correctly. With this, Windows passes all tests except for `931_epub.t`, so it's very close to working out of the box!

On the other hand, this is only a stopgap measure because Perl may still mangle the arguments under certain conditions. My suggestion would be to add a small `LaTeXML::Util::Run` module with say `run` and `capture` in order to deal with this properly.

In the meanwhile, this PR should cover like 99.9% of the situations correctly.


brucemiller:
Thanks!

PR #1594, opened 2021-07-31
Consume output of latexmlc invocation and close pipe in EPUB test

xworld21:
Apparently, Windows is more picky with `open3`: if you don't consume the output, the command hangs and `waitpid` waits forever. So consuming the output and calling `close` is the way to go, rather than `waitpid`.

This is the last Windows fix!

### DIFF
--- a/t/931_epub.t
+++ b/t/931_epub.t
@@ -26,6 +26,8 @@ push(@invocation, $latexmlc, '--css=LaTeXML-epub.css', "--dest=$epub_filename", 
 
 my ($writer_discard, $reader_discard, $error_discard);
 my $pid = open3($writer_discard, $reader_discard, $error_discard, @invocation);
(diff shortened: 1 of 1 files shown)


dginev:
I had no idea that `open3` has a gotcha under Windows, this can't have been fun to debug. Thanks!

xworld21:
> I had no idea that `open3` has a gotcha under Windows, this can't have been fun to debug. Thanks!

And that's not documented! At any rate, I tweaked the PR just a bit, since `IPC::Open3` talks about defunct processes and zombies and it's probably still better to call `waitpid`. I hope I got it right now.

dginev:
Actually that does look better - thanks again!

brucemiller:
OK, let's give this a try.  Thanks so much for the Windows debugging!

PR #1595, opened 2021-08-01
Updates and simplify TEI stylesheet, and test case; validate

brucemiller:
Updated and simplfied TEI stylesheet,  and validated the test case.

Comments and suggestsions from any and all TEI users, or potential ones, welcomed!

(@Thanathan-k ?)

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-tei.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-tei.xsl
@@ -11,6 +11,10 @@
     extension-element-prefixes="str"
     exclude-result-prefixes="ltx str m svg xlink">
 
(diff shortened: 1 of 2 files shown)


dginev:
You may also want to tag @physikerwelt since he may have a toolchain using latexml's TEI output to check against this PR.

I don't have much to add myself, if a validator is happy, and the main current TEI consumers are happy, then I'm happy too.

physikerwelt:
@nmeuschke do you know if we have someone currently working on the hyplag / TEI input format, so that we can not only review but also test the change

nmeuschke:
@physikerwelt Unfortunately, nobody is working on math conversion at the moment.

brucemiller:
Thanks for checking, Moritz. I guess we'll go with this for now

Issue #771, opened 2016-06-24
pgfplots seems not to be supported (?)


kaagithub:
Hi again,

Here is a sample that I think is not supported.
I get:

```
Error:latex:\GenericError Package PGF Math Error: Could not parse input '{1,10,20,30,40,50,60,70,80,90}' as a floating point number, sorry. The unreadable part was near ',10,20,30,40,50,60,70,80,90'.
    at /blablabla/testPgfPlots.tex; line 131 col 13
    See the PGF Math package documentation for explanation.
    In Core::Definition::Primitive[\GenericE... from LaTeX.pool.ltxml line 4172
     <= Core::Stomach[@0x46846f0] <= Core::Gullet[@0x4689958] <= Core::Stomach[@0x46846f0] <= ...
```

[testPgfPlots.zip](https://github.com/brucemiller/LaTeXML/files/332047/testPgfPlots.zip)

dginev:
Thanks for the report!

brucemiller:
Hmm... it seems that it's not recognizing the list of values.  This'll take a bit longer to figure out.

brucemiller:
It *is* supported in principle, but there're a few bugs. I've fixed the main reported issue with it not recognizing lists of values. But your sample still fails with a few different issues. Worst is that ```\addlegendentry``` causes it to go into an infinite loop! Even without that, the blue keyword isn't being recognized, and the calculated viewport is wrong, so that the x axes scales, athough present, aren't visible.

So, still working on it...

brucemiller:
Some slight progress, over time: can produce a rough equivalent by commenting out ```\addlegendentry``` (which loops for some reason); and it still doesn't recognize "blue"; otherwise works...

brucemiller:
Alas, I'm going to have to come back to this after the next release.

brucemiller:
or maybe the next one. In between times, we had some regressions, which hopefully have been fixed.  And upon retesting this case, I see that exactly the comment I made before is still the case (missing blue and `\addlegendentry` fails).  Other than that, the plot works fine :>

dginev:
The original example now seems to have regressed. It hits an infinite loop in the latest master + texlive 2020, adding the respective labels. This is likely also inducing at least some arXiv timeouts, though I am unsure if the loop is still there on texlive 2019.

It appears the regression is in the call to `\legendentry{legend}`, since removing that succeeds smoothly on the following small snippet:
```tex
\documentclass{article}
[...]

PR #1597, opened 2021-08-03
Respect \globaldefs in assignments

dginev:
This fixes the regression piece in #771 , but only that much (e.g. the `/pgf/blue` key is still unrecognized and the color feature is not supported).

The regression was actually an opportunity to add a missing piece from the TeX interpretation side of things -- the internals of `pgfplots.code.tex` rely on `\globaldefs` to make certain assignments globally, which latexml hadn't yet emulated.

I tried my best to "hot code" the upgrade, as it will be checked extremely often in `assign_internal` (thanks to Bruce for pointing me to the right place in State!). 

While doing that I fell into a bunch of common traps, chiefly not realizing that getting the integer value of a Number register requires `->valueOf()->valueOf()` and perl will happily not tell you that you're not looking at an integer if you only call a single one of the two calls. I ended up unwinding that piece of logic to the minimal hash lookup. 

So, reporting times on `make test` (parallel on my 32 thread cpu):

time | master | this PR
:---- | :--- | :---
real	| 1m42.705s | 1m44.802s
user | 8m8.832s   | 8m10.928s
sys	| 0m10.501s | 0m10.340s

So just about unchanged! That's a relief. With his regression out of the way, we may be quite close to starting a full arXiv rerun. Feedback welcome!

### DIFF
--- a/lib/LaTeXML/Core/State.pm
+++ b/lib/LaTeXML/Core/State.pm
@@ -89,15 +89,15 @@ sub new {
   my ($class, %options) = @_;
   my $self = bless {    # table => {},
     value   => {}, meaning  => {}, stash  => {}, stash_active => {},
-    catcode => {}, mathcode => {}, sfcode => {}, lccode => {}, uccode => {}, delcode => {},
+    catcode => {}, mathcode => {}, sfcode => {}, lccode       => {}, uccode => {}, delcode => {},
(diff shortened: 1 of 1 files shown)


dginev:
And for the visuals, here is the HTML5 generated by the PR from my minimal example in #771  : 

Screenshot 2021-08-03 at 19-33-01 Untitled Document

brucemiller:
Cool!

Issue #896, opened 2017-12-04
Document HTML5 output dialect 


dginev:
CC-ing @brucemiller and @kohlhase here.

This issue is more of a question/discussion starter. It is related partially to #835, but is mainly due to a revival of our KWARC effort to bundle together and make available an HTML5 representation of the successful subset of arXiv under latexml.

Namely, it would help to have a level of documentation comparable to the [main schema](http://dlmf.nist.gov/LaTeXML/manual/schema/) manual, that specifically zeroes in on the HTML5 output. We have had this discussion internally at KWARC several times over the years, and based on our experience it turned out that working on a dialect of HTML is a much more adoptable approach for the wider community, than working on LaTeXML's custom XML schema. The main goal I have in mind right now is that when a newcomer contributor obtains the HTML5 arXMLiv corpus, they have a convenient way to gloss the different types of markup and supported constructs. It can be a very powerful starter for designing linguistic analysis tools, search engine indexers, etc.

If we agree having this level of documentation would be useful, another question is whether it would belong in the latexml manual, or should be hosted externally -- i'm open to both, although the manual feels like the natural host.

Happy to contribute to that myself, and it would also be a good time to see if there are some schema.org synergies that could be discussed in relation to latexml's (optional) RDFa output.

kohlhase:
I think a documentation of the HTML5 is extremely useful and would indeed be timely. 

If I understand correctly, then this manual would mostly be a documentation of the generated RDFa and CSS classes, and what they mean correct? And as such it would be much less a "schema manual", but an "annotation lexicon", so I am wondering whether the schema.org synergies will be so important.

kohlhase:
I think the LaTeXML manual would be the right place to put such a documentation with a machine-readable copy of the RDFa part on schema.org for outreach.

dginev:
> mostly be a documentation of the generated RDFa and CSS classes, and what they mean correct?

Actually, I think that may be an oversimplification. But I see and agree with the point that we do not want to rewrite the HTML5 spec as part of documenting the latexml dialect. In this regard my comparison to the intermediate latexml XML is indeed misleading, as that is structured the way a traditional self-contained schema definition would be.

It is possible that the correct way to structure the html docs is to:
 * state that the dialect builds on and respects the main html5 specification
 * proceed to provide a map from "authoring constructs" to the specific html5 produced by latexml. 
[...]

PR #1598, opened 2021-08-04
Add 'role' attribute to 'ltx:resource' elements

tkw1536:
This PR is stems from our discussion last Friday and is first step towards making it easier to customize LaTeXML CSS. I don't expect this PR to be merged right away and am happy to hear feedback. 

In particular, it introduces a distinction between resource files that
- provide some kind of semantic styling (e.g. `ltx-ulem.css` for the `ulem` package)
- provide a sensible default theme that might be overwritten ( e.g. `LaTeXML.css`)

This is done by adding a new "role" attribute the `<ltx:resource>` elements. It for now carries either the value "default" or "semantic". The idea is that we eventually add an argument to `latexml[c]` to omit the `role="default"` resources in the output (to be replaced by user-customizable css), but keep the `role="semantic"` ones. 

This PR only introduces the role attribute, and categorizes existing calls to `RequireResource` into semantic and default ones. It doesn't add a new command line flag - that is for future PRs (if we opt to go with this approach). I might also have gotten some classifications wrong.


dginev:
image

We may want to document an example workflow with contributing kinds of CSS, to figure out if the customization setup helps or hinders. For instance, say either a latex style, or a user/vendor, wanted to *add* 
First reaction: Yikes. Too scary. Can we avoid having an attribute that changes test files?

Also, the "semantic" word is something I've grown allergic to in general, but I also don't believe it makes sense to apply for the CSS stack. "Semantic styling" is almost explainable, but there's bound to be a better way of framing it that doesn't sound near-paradoxical.

A document's chosen modules from the LaTeX stack introduce the style, so some CSS files can be seen as "class CSS" or "style CSS" variants, in the latex sense. Additionally, downstream users of latexml (think arxiv-vanity, or cortex) can provide their own "vendor CSS" files, or - if one prefers - "publisher CSS" files that "brand" documents for the system that deploys them online. Currently most of the "vendor" styling is done externally to latexml, with some HTML rewriting after post-processing is completed.

The LaTeXML.css is a kind of a foundation, on which the rest of the latexml-distributed `.css` files build on. But certain vendors may start from scratch and avoid using that entirely. I think Bruce wants to encourage binding contributors to build on top of LaTeXML.css, and I know at least he enjoys doing so.

Meanwhile, emulating beamer could imply "theme CSS" is possible, and some of the examples you've marked up could be seen as "feature CSS" files that style a particular custom  latex component. I would see these as analogous to styling React components actually, so either "feature" or "component" seem like good words for partial snippets of CSS targeting e.g. a small dialect of HTML elements and attributes.

All of that said, we should check before we make the customization more complicated (=expressive), whether we can instead make it simpler and more transparent, so that one can figure out how/when the stylesheets cascade and interplay with each other. Frontend developers have some history of complaining how hard it is to reason about CSS modularly, so my premise itself is a bit in question.

But in general, if you add new attribute for beamer - that's probably fine and great. If you add new attributes that propagate to all XML test files, there better be a great reason for it 😅

tkw1536:
> First reaction: Yikes. Too scary. Can we avoid having an attribute that changes test files?

Ok, here is a second attempt that is less scary, from 200+ files down to 14 changed files. 
It's essentially the same as the first PR except that I'm omitting "role='default'" and assume it's the default. 

> Also, the "semantic" word is something I've grown allergic to in general, but I also don't believe it makes sense to apply for the CSS stack. "Semantic styling" is almost explainable, but there's bound to be a better way of framing it that doesn't sound near-paradoxical.
>

I somewhat agree the word "semantic" is bad, but it's the best one I could think of in this case. 

> A document's chosen modules from the LaTeX stack introduce the style, so some CSS files can be seen as "class CSS" or "style CSS" variants, in the latex sense. Additionally, downstream users of latexml (think arxiv-vanity, or cortex) can provide their own "vendor CSS" files, or - if one prefers - "publisher CSS" files that "brand" documents for the system that deploys them online. Currently most of the "vendor" styling is done externally to latexml, with some HTML rewriting after post-processing is completed.
>
> The LaTeXML.css is a kind of a foundation, on which the rest of the latexml-distributed .css files build on. But certain vendors may start from scratch and avoid using that entirely. I think Bruce wants to encourage binding contributors to build on top of LaTeXML.css, and I know at least he enjoys doing so.

The more I think about it, the more I have come to following opinion.
I've been typing this out and now realize that it has probably been said already. 


For lack of better names, from what I can see there are three kinds of styles:

1. "inline" styles, hard-coded in the LaTeX source
2. "semantic" CSS for custom "html elements"
3. "thematic" CSS that adjust the overall look and feel of the document

__Inline__ styles are those that are directly coded inside the LaTeX source and are translated as such into the html. 
For example using `\color{blue}` results in `style="color:#0000FF;"`.
They are not realized using html classes, but using inline style attributes.
Technically `\vspace` and `\hspace` also fall into this category, but they are mostly ignored in the conversion.

__Semantic__ are those styles that are created because of a lack of html elements. 
For example, there is no `<footnote>` element in HTML, hence we fake having one using a bunch of `<span>`s with classes attached to them.
Similarly, for `\begin{lstlisting}` environments we build a bunch of `<div>`s that need extra styling to appear properly (although in this case we should be using `<code>` elements with fancy CSS, but that is for another issue).

__Thematic__ styles are all the other styles - things that aren't hard-coded in latex source and have a native html5 element they correspond to.
Assuming for a second we did not include any CSS for them, the default browser CSS would take over - and make them look "ugly" but recognizable. 


> Meanwhile, emulating beamer could imply "theme CSS" is possible, and some of the examples you've marked up could be seen as "feature CSS" files that style a particular custom latex component. I would see these as analogous to styling React components actually, so either "feature" or "component" seem like good words for partial snippets of CSS targeting e.g. a small dialect of HTML elements and attributes.
>
> All of that said, we should check before we make the customization more complicated (=expressive), whether we can instead make it simpler and more transparent, so that one can figure out how/when the stylesheets cascade and interplay with each other. Frontend developers have some history of complaining how hard it is to reason about CSS modularly, so my premise itself is a bit in question.

Maybe indeed "feature" instead of "semantic" is the right name for (2).

However my my main point with this PR was not the changes themselves, but the point of highlighting there is a difference between (2) and (3), and figuring out what that difference is.

dginev:
1. Figuring out the differences is good. I'm sure Bruce has his own model to share as well.

2. There is almost always a more specific (= better) adjective than "semantic" for a technical component. Glad you like "feature", that or an adjacent word could serve us well.

And yes, only 14 files changed is fantastic, thanks!

tkw1536:
Will close by means of #896.

PR #1600, opened 2021-08-06
Add --log option to manual page

flyn-org:
Signed-off-by: W. Michael Petullo <mike@flyn.org>

### DIFF
--- a/bin/latexml
+++ b/bin/latexml
@@ -194,6 +194,7 @@ latexml [options] I<texfile>
                     by default it avoids this.
  --path=dir         adds to the paths searched for files,
                     modules, etc;
+ --log=file         specifies log file (default is file named after job name)
  --documentid=id    assign an id to the document root.
  --quiet            suppress messages (can repeat)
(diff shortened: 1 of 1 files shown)


bin/latexml
```diff
@@ -194,6 +194,7 @@ latexml [options] I<texfile>
                     by default it avoids this.
  --path=dir         adds to the paths searched for files,
                     modules, etc;
+ --log=file         specifies log file (default stderr)
```
brucemiller, line 197: Your perfectly reasonable assumption just goes to show why it needed documentation!
The default is actually `<jobname>.latexml.log`.  And, the log file gets a "full" log, whereas stderr now only gets a brief report.

flyn-org: I just force-pushed a change to the usage message. I think the default *used* to be stderr, but a recent upgrade changed things. Anyway, that is what led me to revise my build system, which led me to the documentation omission.

brucemiller:
Thanks!

PR #1601, opened 2021-08-07
Noexpand fix

brucemiller:
This restricts which tokens `\noexpand` applies to, namely it should *not* apply to cs which are `\let` to tokens. 

I had first "fixed" `\ifcat`,`\if`, and then '\ifx`, until I recognized the common theme.


lib/LaTeXML/Core/State.pm
```diff
@@ -422,6 +422,26 @@ sub lookupExpandable {
     return $defn; }
   return; }
 
+# Whether token must be wrapped as dont_expand
+sub isDontExpandable {
```
dginev, line 426: Maybe name it `can_expansion_be_deferred` ? It's hard to find a good name.

Or my first instinct was: this subroutine will only be used in one caller, can we merge it back with it? Namely, the `with_dont_expand` sub.

t/expansion/noexpand_conditional.tex
```diff
@@ -95,5 +95,7 @@ \section{Active Chars}
 \tester{]}{]}
 \tester{[}{@}
 \tester{^}{@}
+\let\letA=A
+\tester{A}{\letA}
```
dginev, line 99: nice new test!

lib/LaTeXML/Core/State.pm
```diff
@@ -422,6 +422,26 @@ sub lookupExpandable {
...
+  my ($self, $token) = @_;
+  # Basically: a CS or Active token that is either not defined, or is expandable
+  # (but not \let to a token)
+  return unless $token;
+  my $defn;
+  my $entry;
+  #  my $inmath = $self->lookupValue('IN_MATH');
+  my $cc = $$token[1];
+  if ($CATCODE_ACTIVE_OR_CS[$cc]) {
```
dginev, line 435: btw this `$CATCODE_ACTIVE_OR_CS` check is redundant with the outer check from don't expand for
`(($cc == CC_CS) || ($cc == CC_ACTIVE))` . Maybe one of the two should be phased out?

dginev (approved):
Read through and this is an excellent correction!

Very happy that you traced it back to the high-difficulty underlying cause with `dont expand`, allowing for all conditionals to stay as they are.

I think the check whether a token is subject to the special treatment is as accurate as at least I can think of, so this should be good to go for immediate merge 👍🏻

lib/LaTeXML/Core/State.pm
brucemiller: Kinda liked the name, exactly because it is ridiculous; gotta have *some* fun. And liked it in `State` because it accesses so much internal stuff; and leaves `with_dont_expand` readable.

t/expansion/noexpand_conditional.tex
brucemiller: And surprising!

PR #1602, opened 2021-08-08
Avoid adding date to split-off pages

### LINKED ISSUE #1478: Question concerning dates
I would appreciate some advice on the best way to stop document publication dates tags (`<div class="ltx_dates">`) being included on each page (except the topmost page, where I want to keep them) of a collection  of HTML pages obtained by converting to HTML and splitting a document with `latexmlpost` and LaTeXML 0.8.5. I have noticed that the part of XSLT sheet LaTeXML-structure-xhtml.xsl handling dates has changed since LaTeXML 0.8.2, and I thought of reverting to the code from LaTeXML 0.8.2, but am not sure if it would cause some undesirable effects elsewhere. Thanks

brucemiller:
Indeed, the redundant dates wasn't intentional.  Your approach is basically correct, but I think (hope) this test is a more general way of excluding the dates.

fixes #1478

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
@@ -390,10 +390,12 @@
       <xsl:call-template name="authors">
         <xsl:with-param name="context" select="$context"/>
       </xsl:call-template>
-      <xsl:call-template name="dates">
-        <xsl:with-param name="context" select="$context"/>
-        <xsl:with-param name="dates" select="../ltx:date"/>
-      </xsl:call-template>
+      <xsl:if test="not(//ltx:navigation/ltx:ref[@rel='up'])">
+        <xsl:call-template name="dates">
+          <xsl:with-param name="context" select="$context"/>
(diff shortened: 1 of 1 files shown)


dginev (approved):
looks good 👍🏻

tkw1536 (approved):
LGTM

evoludolab:
The patch doesn't seem to work as expected for the following minimal example for `date.tex`:
```
\documentclass{book}
\begin{document}
\chapter{My chapter 1}
Contents of chapter 1.

\chapter{My chapter 2}
Contents of chapter 2.
\end{document}
```
With the current head (0b073bad2a93f44ded01520631a8174b800a2a79) of LaTeXML this still inserts the date after each chapter when I convert to xhtml using the following command:
```
latexml --dest=date.xml date.tex ; latexmlpost --dest=date.xhtml --splitat=chapter date.xml
```
and produces `Ch1.xhtml` that renders like
date

The possibly less sophisticated fix proposed in #1478 works as expected. However, maybe my expectations or workflow are messed up 😄. Any clarifications are most appreciated!

xworld21:
I confirm @evoludolab's behaviour, chapters do not (always?) get a `rel="up"`. But I see `rel="start"`, is that more reliable?

Issue #1216, opened 2019-11-21
Recent ImageMagick policies induce it broken by default for ghostscript


dginev:
As can be seen at:
https://stackoverflow.com/a/53180170

And has been experienced by me in all of Authorea, CorTeX conversions of arXiv, and recently working on generic instructions for a latexml-enabled host machine - the modern policy directives of ImageMagick render it quietly disabled on all ghostscript formats. This is very easy to forget when working on the same physical machine for a longer period of time, and "gets me every time" when I setup a new machine that will require image conversions.

The equivalent versioned file for `/etc/ImageMagick-6/policy.xml` needs to be modified from the recent default:
```xml
<!-- disable ghostscript format types -->
  <policy domain="coder" rights="none" pattern="PS" />
  <policy domain="coder" rights="none" pattern="EPS" />
  <policy domain="coder" rights="none" pattern="PDF" />
  <policy domain="coder" rights="none" pattern="XPS" />
```

to the latexml-friendly:
```xml
  <policy domain="coder" rights="read|write" pattern="PS" />
  <policy domain="coder" rights="read|write" pattern="EPS" />
  <policy domain="coder" rights="read|write" pattern="PDF" />
  <policy domain="coder" rights="read|write" pattern="XPS" />
```

An unavoidable side-effect is that this (to an extent) welcomes the security vulnerabilities reported by https://imagetragick.com/ .

I am unsure what the optimal resolution is here - latexml is already "agnostic" to the image processor, and does not require an imagemagick installation, but it is very easy for an end user (even for a co-developer!) to arrive at a setup that "looks good", but is decidedly broken and needs troubleshooting.

Recording here, since I've hit this one too many times now.

brucemiller:
Wow, lovely.  So the instructions to set up latexml that everybody wants to be sooooo easy, will now include editing XML in /etc.   Somebody shoot the messenger, please!

brucemiller:
Is the issue what may potentially be *in* the image file (I would expect less issue from pdf?), or simply that these types cause a delegation via system to ghostview?

dginev:
No worries, the messenger also wanted to shoot himself when he realized this is the third time he's debugging the same devops issue.

The TL;DR from image tragick is:
> There are multiple vulnerabilities in ImageMagick, a package commonly used by web services to process images. One of the vulnerabilities can lead to remote code execution (RCE) if you process user submitted images. The exploit for this vulnerability is being used in the wild.

[...]

Issue #1213, opened 2019-11-05
Modernize and streamline documentation flows


dginev:
One of the reasons I haven't personally "spurred into action" on the documentation issues (#896 , #1073 ) front is  that I do not necessarily feel comfortable with latexml's current documentation infrastructure, in particular when it comes to "technical appendixes" - be it describing schemas or APIs. Bruce and I recently discussed the need for general improvement on that front, and I have been doing some thinking. 

This issue is not meant to bash on the manual's or perldoc's organization, rather to provide some positive examples+ideas on what would be an upgrade in usability, both in contributing as well as consuming/learning from the docs.

As a best practice example, I have been enjoying the API documentation of mathpix recently:
https://docs.mathpix.com

They have a simple/elegant form factor, generated via the open source slate generator:
https://github.com/slatedocs/slate

That approach uses a responsive three-column documentation layout:
 - the left ToC panel allows for intuitive navigation for new learners of the system
 - the central panel contains the actual technical spec, often using tables for itemizing the parameter details
 - the right column allows for adding examples of usage alongside, with an open-ended set of named tabs for e.g. different programming languages

The main upgrade from the manual being that they keep the examples tightly connected to the documentation for each API route. Having specific examples near the definitions greatly enhances the usability of both API and schema appendixes. E.g. consider having "TeX", "XML" and "HTML" tabs for the schema elements, which could be helpful not only for the schema, but also for binding macros (which can also have a "State" tab that describes side-effects).

I also appreciate that the manual [dogfoods](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) latexml itself, which is a cool commitment. I think it is quite doable to provide a slate HTML dialect for latexml if we chose to - oriented towards technical appendixes - while keeping the TeX syntax for the input. I am definitely not intending to argue for Markdown here.

As to deployment, I wonder if it is possible to keep the latest master's documentation mirrored in a `gh-pages` branch on github, while having the main deployment (https://dlmf.nist.gov/LaTeXML/) lag behind to e.g. major releases. It would both encourage third party contributions to the docs - as they can be previewed on github - while also providing a backup deployment in case one of the two domains experiences downtime. The web manual is still a simple set of static HTML+CSS assets, which keeps things very achievable.

Welcoming any and all comments here, but I think this or similar upgrade could really take latexml's usability to a higher level.

dginev:
[...]

PR #1603, opened 2021-08-09
Manual update

### LINKED ISSUE #1073: Document the LaTeXML Daemon at top-level of LaTeXML manual
The LaTeXML manual only talks about `latexml` and `latexmlpost` in Section/Chapter 2 "Using LaTeXML". This should really have a top-level section about the motivation and practicalities of `latexmls` and `latexmlc`. Nothing like this seems to exist at the moment.

### LINKED ISSUE #1402: latexmlpost --format=xml does not seem to convert <Math> to MathML
The documentation for the `--format=xml` option says:

> For the default, xml, the output is left in LaTeXML’s internal xml, but the math is parsed and converted to presentation MathML.

However this doesn't appear to happen in fact.

For example:

    % cat try.tex
    \documentclass{book}
    \title{A book}
    \author{By me}
    \begin{document}
    \maketitle

    Hello, here is an equation: $E=mc^2$.
    \end{document}
    % latexml --dest=try.lxml try.tex
    latexml (LaTeXML version 0.8.4)
    ...
    % latexmlpost --format=xml --destination=try.xmlout try.lxml
    latexmlpost (LaTeXML version 0.8.4)
    ...
    % grep Math try.xmlout
    <p>Hello, here is an equation: <Math mode="inline" tex="E=mc^{2}" text="E = m * c ^ 2" xml:id="p1.m1" fragid="p1.m1">
        <XMath>
        </XMath>
      </Math>.</p>
    %

Using `--format=xhtml` does work as expected, so I don't think this is pointing to some grievous breakage in my installation (most recent version from macports).

My use-case here is that I (probably) want to use my own stylesheets to convert from the LaTeXML internal format to XHTML, but with the conversion to MathML done for me (the reason for that is that the conversion with `--format=xhtml` does 95% of what I want – LaTeXML is a wonderful piece of work! – but the last 5% is dissolving into a mess of special cases, which are messy enough that I suspect it'd be easier overall if I start my post-post-processing a little earlier in the toolchain).

### LINKED ISSUE #1416: docs: build from source
https://dlmf.nist.gov/LaTeXML/get.html could usefully link to 
https://lists.informatik.uni-erlangen.de/pipermail/latexml/2018-May/002368.html
in the Prerequisites section.

brucemiller:
An update to the Manual ( and site).

First patches, just for smooth building and stability.

fixes #1073
fixes #1402
fixes #1416

### DIFF
--- a/doc/site/index.tex
+++ b/doc/site/index.tex
@@ -3,6 +3,7 @@
 \usepackage{hyperref}
 %\usepackage{../sty/latexmldoc}
 \usepackage{latexmldoc}
+\input{latexmlreleases}
 \usepackage{listings}
 % Should the additional keywords be indexed?
 \lstdefinestyle{shell}{language=bash,escapechar=@,
@@ -12,7 +13,6 @@
    morekeywords={latexml,latexmlpost,latexmlmath},
    moredelim=[is][\itshape]{\%}{\%}}
 
-\input{releases.tex}
 \newcommand{\PDFIcon}{\includegraphics{pdf}}
 
 \title{\LaTeXML\ \emph{A \LaTeX\ to XML/HTML/MathML Converter}}
@@ -33,7 +33,7 @@
 \maketitle
 
 %============================================================
-\emph{Now available}:  \htmlref{\LaTeXML\ \CurrentVersion}{get}!
+\emph{Now available}:  \htmlref{\LaTeXML\ \CurrentRelease}{get}!
 
 In the process of developing the
 \href{https://dlmf.nist.gov/}{Digital Library of Mathematical Functions},
@@ -182,7 +182,7 @@ If you want to dig deeper:
 
 \subsection{Quick Start}\label{get.quick}
 \subsubsection{Current Release}\label{get.current}
-The current release is \textbf{\CurrentVersion}. (see the \href{Changes}{Change Log}).
+The current release is \textbf{\CurrentDownload}. (see the \href{Changes}{Change Log}).
 \emph{Prebuilt packages sometimes lag behind the latest release.}
 
 Most people will want to install a prebuilt release,
@@ -317,7 +317,7 @@ and unpacked:
 unzip LaTeXML-master.zip
 cd LaTeXML-master
 \end{lstlisting}
-\subparagraph{or, a tarball} can be downloaded from \CurrentTarball, and unpacked:
+\subparagraph{or, a tarball} can be downloaded from \CurrentDownload, and unpacked:
 \begin{lstlisting}[style=shell]
 tar zxvf LaTeXML-@\CurrentVersion@.tar.gz
 cd LaTeXML-@\CurrentVersion@
@@ -649,9 +649,10 @@ which inhibited processing of pdf files; the problem appears to be fixed recent 
 If you see \texttt{operation not allowed by the security policy}, see the discussion
 at \href{https://stackoverflow.com/a/53180170}{stackoverflow}.
 
+\expandafter\ifx\expandafter#\AllReleases#\else
 \subsection{Archived Releases:}\label{get.archive}
 \AllReleases.
-
+\fi
 %============================================================
 \section{Contacts \& Support}\label{contact}
 
--- a/doc/site/index.tex
+++ b/doc/site/index.tex
@@ -542,7 +542,7 @@ You may want to try \href{https://www.imagemagick.org/}{ImageMagick}, but
 you're on your own, there!  You may  have better luck with \texttt{Graphics::Magick}.
 
 \paragraph*{Installi \emph{only} prerequisites  under Strawberry,}\label{get.windows.strawberry.prereq}
-except for \TeX, and ImageMagic (see \ref{get.windows.imagemagick}),
+except for \TeX, and ImageMagick (see \ref{get.windows.imagemagick}),
 using:
 \begin{lstlisting}[style=shell]
 cpan Image::Size Parse::RecDescent UUID::Tiny
@@ -570,13 +570,13 @@ $Env:LATEXML_KPSEWHICH_MIKTEX_ADMIN=true ; $Env:LATEXML_TEST_TEXLIVE="0"; cpan l
 \subsection{CPAN installation}\label{get.cpan}
 The following command will install \LaTeXML\ and its Perl prerequisites,
 but you may need to pre-install \texttt{libxml2} and  \texttt{libxslt} (See \ref{get.prerequisites}),
-as well as optionally \TeX\, ImageMagic, if desired.
+as well as optionally \TeX\, ImageMagick, if desired.
 \begin{lstlisting}[style=shell]
 cpan LaTeXML
 \end{lstlisting}
 
 \paragraph*{Install \emph{only}  prerequisites}\label{get.cpan.prereq}
-(excluding  \texttt{libxml2},  \texttt{libxslt}, \TeX, ImageMagic)
+(excluding  \texttt{libxml2},  \texttt{libxslt}, \TeX, ImageMagick)
 using:
 \begin{lstlisting}[style=shell]
 cpan Archive::Zip DB_File File::Which Getopt::Long Image::Size \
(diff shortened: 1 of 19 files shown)


dginev:
Could you also include your recommended command for building the manual locally? That way I can reproduce over here and check the rendered variants as well.

brucemiller:
Using `doc/makesite` will generally make the site pages, as well as the manual pdf and pages; a few options to enable/disable.  You'll want `--webdir=/someplace` to tell it where to put the web pages.

tkw1536:
I'm unable to run makesite on a blank directory:
```
$ mkdir /Users/twiesing/tmp/empty-webdir
$ ./makesite --webdir=/Users/twiesing/tmp/empty-webdir
==================================================
Generating site pages
Couldn't read directory /Users/twiesing/tmp/empty-webdir/releases: No such file or directory at ./makesite line 113.
 at ./makesite line 113.
        main::getReleases() called at ./makesite line 55
```

brucemiller:
more updates;

fixes #1216
fixes #1073 
fixes #1402 
fixes #1416 

And nibbles at the edges of #1213

dginev:
Why/how does this PR fix #1216 ? I thought the only way out of that was some system-level change from the OS, or us using something different altogether.

brucemiller:
As I understood it, #1216 was about documenting how to install 
dependencies when you want to run the development version of LaTeXML.
(But then I didn't understand where mktexlsr came in, so perhaps
I don't understand exactly which situation the linked "recipe" solves)

dginev:
> As I understood it, #1216 was about documenting how to install 
> dependencies when you want to run the development version of LaTeXML.
> (But then I didn't understand where mktexlsr came in, so perhaps
> I don't understand exactly which situation the linked "recipe" solves)
> 

Well that issue is rather annoying for lay users to the point where it is unclear if documenting how to mod a file under /etc really "solves" the usability problem. I'd rather keep it open until we find something a bit more satisfying...

brucemiller:
Oh! Too many issues; I thought you were referring to 1416, not 1216.

Well, the solution there is simply to use newer versions of ImageMagic 
and/or Ghostscript. The documentation change just points that out.

brucemiller:
OK, the last commit should make it a bit more friendly for casual use.

brucemiller:
This PR addresses some documentation issues, but isn't a massive rewrite. I think in the next go round, a separate set of howto documents, as suggested in #1213 would be a good idea (although details of how to write/build it are tbd); this would also simplify the manual.

dginev:
> Well, the solution there is simply to use newer versions of ImageMagic 
> and/or Ghostscript. The documentation change just points that out.

I don't follow. I thought I saw the problematic ghostscript security policy on the latest Ubuntu release as well? Could you point me to the part of the PR explaining the details?

doc/site/index.tex
```diff
@@ -541,67 +592,78 @@ \subsection{CPANM Installation}\label{get.cpanm}
...
-
-% Possibly, eventually, want to expose:
-%   https://www.mathweb.org/wiki/????
-% But, it doesn't have anything in it yet.
+\paragraph*{ImageMagick Security} is also an issue, due to the possibility of embadding
+malicious code in postscript, pdf and other image formats.
+For a certain range of versions of ImageMagic and Ghostscript, a security policy was set
+which inhibited processing of pdf files; the problem appears to be fixed recent versions.
+If you see \texttt{operation not allowed by the security policy}, see the discussion
+at \href{https://stackoverflow.com/a/53180170}{stackoverflow}.
```
brucemiller, line 650: @dginev This is where I address the issue, and suggest using the most recent versions, if you can. I also point to the stack overflow discussion where you can figure out if you need to do something special and how to do it.  It is way out of scope here to get into explaining how to modify the policy files of random OS's, however.

dginev: Ok, so the security problem in ghostscript is fixed in recent versions, but the actual usability problem remains - latexml is broken by default for image conversion in ubuntu/debian systems (at the least).

All I ask is that we keep the issue in question open until the default state of image conversion on those OSes is "working", however we manage to achieve that.

brucemiller:
Two solutions are given: update to a not broken by default system, or 
follow the linked instructions.

dginev:
> Two solutions are given: update to a not broken by default system

To my understanding the latest versions of debian, ubuntu and archlinux all have the security  policy in question that makes image conversion broken by default. So you cant update to a working state from multiple distros. I have not checked others, but to my perspective closing the image conversion issue with this PR is pretty much a "worksforme" resolution.

brucemiller:
Image processing by LaTeXML is clearly documented as optional; And, as 
clearly documented, it *REQUIRES* a *WORKING* ImageMagick or 
Graphics::Magick.

If that makes LaTeXML "broken by default" on systems that are "broken by 
default", well, so be it. If you can't upgrade your "broken by default" 
system to a not "broken by default" system, you can modify
the "broken by default" security policy, as outlined in the linked 
stackoverflow.

Writing our own version of ImageMagick and postscript/pdf interpreter is 
not practical, certainly not "best practice", no other workaround within 
LaTeXML seems feasible. Solutions have been clearly documented in the 
installation guide and in a to-be-closed, but findable, Issue. So 
there's no point in keeping an issue open indefinitely waiting for 
action by some third party. That's not "worksforme".

Done.

Any constructive comments on the PR?

doc/site/index.tex
```diff
@@ -464,37 +541,13 @@ \subsubsection{Windows using Strawberry Perl}\label{get.windows.strawberry}
 You may want to try \href{https://www.imagemagick.org/}{ImageMagick}, but
 you're on your own, there!  You may  have better luck with \texttt{Graphics::Magick}.
 
-\paragraph*{Installing \emph{only} prerequisites  under Strawberry}\label{get.windows.strawberry.prereq}\\
-Install \href{https://strawberryperl.com}{Strawberry Perl}, and
-the \TeX-system of your choice (if desired), 
-ImageMagick (if you want to try; see \ref{get.windows.imagemagick}),
-and then install the additional prerequisites as
+\paragraph*{Installi \emph{only} prerequisites  under Strawberry,}\label{get.windows.strawberry.prereq}
+except for \TeX, and ImageMagic (see \ref{get.windows.imagemagick}),
```
dginev, line 545: a bunch of `ImageMagic` typos for `ImageMagick` with a trailing `k`, probably better to grep than for me to comment on each one.

lib/LaTeXML/Common/Model/RelaxNG.pm
```diff
@@ -551,6 +551,7 @@ sub cleanTeX {
   $string =~ s/\#/\\#/g;
   $string =~ s/<([^>]*)>/\\texttt{$1}/g;    # An apparent convention <sometext> == ttfont?
   $string =~ s/_/\\_/g;
+  $string =~ s/\*/any/g;                    # ????
```
dginev, line 554: unsure about this comment, what is a sample input that has the star?

lib/LaTeXML/Common/XML.pm
```diff
@@ -328,3 +328,103 @@ BEGIN {
...
+
+Returns the list of values for the given attributes on C<$node>
+
+=item C<initialize_catalogs()>
+
+Initialize XML::LibXML to recognize the catalogs given in LaTeXML.catalogs.
+
+=item C<set_RDFa_prefixes($document,$map)>
+
+This needs an explanation.
```
dginev, line 416: Maybe:
"This method scans the document's RDFa attributes, extracting the prefixes used. These prefixes are then filtered through a `$map` of known RDFa prefixes and the ones allowed are declared globally for the document via the "prefix" attribute of its root element."

dginev (changes requested):
Great PR overall!

I spotted some minor edits, good to merge right away apart from those.

lib/LaTeXML/Common/Model/RelaxNG.pm
brucemiller: it's for `rawhtml` which can take *any* xhtml.

doc/site/index.tex
brucemiller: fixed.

lib/LaTeXML/Common/XML.pm
brucemiller: done; thanks!

PR #1604, opened 2021-08-09
add space before keywords

xworld21:
One character patch adding one missing space in a stylesheet between the colon and the following keywords. I grepped the other `<xsl:text>:` and this seems to be the only offender re spaces.


dginev (approved):
Sounds good to me, thanks!

brucemiller:
Looks quite reasonable; Thanks!

Issue #1348, opened 2020-10-06
Problems with \nameref* 


henrikt-ma:
First, try this basic use of `\nameref*`:
```
\documentclass{article}
\usepackage{hyperref}

\begin{document}
\section{Strictly proper system}\label{proper}
Next: \nameref*{biproper}

\section{Biproper system}\label{biproper}
Next: \nameref*{improper}

\section{Improper system}\label{improper}
Back: \nameref*{proper}
\end{document}
```
In this case, latexmlpost reports (and the result contains things like _LABEL:*biproper_):
```
Warning:expected:ids Missing items:
	  Target for Label: LABEL:*
```

Then, put a bit of math in the titles:
```
\documentclass{article}
\usepackage{hyperref}

\begin{document}
\section{Strictly proper system: $n > p$}\label{proper}
Next: \nameref{biproper}

\section{Biproper system: $n = p$}\label{biproper}
Next: \nameref{improper}

\section{Improper system: $n < p$}\label{improper}
Back: \nameref{proper}
\end{document}
```

Then, there's a fatal failure in latexmlpost:
```
(CrossRef mini.html processing...
Fatal:perl:die Perl died
	bad ns attribute! at /opt/local/lib/perl5/vendor_perl/5.28/darwin-thread-multi-2level/XML/LibXML.pm line 1687.
	In Post::Document[mini.html] ->addNodes
	 <= Post::CrossRef[@0x7fda46aae6a0]
1 fatal error
```

While working on this bug report, I also ran into segfaults.

Only tested on version 0.8.4.

henrikt-ma:
I thought it was time I should also start testing issues on _master_, and this just gave me bad output for the case of math in the section titles (no fatal failure).

dginev:
@henrikt-ma do you have an option to test your examples with the latest latexml from this git repository? I retested your reported examples today, and they work without any issues. Here is the second example in HTML:

![Screenshot 2021-08-09 at 18-12-24 1 Strictly proper system n p](https://user-images.githubusercontent.com/348975/128780975-990f6784-0672-4399-a05f-71e756a2585d.png)

dginev:
Clarification: the warning from the first example is still there, we ought to get that patched. Second example looks healthy.

henrikt-ma:
I'm trying to reproduce your results, but I'm afraid that I'm not setting up the _master_ version correctly, as this is what I get (three broken links):
```
<article class="ltx_document">
[...]

Issue #1350, opened 2020-10-07
Getting more than just the name when using \nameref


henrikt-ma:
Try this:
```
\documentclass{article}
\usepackage{hyperref}

\begin{document}
\section{Strictly proper system}\label{proper}
Next: \nameref{biproper}

\section{Biproper system}\label{biproper}
Next: \nameref{improper}

\section{Improper system}\label{improper}
Back: \nameref{proper}
\end{document}
```

In the pdfLaTeX output, the `\nameref` is only expanded to the actual name of the linked section, but LaTeXML produces things like _2 Biproper system_.

Using _cleveref.sty_ and _amsthm.sty_, the problem gets worse:
```
\documentclass{article}
\usepackage{hyperref}
\usepackage{amsthm}
\usepackage{cleveref}

\newtheorem{definition}{Definition}[section]
\crefname{definition}{definition}{definitions}
\Crefname{definition}{Definition}{Definitions}

\begin{document}
\section{Strictly proper system}\label{proper}
Next: \nameref{biproper}

\begin{definition}[Biproper system]\label{biproper}
Next: \nameref{improper}
\end{definition}

\section{Improper system}\label{improper}
Back: \nameref{proper}
\end{document}
```

Here, instead of the expected _Biproper system_, the `\nameref` produces _Definition 1.1 (Biproper system)_.

dginev:
Right, I just independently ran into this and commented on it in #1609 . The extra numbers are spans with a `ltx_tag_ref` class, which we need to briefly discuss with @brucemiller and resolve in either the nameref binding or CrossRef (or both).

But chances are that we can close this issue in tandem with the nameref* report - they're very related.

The examples are great and easy to reproduce, thanks @henrikt-ma !

PR #1609, opened 2021-08-10
Star variants of nameref and autoref

dginev:
Fixes https://github.com/brucemiller/LaTeXML/issues/1348

We were missing the star variants of `\nameref*` and `\autoref*`. This PR does the easy bit of recognizing the argument.

The hard bit is yet to be implemented. It is to avoid "double markup" when the macros are used in nested links. That is helpfully illustrated via:
```tex
\newcommand*{\fullref}[1]{\hyperref[{#1}]{\autoref*{#1} \nameref*{#1}}}
```

as I found out from [SE](https://tex.stackexchange.com/a/121871)


dginev:
The problem with the actual feature is that the timing is a bit shifted in latexml and the fill-in information arrives in post-processing. Core processing would deposit the following for the \fullref example:
```xml
<ref labelref="LABEL:proper">
  <ref class="ltx_refmacro_autoref" labelref="LABEL:proper" show="autoref"/> 
   <ref class="ltx_refmacro_nameref" labelref="LABEL:proper" show="title"/>
</ref>
```

Which can't really be "optimized" at the binding stages, even `afterConstruct` is too early. Instead, it may be that the CrossRef post-processor needs to be on the lookout for nested refs, and support a simple unwrap of the inner ones. Some variant of that is already there actually, since the HTML is respectively:

```html
<a href="#S1" title="1 Strictly proper system" class="ltx_ref">
  <span class="ltx_ref ltx_refmacro_autoref">
    <span class="ltx_text ltx_ref_tag">section 1</span>
  </span> 
  <span class="ltx_ref ltx_refmacro_nameref">
    <span class="ltx_text ltx_ref_title">
      <span class="ltx_tag ltx_tag_ref">1 </span>
      Strictly proper system
    </span>
  </span>
</a>
```

which is almost what the TeX intuition seems to be. The **goal** HTML is likely the simplest:
```html
<a href="#S1" title="1 Strictly proper system" class="ltx_ref">
  section 1 Strictly proper system
</a>
```

or something in between. It may be best to not add any of the spans at all, to avoid strange interplay between CSS rules that didn't expect to be used jointly... Thoughts?

dginev:
Also, strange, I see both `ltx_ref_tag` and `ltx_tag_ref` classes generated. That can't be good/intentional? CrossRef.pm mentions the `ltx_ref_tag` variant, but that's about all I can find.

brucemiller:
So, the `*` simply means don't make it a ref?  We could actually respect that by adding an attribute (or even just a class `ltx_noref`) when the star is present. Then `CrossRef` can do all the usual filling-in, but omit the link ? .... Or perhaps, it's better in the XSLT to convert an `ltx:ref` into a `span` instead of `a`?

dginev:
Well,  `*` does not make it a link in the PDF, that's indeed about all it does.

What exactly the term `ref` means here is a bit slippery to me - clearly it's still a reference to the label it is pointing at, just not a clickable one.

As to `ltx_noref`, sure, maybe `ltx_nolink` is clearer though.

And turning those into spans seems consistent with the PDF, which just deposits plain text in those cases. I'll take a look...

brucemiller:
Oh, and while you're at it can you check if #1350 is covered?

PR #1610, opened 2021-08-10
Fix Typo: Whatit -> Whatsit

tkw1536:


dginev:
I was half-wondering if you renamed `whatsit` to `whatchamacallit`

tkw1536:
I should totally do that for April 1st.

brucemiller:
How about `$waitwhat` ?

PR #1590, opened 2021-07-30
New Testing Docker Images

tkw1536:
This PR moves to a new set of docker images with texlive / perl combinations for testing.
These have been cleanly rebuilt from scratch and are more uniform across different versions. 
It also updates the list of versions being tested to include TeXLive 2021.

PR #1612, opened 2021-08-13
Add Windows CI tests

tkw1536:
This PR adds CI to test LaTeXML on Windows.

It includes the following versions of Perl and TeX:
- Strawberry Perl 5.32.1.1
- MiKTeX 21.6.28 / Strawberry Perl 5.32.1.1
- Strawberry Perl 5.30.3.1
- Strawberry Perl 5.28.2.1
- Strawberry Perl 5.26


tkw1536:
This PR should probably be rebased after #1590 is merged.

tkw1536:
I just realized that becauase of #1214, in particular the issue:

> ```
> #   Failed test 't/daemon/formats/citation'
> #   at C:\Users\Vedran\.cpanm\work\1572992197.7568\LaTeXML-0.8.4\blib\lib/LaTeXML/Util/Test.pm line 93.
> # Difference at line 1 for t/daemon/formats/citation
> #       got : 'kpsewhich: warning: running with administrator privileges'
> #  expected : 'No obvious problems'
> ```
we might have to undo/redo the last commit in this PR. Probably should rename the flag to `KPSEWHICH_MIKTEX_ADMIN` and document it.

dginev:
> we might have to undo/redo the last commit in this PR. Probably should rename the flag to KPSEWHICH_MIKTEX_ADMIN and document it.

Theoretical question: do you think Makefile.PL can do a sample run of say `kpsewhich article.cls`, then check if it got a warning on STDERR, and if so -- **set the miktex flag**, so that by the time we're doing `make` and `make test`, all the guards are properly available?

dginev:
I initially thought this miktex mishap would only ever be a problem in CI, but judging by #1214 real users are experiencing it on their completely ordinary Windows installations. Hence the need to actually check at setup, rather than have a CI-only flag.

tkw1536:
As for the MiKTeX workaround, there isn't a simple solution. For Windows:
- If you're running with admin rights, you need the `--miktex-admin` flag to kpsewhich
- If you're not runnning as an admin, having that flag will harm you, and kpsewhich will complain even more

Maybe the right approach for `make test` is to check:
1. If the installed TeX is MikTeX
2. The tests are being run as an administrator

The former might be detectable in `tex --version`, similarly to how we already check for texlive versions. The latter - I have no idea. 

Yet another workaround might be to tell people to switch the `AutoAdmin` flag of MikTeX on:
```
initexmf --admin --set-config-value [Core]AutoAdmin=t
```

This entire MiKTeX issue is extremely annoying - running kpsewhich as an admin should not cause any errors because by it's very nature it is a read-only program. I am not sure how much effort we should spend on fixing the issue on our end (modulo already existing workarounds) and how much we should delegate to the MiKTeX maintainers.

dginev:
> I am not sure how much effort we should spend on fixing the issue on our end

We need exactly the amount of effort to have working CI and working default installations on Windows (with miktex). And no further effort if possible :>

tkw1536:
For that to work I still think the cleanest approach is to surpress standard error output of kpsewhich during the tests.

xworld21:
> For that to work I still think the cleanest approach is to surpress standard error output of kpsewhich during the tests.

@tkw1536 is this actually giving you problems in the CI? The latest master ignores the standard error of the tests. On my Windows VM, I can see kpsewhich printing out warnings (because of running as administrator, or not doing the updates, etc)  but I still get the PASS at the end.

tkw1536:
It's not giving test failures, just throwing a lot of warnings at me. The current workaround is to manually patch the kpsewhich executable with something that suppresses the miktex errors.

dginev:
in other words, problems of aesthetics in CI are still CI problems. If we ever encounter an error, reading it out through the myriad of kpsewhich warnings will be quite painful, and unnecessarily so.

xworld21:
> in other words, problems of aesthetics in CI are still CI problems. If we ever encounter an error, reading it out through the myriad of kpsewhich warnings will be quite painful, and unnecessarily so.

Ah, ok, I get it. I thought there were some remaining installation bugs on Windows, and I couldn't reproduce.

brucemiller:
There's a conflict, and a mention of a possible need to rebase after #1590 was merged (which it has been)

tkw1536:
I've rebased on top of the latest master, I didn't see any conflict when doing so. 
Once the tests have run though this should be good to merge.

dginev:
22 minutes is an impressive number for CI. And even more impressively, all looks passing and readable, thanks Tom!

image

dginev:
Actually, I just realized this is a miktex setup that doesn't run the tests requiring texlive? Not much of a point of installing miktex if we won't run the expl3 tests at least.

tkw1536:
> Actually, I just realized this is a miktex setup that doesn't run the tests requiring texlive? Not much of a point of installing miktex if we won't run the t/83_expl3.t tests at least.

That's because it's installing  the [basic](https://github.com/brucemiller/LaTeXML/pull/1612/files#diff-5ce5c9ff86f58b1f87b35b5227b2e84cb69f022d6741e1854f3e1e181091773cR86) set of packages. See also [the chocolatry package documentation](https://community.chocolatey.org/packages/miktex). 

I can try doing the complete set, no idea how long that takes to install, or how stable the download is.

tkw1536:
ok, done. This might fail horribly.

brucemiller:
On 8/24/21 11:27 PM, Tom Wiesing wrote:
> ok, done. This might fail horribly.

If it does, backing off is fine; I don't know that we
need to exhaustively test every combination of OS, texlive,
perl, etc.  A good mix of most pairings seems reasonable.

tkw1536:
It hasn't failed - but running with a full MiKTeX does take about an hour to run. I'd fell much better just testing the basic set of packages.

dginev:
bad news: it took 58 minutes
great news: 
```
t/83_expl3.t .............. ok
```

Looks good to merge! 

I don't know if we'll ever have enough cache quota to avoid spending the 58 minutes multiple times, but seeing the full test suite pass on windows CI is a great feeling.

brucemiller:
Nice that it passes! (and would be good to know it in the future). OTOH, 
1hr is long.  I'll leave it to you two to balance the reassurance 
against future CI "costs"; I'm good either way :>

PR #1613, opened 2021-08-13
Fix #1605: Allow \bibliography to include '.bib'

### LINKED ISSUE #1605: \bibliography fails if the argument includes '.bib'
BibTeX is indifferent whether one uses `\bibliography{References}` or `\bibliography{References.bib}`, but LaTeXML is more sensitive about it:
```
Error:missing_file:References.bib.bib Couldn't find Bibliography 'References.bib.bib' at Post::MakeBibliography[@0x7f984fb4b108]
Info:expected:References.bib.bib Couldn't find usable Bibliography for 'References.bib.bib' at Post::MakeBibliography[@0x7f984fb4b108]
```
Offending line:
https://github.com/brucemiller/LaTeXML/blob/145e8799ec69097dd3356f516fb918149c557ede/lib/LaTeXML/Post/MakeBibliography.pm#L110

tkw1536:
Fixes #1605

Previously, when the argument to `\bibliography` ended in `.bib` MakeBibliography would fail to find the requested bibliography file.
This was cause by it adding an extra '.bib' extension, even though there already was one.

This commit updates the behavior to only add a '.bib' extension if not already present.

### DIFF
--- a/lib/LaTeXML/Post/MakeBibliography.pm
+++ b/lib/LaTeXML/Post/MakeBibliography.pm
@@ -99,6 +99,7 @@ sub normalizeBibKey {
 sub getBibliographies {
   my ($self, $doc) = @_;
   my @bibnames = ();
+  my $fromBibliography = 0; # coming from an 'ltx:bibliography'
   # use the commandline bibliographies, if explicitly given.
   if ($$self{bibliographies} && scalar(@{ $$self{bibliographies} })) {
     @bibnames = @{ $$self{bibliographies} }; }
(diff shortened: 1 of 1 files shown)


tkw1536:
This PR takes the approach of fixing the issue within latexmlpost.
An alternative approach would be to strip the `.bib` suffix (if it exists) before it even ends up in the non-postprocessed xml. That would probably be around [`\lx@bibliography` in LaTeX.pool](https://github.com/brucemiller/LaTeXML/blob/4a34c9184b8ec3c095d9783473ab043ba0739ae2/lib/LaTeXML/Package/LaTeX.pool.ltxml#L3719). 

Pros of the approach in this PR is backward-compatibility with any non-postprocessed xml file already generated.
Pros of the alternate approach would be that the xml would technically be more semantic as it would contain the resolved bibliography name. 

Not sure which approach is better - comments @dginev @brucemiller ?

tkw1536:
For future compatibility with biber, I will update this so that we can only add the extension if the file does not exist.

tkw1536:
Ready to be merged.

brucemiller:
Slightly weird, but I guess it has to be :>

PR #1614, opened 2021-08-13
Add Exit Codes to 'latexmlpost' and 'latexmlmath' (Fixes #1417)

### LINKED ISSUE #1417: Add latexmlpost exit codes
Executing `latexmlpost` from a shell script always returns 0, which prevents one from handling failures. This can be trivially resolved by exiting with a status code at the end of `latexmlpost`.

```
diff --git a/bin/latexmlpost b/bin/latexmlpost
index 20eda2fa..7294466c 100755
--- a/bin/latexmlpost
+++ b/bin/latexmlpost
@@ -461,6 +461,7 @@ print STDERR $@ if $@;
 my $code = $latexmlpost->getStatusCode;
 print STDERR "\nPostprocessing " . ($code == 3 ? 'failed' : 'complete') . ": " . $latexmlpost->getStatusMessage . "\n";
 print STDERR "processing " . ($code == 3 ? 'abandoned' : 'finished') . " " . localtime() . "\n" if $verbosity > -1;
+exit $code;

 # helpers
```

tkw1536:
Fixes #1417 

This PR adds exit codes to the 'latexmlpost' and 'latexmlmath' executables, as suggested in https://github.com/brucemiller/LaTeXML/issues/1417#issuecomment-751760914.

The first commit uses a patch suggested by @epifanovsky.

### DIFF
--- a/bin/latexmlmath
+++ b/bin/latexmlmath
@@ -240,12 +240,14 @@ if ($xmath) {
   outputXML($result->findnode('//ltx:XMath'), $xmath); }
 
 my $status = $latexml->getStatusMessage;
+my $code = $latexml->getStatusCode;
 # Should be combined with $post's status!
 # But better approach will be to manage all through LaTeXML.pm!!
 my $runtime = RunTime($starttime);
 NoteLog("Conversion complete: " . $status);
 NoteSTDERR("Conversion complete: " . $status . " (reqd. $runtime)");
 UseLog(undef);
+exit $code;
 
 #======================================================================
 # Helpers
--- a/bin/latexmlmath
+++ b/bin/latexmlmath
@@ -247,7 +247,7 @@ my $runtime = RunTime($starttime);
 NoteLog("Conversion complete: " . $status);
 NoteSTDERR("Conversion complete: " . $status . " (reqd. $runtime)");
 UseLog(undef);
-exit $code;
+exit $code == 3 ? 1 : 0;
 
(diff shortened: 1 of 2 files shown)


bin/latexmlmath
```diff
@@ -240,12 +240,14 @@ if ($xmath) {
...
 
 my $status = $latexml->getStatusMessage;
+my $code = $latexml->getStatusCode;
 # Should be combined with $post's status!
 # But better approach will be to manage all through LaTeXML.pm!!
 my $runtime = RunTime($starttime);
 NoteLog("Conversion complete: " . $status);
 NoteSTDERR("Conversion complete: " . $status . " (reqd. $runtime)");
 UseLog(undef);
+exit $code;
```
dginev, line 250: ah, but that's not the same convention. If I remember correctly the exit code has to be zero in all cases except when the run hits a Fatal. `getStatusCode` returns our own internal levels (0,1,2,3) for (ok, warning, error, fatal). So if you want to use that, you'd likely exit with:

```perl
exit $code == 3 ? 1 : 0;
```

dginev (changes requested):

bin/latexmlmath
tkw1536: Done, fixed both `latexmlmath` and `latexmlpost`

dginev (approved):

brucemiller:
good idea!

Issue #1549, opened 2021-06-13
Supporting arXiv:2105.04026 (ltablex, shadethm, adjustbox)


dginev:
I liked the format of #1536 , so here's another issue in the same style. This [mathematical overview](https://arxiv.org/abs/2105.04026) made the rounds online, and latexml is actually surprisingly close to a correct conversion when using `--includestyles`:

```
Conversion complete: 30 warnings; 5 errors; 5 undefined macros[\csq@thequote@oopen, \csq@thequote@oinit, \setlongtables, \on@line, \csq@thequote@oclose]
```

From what I can see all content is present, and even the SVG figures are converted neatly. There are some metadata errors due to a handful of packages in use that are not yet supported:
- [ltablex.sty](https://ctan.org/tex-archive/macros/latex/contrib/ltablex?lang=en)
- [shadethm.sty](https://ctan.org/tex-archive/macros/latex/contrib/shadethm?lang=en)
- [adjustbox.sty](https://ctan.org/tex-archive/macros/latex/contrib/adjustbox?lang=en) and its subrodinate trimclip.sty

That doesn't imply the complete binding for these packages is easy to achieve, but partial support upto this article ought to be. And maybe this is a friendlier issue format than opening an individual issue for each missing package, given the number of issues. Feedback welcome!

brucemiller:
`ltablex` looks conceptually easy, as it combines the features of `longtable` and `tabularx` (which are both implemented). `shadethm`'s documentation claims that it is obsolete. Neither are that common on arXiv, though they are in this one :>

PR #1616, opened 2021-08-18
arxiv fixes from css sprint

dginev:
A handful of smaller fixes during my CSS prototyping sessions with two books and ~10 articles from the arXiv corpus:

 - an experiment supporting the "psfile" key in `\special` to get graphics into some of the older docs
   - [astro-ph0001188](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/no_problem) is one example
 - slightly cleaner warning text for generic "perl warning" and "redefinition" messages
 - using `$stomach->digest` instead of `Digest` in a couple of obvious places
 - `\inputlineno` register and using it via the `\on@line` internal, 
 - some `\listi` internals for amsbook
 - a stub for `\setlongtables` in etoolbox
 - some internal skips for amsthm
 
There are still a couple of errors in the big books I was testing with, but already some improvement. The heavier lifting in #1549 can be done in the next release.


brucemiller:
Lots of little helpful bits!

PR #1617, opened 2021-08-18
Guard existing macros from \DeclareMathAlphabet

dginev:
I was seeing some rather broken tokenization examples in `\mathrm` that lead to bad rendering in HTML. To reproduce an example, you can try:

```tex
\documentclass{article}
\usepackage{fouriernc}

\begin{document}

$\mathrm{Tribes}_{w,s} $

\end{document}
```

converted with `--includestyles`. The `fouriernc.sty` file was being loaded in an arXiv job as it was requested from within a different `.sty` file, and thus being allowed in raw. Turns out, that package has the definition:
```tex
\DeclareMathAlphabet{\mathrm}{T1}{fnc}{m}{n}
```

which derailed latexml's tokenization so that each letter in the example became its own `XMTok`. I am not sure what the ultimately best upgrade is in this regard, but for now we can guard existing macros with a simple `IsDefined` check, hence this PR.

Edit: to be precise, the bad spacing isn't from the single letter tokenization itself, but the follow-up insertion of invisible time `<mo>` elements after.


dginev:
Compare:

 broken  | fixed 
:----: | :----:
image | Screenshot 2021-08-18 at 11-41-56 Untitled Document

brucemiller:
Looks right. Thanks!

PR #1621, opened 2021-08-18
[arXiv] recover abstract order in raw frontmatter uses

dginev:
As I make the pages better styled, some long-standing bugs get more and more thorny in my heels.

I think it's high time we started grappling with the combinations of "raw" tex content mixed with latex structure frontmatter macros seen on arXiv. As a running example, take [hep-ph0001194](https://corpora.mathweb.org/preview/arxmliv/tex%5Fto%5Fhtml/hep-ph0001194), which I also studied in #1620.

Put simply, the abstract appears first in the document, above the title and author list. This is due to using `{abstract}` while doing the rest of the frontmatter via a raw `{center}` environment.

As the comments in LaTeX.pool aptly elucidate, `{abstract}` is a great handle for a lot of the possible markup mishaps/abuse in the document frontmatter. So I added a new `afterConstruct` clause that revisits the constructed XML and for now does a single check:
 * is the current insertion point a centered logical paragraph?
 * if so, and an abstract is found preceding, move the abstract element after the paragraph.

That patch fixes this specific document, as that was the only bit out of order. But down the line, further metadata fishing can be done in this clause, as all "raw" frontmatter nodes would be available - title, authors, affiliations, submission venue, classification, keywords, etc. I think this is as good a place as any to position the fishing expedition, as the article is usually the last bit before the start of the mainmatter document body.


brucemiller:
This is scary code; will it make sense tomorrow? ... or work; ragged right title?

All made complicated by the misuse by authors & coders about abstract. Default latex puts abstract where it is, ams (I think) includes it with `\maketitle`; authors use `\abstract{}` or `\begin{abstract}`, etc.

It seems it would be more "logical" to figure out when to place it where it is, omitting it from the frontmatter list.  But the timing is off between digestion & insertion.

Hmm...

dginev:
Right, it can be used in quite a variety of ways. Just to flesh out the example I tested fully as one use cases, here is the sample tex:


<details>

```tex
\begin{center}
  {\Large \bf The triple Higgs self-coupling at future $e^+e^-$
    colliders: a signal-to-background study for the standard
    model\footnote{Talk given at the Fourth Workshop of the {\it ``2nd
        Joint ECFA/DESY Study on Physics and Detectors for a Linear
        Electron-Positron
        Collider,''} Oxford, UK, 20-23 March 1999.}}\\[0.4cm]
  D.J. MILLER$^{1}$ and S.~MORETTI$^{2}$\\[0.4 cm]
  {\it 1) Deutches Elektronen-Synchrotron DESY, D-22603 Hamburg, Germany}\\
  {\it 2) Rutherford Appleton Laboratory, Chilton, Didcot, Oxon OX11 0QX, UK.}\\
\end{center}
\begin{abstract}
{\small
\noindent
The experimental reconstruction of the Higgs self-energy potential is
essential to a verification of the Higgs boson's r\^ole in spontaneous
electroweak symmetry breaking. The first step towards this goal, the
measurement of the triple Higgs self-coupling, can possibly be
accomplished at the next generation of linear colliders.  Here we
present a background study of the most promising channel, double
Higgs-strahlung off a $Z$ boson, $e^+e^-\to HHZ$, with the subsequent
decay $H\ar b\bar b$, and evaluate the feasibility of its measurement.}
\end{abstract}
```

</details>


There is no `\maketitle` in the document, so indeed the abstract was intended to be deposited at the exact location, rather than to be flushed up to the frontmatter elements.

In the spirit of some recent PRs, another suggestion would be to try and emulate `{abstract}` a little more natively, defining several individual variants of it, and making sure each class uses the one it was intended for. So less magic, more tedium...

PR #1622, opened 2021-08-19
no XSLT date delimiter

### LINKED ISSUE #1607: too many colons in the date when using AMS classes and a few others
The following
```latex
\documentclass{amsart}
\title{MWE}
\date{9 August 2021}
\begin{document}
  \maketitle
\end{document}
```
becomes
image
because `\datename` contains the colon already. This is with all AMS classes and a few more obscure ones (elegantbook, imsproc...).

The trivial fix is to remove the colon from the `ltx:date` template. However, that will get rid of the colon from all the other classes! Unfortunately the use of `\datename` does not seem completely consistent – some classes even include a trailing space already.

dginev:
Fixes #1607

One can play around with the issue by defining:
```tex
\def\datename{\textit{Date}:}
```

with and without a colon. I think carrying that as-is to the XML `ltx:date` element is correct, and then we can make a provision for it in the XSLT with a simple if. Since I don't think we support using colons inside the date itself, a simple conditional can do the job.

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl
@@ -673,7 +673,7 @@
 
   <xsl:template match="ltx:date" mode="intitle">
     <xsl:param name="context"/>
-    <xsl:if test="@name"><xsl:value-of select="@name"/><xsl:text>: </xsl:text></xsl:if>
+    <xsl:if test="@name"><xsl:value-of select="@name"/><xsl:text> </xsl:text></xsl:if>
     <xsl:apply-templates select="node()">
       <xsl:with-param name="context" select="$context"/>
     </xsl:apply-templates>


xworld21:
Just a comment: adding text and punctuation with XSLT is fragile. At least, it does not cope well with different languages. E.g. in French the colon should have an extra space. That kind of problem worries me a bit.

dginev:
@xworld21 we don't have _any_ internationalization solution for the HTML pipeline at the moment. What else would be needed for French? And is that something you're interested in the short term?

The current approach basically makes it necessary - if the changes are large - to branch into a `...-french.xsl` dialect of the stylesheets.

xworld21:
> @xworld21 we don't have _any_ internationalization solution for the HTML pipeline at the moment. What else would be needed for French? And is that something you're interested in the short term?

Ah, I have no particular need. Internationalising sounds tricky (which makes me think, the few pieces of text like the LaTeXML logo might do well with an extra `lang` and/or `xml:lang` attribute, but I am going off-topic).

I grepped quickly through texlive, and I reckon that almost no class adds a colon after the date, they rely on `\datename` providing one if desired. `facsimile.cls` does add a colon, and I couldn't figure out what koma-script really does. Is the extra colon really necessary? (Disclaimer: I really did a quick grep so I might be wrong here.)

brucemiller:
Honestly, I think you're both giving me too much credit, re: that I had actually thought it all through. Probably the formatting with ":' came first seeming like the Right Thing, and then later I discovered `\datename`, (and it's many friends), and just didn't realize that the class author would be supplying the ":" if they wanted it.

So, I suspect that the right thing to do is simply remove the ":" from the `xsl:text` and it should yield what the author expected.

dginev:
@brucemiller fair enough. If that's how it is, here is a 1 character PR dropping the colon instead.

brucemiller:
While CI is running, note that we actually do have bits & pieces of internationalization. To the extent that we actually read the style files, many packages do set various name macros according to options & language, and we use some of them.

PR #1623, opened 2021-08-19
pgfplotstable.sty: add binding allowing raw interpretation

### LINKED ISSUE #1304: Add support/a binding for pgfplotstable.sty
cc @bfirsh 

Unsure if this is a general issue, or if it is specific to `pgfplotstable`. I have two LaTeX packages imported in the same line in the preamble:

```
\usepackage{pgfplots,pgfplotstable}
```

which returns the error for a tikz figure in the main body of the LaTeX when using arxivvanity/engrafo 

```
Error:latex:\GenericError Package pgfplotstable Error: Please load \usepackage{pgfplotstable} before using \pgfplotstablecreatecol
	at main.tex; line 759 col 0 - line 759 col 64
	See the pgfplotstable package documentation for explanation.
	In Core::Definition::Primitive[\GenericE... /usr/local/share/perl/5.30.0/LaTeXML/Package/LaTeX.pool.ltxml; line 5070
	 <= Core::Stomach[@0x55635732e570] <= Core::Definition::Primitive[Begin] <= Core::Stomach[@0x55635732e570] <= ...
```

Can you confirm if multiple packages can be listed in the same command? If not, could this be implemented?

JG

dginev:
Fixes #1304

latexml can now interpret this style raw.

We just need to avoid a generic warning for backwards compatibility (on my texlive 2020) and we get good outcomes, as described in the issue.

### DIFF
--- a/lib/LaTeXML/Package/pgfplotstable.sty.ltxml
+++ b/lib/LaTeXML/Package/pgfplotstable.sty.ltxml
@@ -18,7 +18,10 @@ use LaTeXML::Package;
 #**********************************************************************
 InputDefinitions('pgfplotstable', type => 'sty', noltxml => 1);
 # Avoid generic warnings:
-RawTeX('\pgfplotsset{compat=1.17}');
+my $compat_cs  = T_CS('\pgfk@/pgfplots/compat/current');
+my $compat_val = IsDefined($compat_cs) && ToString(Expand($compat_cs));
+if (!$compat_val || $compat_val eq 'default') {
+  RawTeX('\pgfplotsset{compat=1.17}'); }
 
 #**********************************************************************
 
--- a/lib/LaTeXML/Package/pgfplotstable.sty.ltxml
+++ b/lib/LaTeXML/Package/pgfplotstable.sty.ltxml
@@ -1,0 +1,25 @@
+# -*- CPERL -*-
+# /=====================================================================\ #
+# |  pgfplotstable.sty                                                  | #
+# | Implementation for LaTeXML                                          | #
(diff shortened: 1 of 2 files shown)


brucemiller:
Whoops, dup! (and my commit is embedded within a lot of others)

lib/LaTeXML/Package/pgfplotstable.sty.ltxml
```diff
@@ -0,0 +1,25 @@
...
+# \=========================================================ooo==U==ooo=/ #
+package LaTeXML::Package::Pool;
+use strict;
+use warnings;
+use LaTeXML::Package;
+
+#**********************************************************************
+InputDefinitions('pgfplotstable', type => 'sty', noltxml => 1);
+# Avoid generic warnings:
+RawTeX('\pgfplotsset{compat=1.17}');
```
brucemiller, line 21: is this safe, even if the author sets it himself (her)?

dginev: Assuming the author sets this *after* their `\usepackage{pgfplotstable}`, it should be a clean override. But I can certainly add a check that only sets it if undefined, one minute...

dginev: a bit of a rabbit hole to find the right macro and figure out its default value (... literally `default`), but I can now claim to have tested this and it only sets the compat mode if it wasn't set before.

dginev:
This PR has found a new home as part of the very scary, yet very successful, #1624 . So closing here 👍🏻

PR #1628, opened 2021-08-21
[arXiv] more frontmatter guards

dginev:
Searching for more trade-offs, this PR introduces:
* lock \maketitle;
* sanitize ltx:personname for punctuation abuse
* allow optional arg for \author and lock
* keep locked definitions locked if they get extended (via bindings) by the `AddToMacro` utility 

---

1. Locking \maketitle avoids doubling the frontmatter when a custom `.sty` file redefines it, as in [cs0010030](https://corpora.mathweb.org/preview/arxmliv/tex%5Fto%5Fhtml/cs0010030)

2. Similarly, locking `\author` avoids issues where an otherwise correct author entry does not make it to the frontmatter infrastructure. But, at least one redefinition introduces an optional argument (example I saw: adds a note marker). So I made provisions to accept and drop that.

3. Lastly, some papers redefining `\author` abused it somewhat, where they started the intermediate ones with a leading comma, as in `\author{ , FirstName  LastName }`. And since such `\W` chars don't belong at the ends of person names, I added a sanitization clause to the `ltx:personname` tag. I did it on the tag since there are multiple venues from where person names are created, and also multiple XSLT sheets that map those elements into other formats.

4. `AddToMacro` could be invoked on a locked definition, which I think is OK, since if one does it from a binding there is usually (always?) a good reason. But what the way it did it dropped the `locked` declaration upon extension. So making sure locked definitions remain locked (for raw tex) after getting extended via a binding.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -2294,7 +2294,7 @@ sub AddToMacro {
   else {
     DefMacroI($cs, undef, Tokens(map { $_->unlist }
           map { (blessed $_ ? $_ : TokenizeInternal($_)) } ($defn->getExpansion, @tokens)),
-      nopackParameters => 1, scope => 'global'); }
+      nopackParameters => 1, scope => 'global', locked => $$defn{locked}); }
   return; }
 
 #======================================================================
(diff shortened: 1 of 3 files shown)


brucemiller:
These seem helpful and safe

PR #1630, opened 2021-08-25
widths in math cleanup should be Glue-capable

### LINKED ISSUE #1618: MuGlue widths need care when used
Currently, a minimal snippet such as:
```tex
\documentclass{article}
\begin{document}
test $\; $ here 
\end{document}
```

produces a warning that shows we're missing a method:
```
Warning:perl:warn Argument "5.0mu plus 5.0mu" isn't numeric in division (/) at test.tex; line 3 col 8
```

I believe the MuGlue object for `\thickmuskip` falls back on Number's ptValue (via Glue->Dimension->Number). I'm not entirely clear what the correct sizing logic is with using `\;` on the internal side, so opening an issue instead of a PR. 

Likely Glue and MuGlue need dedicated sizing methods?

dginev:
Fixes #1618 

After being fooled around by Perl's object orientation and error reporting combo, I finally figured out there was a Glue serialized as a string and then recast as a Dimension which caused the warning.

1. Simply put, the width in an XMHint can be a Glue, so build that preemptively.

2. I also spotted that TeX.pool has the same registers defined twice, so compressed that.

3. Also added some extra spaces between the glue keywords in some skip definitions for extra cleanliness

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -1154,10 +1154,9 @@ sub today {
 #======================================================================
 # MuGlue registers; TeXBook p.274
 {
-  my %mparms = (
-    thinmuskip => '3mu', medmuskip => '4mu plus 2mu minus 4mu', thickmuskip => '5mu plus 5mu');
-  foreach my $p (keys %mparms) {
-    DefRegister("\\$p", Glue($mparms{$p})); }
+  DefRegister('\thinmuskip'  => MuGlue("3mu"));
+  DefRegister('\medmuskip'   => MuGlue("4mu plus 2mu minus 4mu"));
+  DefRegister('\thickmuskip' => MuGlue("5mu plus 5mu"));
 }
 #======================================================================
 # Token registers; TeXBook p.275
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -1154,10 +1154,9 @@ sub today {
 #======================================================================
 # MuGlue registers; TeXBook p.274
 {
-  my %mparms = (
-    thinmuskip => '3mu', medmuskip => '4mu plus 2mu minus 4mu', thickmuskip => '5mu plus 5mu');
-  foreach my $p (keys %mparms) {
-    DefRegister("\\$p", Glue($mparms{$p})); }
+  DefRegister('\thinmuskip'  => MuGlue("3mu"));
+  DefRegister('\medmuskip'   => MuGlue("4mu plus 2mu minus 4mu"));
+  DefRegister('\thickmuskip' => MuGlue("5mu plus 5mu"));
```
dginev, line 1159: I think the explicit form is easier to read -- and search for! So kept these, dropping the ones at lines 5426-5430

brucemiller:
Aha!

PR #1637, opened 2021-08-26
epub: allow all core media types; safer ids; portable href's

xworld21:
As promised, a quick improvement on #1635. In short:
- recognise all 'core media types' listed in the EPUB3.2 spec, not just the arbitrary css/png/svg (incidentally, the `.css` match was still case-sensitive);
- be excessively accurate with `@id`, down to encoding invalid characters as `-xNNN-` so that there can't be collisions or invalid id's, ever ~; files starting with a digit would have caused problems so we might as well do the right thing™; the regex is almost a copy and paste from the XML 1.0 spec (collision is still a risk, but a small one, and easy to fix on the user side)~
- use `URI::file` to make sure the `href`'s are actual URLs; things would have probably exploded on Windows
- fix a tiny bug that causes invalid output when using `--splitnaming=*relative`
- do not include the log in the output

Hopefully I'll get around to test this on Windows over the weekend, just to be sure.

I am still missing a couple of XSLT fixes, I'll get them to you soon.

### DIFF
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -13,6 +13,7 @@ package LaTeXML::Post::Manifest::Epub;
 use strict;
 use warnings;
 use File::Find qw(find);
+use URI::file;
 
 our $uuid_tiny_installed;
 
@@ -180,7 +181,7 @@ sub process {
       my $manifest = $$self{opf_manifest};
       my $item     = $manifest->addNewChild(undef, 'item');
       $item->setAttribute('id',         $file);
-      $item->setAttribute('href',       $relative_destination);
+      $item->setAttribute('href',       URI::file->new($relative_destination));
       $item->setAttribute('media-type', "application/xhtml+xml");
       my @properties;
       push @properties, 'mathml' if $doc->findnode('//*[local-name() = "math"]');
@@ -197,7 +198,7 @@ sub process {
       my $nav_map = $$self{nav_map};
       my $nav_li  = $nav_map->addNewChild(undef, 'li');
       my $nav_a   = $nav_li->addNewChild(undef, 'a');
-      $nav_a->setAttribute('href', $file);
+      $nav_a->setAttribute('href', URI::file->new($file));
       $nav_a->appendText($file); } }
   $self->finalize;
   return; }
@@ -233,7 +234,7 @@ sub finalize {
 
     my $file_item = $manifest->addNewChild(undef, 'item');
     $file_item->setAttribute('id',         $file_id);
-    $file_item->setAttribute('href',       $file);
+    $file_item->setAttribute('href',       URI::file->new($file));
     $file_item->setAttribute('media-type', $file_type); }
 
   # Write the content.opf file to disk
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -167,6 +167,16 @@ sub initialize {
   $$self{nav_map}       = $nav_map;
   return; }
 
+sub file_id {
+  my ($name) = @_;
+  # convert file name to valid NCName for use as id
+  # any invalid character, and -, are converted to -xN- where N is the hex codepoint
+  $name =~ s/([^A-Z_a-z\x{C0}-\x{D6}\x{D8}-\x{F6}\x{F8}-\x{2FF}\x{370}-\x{37D}\x{37F}-\x{1FFF}\x{200C}-\x{200D}\x{2070}-\x{218F}\x{2C00}-\x{2FEF}\x{3001}-\x{D7FF}\x{F900}-\x{FDCF}\x{FDF0}-\x{FFFD}\x{10000}-\x{EFFFF}.0-9\x{B7}\x{0300}-\x{036F}\x{203F}-\x{2040}])/-x${\(sprintf("%X", ord($1)))}-/g;
+  # ensure the starting char is valid and prevent collisions with the other id's
+  $name = 'file-' . $name;
+  return $name;
+}
+
 sub process {
   my ($self, @docs) = @_;
   $self->initialize($docs[0]);
@@ -180,7 +190,8 @@ sub process {
       # Add to manifest
       my $manifest = $$self{opf_manifest};
       my $item     = $manifest->addNewChild(undef, 'item');
-      $item->setAttribute('id',         $file);
+      my $item_id  = file_id($relative_destination);
+      $item->setAttribute('id',         $item_id);
       $item->setAttribute('href',       URI::file->new($relative_destination));
       $item->setAttribute('media-type', "application/xhtml+xml");
       my @properties;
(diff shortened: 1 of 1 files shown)


xworld21:
Oh, and I have made the opinionated decision **not** to include the log in the EPUB3 output. The log is written to `\jobname.latexml.log` (why not `latexmlc`?) where the source files are, so it does not need to be in the final product.

xworld21:
(Sorry for the forced push, I had the wrong gitconfig in place...)

lib/LaTeXML/Post/Manifest/Epub.pm
```diff
@@ -143,6 +167,16 @@ sub initialize {
...
   return; }
 
+sub file_id {
+  my ($name) = @_;
+  # convert file name to valid NCName for use as id
+  # any invalid character, and -, are converted to -xN- where N is the hex codepoint
+  $name =~ s/([^A-Z_a-z\x{C0}-\x{D6}\x{D8}-\x{F6}\x{F8}-\x{2FF}\x{370}-\x{37D}\x{37F}-\x{1FFF}\x{200C}-\x{200D}\x{2070}-\x{218F}\x{2C00}-\x{2FEF}\x{3001}-\x{D7FF}\x{F900}-\x{FDCF}\x{FDF0}-\x{FFFD}\x{10000}-\x{EFFFF}.0-9\x{B7}\x{0300}-\x{036F}\x{203F}-\x{2040}])/-x${\(sprintf("%X", ord($1)))}-/g;
+  # ensure the starting char is valid and prevent collisions with the other id's
+  $name = 'file-' . $name;
+  return $name;
```
dginev, line 177: I converted a large book from arXiv and took a look through the OPF manifest ids. Here are some curious entries:
```xml
<item id="file-images-x2F-GD-x2D-nosuccess-x2D-cont.png" href="images/GD-nosuccess-cont.png" media-type="image/png"/>
```

The href points to the correct file, but I wonder why the id got the `X2F` and `X2D` entries? They look like the hex codes for slash and dash. Well, the dash could have stayed as-is, and the X2F could have become the dash, keeping a slightly easier to read:
```xml
<item id="file-images-GD-nosuccess-cont.png" href="images/GD-nosuccess-cont.png" media-type="image/png"/>
```

which I think is closer to the state from the master branch.

Also, the leading `file-` prefix is an interesting choice, if it is to help avoid clashes, maybe a leading `opf-` instead is both more unique and slightly clearer? So my example may end up:

```xml
<item id="opf-images-GD-nosuccess-cont.png" href="images/GD-nosuccess-cont.png" media-type="image/png"/>
```

dginev (commented):
Almost ready to merge, some minor polish points from me. @xworld21 has the right pragmatic sense here and the PR is clearly putting the OPF generation on a better footing.

At some point during the next release(s) we may want to do more epub generation tests, also using the arXiv docs, and see if the validity and OPF manifests hold up well also on our end.

Most likely I have to hand off the last bit of review back to @brucemiller , as I'm traveling from tomorrow, but it looks like we're almost ready for a merge here.

lib/LaTeXML/Post/Manifest/Epub.pm
xworld21: > why the id got the `X2F` and `X2D` entries?

We would run a tiny risk of collision e.g. `math/0.svg` and `math-0.svg` get the same id (unlikely to happen, I know).

I can do some simplifications to get more readable entries (two lines of code): (1) encode `-` only when ambiguous (2) use `_` instead of `-` for the encoding (3) use the `URI::file` value to make Windows and Unix output the same id's [uh, (3) is a good idea regardless]. How does that sound @brucemiller ?

It should be noted that these id's are not visible because they are not used outside of the manifest (e.g. they are not anchors), so I'd prioritize avoiding collisions over readability.

> maybe a leading `opf-` instead is both more unique and slightly clearer

Actually, these id's are scoped to the OPF document, so `opf-` would be redundant. The other id's we contend with are `pub-id` (not a file), `nav` (...a file, actually). Here `file-` would just mean "actual file", or maybe "file generated before `Epub.pm`", does that make sense?

xworld21: I went for all (1)-(3), now the above example would become
```html
<item id="url-images_x2F_GD-nosuccess-cont.png" href="images/GD-nosuccess-cont.png" media-type="image/png"/>
```
and underscore would normally be preserved unless already in a sequence like `_x2F_`.

The code is slightly more complicated because of the lookahead in the regex.

lib/LaTeXML/Post/Manifest/Epub.pm
```diff
@@ -226,7 +226,7 @@ sub finalize {
         my $OPS_abspath  = $_;
         my $OPS_pathname = pathname_relative($OPS_abspath, $OPS_directory);
         my (undef, $name, $ext) = pathname_split($OPS_pathname);
-        if (-f $OPS_abspath && $ext ne 'xhtml' && "$name.$ext" ne 'LaTeXML.cache') {
+        if (-f $OPS_abspath && $ext ne 'xhtml' && "$name.$ext" ne 'LaTeXML.cache' && $OPS_abspath ne 'content.opf') {
```
xworld21, line 231: This fixes an unintended behaviour from #1639. So much for submitting a PR without testing every single case. Without this fix, if the manifest `content.opf` is supplied by the user, it ends up referenced within itself with type `application/octet-stream`. Not sure what consequences it would have, but better not to do that.

lib/LaTeXML/Post/Manifest/Epub.pm
dginev: I like `url-` a lot more than `file-`, no reason why we can't both avoid conflicts and have a naming convention that everyone enjoys :) Of course, unclear if Bruce enjoys the x2F (compared to e.g. "slash", or double underscore or...), but it's up to him to finish the review here. I'd probably opt for the most human-friendly variant.

Generally happy with a merge here since it cleans things up a bit, can be tweaked as people stress-test it against larger ePub projects.

brucemiller:
I guess I have to find a moment to test things, to see exactly where these ids are showing up. Generally, I'd prefer more recognizable ("meaningful") ids, labels, etc; if they're going to be noise, they ought to be more obviously noise (like Lisp's old (gensym) g13209x0184x).  Along with the preference (even if not always the discipline) for code reuse, rather than cargo or even worse, slightly modified cargo.

I'm inclined to pick up on @dginev 's suggestion of moving all the cleaners out of Package to a new `LaTeXML::Util::Cleaner`` module and fiddling the imports.  If I did that, would `CleanID` serve your purpose?

xworld21:
> `If I did that, would `CleanID` serve your purpose?

Very likely yes – what we need are just valid IDs. Those appear only in the `content.opf` file.

The only risk is generating the same ID for different files because `CleanID` is not one-to-one.

lib/LaTeXML/Post/Manifest/Epub.pm
brucemiller: OK, I finally took a look saw where the "ids" are coming in, and they're really only aliases to the `href` and purely internal to the manifest; barely ids at all, from my perspective :>   You could probably come up with something prettier, and even avoid conflicts, but with messier code... and really, who's ever going to see it except @dginev ? :>

So, yeah, it's ugly, but ok.

brucemiller (approved):

PR #1639, opened 2021-08-28
epub: do not overwrite files supplied by the user

xworld21:
This should be my last last-minute non-XSLT PR about EPUB3.

Should the user ever provide their own `nav.xhtml` or `content.opf` (by means of `RequireResource()`, so definitely not by accident), latexmlc should not overwrite them. So I have added two `if (-f ...)`'s around the relevant code and an Info message.

I do plan to use this to replace the default `nav.xhtml`, which is almost unusable with large documents. Once I figure out how to generate a good `nav` from within latexmlc (which may need a small refactoring), I'll send a PR, I promise!


dginev (approved):
PR looks good, and upgrading the nav long-term sounds good to, thanks!

PR #1640, opened 2021-08-28
Set pgfplots version to the most recent version;

brucemiller:
Avoid warnings by setting pgfplots to its most recent version, rather than a specific one. pgfplotstable doesn't need to set the version, since it loads pgfplots.


dginev (approved):
Nice upgrade, and nice guards! LGTM

PR #1642, opened 2021-08-29
drop dvisvgm pre-1.6 compatibility

xworld21:
dvisvgm 1.6 was released in 2014, is it time to drop the code around detecting the number of digits in the file names? The current code breaks on >99 images.

Alternatively I can add a few lines to detect say up to <10,000 images.


brucemiller:
Too bad we haven't counted the number of images expected beforehand (could be refactored to do so, but...).

Anyway; looks reasonable; Thanks!!

PR #1643, opened 2021-08-30
create proper EPUB3 navigation document

xworld21:
[Edited to match updates.]

This PR:
- removes the context from `<xhtml:title>`
- uses `<xhtml:title>` to populate the navigation document
- makes the navigation document nested
- remove navigation headers and footers from the pages (redundant, and quite an odd sight in EPUB)
- does not use `nav.xhtml` if already present: create `nav-1.xhtml` instead (or `nav-2.xhtml`, ...)

(I have also added a missing space in front of the separator, that was just a typo, right?)

### DIFF
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -15,6 +15,9 @@ use warnings;
 use File::Find qw(find);
 use URI::file;
 
+# ensure that we can use xhtml: in the xpath queries below
+$LaTeXML::Post::Document::XPATH->registerNS('xhtml' => 'http://www.w3.org/1999/xhtml');
+
 our $uuid_tiny_installed;
 
 BEGIN {
@@ -213,7 +216,8 @@ sub process {
       my $nav_li  = $nav_map->addNewChild(undef, 'li');
       my $nav_a   = $nav_li->addNewChild(undef, 'a');
       $nav_a->setAttribute('href', URI::file->new($relative_destination));
-      $nav_a->appendText($file); } }
+      map { $nav_a->appendChild($_->cloneNode(1)) }
+        $doc->findnodes('/xhtml:html/xhtml:head/xhtml:title/node()'); } }
   $self->finalize;
   return; }
 
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -248,7 +248,7 @@ sub finalize {
         my $OPS_abspath  = $_;
         my $OPS_pathname = pathname_relative($OPS_abspath, $OPS_directory);
         my (undef, $name, $ext) = pathname_split($OPS_pathname);
-        if (-f $OPS_abspath && $ext ne 'xhtml' && "$name.$ext" ne 'LaTeXML.cache' && $OPS_abspath ne 'content.opf') {
+        if (-f $OPS_abspath && $ext ne 'xhtml' && "$name.$ext" ne 'LaTeXML.cache' && $OPS_pathname ne 'content.opf') {
           push(@content, $OPS_pathname); }
       } }, $OPS_directory);
 
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -101,7 +101,8 @@ sub initialize {
 
   # Metadata
   my $rootentry         = $$self{db}->lookup('SITE_ROOT');
-  my $document_metadata = $$self{db}->lookup("ID:" . $rootentry->getValue('id'));
+  my $root_id           = $rootentry->getValue('id');
+  my $document_metadata = $$self{db}->lookup("ID:" . $root_id);
   my $document_title    = $document_metadata->getValue('title');
   $document_title = $document_title ? $document_title->textContent : 'No Title';
   my $document_authors = $document_metadata->getValue('authors') || [];
@@ -167,6 +168,7 @@ sub initialize {
   $$self{opf_spine}     = $spine;
   $$self{opf_manifest}  = $manifest;
   $$self{nav}           = $nav;
+  $$self{root_id}       = $root_id;
   $$self{nav_map}       = $nav_map;
   return; }
 
@@ -185,6 +187,9 @@ sub url_id {
 sub process {
   my ($self, @docs) = @_;
   $self->initialize($docs[0]);
+  my $root_id  = $$self{root_id};
+  my %nav_maps = ($root_id => $$self{nav_map});
+  my $nav_li   = undef;
   foreach my $doc (@docs) {
     # Add each document to the spine manifest
     if (my $destination = $doc->getDestination) {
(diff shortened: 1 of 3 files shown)


brucemiller:
OK, I'm playing a bit of catchup. Reminding myself what the nav document is, etc; yeah, having the filename as the link content doesn't seem helpful.

The reason you don't get the "benefits" of CrossRef or the DB is because the `nav.xhtml` is being created explicitly as html;  the DB (and CrossRef) deal with LaTeXML's XML which needs to be xslt'd to html. More for @dginev: is there a reason not to create, at an earlier stage, a `nav.xml` with an appropriately attributed `ltx:TOC`, and let `CrossRef` and `XSLT` do their thing?

Probably needs an extension here or there to deal with adding/detecting the `epub:toc` bit.  But ultimately, this would allow fully customized navigation document based on a `nav.tex`.

xworld21:
In the meanwhile, I found a good way of generating a *nested* navigation document (i.e. like a proper TOC, with sections nested below chapters and so on). [Commit here](https://github.com/xworld21/LaTeXML/commit/be4b4b02e835ce157618b36e5fed15b6de8232c1#diff-bf0620fc822ed17e033592838eb4f491770537c5549b07409f101b054a28a91a). So we can get a good navigation document from `Epub.pm` after all, without a new CrossRef/XSLT mechanism.

Are these changes (titles + nested toc ± `epub:type='toc'` detection) good enough for a last minute PR for 0.8.6, in case the CrossRef/XSLT takes too long to be fleshed out? They should be enough to make EPUB usable and probably solve most use cases.

brucemiller:
Hmm, well it seems a bit klunky (no offense; kinda necessary within the constraints), but it does improve the nav and is isolated to `ePub.pm` which could be upgraded transparently later. If @dginev is ok with it, we could merge this for now, and possibly improve it later.

xworld21:
Ok, in the interest of making everybody's life easier (I hope!), I tidied up the changes. Now this PR adds the titles _and_ creates a nested TOC. This is what most EPUBs should look like, so I will also shelve #1644 about `epub:type="toc"` – user customization clearly need some planning.

Note that I added some slightly opinionated commits, kept separate for ease of cherry picking. I removed headers and footers, except for the logo on the first page. If the user creates `nav.xhtml` (e.g. `latexmlc --dest=nav.epub`), LaTeXML will create a navigation document named `nav-1.xhtml`. Again because customization needs some thinking.

[Minimal example attached](https://github.com/brucemiller/LaTeXML/files/7111500/epubexample.zip), plus before/after screenshots from Play Books:
image image

dginev:
> More for @dginev: is there a reason not to create, at an earlier stage, a nav.xml with an appropriately attributed ltx:TOC, and let CrossRef and XSLT do their thing?

That would make sense. Also for the reason that we need a special sanitization step. I have seen some metadata sneaking in that erroneously contains footnotes/sidenotes, also in titles. There's also the open question of what exactly metadata fields should keep for math, as MathML has a poor chance of ever being catalog-friendly... All of which fit nicely in CrossRef's purview.

Now, as to assembling a `nav.xml`, that depends on whether we believe navigation entries are easier to create via XSLT or via perl. I'm somewhat impartial on that end. But having the right API to grab any metadata of interest from any post-processor *following* CrossRef -- that would be great.

xworld21:
> we believe navigation entries are easier to create via XSLT or via perl

As a user, I am in favour of XSLT-based solutions. If I want to change the navigation document, for instance insert `<li><span>Appendix</span></li>` before the backmatter, I can "just" pass `--stylesheet` or even better create a local `LaTeXML-epub3.xsl`. If you generate the document in perl, that's much harder.

dginev:
> As a user, I am in favour of XSLT-based solutions

I wish it was that simple. The other considerations that go in here are: what other formats are likely to get supported? Are there other post-processors that are similar to Manifest, and are there other manifests? Is this something to keep encouraging :> Do we want to enable those as plug-in modules, which allows some extra freedom without the constraints of the velocity and direction of the main LaTeXML project, etc.

And there is a very subjective toolchain "taste" that goes in when deciding between two turing-complete languages. Why invest in perl post-processors based around CrossRef's current implementation, rather than redo as much as possible in XMP and XSLT? Well, depends on who's working on the codebase, and the specific problems getting solved.

Then we get into a branch of discussions referred to as "plugins and profiles" where we have some unfinished plans about refactoring post-processing in a way where it is possible to easily and predictably add/swap in custom post-processors (as perl class files). Also from external repositories in the `LaTeXML::Plugin::` namespace.

So the more we pull here, the more strands we'll get in our hands, and the more carefully we should be making choices...

xworld21:
> The other considerations that go in here are: what other formats are likely to get supported?

Ah, I think I get it: this `nav.xml` is not being produced by `Epub.pm`, so it should be agonstic w.r.t. the requested format. If you try to produce it from `Epub.pm`, then it needs to travel back into the pipeline. So it doesn't fit right now.

What I am hoping for, is that whatever mechanism you adopt for customising postprocessing, it is just as easy to modify as it currently is for bindings and stylesheets. The fact that I can just drop `LaTeXML-html5.xsl` or some `sty.ltxml` in place is great.

dginev:
> What I am hoping for, is that whatever mechanism you adopt for customising postprocessing, it is just as easy to modify as it currently is for bindings and stylesheets.

We're completely on the same page here 👍🏻

brucemiller:
After the last release, this lay dormant... until we're gradually gearing up for the next release!  So, time to see if we can sort this out.

From digging around the code, I get the impression that the current navigation document was aimed purely at machine readability and only listed the page files, rather than any attempt at being pretty, accessible or human readable, let alone customizable.  It's written directly as xhtml.

To go beyond that, I'd think we want to let CrossRef do what it's good at: filling in a possibly heirarchic table of contents with titles, links and so on.  All that needs, as a start, is to prepend a simple document containing `<TOC ...>` that will be filled in. We'd have to make sure that it ends up satisfying the navigation doc criteria, `epub:type` attribute, etc.

We can omit adding the navigation page, if there seems to already be an appropriate TOC within the document; Question: if `\tableofcontents` were present (perhaps on separate page, used `ol` appropriately, etc), would that be a sufficient navigation doc?

If this approach is workable, it seems to fit the overall design much better; And if so, it would probably be best to frame it this way now, rather than pile on more complications that will be harder to disentangle later (if they need to be).

brucemiller:
Just to make sure I'm understanding what the navigation doc is supposed to do: Would it work, conceptually, if a source document had `\tableofcontents`?  (Provided of course that the TOC was converted into the correct `nav/ol/li`, with `@epub:type`, is pointed to by the spine, etc)

If so, that gives us a starting point for what I expect will be a more maintainable approach.  We can assure that epubs have such a TOC added, if not already present, etc.
Does that nav doc need to be split out as a separate page?

xworld21:
> (Provided of course that the TOC was converted into the correct `nav/ol/li`, with `@epub:type`, is pointed to by the spine, etc)

Yes, that sounds right. Any `<nav>`, wherever it is, can act as navigation document as long as it has the correct HTML structure and attributes. And as far as I remember, the navigation document can refer to fragments inside a file, not just the file itself, so really any TOC should work as long as it reaches every split file. So the machinery powering `--navigationtoc` should do the job.

> Does that nav doc need to be split out as a separate page?

Not at all, as long as it has the appropriate `@epub:type` etc., it can be part of another file and/or can be in the spine.

PR #1644, opened 2021-08-30
detect presence of toc nav in EPUB

xworld21:
This adds three commits on top of #1643. Technically, it is probably 5-6 new lines, but there is some reshuffling of the `nav.xhtml` code to get the right logic, so I'll keep it separate.

The change lets the user customise the navigation document in three ways, in order of priority:
1. supply the file `nav.xhtml` in some way (`RequireResource`, `--splitnaming=label`, ???) – merged in #1639
2. add `<nav epub:type="toc" ...>` in at least one the documents; the first one (in reading order) will be marked as navigation document
3. (default) let latexmlc generate `nav.xhtml` – hopefully with titles instead of filenames #1643

This PR adds 2, which is suggested by the spec itself: any document can do double duty as navigation document, see [§5.1, paragraph 3](https://www.w3.org/publishing/epub3/epub-packages.html#sec-package-nav). `<nav epub:type="toc" ...>` is exactly the tag that must appear in the navigation document, so the code just checks whether it exists or not.

I think this is the best we can do for EPUB without refactoring, so I'll stop the PRs now. Long term, I plan to send some XSLT that generates a valid `<nav epub:type="toc">` in the front page, so that 2 becomes as easy as calling `--navigationtoc=...`. I don't think I can improvise it, though.

PS: setting the destination to `index.xhtml` is necessary, otherwise `--dest=nav.epub` will create `nav.xhtml` and end up with a non-working EPUB!

### DIFF
--- a/lib/LaTeXML.pm
+++ b/lib/LaTeXML.pm
@@ -201,6 +201,8 @@ sub convert {
     $$opts{sitedirectory} = $sandbox_directory;
 
     if ($$opts{format} eq 'epub') {
+      # give main document a predictable name, avoid clash with nav.xhtml
+      $sandbox_destination = 'index.xhtml';
       $$opts{resource_directory} = File::Spec->catdir($sandbox_directory, 'OPS');
       $$opts{destination} = pathname_concat(File::Spec->catdir($sandbox_directory, 'OPS'), $sandbox_destination); }
     else {
@@ -609,7 +611,7 @@ sub convert_post {
     my $manifest_maker = LaTeXML::Post::Manifest->new(db => $DB, format => $format, log => $$opts{log}, %PostOPS);
     $manifest_maker->process(@postdocs); }
   # Archives: when a relative --log is requested, write to sandbox prior packing
-  #     TODO: This can be enhanced, as any path can be relativized. 
+  #     TODO: This can be enhanced, as any path can be relativized.
   #     TODO: ALSO, we now always have a log file, so maybe always add it as default?
   if ($$opts{log} && ($$opts{whatsout} =~ /^archive/) && (!pathname_is_absolute($$opts{log}))) {
     ### We can't rely on the ->getDestinationDirectory method, as Fatal post-processing jobs have UNDEF @postdocs !!!
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -154,6 +154,7 @@ sub initialize {
     my $nav_html = $opf->createElementNS("http://www.w3.org/1999/xhtml", 'html');
     $nav->setDocumentElement($nav_html);
     $nav_html->setNamespace("http://www.idpf.org/2007/ops", "epub", 0);
+    $nav_html->setAttribute('xml:lang', $document_language);
     my $nav_head  = $nav_html->addNewChild(undef, 'head');
     my $nav_title = $nav_head->addNewChild(undef, 'title');
     $nav_title->appendText($document_title);
--- a/lib/LaTeXML/Post/Manifest/Epub.pm
+++ b/lib/LaTeXML/Post/Manifest/Epub.pm
@@ -15,6 +15,9 @@ use warnings;
 use File::Find qw(find);
 use URI::file;
 
+# ensure that we can use xhtml: in the xpath queries below
+$LaTeXML::Post::Document::XPATH->registerNS('xhtml' => 'http://www.w3.org/1999/xhtml');
+
 our $uuid_tiny_installed;
 
 BEGIN {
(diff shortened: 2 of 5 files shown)


lib/LaTeXML.pm
```diff
@@ -201,6 +201,8 @@ sub convert {
     $$opts{sitedirectory} = $sandbox_directory;
 
     if ($$opts{format} eq 'epub') {
+      # give main document a predictable name, avoid clash with nav.xhtml
+      $sandbox_destination = 'index.xhtml';
```
dginev, line 205: that's ... a bit too standard? if the desired effect is to avoid a clash, why not check for a clash and avoid it? 

I'm unsure what to think about every single ePub generated by latexml having the same name for its main xhtml file.

lib/LaTeXML/Post/Manifest/Epub.pm
```diff
@@ -208,12 +222,25 @@ sub process {
...
-      my $nav_li  = $nav_map->addNewChild(undef, 'li');
-      my $nav_a   = $nav_li->addNewChild(undef, 'a');
-      $nav_a->setAttribute('href', URI::file->new($relative_destination));
-      $nav_a->appendText($file); } }
+      # Add to default navigation document
+      if (!$nav_xhtml && !$tocfound) {
+        my $nav_li = $nav_map->addNewChild(undef, 'li');
+        my $nav_a  = $nav_li->addNewChild(undef, 'a');
+        $nav_a->setAttribute('href', URI::file->new($relative_destination));
+        my ($title) = split(/ \x{2023} /, $doc->findnode('/xhtml:html/xhtml:head/xhtml:title')->textContent, 2);
```
dginev, line 230: i suspect this may be a bit more reliable to build from the CrossRef-provided metadata, but I'll let @brucemiller brainstorm that. If the `\x{2023}` delimiter, or its spacing, changes down the road, the regex will fail, which is a bit tricky.

dginev (commented):
Nice to see more and more getting fleshed out for epubs, thanks @xworld21 !

lib/LaTeXML/Post/Manifest/Epub.pm
xworld21: Where do I pull out the CrossRef data? I could only find a tag + text in the db, and it seemed like content supposed to go through XSLT. `<xhtml:title>` is already postprocessed, so it seemed like an easier option. That is to say, I couldn't figure out the correct source for the titles!

Talking about `<xhtml:title>`, we might as well simplify the tag in the EPUB3 stylesheet. The titles don't need to contain the document title and all the parent sections, we are not in a browser in this case.

(If the regex fails, `$title` will just be the full string, AFAIK.)

lib/LaTeXML.pm
xworld21: The split file names are also standardised, `Ch1.xhtml`, `Ch1.S1.xhtml` etc, I don't think it's too bad if the main page is always the same.

To be honest, I also did it because generating a custom TOC is virtually impossible because XSLT does not know the name of the file it's creating. It knows the others via `@href`, but not itself. I meant to bring this up at some point.

dginev: Great piece of information to add here, thanks!

brucemiller: The names are standard only if you use id-based naming (and haven't customized the id generation). Someone might use label-based names, for example (although that's more useful on a website than hidden within an ePub)

lib/LaTeXML/Post/Manifest/Epub.pm
brucemiller: CrossRef fills in an `ltx:TOC` with data, so it may be that it's already there, or could easily be, in some form.
I haven't yet looked at your PR to see exactly what you're after with "custom" TOC. 

Commenting a bit further back: my strategy has been that XSLT isn't creating files, but transforming them; all the paths should already be embedded somewhere as relative paths.

xworld21: Re `<xhtml:title>`, I have adopted a more sensible strategy (forced pushed also in #1643 because that part could be mergeable as it is). I simply removed the context from the HTML title (using a convenient parameter that was already there) and cloned the title into the navigation document. Having navigation entry == HTML document title is simple to reason about, I think, easy to customize for the user, etc.

brucemiller:
While I like the intention, I feel like this is risky for a last-minute merge. It is more likely (than #1643) to establish use patterns and clash with future changes, such as basing the nav on a `nav.xml` (or `.tex`) and leveraging the usual post processing chain.

xworld21:
As I just mentioned, #1643 creates a good navigation document out of the box. So an escape hatch for the user to provide a different navigation is not so urgent anymore.

PR #1647, opened 2021-09-02
Digested refinement

### LINKED ISSUE #1641: Math mode switch breaks on empty Digested args
Minimal example from [arXiv:1905.00240](https://corpora.mathweb.org/preview/sandbox%5Farxiv%5F1905/tex%5Fto%5Fhtml/1905.00240) : 
```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}

Here  the "$\underbrace{}$" indicates that the terms originate from
either $E^H$ or $E^{AD}$, and we have used the expression $\Delta A  =2D\mathcal{M}_1$.

\end{document}
```

From what I can make out, there is a subtle problem with how the `Digested` parameter type is implemented for empty arguments in math mode, which make the definition of `\underbrace` get confused on the mode switch:
```perl
DefMath('\underbrace Digested', "\x{23DF}", operator_role => 'UNDERACCENT',   # BOTTOM CURLY BRACKET
  scriptpos => 'mid');
```

I won't have time to patch this today, but it looks both healthy impact for arXiv (quick 100 errors and Fatal when encountered), and easy to fix? So recording an issue.

brucemiller:
This patch (partly) reverts recent changes to the `Digested` parameter type, but also corrects many misuses of that type to a normal `{}` argument.  And a handful of refinements discovered along the way (graphics box scaling; `\GenericError` output; a couple of swapped Unicode points).

fixes #1641


dginev:
Always delightful when we get parameter kinks ironed out!

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -5280,7 +5280,11 @@ sub make_message {
   my $type    = ToString(shift(@args)) || $cmd;
   $stomach->bgroup;
   Let('\protect', '\string');
+  # Note that the arg really should be digested to get to the underlying text,
+  # but why tempt fate, when we're already making an error message?
```
dginev, line 5284: I do agree with this sentiment

lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -36,43 +36,78 @@ DefParameterType('GraphixDimension', sub {
...
-      depth      => $d->multiply($yscale),
-      xscale     => $xscale,
-      yscale     => $yscale,
-      xtranslate => $w->multiply(($xscale - 1) / 2),
-      ytranslate => $h->add($d)->multiply(($yscale - 1) / 2)); },
+    graphics_scaledbox_props($box, $xscale, $yscale || $xscale); },
   mode => 'text');
 Let('\scalebox', '\Gscale@box');
 
-DefConstructor('\Gscale@@box OptionalMatch:* {GraphixDimension}{GraphixDimension} Digested', sub {
```
dginev, line 65: is the `OptionalMatch:*` variant intentionally not supported anymore?

t/fonts/mathaccents.tex
```diff
@@ -16,4 +16,35 @@
...
+ [\testmacro{ABC}] \quad
+ [\testmacro\nothing A] \quad
+ [\testmacro\relax A] \quad
+ [\testmacro{\verb+$+}] \quad
+ [\testmacro\bgroup A \egroup] \quad
+ [\testmacro\bgroup \verb+$+ \egroup] \]
+Note that this only accents the A
+\[ [\testmacro\abc] \]
+
+Other accent-like macros (underbrace, overbrace) only accept regular style arguments
```
dginev, line 38: kind of a over-protective question but, maybe nice to also have them tested stacked? namely the two:
```tex
\[ \overbrace{\underbrace{ARG}_{under}}^{over} \]
\[ \underbrace{\overbrace{ARG}^{over}}_{under} \]

```

dginev (commented):
Very cool overall! Left some pointed comments.

lib/LaTeXML/Package/graphics.sty.ltxml
brucemiller: Yeah, 'cause `\Gscale@@box` was never supposed to.

t/fonts/mathaccents.tex
brucemiller: I could add that, but it's kinda testing something entirely different. The added tests question what kind of material various cs (in LaTeX) can process as "arguments", and make sure that LaTeXML processes them in the same way.  Since over/underbrace actually turn out to only accept the standard  `#1` style argument (ie a single token or {} delimited), and your sample has the args in braces, it "has to" work.  Of course, something else could go wrong :>

dginev (approved):

lib/LaTeXML/Package/graphics.sty.ltxml
dginev: great, good to merge then!

t/fonts/mathaccents.tex
dginev: Alright, maybe its point is indeed too different for this PR. I tried a few other tests, one of which had a discrepancy with pdflatex, but again in an unrelated way:
```tex
\testmacro\noexpand\relax
```
which produces:
```
Error:expected:Token Missing argument Token for Core::Definition::Expandable[\noexpand Token] at ubrace.tex; line 4 col 24
```

This ran just fine in pdflatex and produced an underbrace over some empty space - for which I wonder how the width was determined. Leaving it to you what to do with this partial information :>

dginev: An extension to my previous note, if you do:
```tex
\[ [\expandafter\underbrace\underbrace\noexpand\relax] \]
```

pdflatex will produce two adjacent empty undebraces fenced in with the square brackets, without any errors. latexml instead emits a warning for a missing argument, and produces HTML with a `missing` word with two vertical underbraces under it. So this may be more directly related to the `{}` argument, even if my test is admittedly tortured.

brucemiller: The last patch deals with the basic `\noexpand\relax`, but I'm not even sure that the last cases are even fixable, given that LaTeXML tries to get "an argument" and process it, whereas TeX just kinda grinds on through as long as it can. That it puts an accent on nothing indicates that it's not a realistically working case; it's kinda surprising that it doesn't create an error.  Anyway: I'm punting (for now).

dginev: Thanks for the `\noexpand` bit. For the `\expandafter` bit I went to check the source, and indeed there is a difference - macros such as `\int`, `\eq`, `\neq`, `\mapsto`, `\cong`, `\longrightarrow`, `\cdots`, `\overbrace`, `\underbrace` ... in `latex/base/fontmath.ltx` are declared as protected, while our DefMath declarations do not have the protection added (for now?)

I think every argument-accepting math macro in `fontmath.ltx` ends up declared as protected.

Indeed, just quickly changing the definition to:
```perl
DefMath('\lx@math@underbrace {}', "\x{23DF}", operator_role => 'UNDERACCENT',     # BOTTOM CURLY BRACKET
  scriptpos => 'mid');
DefMacroI('\underbrace', undef, Tokens(T_CS('\protect'), T_CS('\lx@math@underbrace')));
```

leads to my last example correctly finding the letter `f` as the argument of underbrace, as well as the rest of the details that come up through pdflatex. So that may be the only fix needed to get all my examples working - though what the "Best" way to implement the protection is, I am unsure. Maybe we need a `protected => 1` flag for DefMath, which is turned on by default?

dginev: Also, feel free to merge the PR as-is, I can open an issue for this newly discovered discrepancy.

PR #1650, opened 2021-09-07
Robust prelims

### LINKED ISSUE #1649: Some math macros need to be robust
A follow-up to our recently discovered discrepancy for the argument reads of `\underbrace` in the [#1647 comments](https://github.com/brucemiller/LaTeXML/pull/1647#discussion_r702257568)

It appears in my texlive 2020 `fontmath.ltx` there are 46 definitions using `\DeclareRobustCommand`, among which the `\overbrace` and `\underbrace` examples. In the protected cases, some low-level programmatic uses become a bit subtle, as in the example:

```tex
\[ \expandafter\underbrace\underbrace foo \]
```

where the outer underbrace receives no arguments, while the inner one receives the letter `f`. Without the `\protect` machinery, latexml interprets the example as two nested underbraces over a missing argument, which may end up problematic.

It may be worthwhile adding a `protected => 1` flag to the DefMath API, and maybe even make all DefMaths protected by default? The default irrespective, going through the `.ltx` definitions and marking all protected macros as such in the latexml definitions would be a good upgrade to stay closer to pdflatex parity.

brucemiller:
This PR adds a new `robust` option to `DefMacro`, `DefMath` and several others, as well as extending `protected` to `DefMath`.  Whereas protected models eTeX's `\protected` prefix,  the `robust` option emulates LaTeX's Robust mechanism, which sneaks in a `\protect`.  Well, it's actually a bit more complicated than that and the current PR only does a simplistic emulation. This will be fleshed out and tested further in the future. However, this patch still does help.

fixes #1649


lib/LaTeXML/Package.pm
```diff
@@ -1033,7 +1041,8 @@ sub DefMacroI {
     $STATE->assignMathcode($cs => 0x8000, $options{scope}); }
   $cs        = coerceCS($cs);
   $paramlist = parseParameters($paramlist, $cs) if defined $paramlist && !ref $paramlist;
-  $STATE->installDefinition(LaTeXML::Core::Definition::Expandable->new($cs, $paramlist, $expansion, %options),
+  my $defcs = ($options{robust} ? defRobustCS($cs, $options{scope}) : $cs);
+  $STATE->installDefinition(LaTeXML::Core::Definition::Expandable->new($defcs, $paramlist, $expansion, %options),
     $options{scope});
   AssignValue(ToString($cs) . ":locked" => 1, 'global') if $options{locked};
```
dginev, line 1047: should we also lock `$defcs` here - when available - or is that too much safety for no benefit?

dginev (approved):
Makes sense, nice start!

brucemiller:
> should we also lock |$defcs| here - when available - or is that too much 
> safety for no benefit?

I worry that too much safety sometimes is unsafe

brucemiller:
on second thought, locking only the munged cs is probably pretty pointless.

dginev:
> on second thought, locking only the munged cs is probably pretty pointless.

well, one way or another, PR is good to merge in my book :> Thanks!

PR #1652, opened 2021-09-11
Initial beamer.cls binding

tkw1536:
This is a first attempt at a binding for `beamer.cls` I've been working on over the past month. 
Lots of stuff isn't supported, but the basic use cases produce something sensible.


brucemiller (changes requested):
An impressive piece of work!

MANIFEST
```diff
@@ -388,6 +388,9 @@ lib/LaTeXML/Package/babel.sty.ltxml
 lib/LaTeXML/Package/balance.sty.ltxml
 lib/LaTeXML/Package/bbm.sty.ltxml
 lib/LaTeXML/Package/bbold.sty.ltxml
+lib/LaTeXML/Package/beamer.cls.ltxml
+lib/LaTeXML/Package/beamer_support_preload.sty.ltxml
+lib/LaTeXML/Package/beamer_support_postload.sty.ltxml
```
brucemiller, line 393: Although `beamer.cls.ltxml` would be a bit on the large side, it would be better to combine these three into a single binding.

MANIFEST
```diff
@@ -740,6 +743,7 @@ t/80_complex.t
 t/81_babel.t
 t/82_moderncv.t
 t/83_expl3.t
+t/84_beamer.t
```
brucemiller, line 746: Better to make a test and subdirectory named, perhaps, `slides` and include the new tests, along with a test case for classic LaTeX slides. (I think the schema change will cause a problem with slides)

MANIFEST
```diff
@@ -857,6 +861,24 @@ t/babel/numprints.xml
...
+t/beamer/frames.xml
+t/beamer/fsop.pdf
+t/beamer/fsop.tex
+t/beamer/fsop.xml
+t/beamer/overlayenv.pdf
+t/beamer/overlayenv.tex
+t/beamer/overlayenv.xml
+t/beamer/titles.pdf
+t/beamer/titles.tex
+t/beamer/titles.xml
```
brucemiller, line 881: Too many tests! Good to be thorough, but for long term maintenance it is more convenient to combine into a single (or few) tests. So long as they are compatible as far as options or declarations are concerned, it is nice to work in a single test case through progressively more exotic features.

lib/LaTeXML/Package/beamer.cls.ltxml
```diff
@@ -0,0 +1,150 @@
...
+RelaxNGSchema("LaTeXML");
+# RequireResource('ltx-article.css');
+
+# ignore all the styling
+DefMacro('\setbeamertemplate{}[] Optional {}', undef);
+DefMacro('\setbeamercolor{}{}', undef);
+DefMacro('\setbeamerfont{}{}', undef);
+
+#**********************************************************************
+# Core sectioning support, stolen from article.cls!
```
brucemiller, line 105: Why steal, instead of simply `LoadClass('article');` ?

lib/LaTeXML/Package/beamer_support_preload.sty.ltxml
```diff
@@ -0,0 +1,482 @@
...
+  Warn('todo', $feature, "beamer feature $feature is not yet supported; binding requires more development"); }
+
+#**********************************************************************
+# Custom Beamer Arguments
+#**********************************************************************
+
+# The 'Angled' (and corresponding 'OptionalAngled') parameter types are arguments in angle brackets.
+# The 'Angled' argument type is a literal argument type, meaning is not digested by default.
+# Example: <value>
+DefParameterType('Angled',         \&readAngled, reversion => \&revertAngled);
```
brucemiller, line 31: For shortish `DefParameterType`, unless code is reused, I prefer to just use anonymous subs.

brucemiller, line 31: The names are rather mysterious. Is `Angled` used for anything other than a Beamer action or alert or overlay?  Probably would be better to use something like `BeamerOverlay` as the type name.  You could possibly build the parsing into the type, as well, (if the timing is right), and avoid a lot of boilerplate.

Did you try `[Angled]` instead of `SquareAngled`? Seems like that should work, without a new name.

lib/LaTeXML/resources/RelaxNG/LaTeXML-structure.rnc
```diff
@@ -21,7 +21,7 @@ element document { document_attributes, document_model }
 
 ## The content allowable as the main body of the document.
 document.body.class = Para.model | paragraph | subsubsection
-  | subsection | section | chapter | part | slide | sidebar
+  | subsection | section | chapter | part | frame | sidebar
```
brucemiller, line 24: This change will require even "simple" slides (eg. Classic LaTeX slides) to have each slide wrapped in an `ltx:frame`.  Would probably be better to use `slide | frame`.

Except that `frame` is way too generic and overloaded for the schema.  Perhaps `slide-sequence` would be better?

brucemiller (commented):
OK, that's it for now :>  I'll probably have more comments after things condense and settle a bit.

MANIFEST
tkw1536: I wasn't aware there was something that already used the slides, I'll see about condensing the tests.

MANIFEST
tkw1536: The original intent was to allow future compatibility for `beamerarticle.cls` (and share the helper code), but maybe I am thinking too general.

lib/LaTeXML/Package/beamer_support_preload.sty.ltxml
tkw1536: I originally had is named `BeamerOverlay` but that sounded a bit too verbose for my taste.

I hadn't tried `[Angled]`, will check if that works.

lib/LaTeXML/resources/RelaxNG/LaTeXML-structure.rnc
tkw1536: I used `slide` and `frame` to stay in line with the beamer terminology.
At first glance I don't like `slide-sequence`, it seems to verbose. 
I'll see if I can come up with a better term.

tkw1536:
Thanks for the comments; I am rather busy returning to work this week so the comments are about it for now.
I'll try and work in some the suggestions later, before the weekly-ish Thursday meeting.

brucemiller:
> The original intent was to allow future compatibility for 
> |beamerarticle.cls| (and share the helper code),

That makes sense; maybe a single beamer_support.sty.ltxml that
would get (whatever turns out to be) the common code?

brucemiller:
On 10/4/21 08:25, Tom Wiesing wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In lib/LaTeXML/resources/RelaxNG/LaTeXML-structure.rnc 
> <https://github.com/brucemiller/LaTeXML/pull/1652#discussion_r721312289>:
> 
>> @@ -21,7 +21,7 @@ element document { document_attributes, document_model }
>   
>   ## The content allowable as the main body of the document.
>   document.body.class = Para.model | paragraph | subsubsection
> -  | subsection | section | chapter | part | slide | sidebar
> +  | subsection | section | chapter | part | frame | sidebar
> 
> I used |slide| and |frame| to stay in line with the beamer terminology.
> At first glance I don't like |slide-sequence|, it seems to verbose.

I'm not a fan of verbosity either, but we can't get too wedded to
beamers terminology in the general Schema. And things like "frame" or 
"angled" are too overloaded.

brucemiller:
> I'll try and work in some the suggestions later,

Sounds good; Thanks!

MANIFEST
brucemiller: Still too many tests. While more unit-like tests make good sense while developing, for maintenance (imho) they provide clutter and slowness. In most cases, if anything changes, they'll all fail anyway. Might as well have one (or very few) to focus on.  It looks to me like all the tests (w/ possible exception of fsop) are things that would be in a typical beamer doc, so might as well put them together; possibly separated by sections about what feature is being tested.

lib/LaTeXML/resources/RelaxNG/LaTeXML-structure.rnc
```diff
@@ -21,7 +21,7 @@ element document { document_attributes, document_model }
 
 ## The content allowable as the main body of the document.
 document.body.class = Para.model | paragraph | subsubsection
-  | subsection | section | chapter | part | slide | sidebar
+  | subsection | section | chapter | part | slideorsequence | sidebar
```
brucemiller, line 24: No, I meant literally `slide | slidesequence` here. Defining the pattern is perfectly legit rng, but violates the unstated naming convention of `*.class` and such which greatly helps documentation. In fact, I would have used a class here, to avoid the repetition, but for the slowly accumulation of section, then subsection, then...

lib/LaTeXML/Package/beamer.cls.ltxml
```diff
@@ -0,0 +1,1323 @@
...
+);
+# skip everything until that token
+DefConstructorI(T_CS('\beamer@processline'), undef, '', reversion => '',
+  afterDigest => [sub {
+      my ($stomach, $whatsit) = @_;
+      my $gullet = $stomach->getGullet;
+      $gullet->readRawLine;
+      my ($unused, $line) = readRawUntilMatch($gullet, @BEAMER_PROCESSLINE_MARKS);
+      # put the stop token back (if we saw it!)
+      $gullet->unread(T_CS($line)) if defined($line); }]);
```
brucemiller, line 454: This needs a `return;` (or return something/nothing)

lib/LaTeXML/resources/RelaxNG/LaTeXML-structure.rnc
```diff
@@ -144,21 +144,40 @@ subparagraph_model = SectionalFrontMatter.class*, (subparagraph.body.class | Bac
 
 #======================================================================
 
+slidesequence =
+## A slidesequence within a slideshow.
+## Each slide contains a set slides, typically those that are revealed constructively. 
+element slidesequence { slidesequence_attributes, slidesequence_model }
+
+## The content allowable as the main body of a \elementref{slidesequence}.
+slidesequence.body.class = slide
```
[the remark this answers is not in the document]
tkw1536: Fixed in 018673b2a5af4a6851cc1dd2a13570592d2fddb5.

lib/LaTeXML/Package/beamer.cls.ltxml
tkw1536: Done in a80f79cadd77aa9a70d1aefda389c8ce6ab90d40.

lib/LaTeXML/Package/beamer_support_preload.sty.ltxml
tkw1536: I have renamed `Angled` to the clearer `BeamerAngled`. 

> I've replaced SquareAngled by [Angled] as suggested in https://github.com/brucemiller/LaTeXML/commit/b114fa69dafdbdd9da6728e3cc0fe148d0c4ed97.

Turns out that doesn't actually work, there are subtle differences when a `[BeamerAngled]` is followed by another argument with square brackets (And I managed to re-generate one of the test cases and not even notice that it was horribly broken).
So I've had to re-add the argument type in 6051b578f7d58472eed53a67192b1a58329dbe99.

MANIFEST
```diff
@@ -740,6 +741,7 @@ t/80_complex.t
 t/81_babel.t
 t/82_moderncv.t
 t/83_expl3.t
+t/84_beamer.t
```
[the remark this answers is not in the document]
tkw1536: Done

tkw1536:
I've rebased the branch onto the latest master and also taken care of all of the small leftover fixes.

brucemiller (approved):
Looks great! Thanks!!!

PR #1655, opened 2021-09-16
convert paths to urls

xworld21:
Yet another Windows problem. Some paths need to be converted to URLs, or Windows will end up with lots of `%5C` instead of slashes. This PR seems to fix:
- images
- math images
- resources in subfolders
- `--splitnaming=*relative`

If you know of any other path that is being shoved into an attribute, can you please check and add `pathname_to_url`? I am not sure I fixed all of them.

~PS: I took the opportunity to replace some `URI::file->new` with `pathname_to_url` since it does the same job.~ Edit: with apologies for the forced pushes, I realised that `URI::file` and `pathname_to_url` are probably not equivalent, as the latter does not encode the string.

### DIFF
--- a/lib/LaTeXML/Post/CrossRef.pm
+++ b/lib/LaTeXML/Post/CrossRef.pm
@@ -355,7 +355,7 @@ sub fill_in_refs {
           $ref->setAttribute(title => $titlestring); } }
       if (!$ref->textContent && !element_nodes($ref)
         && !(($tag eq 'ltx:graphics') || ($tag eq 'ltx:picture'))) {
-        my $is_nameref = ($ref->getAttribute('class')||'') =~ 'ltx_refmacro_nameref';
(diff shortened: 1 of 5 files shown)


lib/LaTeXML/Post/CrossRef.pm
```diff
@@ -631,7 +631,7 @@ sub generateURL {
         $url .= '#' . $fragid; }
       elsif ($location eq $doclocation) {
         $url = ''; }
-      return $url; }
+      return pathname_to_url($url); }
```
xworld21, line 634: Big problem here: apparently `pathname_to_url('./') eq ''`. This breaks `--splitnaming=*relative` because certain links are to `./`, they become empty, and the links end up with no `@href`.

brucemiller:
Seems to work; and actually seems to be what should have been the intention all along. Thanks!

PR #1659, opened 2021-09-25
epub: detect scripted content and improve mathml, svg detection

xworld21:
There are situations where an EPUB might contain scripted content (e.g. an interactive SVG, or embedding MathJax). Scripted content must be reported in the manifest with the `scripted` property.

The PR cleans up the existing code setting the `mathml` and `svg` properties, plus it sets the `scripted` property when needed. This also means parsing the SVG files to find if they contain scripts or MathML content.

(Unfortunately this causes a trivial conflict with #1643.)


lib/LaTeXML/Post/Manifest/Epub.pm
```diff
@@ -198,8 +217,9 @@ sub process {
       $item->setAttribute('href',       $item_url);
       $item->setAttribute('media-type', "application/xhtml+xml");
       my @properties;
-      push @properties, 'mathml' if $doc->findnode('//*[local-name() = "math"]');
-      push @properties, 'svg'    if $doc->findnode('//*[local-name() = "svg"]');
+      push @properties, 'scripted' if $doc->findnode($xpath_scripted . ' | ' . $xpath_forms);
+      push @properties, 'mathml'   if $doc->findnode($xpath_mathml);
+      push @properties, 'svg'      if $doc->findnode($xpath_svg);
```
brucemiller, line 222: These changes assume we'll only ever have xhtml in epub.  It's at least something to watch out for.

lib/LaTeXML/Post/Manifest/Epub.pm
```diff
@@ -244,7 +264,22 @@ sub finalize {
...
+        $doc = XML::LibXML->load_xml(location => pathname_concat($OPS_directory, $file)); };
+      if ($@) {
+        Error('parsing', $file, undef, "$@"); }
+      else {
+        push @properties, 'scripted' if $XPATH->findnodes($xpath_scripted, $doc);
+        push @properties, 'mathml  ' if $XPATH->findnodes($xpath_mathml,   $doc);
+        my $properties = join(" ", @properties);
+        $file_item->setAttribute('properties', $properties) if $properties; } }
+
+  }
```
brucemiller, line 282: This test is basically what we'd want to determine whether to use img or object, as discussed in #1444, right?  That suggests that it might be better higher up the chain, such as `Post::Graphics`?

lib/LaTeXML/Post/Manifest/Epub.pm
xworld21: This one is easy to explain: the EPUB specs (both 2 and 3 as far as I know) require HTML5 in XML serialisation, with the appropriate namespaces. So this code should be ok, and more accurate, actually.

lib/LaTeXML/Post/Manifest/Epub.pm
xworld21: The img/object distinction is a bit more subtle because the `` tag disables every external resource as well. I'd say this code is specific to EPUB (the spec just wants to know if there are scripts and MathML content*).

* Actually, one should also declare external resources, e.g. web fonts not embedded in the file. This requires parsing the CSS content of the image, so it seems quite hard to do reliably.

lib/LaTeXML/Post/Manifest/Epub.pm
```diff
@@ -15,6 +15,25 @@ use warnings;
 use File::Find qw(find);
 use URI::file;
 
+# HTML5 namespaces
+our $XPATH = $LaTeXML::Post::Document::XPATH;
```
brucemiller, line 19: The timing of this is off; it will set `$XPATH` to the current value when `LaTeXML::Post::Manifest::Epub` is loaded, rather than when the document is being packaged. That will probably work for a one-off conversion because LaTeXML has defered but will not (necessarily) be correct in more general (eg. daemon) contexts.

The problem is that we haven't got a clear clean API for managing namespaces so that we can declare these mappings once, and use them consistently throughout conversion & postprocessing (ideally, they should only make it into the otuput document if they're needed).

Short of that, it would be best to move these `registerNS` into one of the `Epub` methods; probably `initialize`.

I'm trying to get some of these PR's moving again. @dginev do you have any comments?

lib/LaTeXML/Post/Manifest/Epub.pm
brucemiller: What would be the harm in simply assuming that any svg image contains mathml and scripts?

xworld21: Per spec, the `mathml` and `scripted` properties "MUST be set if and only if the criterion specified in the description is met", so the output would be invalid. To my memory, epubcheck was flagging the properties as errors if no mathml/scripts were in the file.

lib/LaTeXML/Post/Manifest/Epub.pm
xworld21: I moved most of the `$XPATH` code to `initialize`. Once you reach `initialize`, `XPATH` should already exists, right?

xworld21:
Updated to work with latest Epub.pm updates on master.

PR #1662, opened 2021-09-26
ensure tbody is emitted consistently in equationgroup's

### LINKED ISSUE #1656: Subequations produce `<tr>` after `<tbody>`
Consider the following document

    \documentclass{book}
    \title{Book}
    
    \usepackage{amsmath}
    
    \begin{document}
    Here is the galilean transformation:
    \begin{subequations}
    \begin{equation}
    \begin{aligned}
    x'&=x-Vt\\
    y'&=y\\
    z'&=z\\
    t'&=t\\
    \end{aligned}
    \end{equation}
    \end{subequations}
    
    \end{document}

When processed with:

    % latexml --VERSION
    latexml (LaTeXML version 0.8.5)
    % latexml --dest=test-mathml.lxml test-mathml.tex
    % latexmlpost --format=xhtml --destination=t/test-mathml.xhtml test-mathml.lxml 

this produces an output HTML file containing

    <p class="ltx_p">Here is the galilean transformation:</p>
    </div>
    <div id="p2" class="ltx_para">
    <table id="Ch0.E1" class="ltx_equationgroup ltx_eqn_table">
    <tbody id="Ch0.E1.1" class="ltx_eqn_row"><tr><td class="ltx_eqn_cell" colspan="5"/></tr></tbody>
    <tr id="Ch0.E1.1X" class="ltx_equation ltx_eqn_row ltx_align_baseline">
    <td class="ltx_eqn_cell ltx_eqn_center_padleft"/>

Note that here both `<tbody>` and `<tr>` appear as children of the `<table>` element; `<tbody>` can't be a sibling of `<tr>` (see eg the [whatwg spec for this element](https://html.spec.whatwg.org/multipage/tables.html#the-tbody-element); and in practical terms, the W3C epubcheck tool objects to this, too).

(Found it!)

xworld21:
This supposedly fixes #1656 by using the same test whenever a `<tbody>` might be output. I *think* this guarantees that either everything is wrapped in (a single) `<tbody>`, or nothing is.

Note that I changed the test! It now checks that there are no descendant `ltx:equationgroup/ltx:tags`, just in case an equation group contains another one, and both have tags, which would create two very wrong rowspanned tags. I suppose that never happens, but it is not guaranteed by the schema.

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-block-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-block-xhtml.xsl
@@ -522,6 +522,12 @@
   <xsl:template match="*" mode="aligned-begin"/>
   <xsl:template match="*" mode="aligned-end"/>
 
+  <!-- when an equationgroup has tags, but there are no further tags below, the equation number comes from the group -->
+  <func:function name="f:has-group-eqno">
+    <xsl:param name="node" select="."/>
+    <func:result select="ancestor-or-self::ltx:equationgroup[position()=1][ltx:tags][not(descendant::ltx:equationgroup/ltx:tags | descendant::ltx:equation/ltx:tags)]"/>
+  </func:function>
+
   <!-- ======================================================================
        Generate the padding column (td) for a (potentially) numbered row
        in an aligned equationgroup|equation.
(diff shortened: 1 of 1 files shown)


xworld21:
PS: it would be good to add a test. Maybe check the html output of `t/ams/amsdisplay.tex`? How do I do that?

brucemiller:
I've procrastinated on this, since it was rather scary. It wasn't so much that the patch was badly done, but the existing code is a bit too mysterious to predict.  For that reason, I came up with an alternative where I've (hopefully) made the cases & logic a bit clearer and could carefully test.

Could you take a look at #1888?  Thanks for pointing in the right direction!

PR #1664, opened 2021-09-27
allow VERTBAR as anyOp in math parsing

dginev:
Another small PR while I'm warming up, from the [arXiv parse warnings with VERTBAR](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/warning/not%5Fparsed/%3EVERTBAR?all=false), which appear in 8.79% of arXiv docs.

Apparently we currently don't recognize the VERTBAR role in the "anyOp" category, thus failing to parse fragment expressions with vertical bars. Two simple motivating examples from the report that now pass with this PR:

* `$ |_{\cal L} $ `
* `$ J_{\parallel 2} $`

I *hope* this doesn't have side-effects in parsing actual fenced expressions with vertbars, due to the way `anyOp` is used in the AnythingAny and script rules -- it's only employed when the standard rules fail to match.

### DIFF
--- a/lib/LaTeXML/MathGrammar
+++ b/lib/LaTeXML/MathGrammar
@@ -675,6 +675,7 @@ nestOperators :
 AnyOp   : relop | METARELOP | ARROW | AddOp | MulOp | MODIFIEROP
         | preScripted['bigop']
         | OPERATOR addScripts[$item[1]]
+        | VERTBAR addScripts[$item[1]]
 
 # Sub or superscripts on operators;
 # we recognize the structure, not necessarily the meaning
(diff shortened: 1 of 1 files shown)


dginev:
Studying the failed tests, it seems including into `anyOp` was too easy to avoid side-effects.

Trying the next best thing (to my mind), I included dedicated VERTBAR fragment rules in `AnythingAny`, `aSubscript` and `aSuperscript` which ought to do only what I had in mind with the motivating examples in the PR description.

dginev:
Eh, works a bit better but I am not too fond of the parses.

It should really be added to `AnyOp` conceptually, but making that work with RecDescent is probably too much effort for what it's worth. So just closing this PR, and leaving the trace that something was tried :>

PR #1667, opened 2021-09-30
[arXiv] minor upgrades for svmult, ifsym and bbding

dginev:
A basic debugging session over arXiv:1606.00123, upgrading it to a warning status quickly:

- some missing convenience macros from svmult.cls.ltxml
- a very minor convenience macro (which I saw recommended in both bbding and ifsym) for producing a unicode envelope ✉

As usual, tried to be creative in time management. I added the two new binding files without implementing them in their entirety - just the macro I was debugging - which allows making this PR in the same morning as my first encountering the problem.

But, to keep the reporting honest and address the sentiment of "let's not claim we support capabilities we really do not", I left in the **same warning** which would have been produced if the binding wasn't present, just tweaking a little the Details message. I'm hoping this is an acceptable way for having my cake and eating it too - have latexml openly claim the binding is not supported, while incrementally adding individual macros, as we need them in arXiv and other places.

### DIFF
--- a/lib/LaTeXML/Package/svmult.cls.ltxml
+++ b/lib/LaTeXML/Package/svmult.cls.ltxml
@@ -85,6 +85,13 @@ DefMath('\gid',    "\x{2267}", role => 'RELOP', meaning => 'greater-than-or-equa
 DefMath('\grole',  "\x{2277}", role => 'RELOP', meaning => 'greater-than-or-less-than');
 Let('\qedsymbol', '\qed');
 
+#  special signs and characters
+DefMacro('\D', '\mathrm{d}');
+DefMacro('\E', '\mathrm{e}');
+Let('\eul', '\E');
+DefMacro('\I', '{\rm i}');
+Let('\imag', '\I');
+
 DefMacro('\partsize',    '\Large');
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Package/svmult.cls.ltxml
```diff
@@ -85,6 +85,13 @@ DefMath('\gid',    "\x{2267}", role => 'RELOP', meaning => 'greater-than-or-equa
 DefMath('\grole',  "\x{2277}", role => 'RELOP', meaning => 'greater-than-or-less-than');
 Let('\qedsymbol', '\qed');
 
+#  special signs and characters
```
dginev, line 88: The source I used for locating these: https://github.com/openeventdata/Computational-Approaches/blob/master/svmult.cls#L954-L960

The arXiv doc was hitting an error on `\E`

brucemiller:
was the intention to sneak this into 0.8.6 before a large arXiv run?

dginev:
> before a large arXiv run?

this part "yes"

>  into 0.8.6 

this part "agnostic" :> I've already mentally switched over to 0.8.7 work

PR #1677, opened 2021-10-05
Recognize all-caps ".TEX" extensions in ZIP bundles

dginev:
Apparently I had made an oversight when I wrote the ZIP unpacking that only looked at lowercased `.tex` files. 

I ended up examining one of the "invalid" arXiv sources today (since a yet-to-be-named preview service made it very easy) by chance, and I hope this tiny change will yield a lot of extra articles in our arXiv conversion runs (currently 25,000 articles are marked as invalid)

### DIFF
--- a/lib/LaTeXML/Util/Pack.pm
+++ b/lib/LaTeXML/Util/Pack.pm
@@ -40,7 +40,7 @@ sub unpack_source {
   foreach my $member ($zip_handle->memberNames()) {
     $zip_handle->extractMember($member, catfile($sandbox_directory, $member)); }
   # Set $source to point to the main TeX file in that directory
-  my @TeX_file_members = map { $_->fileName() } $zip_handle->membersMatching('\.tex$');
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Util/Pack.pm
```diff
@@ -40,7 +40,7 @@ sub unpack_source {
   foreach my $member ($zip_handle->memberNames()) {
     $zip_handle->extractMember($member, catfile($sandbox_directory, $member)); }
   # Set $source to point to the main TeX file in that directory
-  my @TeX_file_members = map { $_->fileName() } $zip_handle->membersMatching('\.tex$');
+  my @TeX_file_members = map { $_->fileName() } $zip_handle->membersMatching('\.[tT][eE][xX]$');
```
dginev, line 43: i would usually use the `/i` flag here, but since it's passed into a method I ended up dual-casing each letter

PR #1680, opened 2021-10-06
Better timing for deluxetable.sty notes

dginev:
Trying to get just a bit closer to the raw [deluxetable.sty](https://fits.gsfc.nasa.gov/standard30/deluxetable.sty) so that notes work correctly when casually deposited before `\enddata`.

An example affected source from arXiv is [astro-ph/0001016](https://ar5iv.org/html/astro-ph/0001016), which has the misfortune of the note getting deposited in a table cell:
image

By accumulating the notes in `\tblnote@list` and then depositing them at `\end{deluxetable}` via `\spew@tblnotes`, we are guaranteed to get the timing right - at least as right as the original sty file.

This PR upgrades the example in question to:

image

Caveat 1: both screenshots use my arxmliv.css experiment, rather than the default latexml css.
Caveat 2: I haven't really tested this PR extensively, wanted to see if there is obvious feedback to collect at first.

There is a [long list of deluxetable-loading articles](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/info/loaded%5Ffile/deluxetable%2Esty%2Eltxml?all=false) in the CorTeX reports, which I can sample with 5-10 articles and report back.

### DIFF
--- a/lib/LaTeXML/Package/deluxetable.sty.ltxml
+++ b/lib/LaTeXML/Package/deluxetable.sty.ltxml
@@ -32,11 +32,11 @@ DefMacro('\dummytable', '\refstepcounter{table}');
 
 DefMacroI(T_CS('\begin{deluxetable}'), "{}",
   '\set@deluxetable@template{#1}\def\@deluxetable@header{}\begin{table}');
-DefMacroI(T_CS('\end{deluxetable}'), undef, '\end{table}');
+DefMacroI(T_CS('\end{deluxetable}'), undef, '\spew@tblnotes\end{table}');
 
 DefMacroI(T_CS('\begin{deluxetable*}'), "{}",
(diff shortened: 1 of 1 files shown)


dginev:
In case you're wondering why there is a single windows test failure in the PR, it appears our CI use of miktex continues to be fragile.

In case @tkw1536 has time and feels like suffering, the line is:
https://github.com/brucemiller/LaTeXML/runs/3809584120#step:6:1324

I think the best route is to go ahead and email with the miktex team and ask for best practices (and some guarantees) in using their installer for CI.

dginev:
@tkw1536 no need to disable it, since it's only fragile - not fully broken. I reran the CI today, and all tests pass, as miktex was installed correctly this time.

I'll drop an email to the miktex maintainers to ask them about this, but we can keep things as-is for now. Maybe even merge this PR :>

Issue #1665, opened 2021-09-29
Smarter sizing in PDF to PNG image conversion


dginev:
As a running example, take arXiv's [1804.11027](https://arxiv.org/pdf/1804.11027.pdf), in particular Figure 11 on page 8.

The two images in that figure, titled `CUHK03` and `Market-1501` are PDF files in the original paper sources, and the latex for including them is:

```tex
\begin{figure}[t]
\centering
\begin{tabular}{c}
\includegraphics[height=5cm,width=7cm]{cmc-cuhk03}\\
\includegraphics[height=5cm,width=7cm]{cmc-market}
\end{tabular}
\caption{The CMC curves on CUHK03 and Market-1501 datasets.}
\label{fig:cmc-cuhk03-market}
\end{figure}
```

My main focus with this issue is how to avoid harming the image quality, when respecting the sizing arguments (`[height=5cm,width=7cm]`) and creating a new PNG file. Currently, these particular sizing parameters generate an image so small as to be completely illegible:

![image](https://user-images.githubusercontent.com/348975/135242272-a63ee4d9-690e-4eaf-b9e5-e3f8563ba0c6.png)

Converting with the sizing arguments removed creates a legible (but, for arXiv, unconstrained to its maximum size) plot PNG:
![image](https://user-images.githubusercontent.com/348975/135242149-7adadb45-6b87-4575-9184-c423b4bb5cf0.png)

As a brainstorming, half-baked, usability suggestion to conclude with:

Could we rework the width and height parameters in particular as attributes on the image container, rather than as inherent properties of the PNG image? And err on the side of creating higher-quality / larger size PNGs than necessary.

dginev:
And of course something should be said about whether some PDF images can't be converted to SVG rather than PNG, which is a bit more open-ended...

dginev:
In retrospect, having revisited the currently open image issues, this one could be seen as a partial duplicate of #1066 or #902 , since the problem at the core is the same - low quality conversions to PNG from a source PDF figure.

That said, I have a new example and my problem was specifically related to the image size, so it's not _exactly_ a duplicate. Maybe if we collect enough reports the solution becomes more obvious :>

dginev:
Having received some tutoring, and experimenting at length, I think the DPI upgrade combined with some refinement that applies the high DPIs for the various sub-cases, suffices to address the main problems. So I'll close, as the remained of my issue is a duplicate to #1066

PR #1683, opened 2021-10-07
Autocrop PDF images

dginev:
I suspect there may be trade-offs, but autocrop on pdf images has a visible upgrade to the article in #1665

Compare the mainline:

image

to the PR's:

image

In the specific article, I see all figures as improved - though this may be worth additional discussion and testing.

### DIFF
--- a/lib/LaTeXML/Post/Graphics.pm
+++ b/lib/LaTeXML/Post/Graphics.pm
@@ -39,6 +39,8 @@ use base qw(LaTeXML::Post::Processor);
 #                         This is useful for getting the best antialiasing for postscript, eg.
 #          unit= (pixel|point) :  What unit the image size is given in.
 #          autocrop     : if the image should be cropped (trimmed) when loaded.
+#                         This option is sometimes needed for page-oriented formats like pdf,
+#                         where the image size appears to be padded out to page dimensions.
 #          desirability : a number indicating how good of a mapping this entry is.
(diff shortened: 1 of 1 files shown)


brucemiller:
I'm still fuzzy on exactly why ImageMagick comes up with this extra space around the image; A bug in ImageMagick? Are we using it wrongly? Viewers such as atril don't show it, but ImageMagick (eg. display) does.

I'm thinking it's some artifact or misinterpretation of the "page" nature of pdf. Which conceivably could also apply to ps and eps. Did you encounter or look for such cases? Maybe `autocrop` should be enabled for those formats as well?

dginev:
>  special symbol or icon or something

Actually, I would definitely expect a CSS rule expecting icons and thumbnails at the appropriate places in a document. I indeed added a rule last week specifically giving some max-sizing restrictions to images that appear inside footnotes. Which is where CSS is great, since it is very easy to understand what `.ltx_note_content .ltx_graphics` is meant to style.

brucemiller:
Yeah, why don't you add those, as well. And also please add a comment to description of options above there, to the effect "This option is sometimes needed for page-oriented formats like pdf, where the image size appears to be padded out to page dimensions".  Hopefully that's accurate enough, or at least is suggestive, next time we have to ponder this topic.

dginev:
@brucemiller sure thing, done!

PR #1684, opened 2021-10-07
3x the default dpi in image conversion

dginev:
Magic number PR, so I made it separate. I half-expect it to get rejected, but *maybe*, just *maybe* it gives an idea of how to get it transformed into a PR that gets accepted before we start the next arXiv run.

When combined with the autocrop PR ( #1683 ), latexml can deliver a high quality image for each of the problematic figures demoed in #1665 .

image

This is already workable as it is sufficiently high qualtiy+size, and further adaptations could be enforced via CSS, if needed.
I admit this is a very fast-and-loose solution, but given how much better the images look, it really begs the question whether we have some slightly more refined variant of this PR that get us high quality images by default.

### DIFF
--- a/lib/LaTeXML/Util/Image.pm
+++ b/lib/LaTeXML/Util/Image.pm
@@ -310,8 +310,9 @@ sub image_graphicx_complex {
           if ($a1 / $w < $a2 / $h) { $a2 = $h * $a1 / $w; }
           else                     { $a1 = $w * $a2 / $h; } }
         ($w, $h) = (ceil($a1 * $dppt), ceil($a2 * $dppt)); } }
-    my $X = 4;                 # Expansion factor
-    my ($dx, $dy) = (int($X * 3 * 72 * $w / $w0), int($X * 3 * 72 * $h / $h0));
+    my $X = 4;    # Expansion factor
+    my $F = 3;    # Density upscale factor
+    my ($dx, $dy) = (int($X * $F * 72 * $w / $w0), int($X * $F * 72 * $h / $h0));
     Debug("reloading $source to desired size $w x $h (density = $dx x $dy)")
       if $LaTeXML::DEBUG{images};
     $image = image_read($source, antialias => 1, density => $dx . 'x' . $dy) or return;


lib/LaTeXML/Util/Image.pm
```diff
@@ -311,7 +311,7 @@ sub image_graphicx_complex {
           else                     { $a1 = $w * $a2 / $h; } }
         ($w, $h) = (ceil($a1 * $dppt), ceil($a2 * $dppt)); } }
     my $X = 4;                 # Expansion factor
-    my ($dx, $dy) = (int($X * 72 * $w / $w0), int($X * 72 * $h / $h0));
+    my ($dx, $dy) = (int($X * 3 * 72 * $w / $w0), int($X * 3 * 72 * $h / $h0));
```
dginev, line 314: I simply multiplied the density by 3 on each component, the rest is whitespace changes from latexmllint.

brucemiller:
Ah, hmm... If I understand your code & mine...

So the current prescaling is to upscale, then load, then dowscale again, 
apparently purely to give antialiasing (within ImageMagick) more meat to 
work with.

But you actually want to scale the image itself to larger dimensions,
and simply display it "expected" smaller via image width & height.  Is 
that right?

(and if so, isn't there already an option for that lurking in the dark?)

dginev:
Right,  so the core need is that we have a vector PDF that needs to get to PNG without losing quality. Adding more density to the sampling achieves that relatively well, even if it is a tricky question as to exactly how much density is needed. There is a size+runtime trade-off on the other side of quality. But this `3x` increase produced very decent results on the document I tested (and a couple of others since).

The point about the density is the same one about DPI (dots-per-inch) -- we need more dots in an inch's worth of space for these PDF cases :> That's all!

Also, the comments mention a mysterious `$F` factor, but I don't see it defined or used anywhere - maybe this is what I should leverage?

dginev:
@brucemiller [here is a first "cleanup"](https://github.com/brucemiller/LaTeXML/pull/1684/files?w=1)

dginev:
Another question is whether I should only do this for the pdf, ps and eps `{type_properties}` in Graphics.pm? It's a bit hard to predict which of the formats really need this without extensive testing... and starting with PDF is already addressing a lot of the modern papers that tend to get public scrutiny.

dginev:
After reading some more, indeed `$dppt` is a variable that can be leveraged as it appears intended to specify this quantity. But a piece of code acts as a soft block for "extreme" quality requests:
```perl
if ($a3) {             # If keeping aspect ratio, ignore the most extreme request
          if ($a1 / $w < $a2 / $h) { $a2 = $h * $a1 / $w; }
          else                     { $a1 = $w * $a2 / $h; } }
        ($w, $h) = (ceil($a1 * $dppt), ceil($a2 * $dppt)); } }
```

Keeping the aspect ratio is a common situation, so that guard is a bit too defensive. Without it, we can likely even address the #1066 issue which requests a public interface to control the DPI.

brucemiller:
I think the thing with `$F` and `$X` is simply that I wrote the comment intending to use `$F`, but ended up accidentally writing the code using `$X`. The scaling there is exactly to have ImageMagick rasterize it (while reading it in) at a higher resolution, and then downsampling it to the requested size. This improves the antialiasing.

What you seem to be doing is creating an image at higher resolution and returning it at that resolution.  Basically, you're returning a "larger" image.  I wonder if is looking better due to it being (apparently) shrunk by other sizing/css/whoknowswhat, that may be coupled with a modern screen's "pixel" being bigger than pixel?

dginev:
The image at the top of the PR is indeed sized `933x668`. The original with only the autocrop PR is at `311x222` and looks like this:

image

The CSS I have indeed constrains the original image down to `670x480`, which is not a major difference, I think. You can directly compare the two images here without CSS constraints - the smaller size image just has poorer sampling for the original PDF, or at least that's how I am interpreting the reason for the very blurry quality. I am really not well-versed in the area of image manipulation so... all help welcome.

dginev:
Thanks for that elaboration.

>  others may already be essentially full page. Scaling the latter by 3 may be kinda shocking.

Right. So it's more about "minimal usable size" ... On my 4k screen, in the arXiv preview, specifically, it appears having a width of at least 900 for the underling PNG is needed to get a sharp clear figure. But footnote images and thumbnails don't have to get to that height, naturally, just main-flow figures do. Hm.

brucemiller:
On the contrary, I'd say it implies we should stick to what the author 
requested unless we can provide a better quality image w/o disrupting
either the authors intent or the target website (which means it may be
different for sites other than ar5iv).  And even in 2021, some people
care about bandwidth.

brucemiller:
> So there is /some/ suggestion of the available size.

Indeed, or you might look at LaTeXML::Util::Image's `image_size`.

PR #1686, opened 2021-10-09
Ifsym binding

brucemiller:
A binding for `ifsym.sty` (with most but not all of its glyphs), along with a bit of special case trickery to make it easier to define font-specific font maps where the "U" encoding is typically used.

I'm hesitating to add yet-another test case for such a marginal binding; will attach
[ifs.txt](https://github.com/brucemiller/LaTeXML/files/7315513/ifs.txt)

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4945,6 +4945,9 @@ DefPrimitive('\selectfont', sub {
     my $series = ToString(Expand(T_CS('\f@series')));
     my $shape  = ToString(Expand(T_CS('\f@shape')));
     if    (my $sh = LaTeXML::Common::Font::lookupFontFamily($family)) { MergeFont(%$sh); }
+    elsif ((LookupValue('font')->getEncoding eq 'U') && LoadFontMap($family)) {
+      # Special case hack: If encoding is U, tentatively treat family as the encoding!
+      MergeFont(encoding => $family); }
     elsif (!LookupValue("reported_unrecognized_font_family_$family")) {
       AssignValue("reported_unrecognized_font_family_$family", 1, 'global');
       Info('unexpected', $family, $_[0], "Unrecognized font family '$family'."); }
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -4945,6 +4945,9 @@ DefPrimitive('\selectfont', sub {
     my $series = ToString(Expand(T_CS('\f@series')));
     my $shape  = ToString(Expand(T_CS('\f@shape')));
     if    (my $sh = LaTeXML::Common::Font::lookupFontFamily($family)) { MergeFont(%$sh); }
+    elsif ((LookupValue('font')->getEncoding eq 'U') && LoadFontMap($family)) {
+      # Special case hack: If encoding is U, tentatively treat family as the encoding!
```
dginev, line 4949: Unsure how safe this is, but probably safe enough for now.

I had seem some missing_file warnings in arXiv with a small single letter `u`, I wonder if they're related?
https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/warning/missing%5Ffile/u?all=false

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -1600,6 +1600,9 @@ DefPrimitive('\@@@fontencoding{}', sub {
 DefMacroI('\f@encoding',  undef, sub { ExplodeText(LookupValue('font')->getEncoding); });
 DefMacroI('\cf@encoding', undef, sub { ExplodeText(LookupValue('font')->getEncoding); });
 
+## Defined since we "expect" it; Empty since we know nothing about it.
+DeclareFontMap('U', []);
```
dginev, line 1604: I had seem some missing_file warnings in arXiv with a small single letter `u`, I wonder if they're related?
https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/warning/missing%5Ffile/u?all=false

brucemiller: It may be worth an exploration; Previously if you typeset characters from a font with an "unknown" encoding and didn't otherwise end up with a fontmap, you'd get characters from whever the previous encoding was (ie. wrong chars). I think now you'll end up with them silently missing (which I'm inclined to think is what TeX would do?)

brucemiller: Ok, new batch of commits. Seems better to not define a fontmap for U (default to OT1 encoding), define fontmaps in their own files, and make `LoadFontMap` quieter by default. That seems to act a bit more TeX-like.  So, thanks for the disruptive complement! :>

dginev: Happy to be of help, I suppose :>

PR #1687, opened 2021-10-09
add \@footnotetext for arxiv:1807.00181

dginev:
Minor PR (I hope).

An arXiv paper relied directly on `\@footnotetext` in a custom footnote definition. Which we can support just as we do `\footnotetext`, I think.

At least that paper's footnote looked reasonable with this PR.

Edit: FWIW, 0.14% of arXiv articles had this as an undefined macro, so at least somewhat tangible benefit in merging.


brucemiller:
Seems right.

PR #1688, opened 2021-10-10
[arXiv] Prune empty affil notes, improve IEEEtran keywords

dginev:
Fixes improving the display of [1804.11034](https://ar5iv.org/html/1804.11034). Another easy+minor PR

1. Vertical author list caused by a bunch of empty affiliation notes, which can be avoided. I added some general checks that avoid creating notes without any useful content, and a special pruning step at the end of documents using `authblk.sty`.

2. Keywords were not properly announced for IEEEtran.cls. I added and used the macro responsible, and threw in a handful of other simple naming macros I saw in the class file.


brucemiller:
looks like plausible stuff, and a nice cleanup.

PR #1692, opened 2021-10-13
Latexml keys and dpi

brucemiller:
This PR adds a simple capability to process keyvals in class and package options, which is then used by `latexml.sty` to add a new `dpi=xxx` package option.  That dpi is detected by postprocessing to adjust the resolution and size of images being processed.

Perhaps this should be a WiP, as there are a couple of messy details. It is unclear whether this option should set STATE's `DPI`, since that is used to convert points to "pixels", which is currently less about image processing, as for dealing with svg sizing and positioning, especially the attempt to match svg fonts with document fonts. Or maybe they're the same thing after all?

It is also clunky in that the `dpi` option adds a PI to the to-be-created document, but this gets lost when you use `--preload`, since it doesn't respect `@at@document@begin`.  The dpi is, however, present in the options of the package PI, which is manually added by `--preload`. Ugh!

So, maybe it's good enough as it is until we sort out some other issues?

### DIFF
--- a/lib/LaTeXML/Post/Graphics.pm
+++ b/lib/LaTeXML/Post/Graphics.pm
@@ -22,7 +22,7 @@ use base qw(LaTeXML::Post::Processor);
 
 #======================================================================
 # Options:
-#   dpi : dots per inch for target medium.
+#   DPI : dots per inch for target medium.
 #   ignore_options  : list of graphicx options to be ignored.
 #   warn_options    : list of graphicx options to cause warning if used.
 #   trivial_scaling : If true, web images that only need scaling will be used as-is
@@ -50,7 +50,7 @@ use base qw(LaTeXML::Post::Processor);
 sub new {
   my ($class, %options) = @_;
   my $self = $class->SUPER::new(%options);
-  $$self{dppt}            = (($options{dpi} || 90) / 72.0);    # Dots per point.
+  $$self{DPI}             = $options{DPI};
   $$self{ignore_options}  = $options{ignore_options}  || [];
   $$self{trivial_scaling} = $options{trivial_scaling} || 1;
   $$self{graphics_types}  = $options{graphics_types}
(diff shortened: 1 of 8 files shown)


dginev:
Very cool! I just tried a few conversion with `--preload=[dpi=180]latexml.sty` and things work great. Played around with the values as well (45, 360), all works as expected.

I'll take a look at the code next, but I preemptively agree that we can merge as-is and refine it later, when we get to larger thematic refactors.

lib/LaTeXML/Post/Graphics.pm
```diff
@@ -98,6 +98,21 @@ sub process {
...
   return $doc; }
 
+# Need a proper API to query PI's, probably in Post
+# But Core isn't respecting @at@document@begin from preloaded packages,
+# so here we're testing BOTH a straight DPI PI and the options from latexml.sty (ugh)
+sub getDPI {
+  my ($self, $doc) = @_;
+  my $dpi;
+  foreach my $pi (@{ $$doc{processingInstructions} }) {
+    if ($pi =~ /^\s*DPI\s*=\s*([\"\'])(.*?)\2\s*$/) {
```
dginev, line 108: wrote a small script to test this, and it looks as if this `\2` should really be `\1` here.

It's still `\2` in the package regex, which is what ends up matching.

dginev (approved):
I'm quite fond of the PR! Very much appreciate the help.

The code looks quite reasonable, I'm still doing more learning than reviewing.

Only had a single question about a regex. I also enjoyed the `72.27` refinement, and the bump up to `100` default dpi, though finding the perfect numbers is indeed more art than science :>

lib/LaTeXML/Post/Graphics.pm
brucemiller: Good catch! It should be 1. The `elsif` is an ugly fallback, only needed because `--preload` doesn't respect `@at@document@begin`, which would have added a proper PI.

dginev:
Well, your point is well-taken, I can reproduce both the slowdown and the image size. I don't want to make things any larger than they have to be, certainly. But I also want to avoid blurry/squint-worth quality.

Playing around some more, reconverting-squinting-reconverting-squinting ... (on my 200% zoomed in firefox browser on a 4k 27'' screen). I can see that dpi 140 already gives some blurry outcomes that make me squint on my setup. With 150, I think pretty much everything looks nice and readable. With 300 it makes you "wow", since the sharpness is fantastic, but the compute+bandwidth+storage needed are prohibitive.

So, seeing how I am already viewing from one of the largest possible screens and resolutions, I think the smallest dpi that looks legible (and has no obvious painpoints) over here should do. Right now this seems to be 150, since as I said I found issues at 140. I'll keep testing articles, but it's really great to be able to test with the new option.

The PR is good to merge, as far as I'm concerned :>

brucemiller:
Just for the record: I'm thinking that "dpi" is the wrong thing to be tweaking (if it's dpi, it really should also be setting the dpi in state, used to convert pts to pixels, affecting svg and other potentially other stuff). What you'd probably want to adjust would be "magnification", or even "figure-magnification", if you want to distinguish, say, mathimages or something.  In the end, the product would give you the same thing you get now, but it would make more sense in the long run.  But perhaps best to go with what we have now and think it through before further refinement (just be prepared to change `dpi=xxx` to something else later)

PR #1694, opened 2021-10-13
Minor additions to svmult.cls

dginev:
The remaining additions of #1667 , now that the bindings got merged separately.

Looking forward to seeing them all in action :>


brucemiller:
OK, I'll assume those are good

PR #1695, opened 2021-10-13
Experiment: disable clip in image conversion

dginev:
This PR is another request for help/ideas, even if it technically completes the quest for high quality images in the article used in #1066 .

The summary of what the underlying issue is, if I traced it well enough, is in the following lines when 'trim' and 'clip' are handled:

```perl
      # Use the image's dpi for trim & clip!
      my $idppt = (defined $imagedpi ? ($imagedpi / 72.0) : $dppt);
```

To my understanding, as the optional higher DPI from #1692   gets applied during the first `prescale` phase, when this lower DPI (assuming the image had a lower original size) gets used in the trim/clip phase, the conversion never again returns the image to the high DPI. Disabling trim entirely indeed keeps the requested prescale DPI.

Curiously, when the `autocrop` option is enabled -- for only this very specific document -- it seems that trim has no further effect, as I think the authors also used it with a crop intent. But of course, in general there could be a significant difference between what the two options achieve.

The PR just comments out the line that applies the `trim` when `clip` is present, and by avoiding that operation runs away from the issue. I suspect the preferred solution would be to scale the image back to the desired DPI at the end, in something akin to a "postscale" phase? Unsure.

But yes, this is the last remaining problem for the article in #1066 , so that issue may be close to a resolution.

### DIFF
--- a/lib/LaTeXML/Util/Image.pm
+++ b/lib/LaTeXML/Util/Image.pm
@@ -154,7 +154,7 @@ sub image_graphicx_parse {
   # We ignore viewport & trim if clip isn't set, since in that case we shouldn't
   # actually remove anything from the image (and there's no way to have the image
   # overlap neighboring text, etc, in current HTML).
-  push(@transform, [($trim ? 'trim' : 'clip'), @vp]) if (@vp && $clip);
+  #push(@transform, [($trim ? 'trim' : 'clip'), @vp]) if (@vp && $clip);
   push(@transform, ['rotate', $angle]) if ($rotfirst && $angle);    # Rotate before scaling?
   if ($width && $height) { push(@transform, ['scale-to', $mag * $width, $mag * $height, $aspect]); }
   elsif ($width)         { push(@transform, ['scale-to', $mag * $width, 999999, 1]); }


dginev:
> It's not really clip that's the problem; it's more the dance of figuring out the dpi (and what it "means"), along with what sort of scaling and or prescaling is desired or needed for a quality image. I think I've almost got it, just need a bit more testing.

Exciting to hear!

>  Also, I'd like to get it handled more predictably, so as not to expose an awkward API so long that we have to maintain it. (so hold off on arXiv a bit longer)

Sure. Last week got me to change gears, and now I'm in "delay by default" mental mode for the arXiv rerun. I really want to finish up the upgraded subfigures and the cleaned-up frontmatter, before I start any large runs. It's unfortunate those take time, but they're worth the impact.

> Incidentally, I discovered that there are up to 4 distinct "bounding boxes" in pdf, and that ImageMagick and graphic(s|x) choose differently (by default). So, ultimately `autocrop` is not needed.

Wow, 4 distinct bounding boxes? Good grief. It would be quite impressive if latexml's image conversion interprets those correctly.

brucemiller:
> Wow, 4 distinct bounding boxes? Good grief. It would be quite impressive 
> if latexml's image conversion interprets those correctly.

Ha! I don't even want to have to think about them!
ImageMagick uses the MediaBox, which is essentially the page size,
whereas graphics.sty uses the CropBox which is the desired image size.
But it turns out that you can tell ImageMagick to use cropbox, but
you have to be insistent.

Got a moment for a brief brainstorm chat?

PR #1697, opened 2021-10-18
Image resolution

brucemiller:
Improvements to image postprocessing, particularly for pdf images.
It adds 3 new controls (exposed as latexml.sty package options); each takes a multiplicative factor.
* magnify : increases the size of the image.
* upscale : increases the size used during the transformation, to improve accuracy, but the final image size is unchanged (default is 2)
* zoomout : leaves the image size unchanged, but reduces the width & height attributes used in the img tag; this leaves more browser reserve resolution for zooming, or high resolution displays.

These can (probably should) be used in combination for various goals.

To implement this, more care is taking in managing the image size predictably.
Additionally, we've fixed ImageMagick's reading of pdf file's bounding box in that it defaults to the MediaBox, whereas LaTeX's graphics package uses the CropBox.  This fix makes the `autocrop` operation mostly unneeded.  Richer debugging messages were also added (`--debug=images`) to help develop it.

If approved this is probably ready for arXiv processing, but there are interesting further improvements:
* More complete "prescaling" to determine the maximum resolution needed
* Parameters to control maxium or minimum sizes  and to control resource use.

### DIFF
--- a/lib/LaTeXML/Util/Image.pm
+++ b/lib/LaTeXML/Util/Image.pm
@@ -151,16 +151,18 @@ sub image_graphicx_parse {
       elsif (/^magnification$/) { $mag = $v; }
       else                      { push(@unknown, [$op, $v]); } }
     else { } }                                                     # ?
-   # --------------------------------------------------
-   # Now, compile the options into a sequence of `transformations'.
-   # Note: the order of rotation & scaling is significant,
-   # but the order of various clipping options w.r.t rotation or scaling is not.
+      # --------------------------------------------------
+      # Now, compile the options into a sequence of `transformations'.
+      # Note: the order of rotation & scaling is significant,
+      # but the order of various clipping options w.r.t rotation or scaling is not.
   my @transform = ();
-  # We ignore viewport & trim if clip isn't set, since in that case we shouldn't
-  # actually remove anything from the image (and there's no way to have the image
-  # overlap neighboring text, etc, in current HTML).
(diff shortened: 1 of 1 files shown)


brucemiller:
> The three images in Figure 11, "Analysis of runtimes", from 1803.07574 
> <https://ar5iv.org/source/1803.07574.zip> (link to source), have 
> reverted back to its uncropped A4 page size with a small plot in the 
> upper half.

I'm not seeing that; what options are you using?
I'm seeing 3 subfigures, all same, reasonable(?)  size.
I do see a dire need for proper (flex?) layout of subfigures, though.
Do you perhaps have the images cached wherever you're building?

dginev:
> a billion pts is kind of a red flag.

double-checked and this is indeed fixed with the latest commit, thanks!

brucemiller:
> another article I had tested with prior:
what article?

brucemiller:
> get the vertical A4 page PNGs with the default image settings.

Ha, those images have all bounding boxes set to A4. And trims 18cm off 
the bottom. But "Theoretically", trim should *not* clip, unless clip is 
set. I think the idea is to allow the image to overlap neighboring 
content.  Only, not so easy in html.

dginev (approved):
Right, and clip it does -- matching quite closely the PDF behavior I attached before. I'm still wondering whether letting the caller of latexml specify `autocrop` is better than always clipping, but maybe (hopefully?) there are few enough cases where this does visible harm to the PNGs.

image

Things look as expected with the other four arXiv papers I was testing, the magnify+zoomout combination works quite nicely.

PR #1698, opened 2021-10-20
Empty frontmatter

brucemiller:
This PR omits empty front matter, added by `\@add@frontmatter`.

It also fixes a misspelled command in jheppub.

However, it fails a test in `moderncv`, since that binding adds an empty creator with `role='cv'` and then adds separate firstname, lastname contacts, and cobbles together a title in xslt. That will need some adapting.


brucemiller:
Simple fix to `moderncv` gets it to pass the test, but I'm still not fond of the structure of the data in the xml (or the extra xslt). Hopefully that'll get cleaned up some day.

dginev:
I'll seize on the recent positive consensus around #1717 and close here, thanks!

PR #1700, opened 2021-10-23
Subfigures via flexbox

### LINKED ISSUE #1322: Vertical subfigures not working as expected
I am trying to get two subfigures to appear vertically like so
```tex
\documentclass{article}
\usepackage{graphicx}
\usepackage{subcaption}
\begin{document}
\begin{figure*}[t!]
	\centering
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{A.png}
    \caption{A}
    \end{subfigure}%
    
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{B.png}
    \caption{B}
    \end{subfigure}%
\end{figure*}
\end{document}
```

But this results in figures arranged horizontally, along the same row. I would expect this behaviour for the following code, and in fact, it does give the same result.

```tex
\documentclass{article}
\usepackage{graphicx}
\usepackage{subcaption}
\begin{document}
\begin{figure*}[t!]
	\centering
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{A.png}
    \caption{A}
    \end{subfigure}%
    ~
    \begin{subfigure}[t]{0.5\linewidth}
    \centering
    \includegraphics[width=0.8\linewidth]{B.png}
    \caption{B}
    \end{subfigure}%
\end{figure*}
\end{document}
```

Is there any way to force the figures to go into two rows. I am using the development version hosted here on Github.

dginev:
Fixes #1322 and #979 

The PR offers  a rudimentary refactor to using the CSS flexbox layout for subfigures, rather than HTML tables - allowing for responsive reflow. I have been actively learning about the model recently, so it will be of big benefit to have some extra testing and discussion here.

The support is rudimentary, and doesn't answer some harder questions:
1. Some PDF figures rely on TeX's positioning for the subfigure arrangement. E.g. 5 consecutive `\includegraphics` calls, where the produced PDF receives a TeX-deduced line-break after the second, and we get the 5 images positioned as `[2,3]`. Ideally, we would one day be able to infer there is an intended `break` at that position and insert it in the XML. The PR (and also the master branch) will currently position all 5 of those images on the same row (if they are small enough).

2. Some of the design intended by the author's image sizes may get lost in the flex model. This is a publisher problem, more than an author problem (and specifically my artistic problem in styling arXiv). 
   - Specifically, adding the flex attribute to each subfigure cell, e.g. `flex: 1 1 0px;`, or similar combinations of flex-grow, flex-shrink and flex-basis, may be hard to specify as one-size-fits-all. Allowing two subplots to grow into the page width helps the eyes see the data points better; but allowing two "texture images" generated by an AI GAN model to grow can end up looking quite ugly. It's hard to know which is better without understanding the figure content.
    - The PR doesn't directly encounter this problem, as it avoids any fine-tuning of the individual subfigures, so not a pressing issue.
    - Still, to help myself a little, I experimented with at least adding aspect-ratio classes on all images. Namely the three `ltx_img_(portrait|square|landscape)`. That allows some reasonable CSS to stretch only on the appropriate axis, which leads to some additional consistency in the flex cells compared to the original PDFs.

So, in summary, I've tried to add useful anchors, to set up the basic flex arrangement, and to defer additional styling to third-party vendors. Any fishing to improve further is welcome.

---

A good example from the original issue is [1602.07188](https://ar5iv.org/source/1602.07188.zip). One can try the changed LaTeXML.css on it via:
```
latexmlc 1602.07188.zip --dest=test.html --preload=[localrawstyles,ids,nobibtex,magnify=2,zoomout=2]latexml.sty
```

or a different, fixed-width styling approach from arxmliv.css by also adding the options:
```
--nodefaultresources --css=http://cdn.jsdelivr.net/gh/dginev/arxmliv-css@0.5.0/css/arxmliv.css
```


dginev:
Update 1:

I have added the logic that adds a `ltx:break` when subfigures exceed the `\textwidth` limit. I've only tested the 1602.07188 example so far, and it now matches up nicely with the PDF arrangement.

Update 2, re:
> Aside: for figures 3 & 4, it's also curious that those figures end up smaller (in the arxiv version), which doesn't seem right.

Some trade-off choices there, in particular with choosing a max-width. My logic was only sophisticated enough to find the largest row (in number-of-subfigures) and constrain all rows to that sizing. To get the ideal sizing I'd need to move the sizing classes to each individual flex cell (each subfigure), but that seemed tricky to do in XSLT 1.0, as adjacency querying is quite painful. I could add it as a final retouch in `insertBlock`... This reminds me: we should likely discuss the PR on a call, so that we can walk through the details and scratch our heads together what is the best implementation style for each piece.

Update 3, re:
> the non-arxiv style should probably allow the figures to unwrap more to the line.

I think that can be done with strategic `@media` queries in CSS, where the `ltx_break` elements are styled as ignored, rather than as breaks. But it's also a bit tricky, for the same reason a "dark mode" CSS is tricky -- sometimes the captions use text that is very committed to the specific geometry of the subfigures. E.g. in this article, we have a 2x3 grid with the caption:

>  Top row: the source photo (left), the style image (right, [opt]). Bottom row: results of 500 iterations starting from style (left), content (center) and noise (right) images.

Where any flex liberties invalidate the caption. So maybe even on phone displays some visual mark where the break would have been would help... Certainly open design questions on that end.

Done working here for now, removing the WIP, and we can discuss how to finalize the PR, when convenient.

dginev:
I looked into subtables and floats, discovering that also listings and minipages need a rethink if I am to treat all `ltx:float` tags.

Extra articles I tested with to that end:
1307.6769, 1712.01103 and 1911.01537

So far I haven't updated the LaTeXML.css accordingly, prototyping with the [ar5iv.css](https://github.com/dginev/ar5iv-css/blob/main/css/ar5iv.css) first. The PR is again work-in-progress, since there is a lot of territory to cover in the subfloats case that I hadn't explored until today. But I'm getting some really nice arXiv upgrades, and some extra minor polish on some of the sizing bits.

brucemiller:
hmm, something went wrong; I tried it on the 1st mentioned paper and all the figures are lined up vertically...

brucemiller:
Oh, curious; running your latexmlc (from a separate git clone) is picking up the css from *my* latexml...

Seems to work as expected if I copy your css over, but we out to figure out the misdirection.

dginev:
Some small cleanup and looking into details here and I think I've convinced myself I need to always have a wrapping `<div>` element for the subfigure content, be it an `img`, `table` or a listings `div`, as it allows to separate sizing the contents from sizing the subfigure container. It feels a bit redundant at times, but it's needed to get optimal rendering. Commit coming a little later.

Feeling somewhat happy with how I ended up styling listings for now btw, even though there are some very peculiar tag sets getting generated in the arXiv articles on occasion.

dginev:
I ended up convincing myself today that `{minipage}` should really be using percentage widths relative to `\textwidth` already in the bindings - which allows for very dynamic restyling options for the final CSS, much more flexible than the absolute `pt` numbers.

I added enough rules for the expected relative arrangement to hold in LaTeXML.css, which is too wide on my screen, since the CSS is using the full available width. With a fixed-width CSS (ar5iv.css), I feel quite good about the outcome:

<details>

Screenshot 2021-12-15 at 19-10-50 Untitled Document

</details>

dginev:
I think we should try to get a state near the current one merged at this point, since I am starting to look into scary directions where I overhaul the entire default LaTeXML.css to one with a fixed `var(--text-width)` (and maybe associated other variables) to which we can output relative widths via `calc()` as in `width="calc(0.33*var(--text-width))"`.

But that is best postponed until at least 2022, I just want a basic flexbox state to work with subfigures for now. I've tested a sample document that combines minipage, listings, tables and includegraphics, and it seems to be holding up quite reasonably, although I'm sure there are still edge cases needing to get caught.

But I think this is ready for in-depth review.

dginev:
Debugging with [2105.04404](https://ar5iv.org/html/2105.04404) today has revealed some further details that need polishing. Not to discourage anyone from reviewing what's already in the PR :>

dginev:
Ok, the latest state is in - hopefully the final one for this PR. I now enforced a wrapping div with a generic `ltx_flex_cell` class, and styled it in painful detail in ar5iv.css. The LaTeXML.css is left somewhat barebones, and easy to override - we can revise that as we go forward. The hope is that the HTML scaffold itself wouldn't have to change, even as the CSS gets ever more correct for ever more flex combinations. arXiv keeps surprising with the kinds of mashups of various kinds of contents and aspect ratios that show up.

I'm perfectly happy If we merge here as-is, as of now.

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -3204,16 +3205,102 @@ DefConstructor('\@@generic@caption[]{}', "<ltx:text class='ltx_caption'>#2</ltx:
...
+  if ($contents_count > 1) {
+    @contents = add_subfigure_row_breaks($document, $node, $whatsit, @contents);
+    add_subfigure_row_sizes($document, $node, $whatsit, @contents); }
+  # 2. IDs
+  # Note that even without \caption, we'd probably like to have xml:id.
+  return GenerateID($document, $node, $whatsit, $id_kind); }
+
+Tag('ltx:figure', afterClose => sub { SubfigureAndID(@_, 'fig'); });
+Tag('ltx:float',  afterClose => sub { SubfigureAndID(@_, 'tab'); });
+Tag('ltx:table',  afterClose => sub { SubfigureAndID(@_, 'tab'); });
```
brucemiller, line 3297: Just for the record: Tag accumulates the daemons, so you could have just added your subfigure code with separate `Tag`s.

brucemiller:
Fascinatingly scary! Let's give it a try. We can always evolve from here. Thanks for all the work!!

PR #1717, opened 2021-11-20
Drop empty frontmatter; styled frontmatter notes

dginev:
I am back with a new offer in place of the two previous PRs #1698 and #1689 .

1. I am still convinced dropping empty frontmatter nodes at the end of XML construction is the best time for this sanitization action. As Bruce has encountered with moderncv, existing+tested bindings use empty frontmatter nodes as placeholders/state, which receive content before they are finally closed. If checked early, they get pruned too quickly. If checked late, they're seen as filled in, and kept.
    - But more importantly, the construction end has a single hot path in the code, where all cases can get handled together. The alternative of refactoring individual frontmatter macros means changing more than just `\@add@frontmatter`. There's also `\@add@to@frontmatter` and `{keywords}`, and possibly other macros I haven't seen yet, which use the 'frontmatter' storage in state.

2. Separately, I figured out that adding a dedicated class to frontmatter notes allows me to style them uniformly, and sufficiently well. 
  
3. I kept the typo fix in jheppub

Hopefully we find some way to common ground here.


lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -5035,11 +5035,23 @@ sub insertFrontMatter {
       local $LaTeXML::BOX = Box('', $STATE->lookupValue('font'), '', T_SPACE);
       foreach my $item (@$list) {
         my ($tag, $attr, @stuff) = @$item;
+        # add a dedicated class for frontmatter notes,
+        # in the case we want to style those uniformly.
+        if ($tag eq 'ltx:note') {
+          $attr ||= {};
+          $$attr{class} = ($$attr{class} ? $$attr{class} . ' ' : '') . 'ltx_note_frontmatter'; }
```
brucemiller, line 5042: Since everything seen here is frontmatter of some sort, I'm wondering if this should just add `ltx_frontmatter`, or possibly `$tag . '_frontmatter'` (s/:/_/g) ?

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -5035,11 +5035,23 @@ sub insertFrontMatter {
...
         map { $document->absorb($_) } @stuff;
-        $document->closeElement($tag); } } }
+        my $completed_node = $document->closeElement($tag);
+        # At this time, the frontmatter element should really carry the actual literal values intended.
+        # Thus, if we see an empty element, something went wrong -- including our bindings are too verbose,
+        # as e.g. \preprint{} always generates a ltx:note element.
+        #
+        # To solve this in a single location: prune here!
+        if (($tag ne "ltx:rdf") && !scalar($completed_node->childNodes)) {
+          $document->removeNode($completed_node); } } } }
```
brucemiller, line 5054: Yeah, um, perhaps yeah; it's at least simple :>

lib/LaTeXML/Package/TeX.pool.ltxml
dginev: Alright, updated to a generic `ltx_frontmatter` class, which sadly also means I updated tens of tests in the process. 

The styling I'm interested in can now be done via a CSS selector on `.ltx_note.ltx_frontmatter`, which is perfectly usable.

brucemiller: Ugh. and having that class on `ltx:title`, `ltx:creator`, etc is annoyingly pointless. Maybe you were right the first time?

dginev: Luckily reverting is much, much faster :> I've rewinded things back.

brucemiller:
OK, let's go with this!

Issue #1685, opened 2021-10-08
[arXiv] community article upgrades


dginev:
I was doing some good samaritan user support [on twitter](https://twitter.com/BarakShoshany/status/1446572969434796043) today, in order to prioritize some community requests along with the statistical imperatives.

Reports based on arxiv-vanity on october 8, 2021:

articles failed to render:
  - [x] [1811.04360](https://ar5iv.org/html/1811.04360) -- infinite loop in tikz (PR #1701)
  - [x] [1904.06386](https://ar5iv.org/html/1904.06386) -- infinite loop in tikz (PR #1701)
  - [x] [1912.02922](https://ar5iv.org/html/1912.02922) -- infinite loop in tikz (PR #1701)
  - [x] [1911.11590](https://ar5iv.org/html/1911.11590) -- this seems to already be better

code blocks are messed up:
  - [x] [2105.00613](https://ar5iv.org/html/2105.00613)

I'm hoping to debug these in a couple of weeks, so recording them here as a placeholder and getting the assignment flag.

PR #1723, opened 2021-12-01
Upgrades for converting arXiv:2105.00613

dginev:
This addresses the last paper in my latest arXiv debugging session from #1685 , namely [2105.00613](https://ar5iv.org/html/2105.00613).

I know Bruce is not a fan of sweeping changes based on debugging a single paper, but maybe it can be a start to get a complete solution fleshed out.

Changes that were needed:
 * a bit more care in hyperref to make sure `ltx:anchor` descends down to an element that can allow itself to be wrapped by an anchor. The paper was using `\hypertarget` over a section, which is invalid in our schema - likely rightfully so. The patch descends to the text node of the section title and wraps that instead.
 * a number of stub bindings that seemed to do a reasonable job when interpreted raw - but I've only tested on this one paper. They were for: 
     * atveryend.sty, auxhook.sty, bigintcalc.sty, bitset.sty, bookmark.sty, fancyvrb.sty, gettitlestring.sty, ifetex.sty, iftex.sty, infwarerr.sty, intcalc.sty, kvdefinekeys.sty, kvoptions.sty, kvsetkeys.sty, ltxcmds.sty, pdftexcmds.sty, refcount.sty, upquote.sty, xurl.sty
* a handful of minor latex.pool macros - `\@include`, `\@endparenv`, `\@checkend`
* and a handful of empty stubs in etex.pool - `\pdftexcmds@directlua` and `\directlua`.
* a few extra package dependencies added to nameref.sty.ltxml
* `\Urlmuskip` added to url.sty.ltxml

After which the paper passes without any problems, with very nicely rendered highlighted code blocks - based on a custom `\DefineVerbatimEnvironment` definition, with a custom dialect of color macros. So a hand-made listing technique, which was introduced in the paper itself, and ends up working just fine under latexml.


brucemiller:
Massive, yes, but not really sweeping or that specifically tuned to this particular paper; just a collection of plausible patches where this paper gives reason for optimism. :>

So, sure, why not?

Issue #1703, opened 2021-11-03
Position-dependent \label in align


molke-productions:
If the label is before (the alignment character `&`?), then `\eqref` won’t find the ID:
```tex
Equations \eqref{eq:before} and \eqref{eq:after}:
\begin{align}
\label{eq:before}
&x \quad c_1,
\end{align}
and
\begin{align}
&y \quad c_2,
\label{eq:after}
\end{align}
```
In the resulting XML, the IDs and the equation tags  are apparently correct.

molke-productions:
As a follow-up, this seems to be a more general problem with align/aligned regarding the alignment character.
For example,
```latex
\begin{align}
\begin{aligned}
& a\\
& b
\end{aligned}
\end{align}
```
won’t even compile (starting with `Error:misdefined:&`), whereas the same in `equation` instead of `align` works fine.

dginev:
This is definitely suspiciously broken, I reproduced the report exactly as written. I also stumbled on a possibly related issue from arXiv where leading labels of eqnarray do not get recorded. Maybe even the same exact cause, so I'm recording my minimal example here:

```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}
\begin{eqnarray}
\label{eq:this}
\sum_{k=1}^{M}\mathrm{prob}(Y_k|X,I)&=&1.
\end{eqnarray}

Ref \ref{eq:this}
\end{document}
```

PR #1725, opened 2021-12-02
Table and ref fixes for arxiv:1804.11028

dginev:
This is an independent PR from me randomly checking a Fatal-error conversion from arXiv, which had an easy patch. The PR upgrades to a Warning status, where one of the remaining issues is my minimal example in #1703 

So, the main problems encountered here were due to:
 * capitalization typo in some longtable registers, and adding some extras
 * locking the \tabulary definition, as redefinitions using the .sty internals lead to unpredictable errors+loops, which lead too 100+ error fatals.
 * locking the \vref and related macro definitions in varioref.sty.ltxml, since modifying them using the internal .sty code also leads to unpredictable breakages.

That's all! The resulting HTML is very high quality from what I could tell, again with the notable exception of the \label issue in #1703 .

### DIFF
--- a/lib/LaTeXML/Package/longtable.sty.ltxml
+++ b/lib/LaTeXML/Package/longtable.sty.ltxml
@@ -125,8 +125,8 @@ DefPrimitive('\@longtable@@toccaption{}', sub {
 
 #======================================================================
 # Not used, but must be defined.
-DefRegister('\LTleft'      => Glue(0, '1fill'));
-DefRegister('\LTright'     => Glue(0, '1fill'));
+DefRegister('\LTleft'      => Glue('0pt plus 1fill'));
+DefRegister('\LTright'     => Glue('0pt plus 1fill'));
 DefRegister('\LTpre'       => Glue('12pt plus 4pt minus 4pt'));
 DefRegister('\LTpost'      => Glue('12pt plus 4pt minus 4pt'));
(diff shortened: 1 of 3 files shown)


lib/LaTeXML/Package/longtable.sty.ltxml
```diff
@@ -125,12 +125,22 @@ DefPrimitive('\@longtable@@toccaption{}', sub {
...
 #======================================================================
 # Not used, but must be defined.
-DefRegister('\LTLeft'      => Glue(0));
-DefRegister('\LTRight'     => Glue(0));
-DefRegister('\LTPre'       => Glue(0));
-DefRegister('\LTPost'      => Glue(0));
-DefRegister('\LTcapwidth'  => Glue(0));
-DefRegister('\LTchunksize' => Number(20));
+DefRegister('\LTleft'      => Glue(0, '1fill'));
+DefRegister('\LTright'     => Glue(0, '1fill'));
```
brucemiller, line 129: Oooh, bad cargo culting; this is going to fail after #1724. Should write `Glue('0pt plus 1fill')`.

dginev: Exactly as you say, and good that #1724 cleans that up too!

brucemiller:
OK, looks good; Thanks!

PR #1726, opened 2021-12-04
patch \graphicspath for ZIP inputs

dginev:
Minor, but annoying in the arXiv previews - articles that used the `\graphicspath` macro were [not picking up](https://ar5iv.org/html/1804.11028) their images, as the source directory wasn't getting used (and the ZIP inputs change it).

If one wants to reproduce the issue:
```
latexmlc 1804.11028.zip --dest=test.html
```

Once the source directory is properly used, it gets propagated to post-processing via the usual path and the HTML has images again, even in a zip-to-zip conversion.

### DIFF
--- a/lib/LaTeXML/Package/graphics.sty.ltxml
+++ b/lib/LaTeXML/Package/graphics.sty.ltxml
@@ -223,9 +223,7 @@ DefConstructor('\graphicspath DirectoryList', sub {
     my @paths = ();
     my $root  = $STATE->lookupValue('SOURCEDIRECTORY') || '';
     foreach my $dir ($paths->getValues) {
-      $dir = pathname_canonical(ToString($dir));
-      $dir = pathname_concat($root, $dir) if $root && !pathname_is_absolute($dir);
-      my $path = pathname_absolute($dir);
+      my $path = pathname_absolute(pathname_canonical(ToString($dir)), $root);
       push(@paths, $path);
       PushValue(GRAPHICSPATHS => $path); }
     return (paths => [@paths]); });
--- a/lib/LaTeXML/Package/graphics.sty.ltxml
+++ b/lib/LaTeXML/Package/graphics.sty.ltxml
@@ -221,8 +221,11 @@ DefConstructor('\graphicspath DirectoryList', sub {
   properties => sub {
     my ($stomach, $paths) = @_;
     my @paths = ();
+    my $root  = $STATE->lookupValue('SOURCEDIRECTORY') || '';
     foreach my $dir ($paths->getValues) {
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -221,8 +221,11 @@ DefConstructor('\graphicspath DirectoryList', sub {
   properties => sub {
     my ($stomach, $paths) = @_;
     my @paths = ();
+    my $root  = $STATE->lookupValue('SOURCEDIRECTORY') || '';
     foreach my $dir ($paths->getValues) {
-      my $path = pathname_absolute(pathname_canonical(ToString($dir)));
+      $dir = pathname_canonical(ToString($dir));
+      $dir = pathname_concat($root, $dir) if $root;
```
brucemiller, line 227: Isn't it a problem to unconditionally prefix `$root` if the directory already was absolute (and you're *not* in a zip)?

dginev: You're right, I've introduced a bug when using absolute paths in the argument. My treatment should only be done in the relative path case.

Btw in general I also don't think the ZIP flow could recover an absolute path given via `\graphicspath`, so that case is likely best left as-is. I'll make a small test and retry, thanks for catching this.

dginev: Tested and patched, exactly as you described it.

lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -221,8 +221,11 @@ DefConstructor('\graphicspath DirectoryList', sub {
   properties => sub {
     my ($stomach, $paths) = @_;
     my @paths = ();
+    my $root  = $STATE->lookupValue('SOURCEDIRECTORY') || '';
     foreach my $dir ($paths->getValues) {
-      my $path = pathname_absolute(pathname_canonical(ToString($dir)));
+      $dir = pathname_canonical(ToString($dir));
+      $dir = pathname_concat($root, $dir) if $root && !pathname_is_absolute($dir);
+      my $path = pathname_absolute($dir);
```
brucemiller, line 228: Is this better (or different) than simply passing `$root` as the base directory, as in: ```$path=pathname_absolute(pathname_canonical(ToString($dir)), $root);```
since that's what that argument is intended for?

dginev: Oh, my oversight that this was supported so easily - upgraded, thanks!

dginev:
Ah, there was a rebase mistake...

dginev:
There. Too many master branches on my local cloned repo, I rebased to the wrong one by accident.

PR #1728, opened 2021-12-07
Small tracing upgrade to Registers

dginev:
A bit of tracing that helped me debug the infinite loops from #1685 as a standalone PR.

### DIFF
--- a/lib/LaTeXML/Core/Definition/Expandable.pm
+++ b/lib/LaTeXML/Core/Definition/Expandable.pm
@@ -75,7 +75,7 @@ sub invoke {
       Debug($self->tracingArgs(@args)) if @args; } }
   elsif (!$$self{parameters}) {                    # Trivial macro
     LaTeXML::Core::Definition::startProfiling($profiled, 'expand') if $profiled;
-    Debug($self->tracingCSName . ' -> ' . tracetoString($expansion))
+    Debug($self->tracingCSName . ' ->' . tracetoString($expansion))
       if $tracing || $LaTeXML::DEBUG{tracing};
     # For trivial expansion, make sure we don't get \cs or \relax\cs direct recursion!
     if (!$onceonly && $$self{cs}) {
@@ -96,7 +96,7 @@ sub invoke {
           && (($r eq 'LaTeXML::Core::Token') || ($r eq 'LaTeXML::Core::Tokens'))
         ? $_ : Tokens(Revert($_))); } @args;
     if ($tracing || $LaTeXML::DEBUG{tracing}) {    # More involved...
-      Debug($self->tracingCSName . ' -> ' . tracetoString($expansion));
+      Debug($self->tracingCSName . ' ->' . tracetoString($expansion));
       Debug($self->tracingArgs(@targs)) if @args; }
     LaTeXML::Core::Definition::startProfiling($profiled, 'expand') if $profiled;
     $result = $expansion->substituteParameters(@targs); }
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Core/Definition/Expandable.pm
```diff
@@ -96,7 +96,7 @@ sub invoke {
           && (($r eq 'LaTeXML::Core::Token') || ($r eq 'LaTeXML::Core::Tokens'))
         ? $_ : Tokens(Revert($_))); } @args;
     if ($tracing || $LaTeXML::DEBUG{tracing}) {    # More involved...
-      Debug($self->tracingCSName . ' -> ' . tracetoString($expansion));
+      Debug($self->tracingCSName . ' ->' . tracetoString($expansion));
```
dginev, line 99: These space removals are not typos - TeX doesn't have a space after these two arrows, so it's easier to diff without that space added in latexml.

lib/LaTeXML/Core/Definition/Register.pm
```diff
@@ -44,7 +45,15 @@ sub valueOf {
...
 sub setValue {
   my ($self, $value, @args) = @_;
-  &{ $$self{setter} }($value, @args);
+  my $tracing = $STATE->lookupValue('TRACINGCOMMANDS');
+  if ($tracing || $LaTeXML::DEBUG{tracing}) {
+    my $scope  = $STATE->getPrefix('global') ? 'globally ' : '';
+    my $csname = $$self{cs}->toString;
+    Debug("{$scope" . "changing " . $csname . "=" . $self->valueOf()->toString() . "}");
+    &{ $$self{setter} }($value, @args);
+    Debug("{into " . $csname . "=" . $self->valueOf()->toString() . "}"); }
```
brucemiller, line 54: Cool; this looks useful. But since it's a TeX enhancement, wouldn't it be more pleasant to keep on 1 line, like maybe:
```
Debug("{$scope assigning $csname=".$self->valueOf()->toString()."} [previously $prev]");
```
? (with the 1st `Debug` replaced by `my $prev=...`)

dginev: It really depends on how you're planning on using it. The reason I dropped the space for the arrow above for instance is a bit frustrating since the printed output looks clearly *worse*. So if we're optimizing tracing in latexml, assuming a correct latexml implementation, your suggestion is indeed an improvement.

But if the workflow is to open two windows side by side, one with the pdflatex log, the other with the latexml log, having the output on individual lines gets the diffs closer to each other. Similarly if you ran the actual `diff` algorithm.

I'm open to changing it to a single line, just explaining that the 2-line output was also useful for what I was debugging.

brucemiller: Maybe I'm missing what TeX would have been printing; does it print any trace while such assignments are taking place? If so, then getting close to its output is best; otherwise making the (informative) intrusion as short as possible makes it easier to compare.

dginev: Right, if `\tracingall` is set, tex will be printing some messages. For:
```tex
\tracingall
\newcount\foo
\global\foo=2
\bye
```

you'd see:
```
{globally changing \foo=undefined}
{into \foo=\relax}
{globally changing \foo=\relax}
{into \foo=\count27}
%... some further internal details
{globally changing \count27=0}
{into \count27=2}
```

The PR is a little higher level since the latexml log will only show you:
```
{globally changing \foo=0}
{into \foo=2}
```

It can change either way, just an example

brucemiller: Ah, ok; seems surprisingly verbose for Knuth, but let's go with your original.

PR #1730, opened 2021-12-08
Correct right german quote marks 

### LINKED ISSUE #1673: babel.def.ltxml: German quotes
The definitions seem to point to the wrong Unicode:
https://github.com/brucemiller/LaTeXML/blob/be6bbab59dbf48b7834d96db65c49226c27a96b4/lib/LaTeXML/Package/babel.def.ltxml#L41-L44
That is, `\grq` should be U+2018 and `\grqq`should be U+201C (https://en.wikipedia.org/wiki/Quotation_mark#German).
In LaTeX’s babel.def, this is `\textquoteleft` and `\textquotedblleft`, respectively (referring to the left one in English).

brucemiller:
As it says; Yep, that looks better. Thanks molke!

fixes #1673


dginev (approved):
Nice!

PR #1732, opened 2021-12-08
[minor] variety robustness patches

dginev:
In the process of developing with [1804.11021](https://ar5iv.org/html/1804.11021) for flex subfigures, I debugged a few unrelated issues and factored the patches out in this separate PR.

Details of the tidbits:
 - [guard] deduplicate the searchpaths list before writing in the processing instructions in the core XML
 - Make sure counter arguments are expanded so that uses such as `\addtocounter{\@captype}{...}` work as expected
 - handle name arg of `{IEEEbiography}` and also allow for inline-block content - the article has multiple paragraphs
 - a handful of extra small auxiliary macros in LaTeX.pool
 - make sure the `{document}` redefinition in subfiles.sty leaves preamble mode.

This gets 1804.11021 to a "no problem" status and I can now focus on getting just the right subfigure flavor.

### DIFF
--- a/t/structure/IEEE.xml
+++ b/t/structure/IEEE.xml
@@ -1005,7 +1005,9 @@ Stuff, More Stuff
       <tabular>
         <tr>
           <td>Picture of Me</td>
-          <td><p><text font="bold">ME</text> is a really great guy.</p></td>
+          <td><inline-block>
+              <p><text class="ltx_font_bold"><text font="bold">ME</text> is a really great guy.</text></p>
+            </inline-block></td>
         </tr>
       </tabular>
     </float>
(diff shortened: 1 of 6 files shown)


t/structure/IEEE.xml
```diff
@@ -1005,7 +1005,9 @@ Stuff, More Stuff
       <tabular>
         <tr>
           <td>Picture of Me</td>
-          <td><p><text font="bold">ME</text> is a really great guy.</p></td>
+          <td><inline-block>
+              <p><text class="ltx_font_bold"><text font="bold">ME</text> is a really great guy.</text></p>
```
dginev, line 1009: ah, i need to revise the test further here actually...

dginev:
Should be good to review now.

PR #1735, opened 2021-12-12
Lexeme fixes from arXMLiv 2020

### LINKED ISSUE #1465: Further lexeme upgrades
An issue to collect lexeme upgrades for the next time we rerun arXiv, if I may.
One unfortunate observation is that the current code lexes `\underline{A}` and `\overline{A}` the same, namely as:

```
¯ start_ARG italic_A end_ARG
```

They ought to be distinct, and there may be a better syntax. Maybe:
```
¯ overarg start_ARG italic_A end_ARG
¯ underarg start_ARG italic_A end_ARG
```

dginev:
Fixes #1465 .

Addresses the two easy+obvious cases in the issue, deferring more advanced problems (long annotations, color lexemes) for 2022.

This should be all that's needed for the arXMLiv 2021 rerun, lexeme-wise.

### DIFF
--- a/lib/LaTeXML/MathParser.pm
+++ b/lib/LaTeXML/MathParser.pm
@@ -714,7 +714,10 @@ sub node_to_lexeme {
           if (my $value = $$properties_pending{$attr}) {
             push @to_add, $value; } }
         if (@to_add) {
-          my $prefix = join("_", sort(@to_add)) . "_";
+          # avoid e.g. "italic italic" from "\mit \Omega"
+          my %seen          = ();
+          my @unique_to_add = grep { my $new = !$seen{$_}; $seen{$_} = 1; $new; } @to_add;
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/MathParser.pm
```diff
@@ -714,7 +714,10 @@ sub node_to_lexeme {
           if (my $value = $$properties_pending{$attr}) {
             push @to_add, $value; } }
```
brucemiller, line 715: Wouldn't it have been simpler to simply make `@to_add` a hash in the first place, and then `$to_add{$value}=1;` and use `keys %to_add` later ?

dginev: sure, can do!

brucemiller:
cool; thanks!

PR #1738, opened 2021-12-13
Emulate more psfile options via includegraphics

dginev:
This PR is meant to be a small stir of a recent hot pot :>

I was debugging [arXiv:hep-ph/0105138](https://ar5iv.org/html/hep-ph/0105138) which is no longer a timeout in the latest master. But, that allowed to notice the images reached HTML rather skewed, with an artificial A4/Letter-sized white frame. In other words, the same problem we started discussing at #1683 

Unsurprisingly, adding the `autocrop => 1` flag to the `.ps` case of Graphics.pm specifically reaches an optimal looking figure, as you can enjoy in this screenshot:
image

The images in question all come from the `\special` directive which I recently enabled in #1616. In this case the source was:
```tex
\special{psfile=g0.ps   voffset=-200
hoffset=-20  angle=0   vscale=80  hscale=80 }
```

I'm not sure if these options would be useful to emulate to preserve the original aspect ratio, the autocrop is certainly the simplest one-line patch. Maybe they can be translated into a mixture of `trim`, `clip` and `scale` from `\includegraphics` and achieve the same result? Wondering about your preference @brucemiller

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -2392,22 +2392,71 @@ DefPrimitive('\special {}', sub {
     if ($special_str =~ /\bpsfile=(.+?)(?:\s|\})/) {
       my $graphic = $1;
       RequirePackage('graphicx', searchpaths_only => 1);
+      my @kv;
+      for my $prop (qw(voffset hoffset hscale vscale hsize vsize angle)) {
+        if ($special_str =~ /\b$prop=(.+?)(?:\s|\})/) {
+          push(@kv, T_OTHER(',')) if @kv;
+          push(@kv, T_OTHER($prop), T_OTHER("="), T_OTHER($1)); } }
+      @kv = (T_OTHER("["), @kv, T_OTHER("]")) if @kv;
       $stomach->getGullet->unread(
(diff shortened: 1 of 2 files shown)


brucemiller:
Will somebody please grab those scissors before he hurts someone!

The problem, as you suggest, is that the current `\special` is ignoring those keywords; the offsets are probably the biggest thing. Using autocrop would affect *all* ps images, so basically, the images that would don't treat correctly would improve, but the ones we do will be worse.

dginev:
Great (and entertaining) feedback, thanks!

I'm getting close with emulating the various offsets and sizes through includegraphics this morning, but the results are still not as good as the autocrop screenshot above... Will tinker a little more.

dginev:
Ok, I got a little bit closer to the autocrop result, though I had to do something strange - use the `voffset` and `hoffset` values to symmetrically trim on both vertical/horizontal sides, rather than on one only. Since we don't have a direct offset in includegraphics - and I didn't think this is the moment to implement one from scratch - I ended up with trying to use trims.

Not really sure this is more reliable than autocrop, but it does respect individual options better. Here is a screenshot of the same figure for comparison:
image

brucemiller:
OK, sounds plausible; Thanks!

PR #1739, opened 2021-12-13
Unwrap quoted filenames in \input

dginev:
Seemingly harmless(?) addition of unwrapping quoted filenames in input directives, such as `\input{"epsf.sty"}`.

Seen in long-tail cases e.g. in the [arxmliv warning report](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/warning/missing%5Ffile?all=false). There's also [a couple of SE comments](https://tex.stackexchange.com/questions/191930/spaces-in-files-names-in-input-or-includegraphics#comment442865_191930) mentioning that as the preferred method when the filename has spaces.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -2071,6 +2071,7 @@ my $input_options = {};    # [CONSTANT]
 sub Input {
   my ($request, %options) = @_;
   $request = ToString($request);
+  $request =~ s/^"(.+)"$/$1/;    # unwrap if in quotes \input{"file name"}
   CheckOptions("Input ($request)", $input_options, %options);
   # HEURISTIC! First check if equivalent style file, but only under very specific circumstances
   if (pathname_is_literaldata($request)) {
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -2071,7 +2071,7 @@ my $input_options = {};    # [CONSTANT]
 sub Input {
   my ($request, %options) = @_;
   $request = ToString($request);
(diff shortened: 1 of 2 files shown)


brucemiller:
Is it possible that there be multiple pairs of quotes? (and if so, what would be the correct treatment?)

dginev:
Ah, interesting! What you describe is a supported feature (just ran a few tests with my pdflatex) - any number of quotes is accepted, as long as they are properly balanced. E.g. this worked:
```tex
\input{""""""""""""""""""File Name""""""""""""""""""}
```

And on top of that the SE example also showed this is relied on in `\includegraphics` -- and I assume any other mechanism that leverages `\input` internally. I've quickly added a perl "regular" expression that matches this case (not regular any more, since we are using backreferences, but still quite effective).

Curious if you think I should add this to `pathname_canonical` as well, to make sure any input path in latexml unwraps quotes. Though then we would be at risk to misinterpret files that truly have quotes in the filename - but those wouldn't have been used in pdflatex in the first place.

brucemiller:
> Curious if you think I should add this to |pathname_canonical| as well, 

Well, I would, except for wondering what it might break in Windows
(or fix, for that matter!)

dginev:
> Curious if you think I should add this to |pathname_canonical| as well,
> Well, I would, except for wondering what it might break in Windows (or fix, for that matter!)

Maybe we defer that until we're sure we need it *everywhere* then? The PR is very close to the latex intention - bits relying on `\input` specifically have the quotation guards, rather than any latexml asset on the filesystem.

PR #1740, opened 2021-12-14
\qed needs to be expandable for arXiv

dginev:
As we've known for a while now, we have some arxiv articles that get emulated as "fork bombs" due to some latex macros being instead defined as latexml primitives/constructors. Spotted a really simple one today, which feels too simple to go unfixed:

```tex
\edef\qqed{\qed}
\def\qed{\qqed \medskip}
```

Allocates over 4 GB of RAM in 6 minutes when `\qed` is first invoked in the body of the document and dies through the [out-of-memory guard](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/cortex/oom?all=false). If `\qed` is always defined as a macro, that entire busy loop can be avoided.

Tested on [arXiv:math/0109221](https://ar5iv.org/html/math/0109221)

### DIFF
--- a/lib/LaTeXML/Package/IEEEtran.cls.ltxml
+++ b/lib/LaTeXML/Package/IEEEtran.cls.ltxml
@@ -403,8 +403,10 @@ Let('\iedlabeljustifyr', '\IEEEiedlabeljustifyr');
 Let('\QED',       '\IEEEQED');
 Let('\QEDclosed', '\IEEEQEDclosed');
 Let('\QEDopen',   '\IEEEQEDopen');
-DefConstructor('\qed',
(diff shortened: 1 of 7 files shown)


dginev:
A much more far-reaching solution idea here would be to automatically add locks to all latexml primitives/constructors (i.e. `locked => 1`) so that they can never be redefined as macros from within the TeX sources. However, we can be most certain that would lead to various other breakage in arXiv, as a lot of redefinitions change argument signatures, or even conditional structure. So that's no silver bullet either. Perfect emulation is the only perfect solution, sadly.

PR #1741, opened 2021-12-14
New guards: iflimit, absorblimit; subeqn.sty; misc patches

dginev:
I combined a few diverse bits in what is hopefully my last "tour of the territory" PR for arXiv before we start the big run.

* misc macros added: `\vsplit`, `\icmladdress`, `\@killglue`
* added a small binding for subeqn.sty
* locked the math font directives such as `\mathbf`
* silenced a couple of cosmetic perl "undef value" warnings from recent commits
* new guards following `tokenlimit` from #1736 - `iflimit` and `absorblimit`, which allow to catch some out-of-memory errors early, as well as reduce timeouts in those two code paths down to ~5 minutes.

Guard values that I've used that get the large books to pass:
 * tokenlimit: `99,999,999`
 * iflimit: `199,999`
 * absorblimit: `1,099,999`
 * pushbacklimit: `99,999`

The absorb limit value may be a little tight, it is just enough to get the 11 chapter book for arXiv:2105.10386 to pass through successfully.

Using these for e.g. arXiv:math/0408299, I now see:
```
Fatal:timeout:if_limit Conditional limit of 199999 exceeded, infinite loop?
	at TeX.pool.ltxml; line 748
	In Core::Definition::Conditional[\ifdim ... /home/deyan/perl5/lib/perl5/LaTeXML/Package/TeX.pool.ltxml; line 748
Conversion complete: 1 warning; 1 error; 1 fatal error; 1 missing file[diagram] (See /tmp/import_1992965.latexml.log)
Status:conversion:3

real	5m24.539s
user	5m21.697s
sys	0m2.680s
```

for a reasonable early-stop outcome. I am also going to up the latexml RAM limit to 5 GB in the latexml workers, again to get the big books passing without issue.

Edit: I also added a limit to pushback size, which is quite efficient in catching the "fork bomb" cases early, at least the ones which accumulate in the Gullet pushback - and I'm assuming that is most of the cases we're likely to spot in arXiv

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -668,6 +668,7 @@ lib/LaTeXML/Package/standalone.sty.ltxml
 lib/LaTeXML/Package/stfloats.sty.ltxml
 lib/LaTeXML/Package/stmaryrd.sty.ltxml
 lib/LaTeXML/Package/subcaption.sty.ltxml
+lib/LaTeXML/Package/subeqn.sty.ltxml
 lib/LaTeXML/Package/subfig.sty.ltxml
 lib/LaTeXML/Package/subfigure.sty.ltxml
 lib/LaTeXML/Package/subfloat.sty.ltxml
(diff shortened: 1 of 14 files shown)


brucemiller:
Scary, but let's give it a try

PR #1742, opened 2021-12-15
Make \label not disappear on reversion, except for tex strings

### LINKED ISSUE #1703: Position-dependent \label in align

brucemiller:
So, it appears that reversion is getting used for more substantial things than creating tex/content-tex strings for math & images, so we have to be more careful about what disappears and when. Ultimately may need a more fundamental fix, but this works for the most obvious case.

fixes #1703

### DIFF
--- a/lib/LaTeXML/Common/Font.pm
+++ b/lib/LaTeXML/Common/Font.pm
@@ -715,7 +715,7 @@ sub merge {
   $shape     = $$self[2] if (!defined $shape)  || ($oflags & $FLAG_FORCE_SHAPE);
   $size      = $$self[3] if (!defined $size);
   $color     = $$self[4] if (!defined $color);
-  $bg        = $$self[5] if (!defined $bg);
+  $bg        = $$self[5] if (!exists $options{background});
   $opacity   = $$self[6] if (!defined $opacity);
   $encoding  = $$self[7] if (!defined $encoding);
   $language  = $$self[8] if (!defined $language);
--- a/lib/LaTeXML/Common/Font.pm
+++ b/lib/LaTeXML/Common/Font.pm
@@ -502,7 +502,7 @@ sub computeStringSize {
     $w += int($cw * $size);
     if (my $kern = $chars[0] && $$metric{kerns}{ $char . $chars[0] }) {
       $w += int($size * $kern); }
-    if ($ismath) {
+    if ($ismath && $ci) {
       $w += int($size * $ci); }
     $h = max($h, int($ch * $size));
     $d = max($d, int($cd * $size)); }
--- a/lib/LaTeXML/Core/Alignment.pm
+++ b/lib/LaTeXML/Core/Alignment.pm
@@ -415,10 +415,11 @@ sub normalize_cell_sizes {
             . ") size " . showSize($cw, $ch, $cd)
             . " Boxes=" . ToString($boxes)) if $LaTeXML::DEBUG{halign} && $LaTeXML::DEBUG{size};
         my $empty =
-          ((!$cw) || $cw->valueOf < 1)
-          || (((!$ch) || $ch->valueOf < 1)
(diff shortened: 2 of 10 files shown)


dginev:
Very cool! I was going to suggest it's "too good to be true" with a one-liner fix, but indeed it may be -- my example in my [issue comment](https://github.com/brucemiller/LaTeXML/issues/1703#issuecomment-984239513) is still not detecting the label.

Any chance there is a connected second line that gets the eqnarray working as well?

dginev:
Really great! Not to rain on your parade, but I was double-checking all reported examples, and am seeing some new latexml Errors show up at the second entry in #1703 :
```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}

\begin{align}
\begin{aligned}
& a\\
& b
\end{aligned}
\end{align}

\end{document}
```

namely:
```
Error:misdefined:& The token T_ALIGN[&] should never reach Stomach! at asset1.tex; line 7 col 1
Error:unexpected:\endgroup Attempt to close non-boxing group at asset1.tex; line 9 col 0
Error:unexpected:\@hidden@egroup Attempt to close boxing group at asset1.tex; line 10 col 0
Warning:unexpected:\end{document} Attempt to end document with open groups, environments or conditionals at asset1.tex; line 12 col 0
Error:unexpected:<endgroup> Attempt to pop last locked stack frame at String; line 2 col 0
Error:unexpected:<endgroup> Attempt to pop last locked stack frame at String; line 2 col 0
Error:unexpected:<endgroup> Attempt to pop last locked stack frame at String; line 2 col 0
```

Any clues?

dginev:
All other examples from all other mentioned issues appear correct on my local machine!

So the error above is the only bit left from the reported cases, and maybe it can even be deferred to a later date given that `{align}` and `{aligned}` are not really commonly meant to be nested?

dginev (approved):
Code looks great to me, just read through. Also happy to see the amsdisplay.xml test evolving to match the generated PDF more closely.

brucemiller:
Jeez, I keep overlooking the piggybacked cases in #1703!  Seems coincidentally related, but I'll take a look.

dginev:
I am not sure if this is helpful or distracting, but I've tracked down another broken case, using the latest state in the PR here on the same arXiv article I had mentioned as related to the issue (1804.11028).

The example is:
```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}
\begin{eqnarray}
\nonumber
\mathrm{prob}(X|I)&=&1*\mathrm{prob}(X|I);\\
\nonumber
&\stackrel{\ref{eq:Normalization Requirement}}{=}&\sum_{k=1}^{M}\mathrm{prob}(Y_k|X,I)*\mathrm{prob}(X|I);\\
\label{eq:Summed Marginalization}
&\stackrel{\ref{eq:Product Expansion 3}}{=}&\sum_{k=1}^{M}\mathrm{prob}(X,Y_k|I).
\end{eqnarray}
\ref{eq:Summed Marginalization} is a generalization of
\end{document}
```

brucemiller:
OK, I think I've fixed them all (no more please! :> )  Take a last look, and hopefully it's good to go.

dginev:
Yes, this should be final! Writing to confirm the examples work in full, including the full 1804.11028 doc. Will quickly skim the code now...

dginev (approved):
All looks plausible! Couldn't find anything to be picky about in this one. Thanks!!!

PR #1744, opened 2021-12-16
Fallback to case-insensitive path search

dginev:
Debugging with [2105.04404](https://ar5iv.org/html/2105.04404), I noticed an image that was failing to get picked up.

The author wrote:
```tex
\includegraphics[width=0.9\textwidth]{illu_blur_MNIST.pdf}
```

but included the file `illu_blur_mnist.pdf`. Naturally,  pdflatex finds the file despite the discrepancy in case, while latexml didn't - as we're looking for exact matches.

To add a simple and effective patch, I created a fallback sub `pathname_findall_nocase`, which takes the slow route of checking all paths, normalized to lowercase, for the desired one, also normalized to lowercase.

Hopefully this catch improves the visuals for all kinds of arXiv articles that were authored on case-insensitive file systems.

### DIFF
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -327,7 +327,7 @@ sub pathname_findall {
     # To allow arbitrarily cased spellings,
     # lowercase the target path and all candidate paths
     $pathname = lc($pathname);
-    @paths    = grep { -f $_ } grep { lc($_) =~ /$pathname/ } keys %star_candidates; }
+    @paths    = grep { -f $_ } grep { lc($_) =~ /\Q$pathname\E/ } keys %star_candidates; }
   return @paths; }
 
 # It's presumably cheep to concatinate all the pathnames,
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -317,15 +317,16 @@ sub pathname_findall {
   return unless $pathname;
   my @paths = grep { -f $_ } candidate_pathnames($pathname, %options);
   if (!@paths) {    # Fallback:
-     # To allow arbitrarily cased spellings,
-     # lowercase the target path and all candidate paths
-    $pathname = lc($pathname);
+        # This is a bit of a slow call, but really quite OK to run in rare fallback cases
+        # an average arXiv article may return with ~50-100 pathnames when looking for '*' graphics
+    my ($pathdir, $name, $type) = pathname_split($pathname);
     my %star_candidates = ();
-    # This is a bit of a slow call, but really quite OK to run in rare fallback cases
-    # an average arXiv article may return with ~50-100 pathnames when looking for '*' graphics
-    for my $c (candidate_pathnames('*', %options)) {
+    for my $c (candidate_pathnames($pathdir ? "$pathdir/*" : '*', %options)) {
       $star_candidates{$c} = 1; }
-    @paths = grep { -f $_ } grep { lc($_) =~ /$pathname/ } keys %star_candidates; }
+    # To allow arbitrarily cased spellings,
+    # lowercase the target path and all candidate paths
+    $pathname = lc($pathname);
+    @paths    = grep { -f $_ } grep { lc($_) =~ /$pathname/ } keys %star_candidates; }
   return @paths; }
 
 # It's presumably cheep to concatinate all the pathnames,
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -377,18 +377,18 @@ sub candidate_pathnames {
   # Now, combine; precedence to leading directories.
   foreach my $dir (@dirs) {
     opendir(DIR, $dir) or next;
-    for my $local_file (readdir(DIR)) {
+    my @dir_files = readdir(DIR);
+    closedir(DIR);
+    for my $local_file (@dir_files) {
       for my $regex_pair (@regexes) {
         my ($i_regex, $regex) = @$regex_pair;
         if ($local_file =~ m/$i_regex/) {
           my $full_file = pathname_concat($dir, $local_file);
           push(@nocase_paths, $full_file);
-          # Always check if a strict match also exists on disk:
-          if (($local_file =~ m/$regex/) && (-f $full_file)) {
+          if ($local_file =~ m/$regex/) {
             # if we are only interested in the first match, return it:
             return ($full_file) if $options{findfirst};
-            push(@paths, $full_file); } } } }
-    closedir(DIR); }
+            push(@paths, $full_file); } } } } }
   # Fallback: if no strict matches were found, return any existing case-insensitive matches
   # Defer the -f check until we are sure we need it, to keep the usual cases fast.
   return @paths ? @paths : grep { -f $_ } @nocase_paths; }
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -33,7 +33,8 @@ use File::Copy;
 use File::Which;
 use Cwd;
 use base qw(Exporter);
-our @EXPORT = qw( &pathname_find &pathname_findall &pathname_kpsewhich
+our @EXPORT = qw( &pathname_find
+  &pathname_findall &pathname_findall_nocase &pathname_kpsewhich
   &pathname_make &pathname_canonical
   &pathname_split &pathname_directory &pathname_name &pathname_type
   &pathname_timestamp
@@ -318,6 +319,20 @@ sub pathname_findall {
   my @paths = candidate_pathnames($pathname, %options);
   return grep { -f $_ } @paths; }
 
+sub pathname_findall_nocase {
+  my ($pathname, %options) = @_;
+  return unless $pathname;
+  # To allow arbitrarily cased spellings,
+  # lowercase the target path and all candidate paths
+  $pathname = lc($pathname);
+  my %star_candidates = ();
+  # This is a bit of a slow call, but really quite OK to run in rare fallback cases
+  # an average arXiv article may return with ~50-100 pathnames when looking for '*' graphics
+  for my $c (candidate_pathnames('*', %options)) {
+    $star_candidates{$c} = 1; }
+  my @paths = grep { -f $_ } grep { lc($_) =~ /$pathname/ } keys %star_candidates;
+  return @paths; }
+
 # It's presumably cheep to concatinate all the pathnames,
 # relative to the cost of testing for files,
 # and this simplifies overall.
@@ -405,7 +420,7 @@ sub pathname_kpsewhich {
   return; }
 
 sub build_kpse_cache {
-  $kpse_cache = {};            # At least we've tried.
+  $kpse_cache = {};    # At least we've tried.
   return unless $kpsewhich;
   # This finds ALL the directories looked for for any purposes, including docs, fonts, etc
   $kpse_toolchain = "--miktex-admin" if ($ENV{"LATEXML_KPSEWHICH_MIKTEX_ADMIN"});
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -391,7 +391,7 @@ sub candidate_pathnames {
             push(@paths, $full_file); } } } } }
   # Fallback: if no strict matches were found, return any existing case-insensitive matches
   # Defer the -f check until we are sure we need it, to keep the usual cases fast.
-  return @paths ? @paths : grep { -f $_ } @nocase_paths; }
+  return @paths ? @paths : @nocase_paths; }
 
 #======================================================================
 our $kpsewhich      = which($ENV{LATEXML_KPSEWHICH} || 'kpsewhich');
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -307,30 +307,17 @@ sub pathname_installation {
 sub pathname_find {
   my ($pathname, %options) = @_;
   return unless $pathname;
+  $options{findfirst} = 1;
   my @paths = candidate_pathnames($pathname, %options);
-  foreach my $path (@paths) {
-    return $path if -f $path; }
-  return; }
+  return $paths[0]; }
 
 sub pathname_findall {
   my ($pathname, %options) = @_;
   return unless $pathname;
-  my @paths = grep { -f $_ } candidate_pathnames($pathname, %options);
-  if (!@paths) {    # Fallback:
-        # This is a bit of a slow call, but really quite OK to run in rare fallback cases
-        # an average arXiv article may return with ~50-100 pathnames when looking for '*' graphics
-    my ($pathdir, $name, $type) = pathname_split($pathname);
-    my %star_candidates = ();
-    my $star_query      = $pathdir ? pathname_concat($pathdir, "*") : '*';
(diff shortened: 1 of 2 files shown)


dginev:
Fallback is now active for all filename searches, thanks @brucemiller ! Ready for review.

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -315,8 +315,18 @@ sub pathname_find {
...
-  return grep { -f $_ } @paths; }
+  my @paths = grep { -f $_ } candidate_pathnames($pathname, %options);
+  if (!@paths) {    # Fallback:
+     # To allow arbitrarily cased spellings,
+     # lowercase the target path and all candidate paths
+    $pathname = lc($pathname);
+    my %star_candidates = ();
+    # This is a bit of a slow call, but really quite OK to run in rare fallback cases
+    # an average arXiv article may return with ~50-100 pathnames when looking for '*' graphics
+    for my $c (candidate_pathnames('*', %options)) {
```
brucemiller, line 326: hmm....
Actually, as far as the name & extension go the deferred-case-insensitivity could actually be incorporated  into `candidate_pathnames` without any extra fs calls; In the last double loop over dirs and exts, you could (eg) dup each `push(@paths,...` with a `push(@lcpaths,...` that matches case insensitively.  Then at the end, `return(@paths ? @paths : @lcpaths);`.

What that doesn't handle --- but neither does your version! --- is case mismatch with search path directories, or a directory component on the given `$pathname`. But then, w/o testing, not sure whether it even should?  (except probably the latter)

dginev: Right, let's check the last part. For including `{maybe/illu_blur_mnist.pdf}` in a directory I named `mAybe` I see:

```
LaTeX Warning: File `mAybe/illu_blur_MNIST.pdf' not found on input line 6.
```

So I think the directory components are not included in the casing guards (thankfully)

dginev: I'm a bit confused about the idea to move it down to `candidate_pathnames`. Wouldn't that imply *always* computing the fallback list? The location I added it at first checks that a non-empty list of candidates has no actual files (`-f`) on the file system, and only then tries to figure out alternatives. The earliest time we know to need the fallback for certain is after the first batch of file system checks - otherwise we'll always piggyback the files?

dginev: Specifically in my example, the regular/master `candidate_pathnames` come back as:
```
/tmp/maybe/illu_blur_MNIST
/tmp/maybe/illu_blur_MNIST.svg
/tmp/maybe/illu_blur_MNIST.png
/tmp/maybe/illu_blur_MNIST.gif
/tmp/maybe/illu_blur_MNIST.jpg
/tmp/maybe/illu_blur_MNIST.jpeg
/tmp/maybe/illu_blur_MNIST.eps
/tmp/maybe/illu_blur_MNIST.ps
/tmp/maybe/illu_blur_MNIST.postscript
/tmp/maybe/illu_blur_MNIST.ai
/tmp/maybe/illu_blur_MNIST.pdf
/tmp/maybe/illu_blur_MNIST.SVG
/tmp/maybe/illu_blur_MNIST.PNG
/tmp/maybe/illu_blur_MNIST.GIF
/tmp/maybe/illu_blur_MNIST.JPG
/tmp/maybe/illu_blur_MNIST.JPEG
/tmp/maybe/illu_blur_MNIST.EPS
/tmp/maybe/illu_blur_MNIST.PS
/tmp/maybe/illu_blur_MNIST.POSTSCRIPT
/tmp/maybe/illu_blur_MNIST.AI
/tmp/maybe/illu_blur_MNIST.PDF
```

which is very specific to the cased `illu_blur_MNIST` and doesn't need any filesystem checks to be constructed and returned back as `@paths`. Then the `-f` checks whittle it down to the empty array.

 However, if the actual file was named `/tmp/maybe/iLLu_blur_MNIST.PDF` or `/tmp/maybe/illu_blur_MnisT.PDF` etc, there would be no way to know what string to try for efficiently - I'd need to permute all possible cased namings. Instead, going for the `/tmp/maybe/*.pdf`, `/tmp/maybe/*.PDF` , ... variants via `readdir()` in `candidate_pathnames`, and then intersecting with the lowercased variant, ought to be a reasonable way of implementing things.

Since the star variants require a `readdir()`, which is a file-system operation, I think we should really only run it when we are sure none of the usual literal matches worked. Or did I misunderstand your suggestion?

brucemiller:
Good that we can ignore directory case.
I was thinking of something like:
```
my @paths = ();
my @ncpaths = ();
   # Now, combine; precedence to leading directories.
   foreach my $dir ***@***.***) {
     foreach my $ext ***@***.***) {
       if ($name eq '*') {    # Unfortunately, we've got to test the 
file system NOW...
         if ($ext eq '.*') {    # everything
           opendir(DIR, $dir) or next;
           ***@***.***, map { pathname_concat($dir, $_) } grep { !/^\./ 
} readdir(DIR));
           closedir(DIR); }
         else {
           opendir(DIR, $dir) or next;    # ???
           my @f = readdir(DIR);
           ***@***.***, map { pathname_concat($dir, $_) } grep { 
/\Q$ext\E$/ } @f);
           ***@***.***, map { pathname_concat($dir, $_) } grep { 
/\Q$ext\E$/i } @f);
           closedir(DIR); } }
       elsif ($ext eq '.*') {             # Unfortunately, we've got to 
test the file system NOW...
         opendir(DIR, $dir) or next;      # ???
         my @f = readdir(DIR);
         ***@***.***, map { pathname_concat($dir, $_) } grep { 
/^\Q$name\E\.\w+$/ } @f);
         ***@***.***, map { pathname_concat($dir, $_) } grep { 
/^\Q$name\E\.\w+$/i } @f);
         closedir(DIR); }
       else {
         ***@***.***, pathname_concat($dir, $name . $ext)); } } }
   return ***@***.*** ? @paths : @ncpaths); }
```
Untested, but adds *no* additional fs calls; just some extra grepping.

dginev:
Snippet from Bruce's comment (moving to get highlighting, it appears email code blocks don't activate):
```perl
my @paths = ();
my @ncpaths = ();
   # Now, combine; precedence to leading directories.
   foreach my $dir ***@***.***) {
     foreach my $ext ***@***.***) {
       if ($name eq '*') {    # Unfortunately, we've got to test the 
file system NOW...
         if ($ext eq '.*') {    # everything
           opendir(DIR, $dir) or next;
           ***@***.***, map { pathname_concat($dir, $_) } grep { !/^\./ 
} readdir(DIR));
           closedir(DIR); }
         else {
           opendir(DIR, $dir) or next;    # ???
           my @f = readdir(DIR);
           ***@***.***, map { pathname_concat($dir, $_) } grep { 
/\Q$ext\E$/ } @f);
           ***@***.***, map { pathname_concat($dir, $_) } grep { 
/\Q$ext\E$/i } @f);
           closedir(DIR); } }
       elsif ($ext eq '.*') {             # Unfortunately, we've got to 
test the file system NOW...
         opendir(DIR, $dir) or next;      # ???
         my @f = readdir(DIR);
         ***@***.***, map { pathname_concat($dir, $_) } grep { 
/^\Q$name\E\.\w+$/ } @f);
         ***@***.***, map { pathname_concat($dir, $_) } grep { 
/^\Q$name\E\.\w+$/i } @f);
         closedir(DIR); }
       else {
         ***@***.***, pathname_concat($dir, $name . $ext)); } } }
   return ***@***.*** ? @paths : @ncpaths); }
```

dginev:
> Good that we can ignore directory case. I was thinking of something like: ...
> Untested, but adds *no* additional fs calls; just some extra grepping.

Right, so I definitely *wish* we could do that, but that's not the conditional branch we are on with these weird examples. Namely, they hit the last else case, since both `$name` is fixed (e.g. `illu_blur_MNIST`) and the extension `$ext` is fixed (e.g. `.pdf`, `.PDF`, ...) due to using an explicit list of graphics-friendly extensions. So the call to candidate pathnames just returns all possible concatenations *without* ever checking the file system:

```
  foreach my $dir (@dirs) {
    foreach my $ext (@exts) {
# not eligible for these branches
else {
        push(@paths, pathname_concat($dir, $name . $ext)); } } }
  return @paths; }
```

and it is the `-f` check in `pathname_findall` that validates they do or do not exist on disk. So  modifying the branch you mention doesn't particularly help? If anything it makes it a little hard to figure out when the fallback paths are used and when not (I'm still a bit hazy)

dginev:
My regex was a bit naive to have worked with inclusions with subdirectories though, upgraded it now. It indeed feels a bit odd at `pathname_findall`, but `candidate_pathnames` is a bit too large for me to feel comfortable refactoring in. Maybe a new function is in order after all e.g. `pathname_nocase_findall`?

brucemiller:
Right, I was puzzled why there was no `-f` test in the last clause. Was this supposed to be "optimization" that `pathname_find` only makes *one* `-f` test (but potentially many `readdir`) while `pathname_findall` tests all results? Actually, the current code seems perverse; the API uses a weird notion of "candidate" for wild cards (some files exist, some dont)

That last clause could also be modelled after the previous ones by reading the directory and grepping, instead of using `-f` later.  Actually that would allow a refactor reducing the number of calls when there are wildcards (since there often can be multiple extensions?)  But would trade `readdir` for `-f` in the common case of given name & extension.  On the other other hand, it allows for a short circuit when looking for a single file (`findfirst=>1`).

What do you think of this:
```
my @paths = ();
my @ncpaths = ();
  # Now, combine; precedence to leading directories.
  foreach my $dir (@dirs) {
    opendir(DIR, $dir) or next;
    my @files = readdir(DIR);
    closedir(DIR);
    foreach my $ext (@exts) {
      if (($name eq '*') & ($ext eq '.*')) {
        push(@paths, map { pathname_concat($dir, $_) } grep { !/^\./ } @files); }
      elsif($name eq '*') {
        push(@paths, map { pathname_concat($dir, $_) } grep { /\Q$ext\E$/ } @files);
        push(@ncpaths, map { pathname_concat($dir, $_) } grep { /\Q$ext\E$/i } @files); }
      elsif ($ext eq '.*') {             # Unfortunately, we've got to test the file system NOW...
        push(@paths, map { pathname_concat($dir, $_) } grep { /^\Q$name\E\.\w+$/ } @files);
        push(@ncpaths, map { pathname_concat($dir, $_) } grep { /^\Q$name\E\.\w+$/i } @files); }
      else {
        push(@paths, map { pathname_concat($dir, $_) } grep { /^\Q$name\E\Q$ext\E$/ } @files);
        push(@ncpaths, map { pathname_concat($dir, $_) } grep { /^\Q$name\E\Q$ext\E$/i } @files); } }
    return @paths if $options{findfirst} && @paths; }
  return (@paths ? @paths : @ncpaths); }
```
(still untested)

dginev:
Hm... it may be a promising upgrade? A little scary to predict what other pieces may be affected though. I'm wondering if some other piece of code is already relying on `candidate_pathnames` to return the possible pathnames even when they aren't physically present there - which your improved refactor would eliminate.

Btw, you may save a little bit of regex work runtime by only looking for the non-i match after the `/i` match succeeds - since it's a restriction on the other case. As in:
```perl
for my $file(@files) {
  if ($file =~ /^\Q$name\E\Q$ext\E$/i) {
    my $fullfile = pathname_concat($dir, $file);
    push(@ncpaths, $fullfile);
    if ($file =~ /^\Q$name\E\Q$ext\E$/) {
        push(@paths, $fullfile); } } }
```

actually, correct me if I'm wrong, but isn't  `/^\Q$name\E\Q$ext\E$/` the same as equality? `eq "$name$ext"` ?

...though of course I lost the nice grep-map idiom.

brucemiller:
The two `pathname_find{all}` appear to be the only callers, so there shouldn't be any surprises if the code does what I'm thinking it does. :>

Indeed, the next to the last regexp is equivalent to eq. Alternatively, we could iterate over a sequence of regexps created from `@exts` and `$name`, which would (1) replace the entire inner if/elsif mess with a pair of push/map/grep, and (2) which maybe makes your regexp optimization more practical.

dginev:
Ok, let me see...

dginev:
I tried to speed things up by pulling out the regex building *before* the directory rading, and ensure each directory is read only once... but on my SSD drive I didn't see any actual speed. No slowdown either at least :>

The benefit I did get is that the `pathname_find` and `pathname_findall` functions are once again dead simple. But I really couldn't get myself into the mindset to do the chain of map-grep. Feel free to take it from here if you wish, but this works on my subtle arXiv test, and all other tests seem to pass.

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -358,26 +357,41 @@ sub candidate_pathnames {
...
-  return @paths; }
+    opendir(DIR, $dir) or next;
+    for my $local_file (readdir(DIR)) {
+      for my $regex_pair (@regexes) {
+        my ($i_regex, $regex) = @$regex_pair;
+        if ($local_file =~ m/$i_regex/) {
+          my $full_file = pathname_concat($dir, $local_file);
+          push(@nocase_paths, $full_file);
+          # Always check if a strict match also exists on disk:
+          if (($local_file =~ m/$regex/) && (-f $full_file)) {
```
brucemiller, line 387: Doesn't `-f` have to succeed, since it's obtained from a directory listing?

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -358,26 +357,41 @@ sub candidate_pathnames {
...
+            # if we are only interested in the first match, return it:
+            return ($full_file) if $options{findfirst};
+            push(@paths, $full_file); } } } }
+    closedir(DIR); }
```
brucemiller, line 391: Better to `my @files=readdir(DIR);` and then `closedir(DIR);`, since the shortcircuit return will leave it open.

lib/LaTeXML/Util/Pathname.pm
dginev: Ooh, you're right! It is now completely redundant, unlike before, as that else case is no longer present. Nice!

brucemiller:
> But I really couldn't get myself 
> into the mindset to do the chain of map-grep.

The loop's fine, since it allows for the optimizations,
even if they don't matter :>

dginev:
Much tidier now, thanks @brucemiller ! I am almost convinced I did get a speedup, but the slowest expl3 test is so slow that my parallel testing harness only really sees its time. Definitely appears better than the original patch, and ought to be faster *sometimes*.

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -358,26 +357,41 @@ sub candidate_pathnames {
...
+        if ($local_file =~ m/$i_regex/) {
+          my $full_file = pathname_concat($dir, $local_file);
+          push(@nocase_paths, $full_file);
+          if ($local_file =~ m/$regex/) {
+            # if we are only interested in the first match, return it:
+            return ($full_file) if $options{findfirst};
+            push(@paths, $full_file); } } } } }
+  # Fallback: if no strict matches were found, return any existing case-insensitive matches
+  # Defer the -f check until we are sure we need it, to keep the usual cases fast.
+  return @paths ? @paths : grep { -f $_ } @nocase_paths; }
```
brucemiller, line 394: But again, we don't really need the `-f` test, since the files do exist in the directory; it's just that they don't have the same case as the request.

dginev: ... it's now officially my blindspot. patch incoming

dginev: ...patch landed

brucemiller:
Woohoo!

PR #1746, opened 2021-12-18
scaffold of support for \vdotswithin

### LINKED ISSUE #1672: mathtools.sty.ltxml: replace \vdotswithin by \vdots
Currently, `\vdotswithin` is apparently replaced with an empty string:
https://github.com/brucemiller/LaTeXML/blob/be6bbab59dbf48b7834d96db65c49226c27a96b4/lib/LaTeXML/Package/mathtools.sty.ltxml#L642
Maybe, at least `\vdots` would be more appropriate, disregarding the alignment (probably the same for `\shortvdotswithin`).

dginev:
Fixes #1672 .

It is a partial upgrade (the centering of the vdots is not *yet* supported). The issue description seemed to be happy with that as a first step, so here it is.

Minimal example from the mathtools docs:
```tex
\documentclass{article}
\usepackage{mathtools}
\begin{document}

\begin{align*}
a &= b \\
& \vdotswithin{=} \\
& = c \\
\shortvdotswithin{=}
& = d
\end{align*}
\end{document}
```

before:

after:

### DIFF
--- a/lib/LaTeXML/Package/mathtools.sty.ltxml
+++ b/lib/LaTeXML/Package/mathtools.sty.ltxml
@@ -639,11 +639,17 @@ DefMacro('\Aboxed{}', '#1');
 # TODO: Make this actually do something
 DefMacro('\ArrowBetweenLines[]',                   '');
 DefMacro('\csname ArrowBetweenLines*\endcsname[]', '');
-DefMacro('\vdotswithin{}',                         '');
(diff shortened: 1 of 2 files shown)


brucemiller:
Looks plausible; Thanks!

dginev:
> Looks plausible; Thanks!

Sure thing! Did you mean to merge by chance?

PR #1748, opened 2021-12-22
deflate the full ZIP output

dginev:
Between fixing the path bug that was failing to find graphics, adding the `magnify=2` visual upgrade and arXiv getting quite bigger in 2021, I've ended up rather low on space on my server machines.

Maybe this small PR could be a big help, opting to deflate all assets, rather than only the main HTML one. I definitely had missed the `.log` files, but even `png` assets see a 1-3% size reduction, depending.


brucemiller:
OK, I'll pretend that's acceptable.

dginev:
Thanks! It's now live and making a difference 🗜️

PR #1749, opened 2021-12-27
Auto-load graphicx on \includegraphics for arXiv

dginev:
This PR falls fully in the OmniBus philosophy for recovering hard-to-predict old sources into usable HTML.

Specifically, I spotted the issue at [1804.11029](https://ar5iv.org/log/1804.11029), where seemingly unexplicably `\includegraphics` was failing to get defined. It turns out, this error is based on the dual possibility for arXiv submissions to be processed with the classic latex executable, or the newer pdflatex one -- information which is not part of the sources, but is instead submission metadata only available to arXiv itself. Thus, we can't really predict when and where such cases occur, and latexml's best strategy is (usually) to pretend to be pdflatex in some cases and "something unknown" in others.

In this specific case, we have the code:
```tex
% *** GRAPHICS RELATED PACKAGES ***
%
\ifCLASSINFOpdf
  % \usepackage[pdftex]{graphicx}
  % declare the path(s) where your graphic files are
  % \graphicspath{{../pdf/}{../jpeg/}}
  % and their extensions so you won't have to specify these with
  % every instance of \includegraphics
  % \DeclareGraphicsExtensions{.pdf,.jpeg,.png}
\else
  % or other class option (dvipsone, dvipdf, if not using dvips). graphicx
  % will default to the driver specified in the system graphics.cfg if no
  % driver is specified.
   \usepackage[dvips]{graphicx}
  % declare the path(s) where your graphic files are
   \graphicspath{{../eps/}}
  % and their extensions so you won't have to specify these with
  % every instance of \includegraphics
   \DeclareGraphicsExtensions{.eps}
\fi
```

which could have just as well been commented out inversely, if the pdflatex submission route was taken.

So, my best common sense guess at an upgrade for latexml here is to offer an `AutoLoad` directive for the `\includegraphics` macro specifically, given how it is one of the most standard commands in the latex ecosystem. Which works as expected and avoids the errors - so the PR is ready for review.

At time of writing, I can't imagine a way to improve our code for `\ifCLASSINFOpdf`, in the general arXiv case, without getting access to their submission metadata.

P.S. A nice extra trick is that the related DefMacro definitions in graphicx.sty.ltxml need to be made global in scope, as the fallback triggers deep within the document body, with lots of TeX groups open.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3418,6 +3418,7 @@ DefMacro('\@xdblfloat{}[]',     '\@xfloat{#1}[#2]');
 DefMacro('\@floatplacement',    '');
 DefMacro('\@dblfloatplacement', '');
 
+DefAutoload('includegraphics', 'graphicx.sty.ltxml');
 #======================================================================
 # C.9.2 Marginal Notes
 #======================================================================
--- a/lib/LaTeXML/Package/graphicx.sty.ltxml
+++ b/lib/LaTeXML/Package/graphicx.sty.ltxml
@@ -43,12 +43,13 @@ DefKeyVal('Gin', 'magnifiable', '', 'true');
 
 # Redefine
 DefMacro('\includegraphics OptionalMatch:* []',
-  '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}');
+  '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}', scope => 'global');
 
 DefConstructor('\@includegraphicx OptionalMatch:* OptionalKeyVals:Gin Semiverbatim',
   "<ltx:graphics graphic='#path' candidates='#candidates' options='#options'/>",
   alias      => '\includegraphics',
   sizer      => \&image_graphicx_sizer,
+  scope      => 'global',
   properties => sub {
     my ($stomach, $starred, $kv, $path) = @_;
     $path = ToString($path); $path =~ s/^\s+//; $path =~ s/\s+$//;


brucemiller:
The global only affects `\includegraphics`, but not any of the other (less often used, but used) macros in the graphics file.  I'm confused about the pdf comments; are you saying that `pdflatex` includes the graphics style by default?  Otherwise, rather than autoload it, wouldn't it be better to just include `graphicx` in  `OmniBus`?

dginev:
Right, so this behavior happens in a document which doesn't use OmniBus, but a class file we have a binding for - so a patch in OmniBus would not be a fix.

My comments about the PDF behavior relate to the conditional `\ifCLASSINFOpdf` which is the part we have no control over/knowledge of - when arXiv received the article it was instructed whether to use `latex paper.tex` vs `pdflatex paper.tex`, but that metadata is not provided to us currently. So we have no good way to emulate "conditional loading" of graphicx in general, thus the need for an auto-load rule.

Would you like me to add auto-load rules for the other graphicx macros? I can do that.

To answer your pdflatex question - no, nothing in the latex ecosystem autoloads graphicx. It only does so for this snippet I provided that is based on `\ifCLASSINFOpdf`.

brucemiller:
So, this document would fail if run by pdflatex? Seems like trying a bit too hard for a rare occurence.

dginev:
Right, if you think it's beyond the scope of what the main pool files should be doing, that's fine as well. And it's definitely far away in the long tail:


what | tasks % | tasks | messages % | messages
-- | -- | -- | -- | --
\includegraphics | 0.04% | 672 | 0.08% | 673

Might be a good example to put a pin on for now,  until (if) it becomes important to recover a larger range of these executable-specific discrepancies in arXiv

brucemiller:
> Right, if you think it's beyond the scope

Well, it's nice to recover as many as feasible, but it
vaguely worries me to stray too far from (la)tex-like behavior.
Putting the global switch on doesn't quite make it behave like
you had a top-level `\usepackage`. It's hard to see exactly what
problem it could create, but somehow it can come back to bite.

PR #1752, opened 2021-12-31
Column flex fixes

brucemiller:
This PR deals with several issues with the new flex approach to multiple figures/subfigures that turned up with DLMF.

* It adds the bookkeeping to maintain `\columnwidth` (as distinct from `\textwidth`)
* It uses the `\columnwidth` to make decisions about subfigure layout, rather than trying to sniff out an intended size.
* It introduces two helper functions (`beforeFloat` and `afterFloat`) that arrange for the various floats to record the current `\columnwidth` while digesting (since the value at construct-time is no longer the correct one), while distinguishing between single and double column floats.
* It patches the the XSLT to recognize more figure content types (but there is still too much potentially inconsistently duplicated testing in LaTeX.pool and the stylesheets that will cause trouble eventually).

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3209,7 +3209,7 @@ DefConstructor('\@@generic@caption[]{}', "<ltx:text class='ltx_caption'>#2</ltx:
 
 sub add_subfigure_row_breaks {
   my ($document, $node, $whatsit, @contents) = @_;
-  my $textwidth     = $whatsit->getProperty('floatwidth') || Dimension('345pt')->valueOf();
+  my $textwidth     = ($whatsit->getProperty('floatwidth') || Dimension('345pt'))->valueOf();
   my $current_width = 0;
   my @new_contents  = ();
   for my $block (@contents) {
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4636,7 +4636,6 @@ DefEnvironment('{minipage}[][][]{Dimension}', sub {
   sizer      => '#body',
   mode       => 'text',
   properties => sub { (
-      layout  => 'vertical',
       width   => $_[4],
       vattach => translateAttachment($_[1])); },
   beforeDigest => sub {
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3130,18 +3130,21 @@ DefMacro('\format@title@font@figure', '\format@title@font@float');
 DefMacro('\format@title@font@table',  '\format@title@font@float');
 
 # stubs for the latex float internals
-for my $float (qw(@float @dblfloat)) {
-  DefEnvironmentI($float, '[]{}',
-    "<ltx:float xml:id='#id' inlist='#inlist' ?#1(placement='#1') class='ltx_float_#2'>"
-      . "#body"
-      . "</ltx:float>",
-    properties       => { layout => 'vertical' },
-    afterDigestBegin => sub {
-      DefMacroI('\@captype', undef, ToString($_[1]->getArg(2)));
-      return; },
-    afterDigest => sub {
-      RescueCaptionCounters(ToString(Expand(T_CS('\@captype'))), $_[1]);
-      return; }); }
+DefEnvironmentI('@float', '[]{}',
+  "<ltx:float xml:id='#id' inlist='#inlist' ?#1(placement='#1') class='ltx_float_#2'>"
+    . "#body"
+    . "</ltx:float>",
+  properties        => { layout => 'vertical' },
+  beforeDigestBegin => sub { beforeFloat(ToString($_[1]->getArg(2))); },
+  afterDigest       => sub { afterFloat($_[1]); });
+DefEnvironmentI('@dblfloat', '[]{}',
+  "<ltx:float xml:id='#id' inlist='#inlist' ?#1(placement='#1') class='ltx_float_#2'>"
+    . "#body"
+    . "</ltx:float>",
+  properties        => { layout => 'vertical' },
+  beforeDigestBegin => sub { beforeFloat(ToString($_[1]->getArg(2)), double => 1); },
+  afterDigest       => sub { afterFloat($_[1]); });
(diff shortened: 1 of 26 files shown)


dginev:
I'd love to take some time taking a detailed look here (likely tomorrow!)

Before I do: do you expect differences in the flex figures we are just generating for arXiv? Anything to be prepared to adapt CSS rules for?

brucemiller:
Intention is mostly the same maybe even moreso. But likely randomly
different due to nuances

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -3215,25 +3218,7 @@ sub add_subfigure_row_breaks {
...
-      # TODO: A tricky bit to consider here is graphics sized to \columnwidth
-      # which isn't yet accurately emulated in latexml itself.
-      # so where two or more graphics may have been expected in the PDF,
-      # latexml would only allow one as the \columnwidth would be defaulted to \textwidth.
-      my $block_width = $block->getAttribute('width');
-      # If we have an author annotation of the width, it will appear as an XML attribute,
-      # and should take precedence over the underlying asset width of the image blob
-      if (!$block_width) {
-        if (my $options = $block->getAttribute('options')) {
-          if ($options =~ /width=([^ "]+)/) {
```
brucemiller, line 3227: Note that this regexp does not extract a valid width in general, particularly when there are multiple keywords.

brucemiller:
It's probably worth retesting some of your layout test cases to see if they still render acceptably.  But also note that this PR should avoid some errors (invalid dimension?) that likely appeared in the previous version.

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -4635,7 +4636,6 @@ DefEnvironment('{minipage}[][][]{Dimension}', sub {
   sizer      => '#body',
   mode       => 'text',
   properties => sub { (
-      layout  => 'vertical',
```
brucemiller, line 4638: This was wrong; minipage does not (necessarily) lay out its contents vertically.

I'm also concerned about similar properties that snuck in on the various floats. Floats don't (necessarily) lay out contents vertically. (unless the contents don't fit).

What were we trying to accomplish here?

dginev: the way the width gets computed is different when this property is set. 

For the example I understood, we don't want to add the width of a figure image with the width of a caption to get the width of the entire figure. As the figure is vertical, the width of widest component is what determines the overall width.

Maybe I overstepped with minipages and floats - though I recall floats worked rather well in my tests. Minipages were bumpy in arXiv whether I had this attribute or not, since there was a lot more going on there.

dginev: Oh and the reason I needed some "correct" width was to determine if I need to auto-generate a line break, as I was trying to emulate the same subfigure grids that were originally created for the PDF output.

brucemiller: Well, vertical means that *all* the children are stacked vertically, rather than fitting them to the current width, and wrapping when needed. So, of course, it affects both the width and height.

But, captions explains why floats/figures/tables have that property (I think your changes just set them more consistently). I think it's probably the wrong way to deal with captions in the long run; really should treat the caption as a block, but wrap the other components.  But that's tricker to set up.

However, for minipages, it's just wrong. And besides, minipages already have their width preset.

dginev:
Doing some testing today, subfigures seem identical in the previous articles I was testing against. Floats have some interesting changes, e.g. in [1307.6769](https://ar5iv.org/html/1307.6769), Figure 1 we have 4 listings in the same figure that are a nice test.

My code (maybe wrongly and coincidentally) had allocated 2 listings per row in that example, while it appears your PR now allows all 4 of those listings on the same row. I'll investigate that part a little more closely.

brucemiller:
Hmm, with scrollbars and all! Not so great. I would have suspected a bad size computation for the `algorithm` environments, but that should have affected your code as well.

Incidentally, not to push too hard, but this patch is needed for (a late release of) DLMF...

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -3206,7 +3209,7 @@ DefConstructor('\@@generic@caption[]{}', "<ltx:text class='ltx_caption'>#2</ltx:
 
 sub add_subfigure_row_breaks {
   my ($document, $node, $whatsit, @contents) = @_;
-  my $textwidth     = (LookupRegister('\textwidth') || Dimension('345pt'))->valueOf();
+  my $textwidth     = $whatsit->getProperty('floatwidth') || Dimension('345pt')->valueOf();
```
dginev, line 3212: Good news, the issue with the `1307.6769` listings breaks was a small typo on this line. We need to get the numerical value behind the `floatwidth` property, either when setting it, or simply changing this line to:
```perl
   my $textwidth     = ($whatsit->getProperty('floatwidth') || Dimension('345pt'))->valueOf();
```

dginev (changes requested):
I think the PR is good to merge after the numerical value typo is patched, thank you for refactoring the logic further!

I haven't tested minipages, but they'll need some standalone attention in any case, so that can be done separately. floats and subfigures are looking good!

dginev:
> Hmm, with scrollbars and all! 

Btw there are no scrollbars in the ar5iv.css rendering even with 4 listings next to each other, so maybe I should be porting over more of the newer, ironed out, styling. But CSS is (sadly) a whole world of its own, completely separate from this PR.

lib/LaTeXML/Package/LaTeX.pool.ltxml
brucemiller: Hah! Good catch! How'd it even appear to work with that??

dginev (approved):

lib/LaTeXML/Package/LaTeX.pool.ltxml
dginev: perl being perl...

```
$ perl -e 'print "LaTeXML::Common::Dimension(whatever)" > 0 ? "bigger!\n" : "not bigger!\n"'
not bigger!
```

brucemiller: Or: LaTeXML! It works for *all* the wrong reasons!

PR #1753, opened 2022-01-02
Metafonts

### LINKED ISSUE #1731: Index entries are inheriting their style
The document
```tex
\documentclass{book}
\usepackage{makeidx}
\makeindex
\begin{document}
This is a paragraph\index{paragraph}

{\small This is a small paragraph\index{small paragraph}}

\printindex
\end{document}
```
Ends up with the xml
```xml
<para>
    <p>This is a paragraph<indexmark>
        <indexphrase key="paragraph">paragraph</indexphrase>
      </indexmark></p>
</para>
<para>
    <p><text fontsize="90%">This is a small paragraph<indexmark>
          <indexphrase key="small paragraph"><text fontsize="90%">small paragraph</text></indexphrase>
        </indexmark></text></p>
</para>
```
This means that in the resulting index, one entry is smaller than the other.  With no other context, changing styles in the index is confusing.

brucemiller:
This PR defines a new `neutralizeFont` helper to be used in metadata macros like `\index`, `\footnote`, etc, in order to reset the font from the current one. This is more complete than digesting `\normalfont`, which only resets the style, but not size, color, etc.

Fixes #1731

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -490,7 +490,7 @@ DefConstructor('\lx@note[]{}[]{}',
     . "#4"
     . "</ltx:note>",
   mode         => 'text', bounded => 1,
-  beforeDigest => sub { reenterTextMode(1); Digest('\normalfont'); },
+  beforeDigest => sub { reenterTextMode(1); neutralizeFont(); },
   properties   => sub {
     my $type = ToString($_[2]);
     (role => $type,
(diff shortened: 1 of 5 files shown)


dginev:
I think this PR is based off from #1752 rather than master? Makes it a little hard to review the changes. If you could rebase this one on master after #1752 is merged in, I'll take a quick look.

brucemiller:
Whoops; thanks for noticing!

dginev (approved):
Looks good!

My only question is whether you could imagine ever calling `reenterTextMode` *without* calling the new `neutralizeFont`.

If they are always called together, you might as well add `neutralizeFont` at the end of the `reenterTextMode` sub, so that we keep them  bundled in future code.

brucemiller:
Well, `\hbox` does `reenterTextMode`, but shouldn't change fonts. But it's a good observation: I tried the other way (adding reenter to neutralizeFonts) and that also was a problem. So, I guess they're slightly orthogonal.

PR #1754, opened 2022-01-02
new \lxRequireResource

### LINKED ISSUE #1721: Could we add \lxIncludeResource to the latexml package?
I would like to request that we add a command `\lxIncludeResource{}[]{}` to the latexml package, defined as
```perl
DefMacro('\lxIncludeResource{}[]{}',
	 '\AtBeginDocument{\lx@include@resource{#1}?#2([#2]){#3}}');
DefConstructor('\lx@include@resource{}[]{}',
           '<ltx:resource type="#1" src="#3" ?#2(media="#2") />');
```
I tried to define that as
```perl
DefConstructor('\lx@include@resource{}[]{}', sub {
  if ( $_[2] ) {
    RequireResource($_[3], {type=>$_[1],media=>$_[2]} ); }
  else {
    RequireResource($_[3], {type=>$_[1]} ); }
  return; });
```
which seems higher level (and doesn't need AtBeginDocument), but couldn't quite get it to work.

For javascript and css, this would duplicate command line options to latexmlpost (except for media), but this would allow moving those options into the tex document instead of having to remember to have them on the command line (or put into a build script).  This also opens up the possibility of including other resources, although I'm not thinking of any at the moment.

(And `\NewDocumentCommand\lxIncludeResource{mom}{}`)

brucemiller:
New `\lxRequireResource[options]{source}` to require resources, such as javascript, css, potentially others in future.  Typical option `type=text/javascript`, unless it can be inferred from the source extension.

Fixes #1721

### DIFF
--- a/lib/LaTeXML/Package/latexml.sty.ltxml
+++ b/lib/LaTeXML/Package/latexml.sty.ltxml
@@ -145,9 +145,11 @@ DefConstructor('\lxRef Semiverbatim {}',
 
 #======================================================================
 # Resources
-# \lxResource[options]{name}
-#  RequireResource($name,%options);
-
+# \lxRequireResource[options]{name}
+#   options: type (mime-type), media (?), ...
+DefPrimitive('\lxRequireResource OptionalKeyVals {}', sub {
+    my ($stomach, $kv, $path) = @_;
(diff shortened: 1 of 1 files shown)


dginev (approved):
sure, looks good

PR #1755, opened 2022-01-03
Multicol markers

### LINKED ISSUE #1711: add a class for multicolumns
In [multicol.sty.ltxml](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/Package/multicol.sty.ltxml), we intentionally ignore the argument to `\begin{multicols}{}` because it determines the number of columns, which is really a styling question that should be done by css.  This makes sense, but it also suggests that we should add some sort of class so that it can be styled.

It appears that numbers are allowed for class names (just not as the [first character](https://www.w3.org/TR/selectors-3/#grammar)), so I think we could use the class `columns#1`.  We could also try `.columns2 { columns: 2 }` etc, in [LaTeXML.css](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/resources/CSS/LaTeXML.css).

brucemiller:
This PR adds `ltx:pagination` markers when switching the number of columns using the `multicol` package.  It uses markers rather than wrapping the content, since the multicolumn content does not necessarily nest with the document structure tree (see the discussion in #1711).  Although these markers can't be directly used by CSS, hopefully they will be of some use.

Fixes #1711

### DIFF
--- a/lib/LaTeXML/Package/multicol.sty.ltxml
+++ b/lib/LaTeXML/Package/multicol.sty.ltxml
@@ -18,8 +18,16 @@ use LaTeXML::Package;
 # Since this is basically styling, we can ignore the effects.
 
 # Optional arg is sortof a heading, but w/o any particular styling(?)
-DefEnvironment('{multicols}{}[]',  '?#2(<ltx:para><ltx:p>#2</ltx:p><ltx:para>)#body');
-DefEnvironment('{multicols*}{}[]', '?#2(<ltx:para><ltx:p>#2</ltx:p><ltx:para>)#body');
+DefEnvironment('{multicols}{}[]',
+  '?#2(<ltx:para><ltx:p>#2</ltx:p><ltx:para>)'
+    . '<ltx:pagination role="start_#1_columns"/>'
+    . '#body'
+    . '<ltx:pagination role="end_#1_columns"/>');
+DefEnvironment('{multicols*}{}[]',
+  '?#2(<ltx:para><ltx:p>#2</ltx:p><ltx:para>)'
+    . '<ltx:pagination role="start_#1_columns"/>'
+    . '#body'
+    . '<ltx:pagination role="end_#1_columns"/>');
 
 DefMacro('\botmark', Tokens());
 DefMacro('\topmark', Tokens());
--- a/t/structure/columns.xml
+++ b/t/structure/columns.xml
@@ -9,7 +9,9 @@
     <p>HEADING</p>
   </para>
   <para xml:id="p2">
+    <pagination role="start_2_columns"/>
     <p>BODY</p>
+    <pagination role="end_2_columns"/>
   </para>
   <para xml:id="p3">
     <p>one column</p>


dginev:
That's great! But if we don't have a path to use them in CSS, I wonder if the work on the initial issue is complete.

brucemiller:
Well, other than giving up the document structure for layout structure, the alternative is "wont fix";

dginev:
Well, "wontfix" is a bit harsh - I can see various paths where the PR is halfway there.

Especially if we do a rewrite pass at the end of construction (or post-processing, either way) that looks for these pagination bookmarks you're inserting and distributes the multicolumn class to the right "logical structure" nodes. There will be cases where it can *also* insert wrapper nodes to contain the styling to the right scopes. Which is all doable once the full tree is constructed, I think?

brucemiller:
Well, maybe there's some CSS mechanism that I don't know or understand well enough; what seems clear is that you can't wrap a single `<div class="multicolumn">` around all the affected columnar material. But maybe you don't need to?  Why don't you manually mock-up the output of the example I posted in #1711 to show me what you're talking about?

dginev:
Something on the lines of (look for `ltx_multicols_2`):

```html
<article class="ltx_document">
  <section id="S1" class="ltx_section">
    <h2 class="ltx_title ltx_title_section">
      <span class="ltx_tag ltx_tag_section">1 </span>One
    </h2>
    <div id="S1.p1" class="ltx_para">
      <p class="ltx_p">Stuff in One
        <span class="ltx_multicols_2">
          More Stuff in One</span>
      </p>
    </div>
    <div id="S1.p2" class="ltx_para ltx_multicols_2">
      <p class="ltx_p">Even more Stuff in One</p>
    </div>
  </section>
  <section id="S2" class="ltx_section">
    <h2 class="ltx_title ltx_title_section ltx_multicols_2">
      <span class="ltx_tag ltx_tag_section">2 </span>Two
    </h2>

    <div id="S2.p1" class="ltx_para">
      <p class="ltx_p">
        <span class="ltx_multicols_2">Stuff in Two.
        </span>
        A bit more stuff in Two
      </p>
    </div>
  </section>
  <section id="S3" class="ltx_section">
    <h2 class="ltx_title ltx_title_section">
      <span class="ltx_tag ltx_tag_section">3 </span>Three
    </h2>

    <div id="S3.p1" class="ltx_para">
      <p class="ltx_p">And finally stuff in Three.</p>
    </div>
  </section>
</article>
```

Which should work "just right" for any logical subtree that is inside the environment and be "subtly wrong" for the start/end of the column mode switch, as the logical tree structure will create extra vertical breaks (i.e. will flush the columns more often).

I think these new attributes are simplest to add after your markers are inserted and the final XML is completed. And CSS can use them with a direct rule, and even allows designers to be selective e.g. only styling completed structures as in `.ltx_para.ltx_multicols_2`

brucemiller:
Given the weather, I feel like I'm being Snowed. Did you actually try this? You're putting `columns:2` onto the various leaves that should (somehow) end up *in* columns, but I was under the impression that that css `columns` should go onto the *container* that has material that should be divided up into columns.  And the lack of that container is exactly the problem that I'm trying to point out.

dginev:
There's no snow here, so maybe I'm missing some flakes?

If the author uses `{multicols}` high up, the container will be a logical node (such as a section, or paragraph), rather than the crummy spans at the start/end, which I suspect you are objecting to. But does my span work as a two-column visual if one *insisted* on that? Sure, just throw in an extra `display: block;` rule. Not that I would recommend that, but it's something we offer designers choice with.

Here is a (very narrow) screenshot:
nosnow

dginev:
To be clear: I wasn't suggesting "annotate only leaves". It's just that your difficult example only allowed for leaves, as the environments were not fully _within_ the `{multicols}` wrapper.

If you add a big `\section{foo}...` with lots of material inside the multicols environment, the column-specifying attribute ought to attach to the highest parent that is still fully inside the `{multicols}` environment, i.e. the `<section>` one.

brucemiller:
> Here is a (very narrow) screenshot:

Well, that's a nice solution to something, I'm sure;
but it's not the solution to the problem I posed.

dginev:
@brucemiller I can now offer a very practical use case, that happens to be easier than the hard minimal examples we have been discussing.

I have an open issue for a [two column theme](https://github.com/dginev/ar5iv-css/issues/1) in ar5iv, and a recent arXiv article made the rounds for relying on a very witty "drama" genre, with a 2-column line-numbered style for extra effect. That article is [2206.08225])(https://arxiv.org/abs/2206.08225).

I would suggest that we can stay practical by making sure the first stab at a solution from this PR addresses that article (or any other of your choosing for that matter). Does that sound reasonable?

brucemiller:
Picking an article to make "better" is the easy case;
making no other articles worse is the challenge. There's a lot of 
potential to screw up lots of stuff by focusing too narrowly.

Page formatting, such as what multicols usually is used for, is used 
with a specific medium in mind (again usually), and so it isn't at all 
clear that they should apply on the web (or if so, perhaps done quite 
differently).

  So, any proposed solutions will have to fall back to doing nothing in 
most cases.

dginev:
Given such a potential to solve too narrowly, we could collect a diverse list of examples (5? 10? 20?) that use multiple columns in PDF.

Then we could see if we have ideas that apply to all cases. I find it difficult at the moment to imagine all possible uses.

teepeemm:
Would it work to fix things in post processing?  If the `<pagination role='start_2_columns'/>...<pagination role='end_2_columns'/>` are siblings, we could change that to `<div class="ltx_multcols_2">...</div>` using
``` xml
  <xsl:key name="multi-col-data"
      match="*[preceding-sibling::ltx:pagination[@role='start_2_columns']][following-sibling::ltx:pagination[@role='end_2_columns']]"
      use="generate-id(preceding-sibling::ltx:pagination[@role='start_2_columns'][1])" />

  <xsl:template
      match="*[preceding-sibling::ltx:pagination[@role='start_2_columns']][following-sibling::ltx:pagination[@role='end_2_columns']]" />
  <xsl:template mode="copy-column-data"
      match="*[preceding-sibling::ltx:pagination[@role='start_2_columns']][following-sibling::ltx:pagination[@role='end_2_columns']]">
      <xsl:apply-templates />
  </xsl:template>

  <xsl:template match="ltx:pagination[@role='start_2_columns'][following-sibling::ltx:pagination[@role='end_2_columns']]">
      <xsl:element name="div" namespace="{$html_ns}">
          <xsl:attribute name="class">ltx_multcols_2</xsl:attribute>
          <xsl:apply-templates select="key('multi-col-data', generate-id())"
            mode="copy-column-data" />
      </xsl:element>
  </xsl:template>
```
This has the advantage of falling back to doing nothing if they aren't siblings.  This also suggests a more general approach for newenvironment asked for in #1835: we could have `<pagination role='[start|end]_environment' class='environmentname'/>` become `<div class="environmentname">` (but probably change from `pagination`).

To followup Bruce's comment about a specific medium, I would suggest a `LaTeXML-multicols.css` that a user would have to explicitly incorporate.  (For example, `\lipsum[1-150]` would be obnoxious to read in a two column format.)

Here is a tex file with 5 multicol examples of increasing complexity.  The XSLT above fails what I thought the easiest case, but handles the next two cases.  It has trouble when there's not a paragraph break at the beginning and end and doesn't work with Bruce's pathological example.  (2206.08225 is fun and all, but it doesn't actually use multicols - each page is two `{figure}[h]` for the two columns, and each figure is a two column tabular: one for the line number and one for the text.  The column breaks are hard coded.)
[multicol-examples.tex.txt](https://github.com/brucemiller/LaTeXML/files/9155261/multicol-examples.tex.txt)

brucemiller:
Well, the basic scenario is that *anywhere* in vertical mode, you might 
have a start or end of column settings, so I'd worry about that case first.

Frankly, tho' it'd be nice to have nice 2 column layouts, it's a small 
payoff for (probably) much work; just doesn't seem to me a priority 
right now, compared to other things.

brucemiller:
Well, this obviously isn't a completely satisfactory solution, if it is a solution at all. But it at least provides a record of the columns from which a better (or application specific?) solution might be explored.  At any rate, it isn't going to get much better in the short term, so... let's merge.

dginev:
That's unfortunate. I hope no one builds a Javascript-based solution using the markers, which may later end up broken when they are re-tailored to fit a CSS-based approach.

brucemiller:
Whether javascript, xslt or css based, it would seem that this PR is a 
reasonable starting point.

PR #1756, opened 2022-01-08
Add more dependencies for newtxmath.sty

dginev:
Spotted a missing `\binom` for a paper using newtxmath.sty, and subsequently realized the dependencies don't accurately match the ones from the texlive `.sty` file.

So this PR updates them to match - hopefully easy & simple.


brucemiller:
Cool; thanks!

PR #1758, opened 2022-01-09
Variety patchup from arXiv 2021 holiday run

dginev:
I looked into a variety of issues this morning, and rapidly patched a number of very easy missing bits from various report levels. 

Feel free to read through and veto anything that seems too brave - I've left the paper names in the commit messages / code comments. 

Mostly extensions to OmniBus itself, a couple of easy bindings, and some optional arguments to standard macros that help in OmniBus use.

At times I wondered whether I should be writing journal-specific bindings, but the utility of adding constructs such as `{thm}` and `{lem}` to OmniBus seems much higher for arXiv, where all kinds of journals may be rediscovering the same constructs.

### DIFF
--- a/lib/LaTeXML/Package/graphicx.sty.ltxml
+++ b/lib/LaTeXML/Package/graphicx.sty.ltxml
@@ -43,12 +43,13 @@ DefKeyVal('Gin', 'magnifiable', '', 'true');
 
 # Redefine
 DefMacro('\includegraphics OptionalMatch:* []',
-  '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}');
+  '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}', scope => 'global');
 
 DefConstructor('\@includegraphicx OptionalMatch:* OptionalKeyVals:Gin Semiverbatim',
   "<ltx:graphics graphic='#path' candidates='#candidates' options='#options'/>",
   alias      => '\includegraphics',
   sizer      => \&image_graphicx_sizer,
+  scope      => 'global',
   properties => sub {
     my ($stomach, $starred, $kv, $path) = @_;
     $path = ToString($path); $path =~ s/^\s+//; $path =~ s/\s+$//;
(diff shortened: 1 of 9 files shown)


lib/LaTeXML/Package/graphicx.sty.ltxml
```diff
@@ -43,12 +43,13 @@ DefKeyVal('Gin', 'magnifiable', '', 'true');
 
 # Redefine
 DefMacro('\includegraphics OptionalMatch:* []',
-  '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}');
+  '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}', scope => 'global');
```
brucemiller, line 46: Did this sneak in from the autoload experiment? (I'm not sure that it's actually bad, here; but confusing: somehow the top-level of any binding file should always be global)

dginev: Since we didn't merge that PR, and I still think it will be needed when graphicx gets autoloaded, I added it here as well yes.

It would seem like the top-level definitions of any binding *ought to be* global, but the default scope of DefMacro is currently local (since the "scope" option [is undefined](https://github.com/brucemiller/LaTeXML/blob/master/lib/LaTeXML/Package.pm#L1048)), so... they aren't. It's just that in regular loads, definitions rarely, if ever, get loaded from nested groups, so we don't notice.

brucemiller:
> It would seem like the top-level definitions of any binding /ought to 
> be/ global, 
...
> so... they aren't.

Indeed & exactly.
Kinda bugs me not to see a clean way to make it so

brucemiller:
Cool; Thanks!

dginev:
Oh, you merged it? I was brewing (in one of my multiple branch-pots today), a solution that defaulted all top-level binding definitions to global in the DefAutoload case. It is _almost_ working, based on `\globaldefs` getting set to true before the load, and unset after. (I tried the \global prefix first, but that gets bumped a bit often).

Then again such sophistication _can_ wait.

brucemiller:
Oh, sorry; didn't realize you had more.
Can always go with another PR

PR #1760, opened 2022-01-16
Aligned fix

brucemiller:
This PR basically undoes d7df025, but provides a more correct fix; since that fix causes regressions in other nested alignment environments (particularly in my TeXamples testes).

Namely amsmath's aligned environment intentionally reads its optional argument in a special way to avoid triggering the alignment machinery.   Hopefully, this still fixes all the testcases raised in that issue.
I've added extra cases to the eqnarray and split testcases to guard against later misguided fixes.


dginev (approved):
Can't really find a good external test, but the few bits I tested had no obvious regressions - so it's likely a much better approach!

And thank you for adding the new tests, those will be quite helpful.

To my limited understanding, looks good!

PR #1761, opened 2022-01-18
Arxiv malformed

brucemiller:
This PR fixes a variety of problems revealed as Error/Malformed in the recent arXiv run; typically things like `ltx:section` appearing out of place due to mismanagement of lists, itemizations, bibliographies, tables and such.

Still some tempting candidates to look at ...

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -1504,10 +1504,18 @@ sub setEnumerationStyle {
     DefMacroI('\labelenum' . $level, undef, Tokens(T_BEGIN, @out, T_END)); }
   return; }
 
+# End paragraphs, like \par, but only if within an item's text
+DefConstructorI('\preitem@par', undef, sub {
+    my ($document, %props) = @_;
+    if (!$props{inPreamble} && !$document->getNodeQName($document->getElement(), 'ltx:itemize')) {
+      $document->maybeCloseElement('ltx:p');
+      $document->maybeCloseElement('ltx:para'); } },
+  alias => '\par');
+
 # id, but NO refnum (et.al) attributes on itemize \item ...
 # unless the optional tag argument was given!
 # We'll make the <ltx:tag> from either the optional arg, or from \labelitemi..
-DefMacro('\itemize@item', '\par\itemize@item@');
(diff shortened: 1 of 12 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -4029,7 +4033,8 @@ DefConstructor('\lx@mung@bibliography@pre', sub {
     if ($tag =~ /^ltx:(?:itemize|enumerate|description)$/) {
       $document->maybeCloseElement($tag); }    # Or even remove (if empty)?
     return; });
-DefConstructorI('\lx@bibnewblock', undef, "<ltx:bibblock>");
+DefConstructorI('\lx@bibnewblock', undef, sub {
+    $_[0]->openElement('ltx:bibblock') if $_[0]->isOpenable('ltx:bibblock'); });
```
dginev, line 4037: shouldn't we open _something_ when a block is not openable? Just to keep the contents grouped. Maybe a generic block or inline-block ?

brucemiller:
Might as well call this done, unless I find something new before you want it merged.

lib/LaTeXML/Package/LaTeX.pool.ltxml
brucemiller: The scenario seems to be that we've somehow already gotten into a bibitem and block when this is needed.

dginev: Oh, great then, nevermind 👍🏻

t/babel/numprints.xml
```diff
@@ -1352,19 +1352,14 @@ vs.
                 </XMath>
               </Math></td>
           </tr>
-          <tr>
-            <td align="char:." thead="row">before  after</td>
```
dginev, line 1356: ouch, the test was wrong here? Interesting find.

dginev (approved):
Nice, everything looks worth having! 

The `\lx@NAT@parselabel` looks a bit scary, but we can refine it after we hit arXiv with it and see what happens.

brucemiller:
> The ***@***.***@parselabel| looks a bit scary, 

Indeed; the biggest delay was convincing myself to code it,
rather than to figure out what the TeX version was even *trying* to do!

PR #1764, opened 2022-01-18
Ensure integers in width and height attributes

dginev:
A PR that achieves nothing except making the W3C validator a little happier.

`width` and `height` attributes need to be integers, as can be seen for [cond-mat/0103099](https://ar5iv.org/html/cond-mat/0103099) at [its validator check](https://validator.w3.org/nu/?doc=https%3A%2F%2Far5iv.org%2Fhtml%2Fcond-mat%2F0103099)

Just use the `int()` cast in a the final attribute placement of Graphics.pm, and we seem better? Works for the example article at least.


dginev:
The extra "fun" bit is that arXiv would need a full rerun to fix the validity issue. Luckily, it's *just* a validity issue, without material harm.

brucemiller:
Hmm, ok.

PR #1766, opened 2022-01-19
Group fixes for 100-error fatals in arXiv

dginev:
This PR is based on going through [the too-many-errors report](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/too%5Fmany%5Ferrors/100?all=false) and addressing what I could.

Namely:
 - Some math interpretation mismatches were due to the generalized `\over` implementation, which is a bit more eager than TeX's. Namely, TeX makes sure to not digest through alignment markers (T_ALIGN) and alignment ends (T_CS("\cr"), and its hidden friends). This gets documents such as `math/0001062` passing. It's the scariest thing I did, since it modifies the main `digestNextBody` routine. Luckily only in the alignment case.
 - a couple of easy allowable packages - bezier.sty and crckapb.sty (only in arXiv, so added it raw)
 - `\put` has to explicitly match its comma, or it is subtly different than needed
 - `\eqno` also needed to be guarded for `\cr` row endings
 - `\subheading`, similarly to `\abstract`, has uses with a single argument, as well as with an `\endsubheading` marker.
 - guard for custom uses of `\@startsection` in arXiv (e.g. `\newcommand\@app{\@startsection {app}{1}{0ex}...`), by defaulting unknown constructs to a section, and issuing a warning, rather than creating 100 malformed errors. Fixes `nlin/0001001`
- sprinkled in some easy macro additions where they were seen raising errors

I'm disappointed I couldn't fix `hep-th/0001124` which has a peculiar scoping issue. The source alternates multiple scoped definitions of `\put`, and latexml gets some of them incorrectly. One has a mandatory ` at ` keyword, the other doesn't. Leaving a breadcrumb here to find the source again later.

I'm walking through the second half of the report page now, but this PR is likely mostly done with - I'll add easy additional bits, if I find any, so feel free to review today. I'll try to wrap up by early afternoon.


dginev:
Also fixed `astro-ph/0002013`, which needed a little more careful hbox-ing of the `\leftline` argument, (similarly for the right and center case).

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -4926,13 +4926,16 @@ DefMacroI('\medbreak',   undef, '\par');
...
 DefMacroI('\line', undef, '\hbox to \hsize');
-DefConstructor('\leftline{}', sub {
+DefMacro('\leftline Undigested',   '\ltx@leftline{\hbox{#1}}');
+DefMacro('\rightline Undigested',  '\ltx@rightline{\hbox{#1}}');
+DefMacro('\centerline Undigested', '\ltx@centerline{\hbox{#1}}');
+DefConstructor('\ltx@leftline{}', sub {
     alignLine($_[0], $_[1], 'left'); },
   bounded => 1);
-DefConstructor('\rightline{}', sub {
+DefConstructor('\ltx@rightline{}', sub {
```
brucemiller, line 4935: You're going to want to add `alias=>'\rightline'` (and similar) to this kind of construct.

dginev:
Also upgraded `astro-ph/0002091` to Error status by avoiding the 100 errors from redefining `\fig` wrongly to a different number of arguments. It's usually harmless, so sadly another fix that is quite limited in reach.

Locked `\@@tabular` in the process, since I think redefining that can never yield good results.

dginev:
Another brave addition that recovers (the bigger part of) `cond-mat/0002070`, which had a high number of setters of the kind:
```tex
\textfont0=\csname\HEAD rm\@nd         
\scriptfont0=\csname\smHEAD rm\@nd 
```

and was dying with 100+ errors of "unexpected \endcsname".

I coded this as a special exception for the Token branch of `Gullet::readValue`, which seems to be abstract enough as to be correct, and available for all single Token registers.

brucemiller:
I'm not quite sure what's going on with that last commit, and how "Token" register comes into it. You did notice that `\@nd` is a macro expanding to `\endcsname`. The peculiar thing in that document seems to be that the `\csname` is expanding (where it usually doesn't)

dginev:
@brucemiller yes, exactly. My diagnostic was that this peculiar assignment:
```tex
\textfont0=\csname\HEAD rm\@nd         
```

does indeed expand `\csname` to obtain the token. Which, if I extrapolated to "the general case" correctly, ought to mean that whenever we set the value of a token register, and call `$gullet->readValue` with `$type eq 'Token'`, we should check for `\csname` and run the expansion for it.

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -640,8 +640,15 @@ sub readValue {
   elsif ($type eq 'Glue')      { return $self->readGlue; }
   elsif ($type eq 'MuGlue')    { return $self->readMuGlue; }
   elsif ($type eq 'Tokens')    { return $self->readTokensValue; }
-  elsif ($type eq 'Token')     { return $self->readToken; }
-  elsif ($type eq 'any')       { return $self->readArg; }
+  elsif ($type eq 'Token') {
+    my $token = $self->readToken;
+    if (Equals($token, T_CS('\csname'))) {
+      my $cstoken = $STATE->lookupDefinition($token)->invoke($self);
+      $self->unread(@{$cstoken});
```
brucemiller, line 647: better would be `$cstoken->unlist`

lib/LaTeXML/Core/Gullet.pm
```diff
@@ -677,6 +684,10 @@ sub readTokensValue {
       if (my $x = $defn->invoke($self)) {
         $self->unread(@{$x}); }
       return $self->readTokensValue; }
+    elsif (Equals($token, T_CS('\csname'))) {
+      my $cstoken = $defn->invoke($self);
+      $self->unread(@{$cstoken});
```
brucemiller, line 689: ditto

lib/LaTeXML/Core/Gullet.pm
dginev: done! thanks

dginev:
Having a good day over here. I'm on a roll with finding resourceful fixes, even feeling a little impressed. The last commit fixed `cond-mat/0002147` which had a some convoluted low-level latex.ltx judo. I didn't emulate fully, and most importantly I neutralized and locked `\@twocolumna` and `\@onecolumna` which were getting into some crazy loops related to `\vbox`. Once disabled, and sprinkling in another handful of small low-level definitions from latex.ltx, the document converts nicely.

It is late afternoon now, but I'd like to keep going, since I may actually fix through the entire first page of "100 Errors", which has a wholesome sound to it.

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -3418,11 +3436,11 @@ DefMacroI('\floatpagefraction', undef, "0.25");
 NewCounter('dbltopnumber');
 DefMacroI('\dbltopfraction',       undef, "0.7");
 DefMacroI('\dblfloatpagefraction', undef, "0.25");
-DefRegister('\floatsep'         => Glue(0));
-DefRegister('\textfloatsep'     => Glue(0));
-DefRegister('\intextsep'        => Glue(0));
-DefRegister('\dblfloatsep'      => Glue(0));
-DefRegister('\dbltextfloatsep'  => Glue(0));
+DefRegister('\floatsep'         => Glue('12.0pt plus 2.0pt minus 2.0pt'));
+DefRegister('\textfloatsep'     => Glue('20.0pt plus 2.0pt minus 4.0pt'));
```
dginev, line 3440: there was a macro parsing these by looking for `\def\foo #1 plus #2 minus #3 {work with #1 #2 #3} `, which was failing, so I chose to add the glue values for the various seps, as pdflatex reported them.

dginev:
As a motivating comment, I am down to 27 papers from the first page of 100 papers of this Fatal category that are still exhibiting a Fatal error. I went through them and there appear to be about four separate issues that look easily patchable. So, stay tuned :>

brucemiller:
> As a motivating comment, I am down to 27 papers from the first page of 
> 100 

Cool! Only 33223 to go!! :>

dginev:
Last commit should get rid of a certain kind of very annoying noisy error that was common whenever one used `--includestyles` to test some very rare CTAN package. Namely, a lot of the generic reporting methods of latex.ltx, such as `\GenericWarning`, end up with a trailing `\@gobble`, especially when they are added to hooks such as `\AtEndDocument`.

What we are currently "approximating" with our `make_message` subroutine in LaTeX.pool is an sandboxed `Expand()` that doesn't have access to the full gullet context where the native implementation would have gobbled. Thus, disable the gobbles for now. The _ideal_ fix would be to have `Expand` generalized to where it can be asked to use the available token streams to continue in cases where we didn't mean to insist on a sandboxed Mouth... But that's a future note.

dginev:
The last patch is mildly unsatisfying. First, it works! I recovered `math/0003117` from 100+ errors down to 2, and it produces a large manuscript with a range of features, including a glossary.

The glossary is exactly the tricky bit. The document relies on the `\glossary` macro from latex.ltx, and uses it *a lot*, in a lot of contexts where the latexml implementation simply isn't ready for it. An example would be:
```tex
  The Restoration Property requires that at any point of a trajectory, damage
occurs (or persists) only with small probability
($\eps$)\glossary{e.greek@$\eps$}.
  The Computation Property requires that the trajectory obey the
transition function in the absence of damage.
```
in the freely flowing text, which is ultimately a an auto-opened `ltx:p` element. latex.ltx naturally has a buffer through the `.aux` files, so there is no need for any sort of schema-compliance in where one uses the `\glossary` macro. Maybe it should evolve to be a misc-level mark that can be plugged in anywhere? Uncertain, and I'd need to do a lot more work to do it right - but I can switch those (and only those) uses to a Warning, so that the rest of the document can show up.

dginev:
Ok, I am fully done working here! ready for review.

If my counting is right, 95 of the first page of 100 Error articles should now complete without a Fatal. Some of the easier ones even without Error, but most still have errors. Hopefully this has some partial benefit for the other  33223 :>

One of the big changes by the time I reached the end of working on this PR is that, as we brainstormed with Bruce, the more experimental (i.e. dubious) styles and bindings can be  tried in a separate repository. Namely: [arxmliv-bindings](https://github.com/dginev/arxmliv-bindings).

So to run any of the trickier documents, a `--path=/path/to/arxmliv-bindings` needs to be added to the latexml call. And yes, I ended up adding a stub for `xy.tex.ltxml`, since one of the cases of 100 Errors was a document with lots of `\xymatrix` calls.

Anyhow, done with this weird sprint -- and a good reminder where the scariest pieces of TeX can be encountered in arXiv.

dginev:
Lovely: I have rebased on the latest master and retested ~10 of the articles I had worked on, also picking up some from the second page of "100 Errors".
Everything I tried converts to good looking HTML without Fatals.

So the rebase was a success, and this PR is again ready for review. 

For complete context, here is the command I use to test. It has a few arXiv-specific options (yet not **all** options used in cortex, though the missing ones are mostly unrelated, e.g. math lexeme output).

```
latexmlc paper.zip --dest=paper.html \
   --preload=[nobibtex,ids,localrawstyles]latexml.sty  \
   --nodefaultresources --css=ar5iv.css 
   --path=~/git/arxmliv-bindings 
   --inputencoding='iso-8859-1' 
   --timeout=3600
   --verbose --verbose
```

lib/LaTeXML/Package/AmSTeX.pool.ltxml
```diff
@@ -193,6 +194,14 @@ Let('\D',           "\\.");
 
 DefMacro('\.', '. ');
 
+# instead of adding taylor's diagram.tex
+# for now warn that we are aware of those diagrams, but they are not supported
+# and avoid the sea of otherwise inevitable errors during drawing
+DefMacro('\diagram Until:\enddiagram', sub {
+    Warn('missing', 'support', 'The \diagram mechanism of diagram.tex is not currently supported, output is degraded.');
+    return Tokenize("missing diagram")->unlist(); },
+  locked => 1);
```
dginev, line 203: btw I did end up adding `diagram.tex.ltxml` in the arxmliv-bindings repo, where I defined:
```
# see arXiv:math/0002180 for an ambitious use.
DefEnvironment('{diagram}', 'diagram.tex is currently unsupported', locked=>1);
```

Wondering if I should delete the guard in AmSTeX.pool.ltxml, though nothing really wrong with guarding in two separate ways (maybe a different package has an undefined \diagram?)

brucemiller:
Scary due to the volume (hard to see everything), but what I can see looks reasonable, and dlmf seems to survive.

Let me know when you're ready (cause more seems to keep popping up :> )

dginev:
It's ready! As you say, I can keep adding, and the PR is big enough :>

Might as well rerun the 11.2021 sandbox and see if we look good, and I can put any new avant-garde patch ideas into a separate PR.

PR #1767, opened 2022-01-21
More rigour in package fallback

dginev:
We discussed this recently in our group meeting, and I claimed fallbacks were implemented - but indeed rather limited in what they did. I spotted a case that could've used a mildly better application of the regexes, and also a handful of simple cases that could be swept in with an easy rule or two.

Here's a minimal example of what the PR could do that couldn't be done before:
```tex
\documentclass{svjour3-id}
\usepackage{my-emulateapj_ed,amsmath-arxiv-v2021,aastex62_rev}
```
induces:
```
Info:fallback:svjour3-id.cls Interpreted 3-id as a versioned package/class name, falling back to generic svjour.cls at test.tex; line 2 col 11

Info:fallback:my-emulateapj_ed.sty Interpreted my-..._ed as a versioned package/class name, falling back to generic emulateapj.sty at test.tex; line 3 col 6

Info:fallback:amsmath-arxiv-v2021.sty Interpreted -arxiv-v2021 as a versioned package/class name, falling back to generic amsmath.sty at test.tex; line 3 col 6

Info:fallback:aastex62_rev.sty Interpreted 62_rev as a versioned package/class name, falling back to generic aastex.sty at test.tex; line 3 col 6
```

If I wanted to really properly develop this feature today (do I? maybe I do! someone should encourage/discourage me), I would proceed to collect *ALL* missing_file names from [their report](https://corpora.mathweb.org/corpus/arxmliv/tex_to_html/warning/missing_file?all=true), then intersect their names with the names of supported bindings, and turn that smaller list into a unit test. The Perfect fallback function should be able to take any custom name from arXiv and find its latexml equivalent. This PR isn't even half of the way there I fear. But it re-starts the ball rolling 👍🏻

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -2009,14 +2009,17 @@ our @find_fallback_suffixes = (
   'arx', 'arxiv',     'conference', 'workshop',
   'tmp', 'alternate', 'preprint',   'fixed',
   # arbitrary 1,2-letter trailing markers
-  '\w\w?',
+  # tragically this is too general:
+  # -- tikz-cd.sty is a real package
+  # '\w\w?',
+  #
   # version-oriented suffixes
   '[vV]?[-_.\d]+',
   # mixed
   'old',      'new',    'final', 'clean',
   'mine',     'priv',   'rev',   'mod',
   'modified', 'edited', 'custom',
-  'altered',
+  'altered',  'rtx'
 );
 # suffixes without a separator
 our @find_fallback_glued_suffixes = (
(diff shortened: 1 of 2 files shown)


dginev:
Ok, ready for review 👍🏻

brucemiller:
You want encouragement! Yeah!!! (depending on encouragement for *what*! :> )

I actually was contemplating exactly this issue with a "my-" prefixed style file. In the end, I was able to get it through anyway.  This will apply to sty as well as cls, right?

A similar, but distinct, issue I encountered: a local but ancient copy of `algorithmic.sty`; The binding loads the TeX definitions, but finds the local copy before the texmf one.  Which is right sometimes, and wrong others. I wonder if the search order should be modified?

dginev:
Applies to both .sty and .cls files yep - the PR is just adding more cases and getting more consistent.

And oh no, you encouraged me, now it's getting scary :>

PR #1771, opened 2022-01-23
macro \orcidID addition to llncs.sty

dginev:
As stated.

Also neutralized two minor pdf-related macros.

### DIFF
--- a/lib/LaTeXML/Package/llncs.cls.ltxml
+++ b/lib/LaTeXML/Package/llncs.cls.ltxml
@@ -241,6 +241,9 @@ DefMacro('\homedir', "\~{ }");
 DefMacro('\idxquad', '\hskip 10pt\relax');
 #======================================================================
 
+# Newer llncs versions have an ORCID id macro
+DefConstructor('\orcidID{}', '<ltx:sup><ltx:ref href="https://orcid.org/#1">#1</ltx:ref></ltx:sup>');
+
 #  /--------------------------------------------------------------------\
 # | INCOMPLETE IMPLEMENTATION                                            |
 # | remove this comment, when done.                                      |
--- a/lib/LaTeXML/Package/llncs.cls.ltxml
+++ b/lib/LaTeXML/Package/llncs.cls.ltxml
@@ -242,7 +242,9 @@ DefMacro('\idxquad', '\hskip 10pt\relax');
 #======================================================================
 
 # Newer llncs versions have an ORCID id macro
-DefConstructor('\orcidID{}', '<ltx:sup><ltx:ref href="https://orcid.org/#1">#1</ltx:ref></ltx:sup>');
+DefMacro('\orcidID Semiverbatim', '\@add@to@frontmatter{ltx:creator}{\@@@orcid{\@@orcid{#1}}}');
+DefConstructor('\@@orcid{}', '<ltx:ref title="ORCID identifier" href="https://orcid.org/#1">#1</ltx:ref>');
+DefConstructor('\@@@orcid{}', '^<ltx:contact role="orcid">#1</ltx:contact>');
 
 #  /--------------------------------------------------------------------\
 # | INCOMPLETE IMPLEMENTATION                                            |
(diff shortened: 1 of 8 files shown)


lib/LaTeXML/Package/llncs.cls.ltxml
```diff
@@ -241,6 +241,9 @@ DefMacro('\homedir', "\~{ }");
 DefMacro('\idxquad', '\hskip 10pt\relax');
 #======================================================================
 
+# Newer llncs versions have an ORCID id macro
+DefConstructor('\orcidID{}', '<ltx:sup><ltx:ref href="https://orcid.org/#1">#1</ltx:ref></ltx:sup>');
```
brucemiller, line 245: Seems like this should be frontmatter/metadata?

dginev: indeed, it is used as a superscript on the author's name in the main frontmatter author list. It can be enhanced, just not sure what the perfect tag setup would be.

brucemiller:
Probably something similar to email; `ltx:contact`.
(would be even more clever to turn it into a url, but...)

dginev:
Would you be OK with me adding a new role for ORCID identifiers? So in `LaTeXML-structure.rnc`:
```yaml
## Attributes for \elementref{contact}.
contact_attributes = 
  Common.attributes,
  FrontMatter.attributes,
  Refable.attributes,

  ## indicates the type of contact information contained.
  ## Commonly useful values are specified, but is open-ended to support extension.
  attribute role { "affiliation" | "address" | "current_address" | "email" | "url"
                 | "thanks" | "dedicatory" | "orcid" | text }?
```

and regenerate all the files? if I was to explain what ORCID ids are in a phrase, it would maybe be "external_identifier", of some sort.

Otherwise I am not quite sure how to use the `ltx:contact` element, since the available roles are not really appropriate. (and it gets stripped out, respectively)

brucemiller:
The ordid role would be fine, though it is unnecessary since the final "text" means that `@role` is open-ended.  However, it seems that it would have been useful to have a `@href` attribute, which would be handy for depositing the url associated with id (or any of the other contact roles; even email).

Aside: the `role=url` case seems awfully unspecific, but I guess that crops up. I guess in those cases (seems only 3) the url is simply the text content.

dginev:
Ok, updated here. I reluctantly waded into the schema+XSLT domains and added the smallest possible provisions for ORCIDs to be explicitly recognized. It's a good point that the addition to the RelaxNG is optional, but I suspect `orcid` and `doi` metadata will be important to formalize in future releases, as they are getting a lot of recent adoption.

The PR now generates the XML:
```xml
<contact role="orcid">
    <ref href="https://orcid.org/0000-0002-4021-2476" title="ORCID identifier">
       0000-0002-4021-2476
    </ref>
</contact>
```

and the follow-up HTML:
```html
<span class="ltx_contact ltx_role_orcid">
  <a href="https://orcid.org/0000-0002-4021-2476" title="ORCID identifier" class="ltx_ref">
      0000-0002-4021-2476
  </a>
</span>

File: lib/LaTeXML/resources/XSLT/LaTeXML-structure-xhtml.xsl [hunk 2]
```diff
@@ -593,6 +595,27 @@
...
+      </xsl:apply-templates>
+      <xsl:apply-templates>
+        <xsl:with-param name="context" select="$innercontext"/>
+      </xsl:apply-templates>
+      <xsl:apply-templates select="." mode="end">
+        <xsl:with-param name="context" select="$innercontext"/>
+      </xsl:apply-templates>
+    </xsl:element>
+    <xsl:text>&#x0A;</xsl:text>
+  </xsl:template>
```
dginev, line 617: this copy-pasta was rather bothersome for me to create... I wonder if we could get a "Generic" contact rule that simply copies over the content and add *ALL* generics under it via `@role='X' or @role='Y' or...`. With a comment explicitly inviting developers from the future to add more generics to it :> 

I don't know enough about the code here at the moment, so I stayed away from such sweeping changes.

dginev:
Did a quick check and spotted an `\orcid` macro in acmart.cls, which could benefit from the link -- so I've also added the ltx:ref in there now. Worked nicely on `arXiv:2111.00824`

brucemiller:
Cool! looks good!!

PR #1779, opened 2022-01-23
revisit authblk.sty model

dginev:
authblk.sty was a bit too carefree about the many-to-many relationship between authors and affiliations, and the way the marks were written down.

I rewrote the code to do the rearrangement in a single pass, at the end of the document, cloning each affiliation note under each author that uses it. I am a bit uncertain about the intended XML markup for multiple affiliations, so this is still a work in progress, until I get some help from @brucemiller . But once we get that in, I can style it in the CSS and rerun all authblk articles.

Pretty much any authblk.sty paper would be a good test, the one I had in front of me was arXiv:2111.15452 .

The 2021 texlive style file had some extra macros (in particular the \affilsep was used), so I updated the binding coverage as well.

### DIFF
--- a/lib/LaTeXML/Package/authblk.sty.ltxml
+++ b/lib/LaTeXML/Package/authblk.sty.ltxml
@@ -43,7 +43,15 @@ DefMacro('\author[]{}', sub {
 
 DefMacro('\lx@ab@author[]{}',
   '\@add@frontmatter{ltx:creator}[role=author]'
-    . '{\@personname{#2}\lx@authormark{\if.#1.\the@affil\else#1\fi}}');
+    . '{\@personname{#2}\lx@split@authormark{#1}}');
+
+# Preserve the order, where we add an "affiliationmark" element
+# for each numeric mark in the \author[] optional argument
+DefMacro('\lx@split@authormark{}', sub {
+    if (my @marks = split(/\s*,\s*/, ToString($_[1]))) {
+      return (map { T_CS('\lx@authormark'), T_OTHER($_) } @marks); }
+    else {
+      return (T_CS('\lx@authormark'), T_CS('\the@affil')); } });
 
 DefConstructor('\lx@authormark{}',
   "^?#mark(<ltx:contact role='affiliationmark' _mark='#mark'></ltx:contact>)()",
@@ -74,21 +82,12 @@ sub authblkRelocateAffil {
     # the last node wins
     if (my $mark = $affil_node->getAttribute('mark')) {
       $mark_to_affil{$mark} = $affil_node; } }
-  # The most accurate approach here is to go author-by-author,
-  # and attach the affiliations in the *exact order* they were requested
-  # \author[1,7]{A} and \author[7,1]{B} are intended to be presented in that order.
+  # Now process all affiliationmarks
   for my $author_node (@author_nodes) {
-    my @author_marks = split(/\s*,\s*/, ($author_node->getAttribute('_mark') || ''));
-    next unless @author_marks;
     $author_node->setAttribute(role => 'affiliation');
-    my $author_parent = $author_node->parentNode;
-    for my $mark (reverse(@author_marks)) {
-      if (my $affil_node = $mark_to_affil{$mark}) {
-        my $new_contact = $author_node->cloneNode(0);
-        $author_parent->insertAfter($new_contact, $author_node);
-        $document->appendClone($new_contact, $affil_node->childNodes); } }
-    # at the end, unbind each author-mark template node
-    $author_parent->removeChild($author_node); }
+    my $mark = $author_node->getAttribute('_mark') || '';
+    if (my $affil_node = $mark_to_affil{$mark}) {
+      $document->appendClone($author_node, $affil_node->childNodes); } }
   return; }
 
 # we shouldn't need these, but if someone is using them low-level
--- a/lib/LaTeXML/Package/authblk.sty.ltxml
+++ b/lib/LaTeXML/Package/authblk.sty.ltxml
@@ -64,21 +64,31 @@ Tag('ltx:document', afterClose => \&authblkRelocateAffil);
 
 sub authblkRelocateAffil {
   my ($document)   = @_;
-  my @affil_nodes  = $document->findnodes(".//ltx:note[\@role='affiliationtext']");
   my @author_nodes = $document->findnodes(".//ltx:contact[\@role='affiliationmark' and \@_mark]");
+  my @affil_nodes  = $document->findnodes(".//ltx:note[\@role='affiliationtext']");
+  # first, unbind all affiliation text nodes - they get reparented or discarded.
+  my %mark_to_affil = ();
   for my $affil_node (@affil_nodes) {
-    # first, unbind all nodes - they get reparented or discarded.
     $affil_node->parentNode->removeChild($affil_node);
-    # next, see if we can find a new home.
-    if (my $affil_mark = $affil_node->getAttribute('mark')) {
-      # Find authors with same mark.
-      for my $author_node (@author_nodes) {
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Package/authblk.sty.ltxml
```diff
@@ -51,19 +60,25 @@ DefConstructor('\affil[]{}',
...
-        $document->appendClone($author, $affilnode->childNodes);
-        $author->setAttribute(role => 'affiliation'); }
-      $affilnode->parentNode->removeChild($affilnode); } }
+      for my $author_node (@author_nodes) {
+        my @author_marks = split(/\s*,\s*/, ($author_node->getAttribute('_mark') || ''));
+        next unless (grep { $_ eq $affil_mark } @author_marks);
+        # An affiliation node may be reused across authors,
+        # but should still get unlinked the first time we reparent it
+        $document->appendClone($author_node, $affil_node);
+        $author_node->setAttribute(role => 'affiliation'); } } }
```
brucemiller, line 81: So, the affiliations are created as `ltx:note` because they haven't (yet) got an author to contain them (in which case they would just be `ltx:contact`); right? But when you're copying the affiliation into the author contact, it should be the contents of the note, not the whole note.

brucemiller:
If there are multiple affiations, there should just be multiple `ltx:contact` within the `ltx:creator`.  Ideally, there should be roles to distinguish the different affiliations (if we know of any difference).

dginev:
Ok, I got the multiple affiliation case working, thanks for the guidance! I had to peel back several layers of assumptions I had had in that code:

- a single pass, in a single loop, is only possible at the end of document construction, where all such metadata has been collected (or end of preamble, but just in case). If I re-parent nodes midway, thinking of what has and has not been encountered and absorbed already is rather confusing.
- the other bit is to remember the many-to-many relationship between authors and affiliations, and create a datastructure keyed into the marks that can be asked for lookups.
- and finally, the affiliations in \author are actually ordered meaningfully - `\author[7,1]` is different than `\author[1,7]`, as the first number is more "primary" than the second. This example is from the paper mentioned above.
- thus, loop through all authors, and look up their affiliation nodes in order, clone+reparent, and in the end - erase all unused nodes.

Sample XML:
```xml
<creator role="author">
  <personname>Katja Weigel</personname>
  <contact role="affiliation">
              University of Bremen, Institute of Environmental Physics (IUP),
              Bremen, Germany
  </contact>
  <contact role="affiliation">
              Deutsches Zentrum für Luft- und Raumfahrt (DLR), 
              Institut für Physik der Atmosphäre, Oberpfaffenhofen, Germany
   </contact>
</creator>
```

Sample HTML:
```html
<span class="ltx_creator ltx_role_author">
  <span class="ltx_personname">Katja Weigel</span>
  <span class="ltx_author_notes">
    <span>
      <span class="ltx_contact ltx_role_affiliation">University of Bremen, Institute of Environmental Physics (IUP), Bremen, Germany</span>
      <span class="ltx_contact ltx_role_affiliation">Deutsches Zentrum für Luft- und Raumfahrt (DLR), Institut für Physik der Atmosphäre, Oberpfaffenhofen, Germany</span>
    </span>
  </span>
</span>
```
Still a little irked by the unmarked span, since it would dodge possible styling rules for `ltx_personname` followed by `ltx_role_affiliation`. Luckily I don't have such a rule at the moment, but something to get back to.

The adjacent contact elements have now enabled me to write the ar5iv CSS rule:
```css
span.ltx_role_affiliation + span.ltx_role_affiliation {
  margin-top: 0.5rem;
  display: block;
}
```

which in combination with my experimental flex model for authors ends up looking as the following screenshot:



All in all, relieved to have the upgrade, ready for review!

dginev:
I'm not sure why but my commit is not showing up in the PR... yet it is visible in the [branch commits history](https://github.com/dginev/LaTeXML/commits/affilsep). Strange. Maybe it shows up by tomorrow...

lib/LaTeXML/Package/authblk.sty.ltxml
```diff
@@ -51,27 +60,41 @@ DefConstructor('\affil[]{}',
...
+    $affil_node->parentNode->removeChild($affil_node);
+    # if we get conflicting marks -- which we shouldn't
+    # the last node wins
+    if (my $mark = $affil_node->getAttribute('mark')) {
+      $mark_to_affil{$mark} = $affil_node; } }
+  # The most accurate approach here is to go author-by-author,
+  # and attach the affiliations in the *exact order* they were requested
+  # \author[1,7]{A} and \author[7,1]{B} are intended to be presented in that order.
+  for my $author_node (@author_nodes) {
+    my @author_marks = split(/\s*,\s*/, ($author_node->getAttribute('_mark') || ''));
```
brucemiller, line 81: The contents of this loop could be much simpler (just change the role attribute, then
```
$document->appendClone($author_node,$affil_node->childNodes));
```
(no new node, etc)
Except for this packing of multiple marks on a node; Where does this arise?  It seems counter to my intuition of what a mark is supposed to be.

dginev: Right, so the `$affil_node->childNodes` contains text nodes and text nodes only.

 What we want are sibling `<contact role="affiliation">` nodes,
 which is what I get by cloning the empty template node in  `$author_node`.

I would have preferred to make the new nodes from scratch, without any cloning, but I repeatedly couldn't find a way to do so without creating tens of lines of boilerplate. But anything that creates this tree would be OK by me.

dginev: As to multiple marks, `\author[1,7]{A}` would initially create the author node:
```xml
<ltx:contact role='affiliationmark' _mark='1,7'></ltx:contact>
```

at the right position in the document tree. Then the subroutine would parse the `'1,7'` in the attribute and deposit the two respective `<contact role="affiliation">` nodes next to it. And in the end, it will unbind the template `affiliationmark` node.

brucemiller: Exactly; that seems to me to be wrong. The `\author` should have created sibling contacts for each mark. Then the rest would be simple.

dginev:
I'm also a bit unsatisfied since it feels like we're losing some of the abstract capability of having the numbers pointing to a database-like collection of affiliation entries in the XML.

This is especially painful in cases similar to the CERN-like articles, where 100+ authors share an affiliation. Rather than  show a single marking number in the doc, and display the affiliation text once underneath, the current approach will copy the CERN information inside every `<creator role="author">` node.

Switching to a model where we ink the numeric marks, disconnected from the affiliation text, and then come back to post-process an href attribute, stays closer to the original model... Though it means redoing the PR again.

lib/LaTeXML/Package/authblk.sty.ltxml
dginev: I see, reasonable alternative -- I'll send a commit in a few minutes.

brucemiller:
Yeah, I can sympathize with that last point; I've wondered the same and it also applies to `ltx:note`.  In either approach, at *some* point you may find yourself wanting the note/contact content (which may be either local or remote), AND(OR) collecting up all such objects (eg. an endnotes section).  On top of that, you usually get `\footnote` or a simple affiliation, but occassionally separate `\footnotemark`,`\footnotetext` and the analogous author/affiliation linkages.

It seems that the 1st operation and implied model was more common, so I took the current approach of rejoining things. But it isn't necessarily the ultimate answer or cast in stone.  That said, I'm not so sure I'm anxious at this moment to dive into teasing these apart and tracking the repercussions :>

lib/LaTeXML/Package/authblk.sty.ltxml
dginev: Alright, refactored. Seeing the same final XML as in my overview comment. See if you like it :>

brucemiller:
Looks good! Thanks!!

PR #1780, opened 2022-01-23
stub in \DeclareRelease

dginev:
We load `glossaries.sty` raw from the ltxml binding, and in the 2021 texlive that raw file uses:
```tex
\DeclareRelease{v4.46}{2020-03-19}{glossaries-2020-03-19.sty}
\DeclareCurrentRelease{v4.49}{2021-11-01}
```

Maybe we can do something useful with that metadata? Until we figure out what, I suggest a stub is better than an error.


brucemiller:
If anything, we could `AssignValue` to something about the file (or even a hash of info about each file we've loaded)

dginev:
> If anything, we could `AssignValue` to something about the file (or even a hash of info about each file we've loaded)

Right, but I am unsure what, and for which further feature. So maybe we leave them empty for now?

PR #1781, opened 2022-01-24
Correctly define \arraybackslash

brucemiller:
as it says


dginev (approved):
Great! One-liner PRs are worth celebrating. 

Bookkeeping an intended article to get fixed by this was `arxiv:1507.07727`

PR #1783, opened 2022-01-24
Avoid divisions by zero in Number, 2022 edition

dginev:
This PR does **not** claim we shouldn't be fixing the underlying problems that lead to these divisions by zero. We should!

But in the meantime, we may get some Error-ing documents convert to their end, without perl dies.


brucemiller:
That's a pretty dubious fallback, but... can't think of anything better

PR #1784, opened 2022-01-25
add \csarg from comment.sty

dginev:
One-line PR that recovers `arXiv:1905.05970` back from the land of the Fatals to an almost fully functioning Error-status document.

I copied the macro definition as-is from its .sty file


dginev:
I snuck in another one-liner, fixing `1905.08869` by introducing `\boldsymbol` to IEEEtran.cls, but only with the `comsoc` option required.

This follows the raw class file directive somewhat strictly, and should avoid unsuspecting collisions with documents loading IEEEtran without that option set. Tested \boldsymbol with it toggled on/off, behaves as expected.

Feel free to merge 👍🏻

PR #1786, opened 2022-01-25
avoid tripping over on images ghostscript cant size

dginev:
Back to normal work mode - this PR is not urgent, and we can discuss at our leisure.

I have now isolated a single article which exhibits a range of problems when processed with ghostscript.

The first, and in a way mildest one, is easy to fix. Namely, we need to guard against the sizing estimate returning empty/undefined. 

The latexml log informs us:
```
Warning:uninitialized:$w0 Use of uninitialized value $w0 in division (/)

at at /usr/local/share/perl/5.32.1/LaTeXML/Util/Image.pm line 309

In Post::Graphics[@0x5609af65ffc8] ->transformGraphic

Fatal:perl:die Perl died

Illegal division by zero at /usr/local/share/perl/5.32.1/LaTeXML/Util/Image.pm line 309.
```

The article in question is `1804.00311`, and I have also tied it together with the #902 issue. Some of the gs conversions take quite long (more than 5 minutes), but - to my surprise - if we avoid this Fatal error we actually get a Warning status conversion for the article.

### DIFF
--- a/lib/LaTeXML/Util/Image.pm
+++ b/lib/LaTeXML/Util/Image.pm
@@ -288,8 +288,15 @@ sub image_graphicx_complex {
   my $zoomout  = $properties{zoomout}  || 1;
   my ($xprescale, $yprescale) = (1, 1);
 
-  # # Wastefully preread the image to get it's initial size
+  # Wastefully preread the image to get it's initial size
   my ($w0, $h0) = image_size($source);
+  # 16x16 pixels (i.e. icon size) is the minimal viable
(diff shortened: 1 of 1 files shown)


brucemiller:
The alternative would be to (robustly) propagate (some flavor of) undef, or otherwise assert that there is no image, and we shouldn't try to do anything with it at all.  Of course that would mean verifying that all callers can cope.

dginev:
Closing in favor of #1811

PR #1787, opened 2022-01-26
Add missing macro to acmart.cls

dginev:
This is our first externally contributed error report for ar5iv that pertains to a latexml binding.

Nice and easy too!


brucemiller:
Seems straightforward

PR #1793, opened 2022-02-05
Allow standard fences as author name embellishments

dginev:
This PR allows standard fences at the start/end positions of `ltx:personname` content.

In international author names in arXiv, sometimes there are cases where a second language is written as a parenthetical. An example is:

```tex
\author{Jian-Ming Tang (湯健銘)}
```

The current regex I had written that avoids issues with commas and other punctuation, was also removing the trailing fence, creating a rather unpleasant string that is only open but not closed.

I keep wondering whether a blacklist is better than a whitelist, and switched again. In a simple world I would only remove comma characters, but there is more possible noise... For now let's just allow the fences?

Ready for review.


brucemiller:
All this fishing is fishy, but seems tolerable for now.

PR #1794, opened 2022-02-05
allow x-color spec in \fcolorbox

dginev:
We are reusing the `\fcolorbox` of color.sty, but the xcolor.sty one allows the extended color specification as well. Example input:
```tex
\documentclass{article}
\usepackage{xcolor}
\begin{document}

\fcolorbox{black!5}{orange!30}{My text}

\end{document}
```

I copied over the definition of `\fcolorbox` and allowed the extended by spec by using `ParseXColor` instead of the usual `ParseColor`.


brucemiller:
OK, seems right.

PR #1795, opened 2022-02-05
Disabled todonotes should not create notes

dginev:
This was reported for ar5iv [here](https://github.com/dginev/ar5iv/issues/24), as potentially harmful -- published articles with "disabled" todonotes should have the notes hidden.

Easy patch I think, ready for review.


brucemiller:
looks right.

PR #1796, opened 2022-02-09
allow bibliography post-convert to see the custom paths

dginev:
Now that the arXiv build is using a custom `--path` to extended bindings, it has become apparent that paths weren't passed into the MakeBibliography inner `->convert` call.

Which is quite important if one was working on say a custom biblatex binding...

perltidy reindented too many lines, my only actual change is on line 219.


brucemiller:
Looks straightforward!

dginev:
Fantastic, thank you @brucemiller !

PR #1799, opened 2022-02-11
[arXiv] avoid malformed \nocite, short titles in OmniBus

dginev:
Fixes for 2110.02279 and 1811.01740, also reported [in ar5iv](https://github.com/dginev/ar5iv/issues/38).

The changed pieces are:

* `\begin{thebibliography}{} \nocite{*}` was causing malformed errors, so I deferred the XML construction for `\nocite` to the document end, where a paragraph should be easy to auto-open. I also tried floating up, but that felt more fragile.

* I hit two other unsupported cases where `\title[]{}` was used with an optional argument. One was due to an unrecognized class, so I added the argument to OmniBus. But the other was due to a raw-interpreted unknown package that redefines `\title`. Since we are using `locked => 1` in LaTeX.pool (which is reasonable+mostly robust), I think we should allow the optional argument for any latex article. Nothing is really lost, as the argument is well - optional.

* some small copy-over additions to natbib.sty and article.cls for raw interpretations
* the special arxiv treatment for bbl-only interpretation was not passed in chapterbib.sty, so I did so - and recovered the references in 1811.01740.

That's all! Fixes the two articles.

### DIFF
--- a/lib/LaTeXML/Package/OmniBus.cls.ltxml
+++ b/lib/LaTeXML/Package/OmniBus.cls.ltxml
@@ -59,8 +59,12 @@ DefEnvironment('{frontmatter}', '#body');
 DefEnvironment('{mainmatter}',  '#body');
 DefEnvironment('{backmatter}',  '#body');
 
-DefMacro('\shorttitle{}',  '\@add@frontmatter{ltx:toctitle}{#1}');
-DefMacro('\subtitle{}',    '\@add@frontmatter{ltx:subtitle}{#1}');
+DefMacro('\shorttitle{}', '\@add@frontmatter{ltx:toctitle}{#1}');
+DefMacro('\subtitle{}',   '\@add@frontmatter{ltx:subtitle}{#1}');
+DefMacro('\title[]{}',
+  '\if.#1.\else\def\shorttitle{#1}\@add@frontmatter{ltx:toctitle}{#1}\fi'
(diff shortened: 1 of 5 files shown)


lib/LaTeXML/Package/OmniBus.cls.ltxml
```diff
@@ -59,8 +59,12 @@ DefEnvironment('{frontmatter}', '#body');
...
 DefEnvironment('{backmatter}',  '#body');
 
-DefMacro('\shorttitle{}',  '\@add@frontmatter{ltx:toctitle}{#1}');
-DefMacro('\subtitle{}',    '\@add@frontmatter{ltx:subtitle}{#1}');
+DefMacro('\shorttitle{}', '\@add@frontmatter{ltx:toctitle}{#1}');
+DefMacro('\subtitle{}',   '\@add@frontmatter{ltx:subtitle}{#1}');
+DefMacro('\title[]{}',
+  '\if.#1.\else\def\shorttitle{#1}\@add@frontmatter{ltx:toctitle}{#1}\fi'
+    . '\def\@title{#2}\@add@frontmatter{ltx:title}{#2}', locked => 1);
+
```
brucemiller, line 67: Shouldn't we just end up with LaTeX.pool's (newly enhanced) `\title`?

dginev: Agreed, that suffices. I've committed the change.

PR #1803, opened 2022-02-19
Fix expl3 regressions from 2021

### LINKED ISSUE #1762: latex 3 / lipsum.sty regression
Unsure if specific to my texlive 2021, but here is a minimal snippet that no longer generates output under latexml:
```tex
\documentclass{article}
\usepackage{lipsum}

\begin{document}
\lipsum[1-12]
\end{document}
```

Even though it also doesn't generate any errors (which is good at least).
There is something about the latest lipsum that is causing a completely unrelated error to the lack of output in a different context, which I'm currently debugging. I am not sure how quickly I'll be able to find a patch, so filing the issue in case it takes longer.

dginev:
Fixes #1762 .

* implements `\pdffilesize` to enable lipsum.sty loading correctly
* adding more of the most recent `latex.ltx` boilerplate available in texlive 2021 to a new `expl3.pool.ltxml` file
* adding a reliance on `expl3.ltx` for the supported packages I know already need it
* closer to the intended LaTeX 3 loading timing (expl3.ltx needs to load early+first)

Aside:
* Add a missing `\subref` to subcaption.sty and also fix the argument number. There is still a CrossRef inaccuracy with the numbering (off-by-one error), but that can be separated out in a later PR.


dginev:
Sadly, it appears I have over-expressed the concrete changes of texlive 2021, where tests now only pass on that latest texlive version. I wonder if this is mostly due to the change in `\fmtversion`, or there are deeper problems to align...

I'll see if I can install a texlive 2018 and find a patch working on both versions.

dginev:
Once again let me thank and praise @tkw1536 for creating the CI docker images. I am now debugging locally as easy as:
```
$ docker pull ghcr.io/tkw1536/latexml-test-runtime:2018-5.28
$ docker run -it ghcr.io/tkw1536/latexml-test-runtime:2018-5.28
root@5895417c01d4:/# wget https://github.com/dginev/LaTeXML/tarball/32d073c86dad624bfe4762fa8aec6c4631159e53    
root@5895417c01d4:/# tar -xvf 32d073c86dad624bfe4762fa8aec6c4631159e53 
root@5895417c01d4:/# cd dginev-LaTeXML-32d073c/
root@5895417c01d4:/dginev-LaTeXML-32d073c# perl Makefile.PL 
root@5895417c01d4:/dginev-LaTeXML-32d073c# make test
```

At which point I get the detailed test failures, and can continue debugging.

dginev:
Small oversight in looking for the raw `expl3.ltx` gets the older texlives passing again.

The last bit to fix is [xparse being fragile between versions](https://github.com/latex3/latex2e/issues/504#issuecomment-779738394).

Looking for the most helpful patch...

dginev:
What exactly does this "fragility" I mention look like at the macro level? Well, here is one of the cruxes, in expl3.sty.

texlive 2020:
```
\cs_gset_protected:Npn \__expl_status_pop:w #1#2 \q_stop
```

texlive 2021:
```
\cs_gset_protected:Npn \__expl_status_pop:w #1#2 \s__expl_stop
```

If the macro gets used in code that expects `\q_stop` but is defined expecting `\s__expl_stop`, the expansion fails to find its argument signature.

dginev:
Ok, I have successfully ported the code for all texlives.

Although there are still a couple of no-op errors in texlive 2020. I will keep looking at those, but I also welcome a review from @brucemiller at this point.

dginev:
I've removed most of the new latex.ltx boilerplate code loaded via RawTeX -- and I think I can remove the rest as well if needed. @brucemiller 

The bits I left in seemed simple enough to be predictable and harmless. Tested extensively with the examples in the issue, and another couple of arXiv papers, all looks good -- the regression symptoms are no longer encountered.

lib/LaTeXML/Package/siunitx.sty.ltxml
```diff
@@ -24,9 +24,10 @@ use LaTeXML::Package;
 #  * table alignments!
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 #======================================================================
+# recent siunitx have an expectation for latex3 capability.
+RequirePackage('expl3');
```
brucemiller, line 28: I'd hoped to avoid this, since straightforward use of siunitx should work without. However, the loading of expl3 may indeed have been expected, so...

brucemiller:
A bit scary, but seems right.

PR #1806, opened 2022-02-25
[WIP] Use U+2016 for fences in norms (and similar)

dginev:
### Description 
An interesting, and surprising, discussion transpired online about what the appropriate Unicode character is for norms written with double vertical lines. It's a bit non-linear to follow, but it [starts here](https://twitter.com/gro_tsen/status/1497209167949488136).

A helpful reference as part of it was the [Unicode Technical report #25](https://www.unicode.org/reports/tr25/), which makes clear (screenshot in the details).

<details> 

</details>

In summary, we should be using U+2016 when we recognize the "circumfix", i.e. fenced, cases. Currently we only use the U+2225 character that is only meant for (infix) operators.

### Approach

1. Rewrite after parse
    The most important missing piece was the ability to define Rewrite rules *after* MathParsing has completed. We are only aware of a "norm" use of these double-vertical-line characters (usually via the `\|` macro) when MathParsing has *completed* fully.
    To this end, I added a new option to rewrites, called `after_parse`, and postponed executing all rewrite rules with that option set to the end of the core processing, just before finalizing.

2. A simple rewrite rule can find the "norm" duals and swap the XMTok elements holding the fences.
3. I needed a guard in Font.pm that avoids scaling characters without known size in the metric tables. 
    - Maybe there is a better way to actually include metrics for the new use of U+2016 ?
4. Finally, I updated any tests using these notations from expecting `∥` to expecting the new `‖`

### DIFF
--- a/lib/LaTeXML/Common/Font.pm
+++ b/lib/LaTeXML/Common/Font.pm
@@ -723,7 +723,7 @@ sub merge {
   $flags     = ($$self[10] || 0) | $flags;
 
   if (my $scale = $options{scale}) {
-    $size = $scale * $size; }
+    $size = $scale * $size if $size; }
   # Set the mathstyle, and also the size from the mathstyle
   # But we may need to scale that size against the existing or requested size.
   my $stylescale = ($$self[3] ? $$self[3] / $stylesize{ $$self[9] || 'display' } : 1);
--- a/lib/LaTeXML/Core.pm
+++ b/lib/LaTeXML/Core.pm
@@ -227,14 +227,24 @@ sub convertDocument {
         $document->absorb($digested); }
       ProgressSpindown("Building");
 
-      if (my $rules = $state->lookupValue('DOCUMENT_REWRITE_RULES')) {
+      my $rules = $state->lookupValue('DOCUMENT_REWRITE_RULES');
+      if ($rules) {
         ProgressSpinup("Rewriting");
         $document->markXMNodeVisibility;
         foreach my $rule (@$rules) {
+          next if $$rule{after_parse};
           $rule->rewrite($document, $document->getDocument->documentElement); }
         ProgressSpindown("Rewriting"); }
 
       LaTeXML::MathParser->new(lexematize => $state->lookupValue('LEXEMATIZE_MATH'))->parseMath($document) unless $$self{nomathparse};
+
+      if ($rules) {
+        ProgressSpinup("Post-parse Rewriting");
+        foreach my $rule (@$rules) {
+          next unless $$rule{after_parse};
+          $rule->rewrite($document, $document->getDocument->documentElement); }
+        ProgressSpindown("Post-parse Rewriting"); }
+
       ProgressSpinup("Finalizing");
       my $xmldoc = $document->finalize();
       ProgressSpindown("Finalizing");
--- a/lib/LaTeXML/Core/Rewrite.pm
+++ b/lib/LaTeXML/Core/Rewrite.pm
@@ -19,12 +19,16 @@ use LaTeXML::Common::XML;
 
 sub new {
   my ($class, $mode, @specs) = @_;
-  my @clauses = ();
+  my @clauses     = ();
+  my $after_parse = 0;
   while (@specs) {
     my ($op, $pattern) = (shift(@specs), shift(@specs));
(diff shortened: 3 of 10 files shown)


dginev:
A related comment is that if this PR looks reasonable, I can also sneak in these rewrites for the infix single bar cases, such as with the `conditional` in `x | x \in S`.

On that front we are currently always using U+007C, where the Unicode spec recommends the unfortunately named "divides" at U+2223.

brucemiller:
I kinda like the idea of post-parse rewrites and have wanted it in the past, but I need to think about it.

brucemiller:
Nice ideas, but I'm thinking it's too narrowly focussed and missing an opportunity. Using rewrites is more flexible and extensible, particularly compared to the (misguided) `Fenced` procedure in `MathParser`, but they need to cooperate better. 

Indeed, this is a good opportunity to fix `Fenced`: it should probably do no more than construct a dual, change the roles of the fences to `OPEN`, `CLOSE` (why was I too paranoid to change `VERTBAR` to `OPEN|CLOSE`?  And why did *you* remove the `meaning='norm'` ?), probably encode the pattern (number of args, punctuation?) as the `meaning`.

Then, the post-parse rewrite rules would be responsible for correcting the meanings (norm, interval!!, etc) , unicode glyphs, etc.

There may be other special cases embedded in the grammar that could benefit.

The possibility of encoding some simplistic (but better than current) semantic analysis suggests there may be more phases than pre/post parse, so maybe things should be more open-ended (eg. `phase=>'post-parse'` instead of `post-parse=>1`).

And for post-parse rewrites to be palatable, they'll need to accept XML patterns, which will need to be not only digested and ingested, but also parsed before constructing the xpath.  This is perhaps the tricky bit.

dginev:
We can certainly extend and add more use cases here, I'm all ears. Changing to [WIP].

Comments on the pieces I already understand:
 - the `meaning` attribute on the fences in a presentation branch of a dual seems to be misleading more than it is helpful. The meaning is on the first token in the content branch of the dual, the presentation one can be drained from over-interpreting individual fences of a fenced notation.
 - I should mention this PR does not concern itself with semantics at all - as you say, it is very narrow. It concerns itself with following a presentation convention dictated by Unicode (which apparently readers claim to easily notice).

I need to look deeper into the XML pattern aspects, I would need a couple of examples to learn where the challenges are.

brucemiller:
Right, that was option 2B; Ideally the "intent" should be determinable 
from the parallel markup.  In the example at hand, although the verbars 
in presentation carry the norm-ness, perhaps your meaning is that it is 
misleading to say that their meaning is norm?

brucemiller:
I came back to this issue while pursuing a related problem from the opposite end. Namely `role=VERTBAR` should turn to `OPEN|CLOSE` when parsed as fences.  That much alone is rather easy; just replace a few lines in the grammar, for example:
```
     | VERTBAR absExpression VERTBAR addScripts[Fence($item[1],$item[2],$item[3])]
```
becomes
```
        | VERTBAR absExpression VERTBAR
            addScripts[Fence(Annotate($item[1],role=>'OPEN'),$item[2],Annotate($item[3],role=>'CLOSE'))]
```
Doing that quickly reveals the silly fact that `\|` is *not* "parallel-to" when used as fences, although it's plausible for infix.  And it doesn't address the issue you've raised here of adjusting the codepoints.

Well, both those points could be addressed by using some kind of variant of `New` instead of `Annotate`, which would copy/adjust various attributes from the token.  And it feels to me more correct to deal with this *during* parsing, rather than *after*.  BUT...

During parsing has the danger of sneaking things into the grammar (or MathParser) that are hard to adjust. OTOH, although doing it after with a rewrite is conceivably more flexible/correctable/extensible, the rewrite rule you wrote is so complex as to defeat that purpose :>  For example, by the time the rewrite sees the dual structure created by `|x|`, the (potential) absolute-value operator is so far from the vertbars, it becomes awkward to patch, while it's trivial at parse time.

Alternatively, maybe the `Annotate` can prime things for a later rewrite to simplify?

So, a post-rewrite sounds like a good feature, but not sure if it completely solves the problem.  Should I make a PR for `Annotate`, or do you want to try to incorporate that here?

dginev:
Closed in favor of #1868.

We can add the post-parse rewrite machinery in a separate pass.

PR #1810, opened 2022-03-05
Psfrag

brucemiller:
This PR arranges for empty `ltx:picture` elements (that have a tex attribute) to generate an image via LaTeX, whether or not the `--pictureimages` commandline switch is used. (If they are non-empty, the content determines the conversion, depending on (possibly) other post-processors).

This technique allows us to improve the `psfrag` bindng by collecting `\psfrag` calls to be used with an `\includegraphics` (or `\epsfbox`) and get LaTeX to combine them into a single image, rather than discard them.

It also allows us to write a binding for the `overpic` package.

Along the way, a few other random bugs were fixed.

### DIFF
--- a/lib/LaTeXML/Package/graphics.sty.ltxml
+++ b/lib/LaTeXML/Package/graphics.sty.ltxml
@@ -71,6 +71,7 @@ sub graphics_scaledbox_insert {
 DefConstructor('\Gscale@box {Float} [Float] {}', sub {
     my ($document, $scale, $yscale, $box, %props) = @_;
     graphics_scaledbox_insert($document, %props); },
+  alias      => '\scalebox',
   properties => sub {
     my ($stomach, $xscale, $yscale, $box) = @_;
     graphics_scaledbox_props($box, $xscale, $yscale || $xscale); },
@@ -98,6 +99,7 @@ DefConstructor('\Gscale@box@dddd {Dimension}{Dimension}{Dimension}{Dimension} {}
 DefConstructor('\Gscale@@box {} {GraphixDimension}{GraphixDimension} {}', sub {
     my ($document, $heighttype, $width, $height, $box, %props) = @_;
     graphics_scaledbox_insert($document, %props); },
+  reversion  => '\resizebox{#2}{#3}{#4}',
   properties => sub {
     my ($stomach, $heighttype, $width, $height, $box) = @_;
     my ($xscale, $yscale)                             = (1, 1);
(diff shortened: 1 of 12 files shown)


dginev:
Fascinating! The code reads very nice to me, the upgrades are quite sensible. I'll do a couple of tests and write back.

dginev:
I drew a random test (for extra fairness), with `math-ph/0011028`. Conversion succeeds without any issues, and I see a good match between the HTML and PDF. 

So this looks good from all sides! I'll go nitpick the code some more, but I couldn't find any obvious bits to protest on my first read.

Details left foldable:
<details>

```tex
\begin{figure}[h]
\psfrag{a}{$\tau_{1,2}$}
\begin{center}
\includegraphics[width=10cm]{permutinglegs.eps}
\end{center}
\caption{The action of a transposition}
\label{transposition}
\end{figure}
```

HTML:
image


PDF:
image

</details>

dginev:
I guess one could protest the sizing being mildly peculiar, but that's minor. The figure I mentioned above looks a bit small in the HTML, due to the narrow width/height in the style attribute.

Should we have both a `width` and `height` attribute for `` _and_ width and height set from within `style`? Maybe we can choose one over the other (at some point).

```html
<img id="S6.F4.pic1" class="ltx_picture ltx_centering" 
   style="width:284.0pt;height:59.3pt;vertical-align:-2px" src="pic/pic1.png"
   alt="\psfrag{a}{$\tau_{1,2}$}\includegraphics[width=284.52756pt]{permutinglegs.eps}" width="529" height="113">
```

dginev (approved):

lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -232,19 +234,47 @@ DefConstructor('\graphicspath DirectoryList', sub {
 DefMacro('\includegraphics OptionalMatch:* [][] Semiverbatim',
   '\@includegraphics#1[#2][#3]{#4}');
 
+sub graphics_get_candidates {
+  my ($path, $kv) = @_;
+  $path = ToString($path); $path =~ s/^\s+//; $path =~ s/\s+$//;
+  $path =~ s/^("+)(.+)\g1$/$2/;                           # unwrap if in quotes
```
dginev, line 240: One of these days we should have a `LaTeXML::Util::Pathname` helper for these path cleaning tricks.

brucemiller:
Yeah, I was wondering if the "missizing" was due to my code, or the lack 
of your beautification. Having style and size is certainly awkward, 
probably a resize?

PR #1812, opened 2022-03-06
Give XSLT more depth

dginev:
Example where that was needed is `arXiv:1612.02710`.

With the 500 depth, post-processing died with a Fatal error. With 1000, it succeeded smoothly.


brucemiller:
wow!

PR #1815, opened 2022-03-06
Epsf

brucemiller:
This fixes a problem with epsfbox not respecting the requested size (`\epsfxsize,\epsfysize`), and also that apparently emulateapj includes epsf.  There's also a gratuitous new binding for `nomencl`; it's not heavily used, but it optionally loads several other packages, so simply loading "dependencies" is error prone.

@dginev; you can go ahead & merge if you're happy with it, and need it.


dginev (approved):
Nice and direct, easy merge!

PR #1817, opened 2022-03-07
Add missing use to epsf.sty.ltxml

dginev:
Tested with `hep-ph/0009104`, avoids the Fatal
```
Undefined subroutine '&LaTeXML::Package::Pool::image_candidates' called at at /usr/local/share/perl/5.34.0/LaTeXML/Package/epsf.sty.ltxml line 51
```
[found here](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/misdefined/LaTeXML%3A%3APackage%3A%3APool%3A%3A%5F%5FANON%5F%5F?all=false)

It's fine if we catch more of these - easy to rerun the ones that leave Fatal traces.

Since we are rerunning a whole lot of epsf.sty, I am happy to merge here quickly and update the reruns.

### DIFF
--- a/lib/LaTeXML/Package/epsf.sty.ltxml
+++ b/lib/LaTeXML/Package/epsf.sty.ltxml
@@ -14,6 +14,7 @@ package LaTeXML::Package::Pool;
 use strict;
 use warnings;
 use LaTeXML::Package;
+use LaTeXML::Util::Image;
 
 #**********************************************************************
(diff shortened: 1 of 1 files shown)


brucemiller:
It was in my head, why didn't I type it!?!?! And I tested it, besides!?!?!  Thanks!

PR #1820, opened 2022-03-09
Enhancements to the LaTeXImage feature

dginev:
First, the defeating conclusion of my investigation here. It appears that providing a robust, predictable and full-featured capability for latex-based image production for arXiv, is essentially equally difficult to writing the bindings for that in Perl.

The intricacies I can enumerate after digging my teeth into this for a day:
 - establishing the exact package setup a paper expects,
 - ensure all custom arXiv assets are available on disk, 
 - collecting all relevant macro definitions and auxiliary files (e.g. `\begin{document}\input{mymacros}`) is undecidable in general
 - which executable should be used also needs robust heuristics. Currently we only use `latex`, but any use of PNG, JPG or PDF image assets requires `pdflatex`.
 - making sure UnTeX produce a TeX string that is as usable as the original input string is also tricky

---
On the mildly positive side, I did fix a handful of documents. For ~20 I tested, I fixed ~50-60%, when the problems were simple.
 - Adding some obvious assets - see the contents of [arxmliv-bindings/supported_originals](https://github.com/dginev/arxmliv-bindings/tree/main/supported_originals)
 - Avoiding empty optional arguments in reversions -- by actually checking if the reverted tokens were empty. `[]` can actually break in some old .sty packages which only expect the square brackets when they hold content.
 - protecting some spaces in UnTeX. I saw a case where `\\scriptscriptstyle\nx` lost the `\n` newline char and lead to a single macro `\scriptscriptstylex` which was undefined.
 - Some internal `\@@` macros didn't have a meaningful part to play in the picture. Or were impossible to emulate in vacuum - such as \@@cite.
 - Reversions with empy math could be stripped out for \psfrag, i.e. `{$$}` could instead be a simple empty arg `{}`
 - certain packages require even an empty `\abstract{}` or `\title{}` to be placed after `\begin{document}`, or the raise a latex error. So I made a provision for "in_body" definitions, when preparing the ltxmlimg.tex file.
 - loading english.sty raw via latex in texlive 2021 raises an error, so I mapped that to `\usepackage[english]{babel}`
 - using `JHEP.cls` has a whole lot of required macros that must exist, or errors are raised, so I instead mapped that back to article.cls - which works just fine.
 - I moved some latex safeguards that are for a specific class to only be added when that class is used.

Overall, every second paper I tested with seemed to have some very specific discrepancy. I kept thinking that each fix I landed only fixes a handful of articles in arXiv, and there will be many more articles left unfixed that need something similar - but impossible to generalize.

So, after all this, I think I will stop. LaTeX images appear equally difficult/hopeless as writing enough perl to emulate that process well in latexml itself.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4178,7 +4178,8 @@ AssignValue(CITE_UNIT           => undef);
 
 DefMacro('\@cite{}{}', '[{#1\if@tempswa , #2\fi}]');
 DefConstructor('\@@cite []{}', "<ltx:cite ?#1(class='ltx_citemacro_#1')>#2</ltx:cite>",
-  mode => 'text');
+  reversion => '\cite[#1]{#2}',
+  mode      => 'text');
 
 # \@@bibref{what to show}{bibkeys}{phrase1}{phrase2}
 DefConstructor('\@@bibref Semiverbatim Semiverbatim {}{}',
--- a/lib/LaTeXML/Package/psfrag.sty.ltxml
+++ b/lib/LaTeXML/Package/psfrag.sty.ltxml
@@ -33,6 +33,8 @@ ProcessOptions();
 
 sub save_psfrag {
   my (@stuff) = @_;
+  # avoid saving if nothing given.
+  return unless @stuff;
   # NOT global!!!  And defer expansion of the leading CS
   AssignValue(saved_psfragments =>
       Tokens(LookupValue('saved_psfragments')->unlist, T_CS('\noexpand'), @stuff));
@@ -44,6 +46,12 @@ sub unsave_psfrag {
 # NOT a constructor since probably args should not be digested yet(?)
 DefPrimitive('\psfrag OptionalMatch:* Semiverbatim [][][][]{}', sub {
     my ($stomach, $star, $a, $b, $c, $d, $e, $f) = @_;
+
+    # Guard against empty $$ argument.
+    my @argt = $f->unlist;
+    if (scalar(@argt) == 2 && $argt[0][1] == CC_MATH && $argt[1][1]) {
+      $f = Tokens(); }
+
     save_psfrag(Invocation(T_CS('\lx@delayed@psfrag'), $star, $a, $b, $c, $d, $e, $f));
     return; });
 DefConstructor('\lx@delayed@psfrag OptionalMatch:* Semiverbatim [][][][]{}', '',
@@ -156,4 +164,3 @@ EoTeX
 
 #**********************************************************************
 1;
-
--- a/lib/LaTeXML/Package/psfrag.sty.ltxml
+++ b/lib/LaTeXML/Package/psfrag.sty.ltxml
@@ -34,6 +34,7 @@ ProcessOptions();
 sub save_psfrag {
   my (@stuff) = @_;
   # avoid saving if nothing given.
+  @stuff = @stuff && Expand(@stuff);
   return unless @stuff;
   # NOT global!!!  And defer expansion of the leading CS
   AssignValue(saved_psfragments =>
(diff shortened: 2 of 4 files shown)


dginev:
Oh I missed one point: another bit of good news is that the available graphics paths seem to be reliably accurate. 

I made doubly sure `TEXINPUTS` is set locally and just before a system() call to that end, and everything clicks together even with ZIP inputs.

So at least I can say there shouldn't be any particular regression in the currently proceeding arXiv run that is specific to the CorTeX setup. Paths work well.

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -4178,7 +4178,8 @@ AssignValue(CITE_UNIT           => undef);
 
 DefMacro('\@cite{}{}', '[{#1\if@tempswa , #2\fi}]');
 DefConstructor('\@@cite []{}', "<ltx:cite ?#1(class='ltx_citemacro_#1')>#2</ltx:cite>",
-  mode => 'text');
+  reversion => '\cite[#1]{#2}',
```
brucemiller, line 4181: Won't this leave an empty `[]` if there's no optional arg? `alias=>'\cite'` might be better?

lib/LaTeXML/Package/psfrag.sty.ltxml
```diff
@@ -33,6 +33,9 @@ ProcessOptions();
 
 sub save_psfrag {
   my (@stuff) = @_;
+  # avoid saving if nothing given.
+  @stuff = @stuff && Expand(@stuff);
```
brucemiller, line 37: I guess `Expand` is ok here? The stuff gets digested later.  But I suppose you're trying to avoid having pointless fragments so as to avoid unnecessary LaTeXImages, right?

brucemiller:
Pretty dramatic summary! Glad you enjoyed yourself :>!!!!

lib/LaTeXML/Package/psfrag.sty.ltxml
dginev: My hope is that the Expand helps a more often than it breaks here.

The reason I added it was to expand custom macros that weren't otherwise available "in vacuum" for the call to `latex ltximg.tex`. From what I remember, the example I had was a macro introduced via `\DeclareMathSymbol` inside a psfrag argument, which was copied as-is and was unknown at the time latex was called.

dginev:
There is another upgrade needed that I spotted this morning, which I won't add to the PR here, but making a note.  In a way it was included in the generalities of my dramatic description :> Concretely:

Old-style code that does:
```tex
\documentclass[prl,aps,amssymb,amsmath,epsf,psfrag,multicol]{revtex}
```

gets recorded as:
```xml
<?latexml class="revtex" options="prl,aps,amssymb,amsmath,epsf,psfrag,multicol" oldstyle="true"?>
```

These options are never used by LaTeXImages, and they should. Or - even better - separate PI directives can be added when the implied files are actually loaded. Which leads me to point 2:

```tex
\input epsfig
\input psfrag.sty
```

Do not leave any trace in the PI instructions, and are thus invisible to post-processing.
I think at the least we should have the two Package methods `LoadLTXML` and `loadTeXDefinitions` double-check if the files they are loading were added already as PIs, and if not - add them then and there.

lib/LaTeXML/Package/psfrag.sty.ltxml
brucemiller: But all user macros should have been digested out later on (which also expands), before it ends up in the TeX attribute and ltximg.  Expand can be dangerous, because it defeats the usual interleaving of expansion/digestion.

dginev: I guess I can remove it, not much gained or lost either way.

dginev:
Thanks for the feedback! All comments were spot on, I have incorporated the changes.

Now that I have seen that LaTeXImages hasn't caused a major regression over arXiv, I have become a lot less ambitious about improving it in general. It's possible that such heavy effort is spent better elsewhere, at least at the moment :>

brucemiller:
The document prep for LaTeXMLImages is pretty scary, but I guess it's inevitable.
Thanks!

PR #1821, opened 2022-03-10
avoid Content MathML output for kludge parses

dginev:
In the interest of:
 * Reducing the size of the search indexes of Content MathML-based engines, such as MathWebSearch
 * keeping us a little more honest about what we claim to be *Content* MathML

I am proposing to completely abstain from generating Content trees for expressions where the latexml `MathGrammar` failed to find a grammatical parse.

Achieving that in practice is fairly simple - this PR offers one way to do that, by adding a `kludge_parse` class to Math elements using the kludge fallback, and the skipping over the Content MathML post-processor for them. 

I only added that for parallel processing -- which is also the mainstream configuration with which Content is generated with latexml today.

---

The argument here is to try and make a meaningful distinction between the branches of MathML. There are examples where an expression is not easily seen as meaningful. Those include:
 * Markup abuse and typos, such as `$(1+$\textbf{x}$=0)$`
 * expressions outside of the current MathGrammar coverage
 * expressions split via e.g. TeX alignments in ways that are hard to reconstruct by latexml (such as 2D diagrams)

Just an idea, happy to discuss further!


dginev:
I think this is good to merge on my end 👍🏻

brucemiller:
Firstly, I would prefer the negative `unparsed` over the ambiguously positive `kludge-parsed`.  Secondly, I think you only mark the top-level expression?  Since parsing works recursively on sub-expressions, you can end up with a parsed top-level, but signficant unparsed subexpressions.  So, to check whether you can generate "proper" content MathML you'd need to descend through the tree following the content branches of `XMDual` and look for unparsed.

And of course, the same question applies to OpenMath. So, rather than check the top-level `Math` element for unparsed and assume that only the 2nd processor can be ContentMathML, you really want a bit more API to the effect of `$processor->canConvert($node)` (or similar).  cMML & OM would get basically the same code.  Then in Post, you'd want to check whether all the processors can do their conversion; maybe even decline to set up the parallel markup if all secondary ones fail.

brucemiller:
And, if you do recognize unparsed subexpressions, you'll want to distinguish the content & presentation branches of `XMDual`; only parse failures within the content branch should cause the expression to be considered as unparsed.

brucemiller:
Currently, unsuccessful parses in *subexpressions* are reflected by `ltx:XMWrap` (or `ltx:XMArg`) remaining in the tree; however, the top-level expressions do not currently have this wrapper.  I had suggested off-line that perhaps we should be adding this wrapper to all expressions upon creating them, in advance of parsing. The wrapper would be left in place when parsing the top-level expression failed (as it does with subexpressions); this would be more consistent representation than introducing a distinct attribute. But I'm having mixed feelings about that; there are lots of places where top-level expressions are both created, and inspected, that need changes.

Maybe a better approach (at least it's more local change, if perhaps less consistent) would be for MathParser to add that wrapper when a (top-level) parse failed.  Then, at least, you have a consistent test for whether an expression is convertible to a content form.

dginev:
I've implemented most of the feedback here - thanks! Commits incoming shortly. One comment.

> only parse failures within the content branch should cause the expression to be considered as unparsed.

This was a bit underspecified to me, so here's one minimal example that comes to mind:
```tex
\binom{ a) }{b}
```

```xml
<Math mode="inline" tex="\binom{a)}{b}" text="binomial@([a)], b)" xml:id="p1.m1">
  <XMath>
    <XMDual>
      <XMApp>
        <XMTok meaning="binomial" name="binom" role="UNKNOWN"/>
        <XMRef idref="p1.m1.1"/>
        <XMRef idref="p1.m1.2"/>
      </XMApp>
      <XMWrap>
        <XMTok role="OPEN" stretchy="true">(</XMTok>
        <XMApp xml:id="p1.m1.3">
          <XMTok mathstyle="text" role="FRACOP" thickness="0pt"/>
          <XMArg xml:id="p1.m1.1">
            <XMWrap class="unparsed">
              <XMTok font="italic" fontsize="70%" role="UNKNOWN">a</XMTok>
              <XMTok fontsize="70%" role="CLOSE" stretchy="false">)</XMTok>
            </XMWrap>
          </XMArg>
          <XMTok font="italic" fontsize="70%" role="UNKNOWN" xml:id="p1.m1.2">b</XMTok>
        </XMApp>
        <XMTok role="CLOSE" stretchy="true">)</XMTok>
      </XMWrap>
    </XMDual>
  </XMath>
</Math>
```

There is indeed a nested `unparsed` annotation in the presentation branch of the dual, but deeply so, and I think that should still mark the overall expression as unparsed *IF* we take the strict interpretation. And so far I'm leaning strict.

brucemiller:
Exactly; sloppy language on my part. What I should have said was: parse failures in the *traversal* of the content branch (ie. following `XMRef`d children).

dginev:
@brucemiller ready for the next round of review. Maybe there are some tricky examples that I haven't thought of yet - should handle the ones I've already mentioned above.

A fun command that tests most math post-processors on a failing parse is:
```
latexmlc --whatsin=math 'literal:a---b' --pmml --om --cmml --keepXMath --mathtex --unicodemath  --preload=[mathlexemes]latexml.sty
```

Changing that into a successful parse with a simple `a-b` input should show an extra two math format annotations - Content MathML and OpenMath.

t/alignment/plainmath.xml
```diff
@@ -66,8 +68,10 @@
               </XMCell>
               <XMCell rowspan="2">
                 <XMWrap>
-                  <XMTok depth="6.8pt" font="italic" height="0" role="CLOSE" yoffset="-6.8pt">)</XMTok>
-                  <XMTok font="italic" height="6.8pt" yoffset="-6.8pt"> </XMTok>
+                  <XMWrap class="unparsed">
```
dginev, line 71: This is something I still want to fix - I'd like to reuse existing XMWrap elements rather than creating a new one. Let me see...

dginev: Hm, on the other hand my check is happening at a point where the failure to parse here happens on the single token holding the `)`. The outer XMWrap is unrelated to the specific pass through MathParser. So maybe this is OK as it is.

lib/LaTeXML/MathParser.pm
```diff
@@ -556,7 +556,13 @@ sub parse_kludge {
     my $kludge = $$pair[0];
     push(@replacements, (ref $kludge eq 'ARRAY') && ($$kludge[0] eq 'ltx:XMWrap')
       ? @$kludge[2 .. $#$kludge] : ($kludge)); }
-  $document->appendTree($mathnode, @replacements);
+  my $wrapped_replacements;
+  if (scalar(@replacements) == 1 and p_getQName($replacements[0]) eq 'ltx:XMWrap') {
+    $wrapped_replacements = $replacements[0];
+    p_setAttribute($wrapped_replacements, 'class', 'unparsed'); }
+  else {
```
brucemiller, line 563: The `@class` is not needed, since the presence of `XMWrap` or `XMArg` already signifies that it's unparsed.

lib/LaTeXML/MathParser.pm
```diff
@@ -1065,6 +1071,15 @@ sub p_getAttribute {
   elsif (ref $item eq 'XML::LibXML::Element') {
     return $item->getAttribute($key); } }
 
+sub p_setAttribute {
```
brucemiller, line 1074: Not really needed (at least not yet)

lib/LaTeXML/Post.pm
```diff
@@ -621,6 +623,28 @@ sub addCrossrefs {
...
+    my $content_branch = $doc->findnode('ancestor::ltx:XMDual/*[1]');
+    return 0 unless mathIsParsed($doc, $content_branch);
+    my @ref_nodes = $doc->findnodes('descendant::*[self::ltx:XMRef]', $content_branch);
+    for my $ref_node (@ref_nodes) {
+      my $rnode = $doc->realizeXMNode($ref_node);
+      # recursively check if the ref node is parsed, using the same rules, and fail if not
+      return 0 unless mathIsParsed($doc, $rnode); } }
+  # all looks nominal
+  return 1; }
+
```
brucemiller, line 647: Does that actually work? :>  I had in mind a more pedestrian:
```
sub mathIsParsed {
  my ($doc, $node) = @_;
  my $tag = $doc->getNodeQName($node);
  if (($tag eq 'ltx:XMWrap') || ($tag eq 'ltx:XMArg')) {
    return 0; }
  elsif (($tag eq 'ltx:XMTok') || ($tag eq 'ltx:XMText') || ($tag eq 'ltx:XMHint')) {
    return 1; }
  elsif ($tag eq 'ltx:XMRef') {
    return mathIsParsed($doc, $doc->realizeXMNode($node)); }
  elsif ($tag eq 'ltx:XMDual') {
    my ($content, $presentation) = element_nodes($node);
    return mathIsParsed($doc, $content); }
  else {
    foreach my $child (element_nodes($node)) {
      return 0 unless mathIsParsed($doc, $child); }
    return 1; } }
```
Untested, but should be close.

lib/LaTeXML/Post/LexMath.pm
```diff
@@ -32,6 +32,8 @@ sub convertNode {
 sub rawIDSuffix {
   return '.lm'; }
 
+sub canConvert { return 1; }
+
```
brucemiller, line 36: Use the OOPs, son :>
Let's assume that MathProcessors *can* convert, unless they claim otherwise.  So put this method on `LaTeXML::Post::MathProcessor`, and then it only needs to be overridden for `MathML::Content` and `OpenMath`.

t/alignment/plainmath.xml
brucemiller: It's kinda unpleasant, but I've got to look at the parse logic a bit more closely to see if/where the `XMWrap` needs to be added.

lib/LaTeXML/Post.pm
dginev: Wouldn't this perl recursion get pretty slow on large formulas in even larger documents? arXiv even has examples of documents that time out in some of those cases.

My initial attempt was to get a handle on all possible paths for a parse failure already in MathParser and just mark the top-level math element with a class, so that the check in post-processing is instant. Although I didn't manage that, since the dual cases where hard to get a handle on...

lib/LaTeXML/MathParser.pm
dginev: I wanted to at least have the option of it reaching the HTML, so that formulas that failed to parse _could_ be styled differently, e.g. for web editing apps that use latexml behind the scenes and want all of their formulas to be parseable (and thus indexable).

lib/LaTeXML/Post/LexMath.pm
dginev: Sure, done.

lib/LaTeXML/MathParser.pm
brucemiller: Hmm, that's getting ahead of yourself. But if so, then the class name doesn't follow our conventions.

brucemiller: And, then probably should be integrated into post processing, so that pmml would add a class to the top-level, rather than redundantly carrying it around.

dginev:
@brucemiller got the PR updated to the latest approach we discussed. Some cases that change in the tests are somewhat borderline/hard to detect, or - worse - I am not that well-studied on the specifics.

Say the `plainmath.tex` test, has:
```tex
$$ \bordermatrix{ & A & B \cr C & 0 & 1 \cr D & 2 & 3} $$
```

which produces an open and close aligned parenthetical which now gets marked as unparsed, namely the XML:
```xml
            <XMRow>
              <XMCell align="center">
                <XMTok font="italic" role="UNKNOWN">C</XMTok>
              </XMCell>
              <XMCell rowspan="2">
                <XMWrap depth="6.8pt">
                  <XMWrap class="unparsed">
                    <XMTok depth="6.8pt" font="italic" height="0" role="OPEN" yoffset="-6.8pt">(</XMTok>
                    <XMTok font="italic" height="6.8pt" yoffset="-6.8pt"> </XMTok>
                  </XMWrap>
                </XMWrap>
              </XMCell>
              <XMCell align="center">
                <XMTok meaning="0" role="NUMBER">0</XMTok>
              </XMCell>
              <XMCell align="center">
                <XMTok meaning="1" role="NUMBER">1</XMTok>
              </XMCell>
              <XMCell rowspan="2">
                <XMWrap>
                  <XMWrap class="unparsed">
                    <XMTok depth="6.8pt" font="italic" height="0" role="CLOSE" yoffset="-6.8pt">)</XMTok>
                    <XMTok font="italic" height="6.8pt" yoffset="-6.8pt"> </XMTok>
                  </XMWrap>
                </XMWrap>
              </XMCell>
            </XMRow>
```

Should I tentatively allow parse failures in array cells (simple: skip over that element when checking), or is there some better strategy to recognize these cases?

brucemiller:
Cool! The problem is starting to clarify itself, and as it does, we can spiral in on the Best solution.

Since the tricky part is now localized in MathParser, and assuming there's no real value in permanently marking unparsed subexpressions, the next suggestion may actually be closer to your original code, but should avoid lots of cruft in the output. 

You can use an internal attribute `_unparsed="1"` to mark unparsed nodes; it's always allowed and gets removed before serializing.  So, then you don't need to add the extra `XMWrap` wrapper, just set the attribute.  When you test in `is_genuinely_unparsed`, you can   use
```
   if (($tag eq 'ltx:XMWrap') || ($tag eq 'ltx:XMArg') || $node->hasAttribute('_unparsed')) {
```
Then, at the end, set the `XMath` attribute directly to `ltx_math_unparsed` and you wont have to translate it later in MathML.pm.

All that should reduce the diffs in the tests to exactly what we're looking for.

As for the border array; we might avoid trying to parse the weird bits by setting `role="kludge"`, but in the end, we don't (yet) actually have a semantically useful representation for the object; some kind of labelled array?

dginev:
@brucemiller Thanks, changes have been added. Retried the tests, all look good - now only with a top-level class on the main `ltx:Math` element in XMath, or `<math>` element in HTML.

The last question to settle is what to do with these special alignments like `\bordermatrix`. Currently no Content MathML will ever be generated for them, as they get marked as unparsed. If that is OK for now with you, happy to go with it.

lib/LaTeXML/MathParser.pm
```diff
@@ -561,9 +563,9 @@ sub parse_kludge {
   my $wrapped_replacements;
```
brucemiller, line 563: why not just add `_unparsed` to all the `@replacements` and not bother wrapping them at all?

dginev: Oh, sure.

dginev: Done.

brucemiller:
That looks pretty much optimal! Cool!! Thanks

PR #1822, opened 2022-03-10
neutralize \scalebox outside of pspicture

dginev:
`arXiv:math/0109128` was dying with 100 Errors for a curious reason - `\scalebox` from pstricks.sty.

It turns out that macro is fully usable without a pspicture, and LaTeXML kept firing schema errors each time an `ltx:g` was attempted to be inserted into an XMTok. One example being:

```tex
\newcommand{\simplex}{{\scalebox{1.3}{\color{gray}\blacktriangle}}}
```

This PR simply disables/neutralizes the scaling in non-pspicture contexts, which recovers that article to an Error status.

### DIFF
--- a/lib/LaTeXML/Package/pstricks.sty.ltxml
+++ b/lib/LaTeXML/Package/pstricks.sty.ltxml
@@ -554,7 +554,8 @@ DefEnvironment('{pspicture} [Float] PSCoord OptionalPSCoord',
     $whatsit->setProperties(need => 1, transform => $org->ptValue,
       pxwidth  => $c1->getX->subtract($c0->getX)->pxValue,
       pxheight => $c1->getY->subtract($c0->getY)->pxValue);
-    Let('\par', '\relax');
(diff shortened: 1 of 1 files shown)


brucemiller:
Hmm... probably should default to something like graphics' `\resizebox`? But maybe not worth the trouble figuring out if graphics is available?

dginev:
I could copy the `\resizebox` definition as `\text@scalebox` in pstricks.sty.ltxml if that sounds good?

brucemiller:
yeah, might as well.

brucemiller:
I don't necessarily think copying the whole `\resizebox` machinery is necessary, but probably adopting an approach like has been used in getting xy working. Namely, checking if we're within an svg context and then using either creating an `svg:g` or `ltx:inline-block` (with scaling attributes).  Alas, your example shows that maybe there's a 3rd case: inside math?  Maybe adjusting font size would be sufficient?

PR #1823, opened 2022-03-11
minor robustness patches for autoloading deluxetable.sty

dginev:
Minor. The `\deluxetable` was the one that flagged to me this was missing.

I also threw in a discardable `flushrt` while I was at the same article.

Recovers a Fatal reached due to 100 unexpected Alignment `&` errors, which were all due to the `\deluxetable` macro not being defined.

### DIFF
--- a/lib/LaTeXML/Package/deluxetable.sty.ltxml
+++ b/lib/LaTeXML/Package/deluxetable.sty.ltxml
@@ -34,9 +34,15 @@ DefMacroI(T_CS('\begin{deluxetable}'), "{}",
   '\set@deluxetable@template{#1}\def\@deluxetable@header{}\begin{table}');
 DefMacroI(T_CS('\end{deluxetable}'), undef, '\spew@tblnotes\end{table}');
 
+DefMacroI(T_CS('\deluxetable'),    undef, T_CS('\begin{deluxetable}'));
+DefMacroI(T_CS('\enddeluxetable'), undef, T_CS('\end{deluxetable}'));
+
 DefMacroI(T_CS('\begin{deluxetable*}'), "{}",
   '\set@deluxetable@template{#1}\def\@deluxetable@header{}\begin{table}');
 DefMacroI(T_CS('\end{deluxetable*}'), undef, '\spew@tblnotes\end{table}');
+
+DefMacroI(T_CS('\deluxetable*'),    undef, T_CS('\begin{deluxetable*}'));
+DefMacroI(T_CS('\enddeluxetable*'), undef, T_CS('\end{deluxetable*}'));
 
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Package/deluxetable.sty.ltxml
```diff
@@ -34,10 +34,16 @@ DefMacroI(T_CS('\begin{deluxetable}'), "{}",
   '\set@deluxetable@template{#1}\def\@deluxetable@header{}\begin{table}');
 DefMacroI(T_CS('\end{deluxetable}'), undef, '\spew@tblnotes\end{table}');
 
+DefMacroI(T_CS('\deluxetable'),    undef, T_CS('\begin{deluxetable}'));
+DefMacroI(T_CS('\enddeluxetable'), undef, T_CS('\end{deluxetable}'));
+
```
brucemiller, line 39: The original definition looks a little fishy, since it sets variables *outside* of the grouping.  Seems like `\deluxetable` and `\enddeluxetable` should have been defined as `\begin{deluxetable}`and `\end{deluxetable}` were (same for `\csname deluxetable*\endcsname`, etc).  Then `\begin{deluxetable}` would get it's `\begingroup` and all should be fine. I wonder why it wasn't?  Was there some previous bug? Or didn't yet know how to do the "*"? Or strange way the environment works that needs this?  Would have to look at  deluxetable uses to see.

brucemiller: Do you still have pointers to any documents that actually *use* (or misuse) deluxetable?  The package is included by aas, so it's getting loaded by very many documents, but seldom actually used. And the only deluxetable errors I come across are due to odd aas class/package that aren't recognized as being variants of it!

*IF* it works, it would be better to just define `\deluxetable` (and *'d) as the `\begin{deluxetable}` were defined.

brucemiller: Hmm. There's something oddly broken in that document. The first deluxetable works fine, but the 2nd is undefined; as if the stack disappeared!?!

dginev: > Hmm. There's something oddly broken in that document. The first deluxetable works fine, but the 2nd is undefined; as if the stack disappeared!?!

I think this is due to `DefAutoload` loading a binding's definitions into the local scope, so whenever it is triggered from deeper within the document, the definitions expire too early.

Needing DefAutoload in that document is in the first place a deficiency of the fallback, since `myemulateapj.cls` isn't getting recognized, due to the regex only looking at `my-`, `my.`, `my_` clearly separated prefixes. Maybe that's worth a separate PR change.

dginev:
Closing for #1875

PR #1824, opened 2022-03-11
allow absorbing active tokens

dginev:
This one is highly curious, and may require some mulling over -- since why would it have escaped us so far if what I am about to say is correct?

in `arXiv:cs/0010009`, [download link](https://get.ar5iv.org/source/cs/0010009.zip) we have a Fatal due to 100 Errors of the kind:
```
Error:misdefined:  The token T_ACTIVE[ ] should never reach Stomach! at paper.tex; line 229 col 10
```

But that seems odd? An active token may carry a definition, and the stomach should be perfectly capable of invoking that definition - unless I am misunderstanding the exact timing of that TeX operation.

The simple change of allowing the active tokens to be "absorbable" gets this arXiv article from a Fatal to a "no problem", and the respective active spaces print out the following in HTML:

Leaving it here for your consideration.

### DIFF
--- a/lib/LaTeXML/Core/Stomach.pm
+++ b/lib/LaTeXML/Core/Stomach.pm
@@ -149,7 +149,7 @@ our @CATCODE_ABSORBABLE = (    # [CONSTANT]
   0, 0, 0, 0,
   0, 0, 0, 0,
   0, 0, 1, 1,
-  1, 0, 1, 0,
+  1, 1, 1, 0,
   0, 0, 0, 0);
 
 sub invokeToken {


brucemiller:
Hmm, you'd think, but since CS is also excluded, whatever should happen with active should already have happened.

dginev:
Hmmm, this is extra mysterious then. May be worth giving a closer study.

dginev:
The bits in question are:
```tex
\catcode`\^^M=\active
\let^^M=\@xobeysp
```

Aha! That latex.ltx internal macro was never defined in our pool file. Ok, PR has changed to now include that definition - paper converts correctly again.

I have to wonder if we should also upgrade that error message... Maybe tomorrow.

brucemiller:
Adding
```
  elsif($CATCODE_ACTIVE_OR_CS[$cc]){
    return; }
```
just before the return of `$state->lookupDigestableDefinition` gets Error:undefined, which is more sensible. There are 23, since it's not globally defining it to ltx:ERROR; better than >100 :>. 

That probably allows eliminating the `if($meaning->isaToken) if($cc==CC_CS)` branch in `$stomach->invokeToken` as well.  (Which would be the alternative way to fix the problem changing to `($cc=CC_CS)||($cc==CC_ACTIVE)`.

brucemiller:
OBTW: I think you do *not* want to make `Let` give warnings for undefined.

PR #1827, opened 2022-03-11
load \jobname.rty for revtex4-1, when present

dginev:
Today I found out that revtex4-1.cls unconditionally loads a `\jobname.rty` file with custom definitions, when it is available on disk.

And of course I found that out by investigating an arXiv document's Fatal status, namely `1305.1045`, which has the files:
```
 chgauge.pdf             
 smodel.bbl              
 smodel.rty              
 smodel.tex   
```

This PR simply adds the check to revtex4-1.cls.ltxml and loads the `.rty` file when it is available.


brucemiller:
Finally, an easy one :>

PR #1829, opened 2022-03-12
text mode \footnote in TeX.pool

dginev:
Better fix to #1826, tested with `arXiv:math/0205199`.

Thanks to Bruce for teaching me the best way to upgrade.


dginev:
I had myself wonder whether we shouldn't make `\lx@note`, defined in LaTeX.pool, generically available from TeX.pool, and reorganize everything around that constructor.

What I have in TeX.pool is backward, since `\vfootnote` is what `\footnote` is defined with, rather than the other way around. The current code seems like a fast patch that does the same thing - until it doesn't :>

brucemiller:
Indeed, this is another one of those cases where you'd like to be reusing common infrastructure w/o necessarily require/use'ing a package.  However, the whole collection of footnote stuff in LaTeX relies a bit too heavily on very latexy stuff like a footnote counter.  Maybe that's not really a problem?

And also, you're right that we've got things a bit backwards: in fact `\vfootnote` is actually more like `\footnotetext` (`lx@notetext`) than `\footnote`.

brucemiller:
I somehow ended up with part of this fix already in master... but thanks!

PR #1832, opened 2022-03-17
Fix #1831: Make \hyperref expand first argument

### LINKED ISSUE #1831: \href does not expand first argument
Small bug: `\href` (from `hyperref.sty.ltxml`) treats its first argument as `Verbatim`, even though the argument may need expanding.

MWE:
```latex
\documentclass{article}
\usepackage{hyperref}
\def\myurl{https://example.com/}
\begin{document}
\href{\myurl}{Link}.
\end{document}
```
which gives you back
```xml
<ref class="ltx_href" href="\myurl">Link</ref>
```

tkw1536:
As reported in #1831, hyperref treats the first argument as 'Verbatim' and does not expand it.
This PR fixes the issue by making it a regular argument.

### DIFF
--- a/lib/LaTeXML/Package/hyperref.sty.ltxml
+++ b/lib/LaTeXML/Package/hyperref.sty.ltxml
@@ -151,7 +151,7 @@ DefRegister('\pdfcompresslevel', Number(0));
 # Additional User Macros
 
 # \href{url}{text}
-DefMacro('\href Verbatim {}', '\@@Url\href{}{}{#1}{#2}');
+DefMacro('\href {} {}', '\@@Url\href{}{}{#1}{#2}');
 
 # \url{url} from url.sty... well sorta
 # It's slightly different in that it expands the argument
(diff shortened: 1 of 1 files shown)


tkw1536:
Seems like [the tests](https://github.com/brucemiller/LaTeXML/actions/runs/2000146034) failed because:

> GitHub Actions 8 lost communication with the server

I think this is an internal error (maybe related to the outage earlier today?) given that they passed without issue [on my fork](https://github.com/brucemiller/LaTeXML/actions/runs/2000146025).

dginev:
The usual problem that lead to this argument being Verbatim is neutralizing all characters that would typically appear inside a URL, notably `#` and `_`.

I believe these are not in the test suite at the moment. Try using the macro on e.g. 
```tex
\documentclass{article}
\usepackage{hyperref}
\begin{document}
\href{https://example.com/a_b_c#d^e^f}{torture test}
\end{document}
```
which works in pdflatex

brucemiller:
Changing from `Verbatim` to a plain `{}` goes too far; it needs to neutralize most chars. Probably `Semiverbatim` is the right parameter type here? It will still allow control sequences.

Sadly, we don't really test `\href` in the tests, currently. Probably should add these tests to `t/expansion/hyperurls.tex`.

brucemiller:
I tried to make a PR to the PR from within the github interface, but it wouldn't take, so I did the brutal thing and just committed to your branch.

dginev:
If the two example tests I gave above work, LGTM

PR #1840, opened 2022-04-01
Rename WhatsIt -> WhatchamaCallIt

tkw1536:
This PR renames the `WhatsIt` object to the much more useful `WhatchamaCallIt`.


brucemiller:
This would be fine, but you've misspelled it. Not to mention inconsistent caMel casing. Except that I just did.

tkw1536:
Maybe next year then 😂

PR #1854, opened 2022-05-09
Fix core

brucemiller:
Several fixes to core functionality
* Conditionals need an equality test as well, eg `\ifx\iftrue\iffalse...` (and enhanced testcase)
* Fix bug when tracing registers that take arguments
* Active characters are macros (expandable) not (directly) digestable
* when unboxing a box, it inherits the current color (and enhanced testcase)
* (TeX's) \footnote goes in text mode.


dginev:
I took a detailed read through the code and it looks highly reasonable. 

The PR adds some very rewarding upgrades to the core TeX interpretation -- they're bound to have compound returns in a variety of ecosystem applications. LGTM.

dginev (approved):

PR #1855, opened 2022-05-10
Fix input

brucemiller:
This PR fixes a peculiar situation where an `\input` before the `\documentclass` loads `LaTeX.pool` but (somehow) broke input. And adds a testcase.

 It also provides options to `InputDefinitions` to allow inputing a TeX/LaTeX file for definitions, but separately controlling whether `@` is made letter or not.


dginev (approved):
Both changes and the new tests read really well to me.

And thank you for fixing the problem I created with wrongly unwrapping the `\input` argument in the parameter type.

PR #1856, opened 2022-05-10
Upgrade SVG RelaxNG Schema

brucemiller:
Upgrade SVG RelaxNG Schema to SVG 1.1 2nd Ed+ from https://github.com/validator/validator.

This schema incorporates the changes from 2nd edition, and some (?) changes from SVG 2.
In particular, it relaxes the (mistaken?) requirement that svg:foreignObject must be contained within an svg:switch that the earlier (unofficial) schema enforced.


dginev (approved):
Strategically this is a great PR, both for staying closer to the official validator, for getting a recent version of SVG, and for enabling lighter foreignObject markup.

Can't evaluate in substance, but I'm sure the graphics examples will do that for us.

Issue #1490, opened 2021-03-14
add XSLT linter


xworld21:
As per subject, I would like to add an XSLT linter, because I am manipulating the stylesheets with XSLT, and diff'ing the changes gets tricky with all the non-semantic differences caused by libxslt.

My idea would be to run the stylesheets through an (almost) identity XSLT transform, then postprocess the output to recover some of the original style:
- convert literal newlines to `&#x0A;` inside `<xsl:text>`;
- convert decimal character references to uppercase hexadecimal;
- reinsert literal newlines inside the attribute values spanning several lines (the XML parser changes newlines into spaces, so one must use a hack here, e.g. restore newline + indentation whenever there is a string of consecutive spaces);
- move the attributes of `<xsl:stylesheet>` and `<xsl:output>` to separate lines.

This can be done while keeping the stylesheets XML-equivalent to the original ones, down to the text nodes, although this means one cannot reindent tags. Things that are lost because of XML parsing:
- how attributes and namespaces are laid out within each tag, except for `<xsl:stylesheet>` and `<xsl:output>` which are done manually from Perl;
- you cannot use two or ore consecutive spaces inside an attribute value except for indentation; consecutive spaces are never needed, though, so that is a reasonable compromise, I hope;
- you cannot use tabs within attributes, those are converted to spaces and cannot be recovered;
- attributes that are intentionally within single quotes for readability, and there aren't any at the moment.

I have made a simple implementation in [this branch](https://github.com/xworld21/LaTeXML/tree/xslt-linter). The linter croaks if the reformatted output is not equivalent to the original file, and it seems to work well.

How to read the branch: the pre-linter commits are minor adjustments that are *supposed* to have no consequence (of course, I submitted them as #1489, and they fail a test with TeXLive 2016, so that's not true! I'll investigate); the linter appears in a single commit; the final commit is how the files look like after linting.

PR #1862, opened 2022-05-21
add XSLT linter

xworld21:
Proof of concept for the XSLT linter #1490, which I have had sitting around for a while. It is fairly simplistic and does not indent the attribute values.


brucemiller:
I very much appreciate the effort you put into this but have reservations. Some regularity does need to be imposed on the XSLT, and some of the changes enhance readability, but some inhibit it.   As you noted, aligning groups of attribute values (like perltidy does with assignments) greatly enhances being able to quickly scan code to find things.

I'm thinking....

PR #1865, opened 2022-06-04
MUST make @ other during latex image processing

brucemiller:
As it says; leaving "@" as letter breaks cases where it's expected to be other.  If there are "@" letter cases leaking into the image sources, those are bugs that need to be fixed.

[I'll probably merge this in shortly, but making a PR to draw attention to it]


dginev:
Bookkeeping that this was introduced as part of a debugging session in #1820 . 

When the error reports related to the `@` in image conversions return in the next arXiv rerun, I will check if I can find better fixes for the internal macros that get passed to latex-proper.

PR #1866, opened 2022-06-05
Logos

brucemiller:
A sudden urge to make the various logos (`\TeX`, `\LaTeX`, `\LaTeXe`, and of course `\LaTeXML`) look Pretty!  And while I was at it, defined a couple of new macros to get the current LaTeXL version (`\LaTeXMLversion,\LaTeXMLrevision,\LaTeXLfullversion`)

Alas, the logo were used in several test cases, so some gratuitous updating there.  But it was fun, so there.


dginev:
Interesting! Don't have time to review in any significant detail, but I can throw a small curve ball your way - I wonder if the new markup looks identical (or at least "equally good") in both Firefox and Chrome.

If yes, probably a nice upgrade.

The test file that repeats `\TeX` tens of times now has some gnarly looking XML, making me wonder if some other macro may now become a better fit...

brucemiller:
Yup, looks pretty much the same ... and pretty close to pdf as well.

PR #1868, opened 2022-06-14
Vertbars

brucemiller:
This is a simpler alternative to stated goal of #1806.

If we don't put too much stress on the semantics of the divides and parallel-to Unicode codepoints, the main point of the Unicode recommendation referenced in #1806 is about grammar or syntax: use these for fences, those for infix operators.  That distinction is exactly what the math parser has figured out, so it seems appropriate that once the parser has determined that a generic `VERTBAR` actually is an OPEN, CLOSE or whatever, it should at that point change the grammatical role *and* the associated Unicode codepoint.

Rewrite rules seem a bit heavy for this task, but I do think that post-parsing rewrite rules will be an important followup for dealing with the actual semantics of various constructs. They'll probably be slightly easier with this normalization in place?


brucemiller:
@dginev would be good to merge, if possible

dginev (approved):
Thanks for explaining the Annotate approach in our call this week! With that in mind, the code here looks solid, and I'm curious to try it out on arXiv.

Issue #844, opened 2017-06-21
A first attempt at documenting latexml.sty and lxRDFa.sty


teepeemm:
I’ve been working on this in my spare time, and the latest email thread has caused me to finally get this out the door.  This issue consists of six files: latexml.dtx, lxRDFa.dtx, ltxdoc.cls.ltxml, doc.sty.ltxml, script.js, and style.css.  They constitute a first attempt at documenting the LaTeXML style files, mostly by copying the existing comments.  (I suppose this would be better as a pull request, but then I’d have to figure out where it fits into the file hierarchy.  And I’m not that good at Git.)

[dist.zip](https://github.com/brucemiller/LaTeXML/files/1091607/dist.zip)

The first two are standard(ish) dtx files (the next section will assume you’ve never seen such a file), which compile to their .sty(.ltxml) and dvi (or pdf or xml) counterparts for documentation.  The .ltxml are my first pass at the corresponding binding, so that the former two can become xml.  They are only complete enough so that these two dtx files can have meaningful output.  The script and style files adjust the resulting html in a few ways.

Several of the files have `% NOTE` where something is incomplete.

**How dtx files work (if you’ve never used them before)**

Running
```
(pdf)latex latexml.dtx
makeindex -s gind.ist latexml.idx
makeindex -s gglo.ist -o latexml.gls latexml.glo
(pdf)latex latexml.dtx
```
will create three files: latexml.sty, latexml.sty.ltxml and latexml.dvi (or pdf) (along with the usual supporting files).  The former two differ from their counterparts only in comments and blank lines (except as noted below).  This can be seen by using the commands
```
diff latexml.sty latexml.orig.sty | egrep -v '^> %' | egrep -v '^>\s*$' | egrep -v '^---$' | egrep -v '^\d\d?a\d\d'
```
and
```
diff latexml.sty.ltxml latexml.orig.ltxml | egrep -v '^> #' | egrep -v '^>\s*$' | egrep -v '^\d\d?\d?a\d'
```
The latter file latexml.dvi (or pdf) is a rough first draft at documenting the package latexml.sty.

How does this happen?  If you look through latexml.dtx, on the first pass it is a regular tex file.  Consequently, many of the lines are commented out (and everything after \end{document} is ignored).  Two parts that aren’t commented out are
```
\generate{
 \file{\jobname.sty}{\from{\jobname.dtx}{package}}
 \file{\jobname.sty.ltxml}{\from{\jobname.dtx}{perl}}
}
[...]

PR #1871, opened 2022-06-20
Generate documentation for packages

teepeemm:
This PR generates the four files blib/lib/LaTeXML/texmf/(latexml|lxRDFa).sty(.ltxml)? from blib/lib/LaTeXML/texmf/(latexml|lxRDFa).pm by using blib/lib/LaTeXML/texmf/sty_ltxml.PL.  I have adjusted Makefile.PL so that this happens during the `make` step of installation (after things get copied over to blib).  The `make` step also creates blib/man3/LaTeXML//texmf//latexml.3 and blib/man3/LaTeXML//texmf//lxRDFa.3.

I have also adjusted genpods so that it creates tex versions of the documentation, and manual.tex will include the documentation in the new appendix chapter B, resolving issue #844.  (If we want to have it in the actual manual, we could put latexml as subsection 4.0.1 and lxRDFa as 8.1.1.)

Because the new pm is a mixture of the old sty and ltxml, I don't see a good way to diff them between the versions.  The best I can come up with is to generate the new sty and ltxml and diff that against the old (and ` | grep -v '> %'` or ` | grep -v '> #'` to ignore the comments in the old versions).  Eventually, you'll see that the only non-comment changes are that I've added
* `\NeedsTeXFormat` (or should we aim for a pure TeX package?),
* `\ProvidesPackage`, and
* `\providecommand{\lxRequireResource}[2][]{}`

Todo:
1. Fix any mistakes in my interpretation of the comments/documentation
2. During `make install`, we should also install the documentation (assuming they have a TeX installation).  From https://tex.stackexchange.com/a/146196/107497, it appears we would want to copy the documentation to $TEXMF/doc/latex/latexml/.  But I'm not sure what we would want to use:
a. If they have Pod::LaTeX, then we can use the tex that genpods creates to create pdf files in doc/manual/texdoc
b. Or we could include those tex (or pdf) files in the git distribution
c. Otherwise, we might be able to use the blib/man3/LaTeXML//texmf//*.3 files (I don't know if texdoc can use these)
d. Otherwise, we should convert the pm files to txt


dginev:
This is definitely ambitious and for Bruce to review, given the level of strategic change to the way we organize our documentation. Interesting!

teepeemm:
I am not a fan of git.

teepeemm:
If I can figure things out, this should be changing six of the original dozenish files from the pull request in order to match up what happened with PRs #1979 and #1906.  If I can't, I might just close this one and start anew.

dginev:
Yikes, I am quite sorry for what happened here @teepeemm . It can get very hard to undo the diverging once it happens, especially if a merge commit sneaked into the mix somewhere. 

And actually yes - I see that you have "merged branch master" already in [2424c54](https://github.com/brucemiller/LaTeXML/pull/1871/commits/2424c54d37affa4d05871b4f8bb5265d56978a3f), which you'd have to backtrack from to recover the PR to a minimal state.

As a rule I always use `git rebase` when the official master branch moves, which on a feature branch can look like:
```
$ git fetch upstream master
$ git rebase upstream/master
$ git push
```

There will be cases where the github remote of your feature branch will be impossible to reconcile with the newly rebased branch, in which case you will have to add some force, once you are certain your local copy is correct, as in:
```
$ git push origin my-feature-branch --force
```

For this PR, it may already be easier to close it fully, and start fresh on a new branch.

teepeemm:
`git rebase upstream/master` kept creating merge conflicts between my code and upstream.  Following https://stackoverflow.com/a/50880042/2336725, I eventually went with:
```
$ git fetch upstream master
$ git reset --soft upstream/master
$ git add --all
```
and I think that's worked.

PR #1874, opened 2022-06-23
Fix ordering bug in \InputIfFileExists

brucemiller:
The "if" clause should appear *before* the file is input!  This fixes the 2020 babel greek regression.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4233,8 +4233,10 @@ DefConstructor('\lx@mark@nocite Semiverbatim',
 #======================================================================
 Let('\@@input', '\input');    # Save TeX's version.
                               # LaTeX's \input is a bit different...
+# Input, now
+DefPrimitive('\ltx@input {}', sub { Input(Expand($_[1])); });
 DefMacroI('\input', undef, '\@ifnextchar\bgroup\@iinput\@@input');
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -5967,11 +5967,12 @@ DefMacro('\IfFileExists{}{}{}', sub {
...
     my ($gullet, $file, $if, $else) = @_;
-    my $file_string = ToString(Expand($file));
+    $file = Expand($file);
+    my $file_string = ToString($file);
     if (FindFile($file_string)) {
       DefMacro('\@filef@und', '"' . $file_string . '" ');
-      Input($file_string);
-      return ($if->unlist); }
+      return Tokens($if, T_CS('\@addtofilelist'), T_BEGIN, $file, T_END,
+        T_CS('\@iinput'), T_BEGIN, $file, T_END); }
```
dginev, line 5975: Very nice find! (and unfortunate bug)

I'm a tiny bit spooked by using `\@iinput` here, since the 2021 latex.ltx has the flipped dependency:

```tex
\def\@iinput#1{%
  \InputIfFileExists{#1}{}%
  { ... 
```

dginev: If this is the most sensible setup in LaTeX.pool for now, that's fine - LGTM!

dginev: Nice! Looks great to me.

PR #1875, opened 2022-06-24
Deluxetable alt

brucemiller:
Rude to usurp #1823, but expedient since by the time I'd sorted out all the false clues, it was basically done.

Once `coerceCS` (used by `DefMacroI` etc) properly recognized the `\csname...\endcsname` idiom, it's less mysterious to just define `\deluxetable` and `\enddeluxetable` (and all the variants), rather than go through the `\begin{deluxetable}` route and then try to patch on top of that.

I didn't do anything about `flushrt`; not sure what that was.

And, I'm assuming you have ar5iv sty.ltxml bindings for   aasms4 aaspp4 aas2pp4 aj_pt4 apjpt4 astro ? Many of the documents referenced in #1823 start off `\documentsyle[aasms4]{article}`  (they work fine if you change article to aastex). I assume the bindings are just like aasms.sty.ltxml.  We can include them in distribution if you like, though they're kinda piddly.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -159,8 +159,13 @@ sub UTF {
 
 sub coerceCS {
   my ($cs) = @_;
-  $cs = T_CS($cs)           unless ref $cs;
-  $cs = T_CS(ToString($cs)) unless ref $cs eq 'LaTeXML::Core::Token';
+  if ((ref $cs) && (ref $cs ne 'LaTeXML::Core::Token')) {
+    $cs = ToString($cs); }
+  if    (ref $cs) { }
+  elsif ($cs =~ s/^\\csname\s+(.*)\\endcsname//) {
+    $cs = T_CS('\\' . $1); }
+  else {
+    $cs = T_CS($cs); }
   return $cs; }
 
 sub parsePrototype {
--- a/lib/LaTeXML/Package/aas_support.sty.ltxml
+++ b/lib/LaTeXML/Package/aas_support.sty.ltxml
@@ -326,8 +326,8 @@ DefMacro('\figcaption OptionalSemiverbatim', sub {
 #======================================================================
 # 2.15 Tables
 RequirePackage('deluxetable');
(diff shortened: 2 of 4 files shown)


brucemiller:
rebased to force a CI re-check and encourage a review :>

dginev:
> I'm assuming you have ar5iv sty.ltxml bindings for aasms4 aaspp4 aas2pp4 aj_pt4 apjpt4 astro ? 

Not really,  none of those have separate bindings available for ar5iv. And indeed - loading OmniBus instead of them and having `DefAutoload` drop the `{deluxetable}` definition is the reason behind a good chunk of the arXiv reports.

So adding the piddly bindings for the rough equivalents of `aaspp.sty.ltxml` seems warranted. Would you like them in latexml or the external arxmliv bindings? Can make a PR in either place.

dginev (approved):
The changes here look more principled than my original attempt, thanks.

I tested on a couple of arXiv documents that had the variants, seemed to go well. So if CI is happy here - LGTM!

brucemiller:
On 7/9/22 10:00, Deyan Ginev wrote:
>     I'm assuming you have ar5iv sty.ltxml bindings for aasms4 aaspp4
>     aas2pp4 aj_pt4 apjpt4 astro ?
> 
> Not really, none of those have separate bindings available for ar5iv. 
> And indeed - loading OmniBus instead of them and having |DefAutoload| 
> drop the |{deluxetable}| definition is the reason behind a good chunk of 
> the arXiv reports.

Maybe we should open a separate issue to figure it out. (I'm actually 
confused about where I found some of those names)  In missing_files 
there are many weird things like "aas2pp4", "myaastex62", 
"aeb_emulateapj_2010".  Easy for a human to guess at, but not so clear 
what regexp to use. ("emulateapj" is such a recognizable base, it might 
suggest more forgiveness in the pre/suffix?; maybe "aastex" too?)

And then I also wonder that in old style cases 
`\documentstyle[aasms4]{article}`, since "aasms4" could be an option, it 
doesn't show up in missing file?

PR #1878, opened 2022-06-29
ignore .DS_Store

teepeemm:
The Mac OS X Finder app (the file explorer GUI) adds a .DS_Store file to every directory the user interacts with.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,5 @@ auto
 tools/travis/dist
 # Various editor workspaces
 .vscode/
+# from Mac OS
+.DS_Store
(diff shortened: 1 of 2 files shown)


tkw1536:
Shouldn't this go into a [global gitignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) instead? I doubt that there is any git repository where one actually wants to commit a `.DS_Store` into.

brucemiller:
You're probably right @tkw1536 , but might as well exclude it anyway.

Thanks for the patch!

PR #1879, opened 2022-07-01
Fix #1872: Generate executable pods with prefix

### LINKED ISSUE #1872: filenames clashing?
Something I came across recently: Windows and Mac OS X have case insensitive file systems.  This means that when genpods converts bin/latexml to doc/manual/pods/latexml.tex and lib/LaTeXML.pm to doc/manual/pods/LaTeXML.tex, the latter is overwriting the former.  The easy fix is for the bins to go to a different file, but this has me wondering if I'm overlooking some way to avoid this happening, or if only Unixy people have been making the manual beforehand and no one really noticed this was happening.

tkw1536:
As noted in #1872, the `./doc/manual/genpods` generates distinct files `doc/manual/pods/latexml.tex` and `doc/manual/pods/LaTeXML.tex`. These file names clash on case-insensitive filesystems, potentially overwriting one another and preventing a proper generation of the manual.

This PR fixes the issue by generating pods for the executables with a new `exe_` prefix. This commit furthermore adjusts their inclusion in `doc/manual/manual.tex`.

### DIFF
--- a/doc/manual/genpods
+++ b/doc/manual/genpods
@@ -112,7 +112,7 @@ if (!-d $GENDIR) {
 
 foreach my $name (@exes) {
   my $src  = "$SRCDIR/bin/$name";
-  my $dest = "$GENDIR/$name.tex";
+  my $dest = "$GENDIR/exe_$name.tex";
   if ($force || (!-f $dest) || (-M $src < -M $dest)) {
     print "Converting POD for $name to LaTeX\n";
     local $::PODDOC = $name;    # $::PODDOC =~ s/::/_/g;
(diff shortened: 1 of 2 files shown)


brucemiller:
Nice solution, but how about putting `_exe` at the end, instead of prefix? Seems slightly more intuitive.

tkw1536:
Sounds reasonable. But it'll take a bit for me to get to it.

tkw1536:
Done.

brucemiller:
Cool! Thanks for the quick response!

PR #1881, opened 2022-07-02
Change marginpar position from fixed to absolute

teepeemm:
In LaTeXML-marginpar.css, a margin note's content should be positioned with `absolute`, so that it can scroll with its anchor point.


brucemiller:
You're right; Thanks!

Issue #1861, opened 2022-05-20
disappearing makebox of zero width


teepeemm:
The tex file:
```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}
\begin{gather*}
\mbox{now you see me}\\
\makebox[1pt]{now you still see me}
\end{gather*}

\begin{gather*}
\mbox{now you see me}\\
\makebox[0pt]{now you don't}
\end{gather*}
\end{document}
```
displays four lines of text in its pdf.  With LaTeXML, I get the xml:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<?latexml searchpaths="/Users/teepeemm/absolute/path/makebox"?>
<?latexml class="article"?>
<?latexml package="amsmath"?>
<?latexml RelaxNGSchema="LaTeXML"?>
<document xmlns="http://dlmf.nist.gov/LaTeXML">
  <resource src="LaTeXML.css" type="text/css"/>
  <resource src="ltx-article.css" type="text/css"/>
  <para xml:id="p1">
    <equationgroup class="ltx_eqn_gather" xml:id="S0.EGx1">
      <equation xml:id="S0.Ex1">
        <text class="ltx_markedasmath">now you see me</text>
      </equation>
      <equation xml:id="S0.Ex2">
        <text class="ltx_markedasmath" width="1.0pt">now you still see me</text>
      </equation>
    </equationgroup>
  </para>
  <para xml:id="p2">
    <equationgroup class="ltx_eqn_gather" xml:id="S0.EGx2">
      <equation xml:id="S0.Ex3">
        <text class="ltx_markedasmath">now you see me</text>
      </equation>
    </equationgroup>
  </para>
</document>
```
and the fourth line of text has completely disappeared.  Adding anything (such as a spacing character) outside of the makebox allows the last line to reappear.

PR #1884, opened 2022-07-08
Post xy cleanup

brucemiller:
As it says, various little bits of cleanup revealed by the arXiv sandbox run.

Ready to merge & re-run the sandbox if you like it.

### DIFF
--- a/lib/LaTeXML/Common/Glue.pm
+++ b/lib/LaTeXML/Common/Glue.pm
@@ -69,12 +69,12 @@ sub new {
       elsif ($mu)    { $f = fixpoint($f);    # in mu
         Warn('unexpected', $unit, undef, "Assumed mu") unless $unit eq 'mu'; }
       else                      { $f    = fixpoint($f, $STATE->convertUnit($unit)); }
-      if (!$punit)              { $plus = $punit = 0; }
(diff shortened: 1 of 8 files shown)


brucemiller:
Oh! Could you take a look at Warning/Perl/warn for redefinitions of ams stuff. It appears that when a versioned name is found, it ends up loading amsmath.sty.ltxml even if it's already been loaded, which cases redefinition warnings.

brucemiller:
Nevermind; I think I've got it fixed in the next commit. Some trickery due to file fallback handling.

dginev (approved):
I was just finishing my coffee preparing to read here when your commit came in - I guess it was lucky I waited :>

From reading through the individual lines, all looks good in my book! LGTM.

(with the usual caveat that some strategic considerations in the alignment code are still beyond my level of understanding, so I'm reading them in a rather shallow manner)

brucemiller:
There's an underlying problem with alignments involving the notion of "empty" cells, and thus empty rows and/or columns. TeX doesn't need the concept; we seemingly do at least by the time we're generating xml or analyzing math structure. Unfortunately, there are 2 problems: you can't prune too early if TeX (or xy) want to access a cell, and secondly I get emptiness wrong often enough (see #1861)

PR #1885, opened 2022-07-09
be consistent in adding class to TOC Contents heading

brucemiller:
Simple fix to a lack of css class consistency


dginev:
This indeed addresses a comment I had in the ar5iv-css:
https://github.com/dginev/ar5iv-css/blob/main/css/ar5iv.css#L696-L697

I will need to double-check how an ar5iv document looks with the change, luckily I have a couple of tests - good to know, good to merge!

dginev (approved):

PR #1886, opened 2022-07-09
Allow optional separator for the prefixes in FindFile_fallback

dginev:
LaTeXML should now be able to recognize that a load of `myemulateapj.cls` can directly use `emulateapj.cls.ltxml`.

The PR allows the 3 currently registered prefixes - `my`, `rx` and `preprint` - to be glued to the file names.


brucemiller:
Looks good; thanks!

PR #1893, opened 2022-07-21
Give \ref a size, not the size of its args

### LINKED ISSUE #1818: \smashoperator bad conversion
I once had the issue, of aligning equal signs when somehting is written on top of them. If the thing on top of it is larger than the equal sign it would result in a weird alignment. For this reason I created an `\lstackrel{"explanation"}{=}` which is defined like this

```tex
\newcommand*{\lstackrel}[2]{ %uses leftsides smashoperator thus ignoring overlap towards the left
    \mathrel{% makes the resulting stack a relation class (like =,<,>,...)
        \smashoperator[l]{\mathop{#2}^{#1}}
    }
}% use when aligning to the left
```
Using `\smashoperator` from the mathtools package.
In LaTeX this looks like this:
Screenshot 2022-03-09 at 11 54 41

When applying LaTeXML it on the other hand comes out like this:
Screenshot 2022-03-09 at 11 55 39

The code is
```LaTeX
\begin{align*}
	\E\big[ \V(X\,|\,\G)\big] + \V\big( \E[X\,|\,\G]\big)
	&= \E\big[ \V(X\,|\,\G)\big] + \E\big[\E[X\,|\,\G]^2\big] -  \E\big[\E[X\,|\,\G]\big]^2 \\
	&= \E \big[ \V(X\,|\,\G)+\E[X\,|\,\G]^2\big] - \E\big[\E[X\,|\,\G]\big]^2 \\
	&\lstackrel{(\ref{conditional var repr})}{=}  \E\big[\E[X^2 \,|\,\G]\big] - \E\big[\E[X\,|\,\G]\big]^2 \\
	&= \E[X^2] - \E[X]^2 \\
	&= \V(X) \qedhere
\end{align*}
```

A possible reason might be, that the url size of the reference is somehow counted towards the space it takes

brucemiller:
fixes #1818

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3853,6 +3853,7 @@ Tag('ltx:*', 'afterClose:late' => sub {
 # These will get filled in during postprocessing.
 # * comes from hyperref
 DefConstructor('\ref OptionalMatch:* Semiverbatim', "<ltx:ref ?#1(class='ltx_nolink')() labelref='#label' _force_font='true'/>",
+  sizer      => '()',                                     # Don't actually know how big this will be!
   properties => sub { (label => CleanLabel($_[2])); });
 # "page" does not make sense in xml.  If the user really wants, they will need:
 # \usepackage{latexml} ... \iflatexml alternate\else page \pageref{label}\fi


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -3853,6 +3853,7 @@ Tag('ltx:*', 'afterClose:late' => sub {
 # These will get filled in during postprocessing.
 # * comes from hyperref
 DefConstructor('\ref OptionalMatch:* Semiverbatim', "<ltx:ref ?#1(class='ltx_nolink')() labelref='#label' _force_font='true'/>",
+  sizer      => '()',                                     # Don't actually know how big this will be!
```
dginev, line 3856: How do the empty parens get interpreted here? I haven't seen them used for sizer before.

brucemiller: Not really interpreted at all; If the sizer is a string, its accumulated size is the size of the Whatsit; if  it contains any  '#digit`, the corresponding arg is "inserted".  So, basically the `\ref` has the same size as "()".

dginev (approved):

PR #1896, opened 2022-07-23
set more reasonable default page sizes

### LINKED ISSUE #1759: use better \textheight estimate
This is a fun+easy upgrade, so it may be motivational to record it in a standalone issue.

I encountered a nearly perfect article: [arXiv:2107.02331](https://ar5iv.org/html/2107.02331) which had a very odd skew in the height of the subfigures. It turns out this was due to using `\textheight` for scaling, e.g.:

```tex
\begin{subfigure}[b]{0.31\textwidth}
        \centering
        \includegraphics[width=\textwidth, height=0.2\textheight]{supplemental-figures/butd-maps/butd-gqa-all.png}
    \end{subfigure}
```

which leads to surprisingly smaller heights than one would have anticipated in the latexml conversion. Should be easy to correct?

brucemiller:
Fixes #1759

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -1094,9 +1094,9 @@ DefMacro('\@dbltoplist',   '');
 DefMacro('\@dbldeferlist', '');
 DefMacro('\@startcolumn',  '');
 # Style parameters from Fig. C.3, p.182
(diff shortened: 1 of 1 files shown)


dginev (approved):
Looks good. Should we document which style is assumed default by LaTeX.pool? E.g. "article.cls letterpaper" vs "article.cls a4paper"...

brucemiller:
Sorta artlcle-ish.

PR #1897, opened 2022-07-23
add native support for tikzbricks.sty

dginev:
Following samcarter's lovely talk about tikzbricks.sty, I discovered latexml can already interpret it raw with good outcomes. For example:

```tex
\documentclass{article}
\usepackage{tikzbricks}
\begin{document}

\begin{wall}
\wallbrick[color=blue]{2}{1}
\wallbrick[color=red]{1}{1}
\addtocounter{brickx}{1}
\wallbrick[color=orange]{1}{1}
\newrow
\addtocounter{brickx}{1}{1}
\wallbrick[color=cyan]{4}{1}
\end{wall}

\end{document}
```

produces:

image

This PR just lets latexml use the raw definition, seeing that it already knows how to.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -717,6 +717,7 @@ lib/LaTeXML/Package/titlesec.sty.ltxml
 lib/LaTeXML/Package/titling.sty.ltxml
 lib/LaTeXML/Package/tikz-3dplot.sty.ltxml
 lib/LaTeXML/Package/tikz.sty.ltxml
+lib/LaTeXML/Package/tikzbricks.sty.ltxml
 lib/LaTeXML/Package/times.sty.ltxml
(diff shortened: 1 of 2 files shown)


dginev:
FYI: `jigsaw.sty` also _almost_ works:
 - in some cases it produces the exact right jigsaw puzzle
 - but encounters spurious errors related to `\pgfmath@temp` and `\pgfmath@next` in a \pgfmathparse call
 - and some other issues with closing `svg:g` in the colored tiles example.

So I won't add a binding for jigsaw.sty, but marking with a comment that it should be relatively little work to add the fixes needed to fully support it.

brucemiller:
doesn't seem to be used on arXiv, sadly. But probably should be.

PR #1898, opened 2022-07-23
Use optional argument to pagebreak

### LINKED ISSUE #1802: Using the optional argument for `\pagebreak`
The `\pagebreak` and `\nopagebreak` macros take an optional argument as to how insistent the macro should be (from 0 to 4, with 4 the default).  If I'm understanding things correctly, `\pagebreak[#]` and `\nopagebreak[#]` complement each other, so that `\pagebreak[x]==\nopagebreak[-x]`.

We ignore the optional argument of `\pagebreak` to realize the xml `<pagination role="newpage"/>` and end up with the empty `<div class="ltx_pagination ltx_role_newpage"></div>` for the html.  With the standard styling, the only effect this has is `height:2em;` (LaTeXML-epub.css also has `page-break-after:always;`).

Since page breaking is definitely more on the pdf end of things, I can agree with mostly ignoring the effect.  But it seems that `\pagebreak[0]` should cause less of an effect than `\pagebreak[4]`.  I could see two ways to go:
* `\pagebreak[x]` should eventually lead to a div with `height:(x/2)em;`, or
* `\pagebreak[x]` should only create the xml when its argument is 3, 4, or missing (and hence the default 4).

This command was discussed back in issues #424 and #1068.

teepeemm:
Examines the optional argument to `\pagebreak[]`, and outputs the XML only when the argument is 3 or 4.  (The alternative would be to end up with vertical spacing equal to `(argument/2)em`, but this would require more complicated XML and also changing the CSS, not to mention still needing to decide whether to break the page in epub.)

Fixes #1802.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4540,11 +4540,16 @@ DefMacro('\sloppypar',    '\par\sloppy');
 DefMacro('\endsloppypar', '\par');
 DefMacroI('\nobreakdashes', undef, T_OTHER('-'));
 
-DefMacro('\showhyphens{}', '#1');     # ?
+DefMacro('\showhyphens{}', '#1');    # ?
     #======================================================================
     # C.12.2 Page Breaking
     #======================================================================
-DefMacro('\pagebreak[]', '\vadjust{\clearpage}');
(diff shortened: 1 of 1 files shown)


brucemiller:
I guess I'm not really getting it, but if it helps, it looks harmless.  Thanks for taking the effort!

PR #1900, opened 2022-07-27
prune single XMTok math to text, if PUNCT or PERIOD

dginev:
This is a discussion starter PR (but maybe already useful?)

I was quickly trying to tackle a timeout from arXiv, using [arXiv:math/0601709](https://arxiv.org/abs/math/0601709), [ar5iv log](https://ar5iv.labs.arxiv.org/log/math/0601709).

It turns out that document has 2162 uses of `$,$` to control comma spacing, which creates - as expected - 2162 math elements with a single comma in them.

I think the markup is cleaner if these kinds of "typesetting abuse" uses of math modes start to gradually get teased away back into text mode. And since latexml is "filled to the brim with heuristics", might as well keep adding more of those.

Not necessarily the best strategy, but I quite like that we now have the `cleanup_Math` subroutine to host such considerations.

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -3843,9 +3843,12 @@ Tag('ltx:Math', afterClose => \&cleanup_Math);
 sub cleanup_Math {
   my ($document, $mathnode) = @_;
   # If the Math ONLY contains XMath/XMText, it apparently isn't math at all!?!
+  # Single token PUNCTs can also be taken out of math.
   if (!$document->findnodes('ltx:XMath/ltx:*'
-        . '[local-name() != "XMText"'
-        . ' and local-name() != "XMHint"]', $mathnode)) {
+        . '[local-name() != "XMText" and local-name() != "XMHint" '
+        . 'and not('
+        . 'local-name() = "XMTok" and (@role="PUNCT" or @role="PERIOD") '
+        . 'and not(preceding-sibling::*) and not(following-sibling::*) )]', $mathnode)) {
     # So unwrap down to the contents of the XMText's.
     my @texts = ();
     foreach my $xmnode (map { $_->childNodes } $mathnode->childNodes) {
@@ -3860,17 +3863,6 @@ sub cleanup_Math {
           else {
             $document->addClass($child, 'ltx_markedasmath');
             push(@texts, $child); } } } }
-    # and replace the whole Math with the pieces
-    $document->replaceTree([undef, undef, @texts], $mathnode); }
-  # Single token PUNCTs can be taken out of math.
-  elsif (my $xmtok = $document->findnode('ltx:XMath[count(./*) = 1]/ltx:XMTok[@role="PUNCT" or @role="PERIOD"]', $mathnode)) {
-    my @texts = ();
-    for my $child ($xmtok->childNodes) {
-      if ($child->nodeType != XML_ELEMENT_NODE) {    # Make sure we've got an element
-        push(@texts, ['ltx:text', { class => 'ltx_markedasmath' }, $child]); }
-      else {
-        $document->addClass($child, 'ltx_markedasmath');
-        push(@texts, $child); } }
     # and replace the whole Math with the pieces
     $document->replaceTree([undef, undef, @texts], $mathnode); }
   else {    # Cleanup any remaining XMTexts
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -3862,6 +3862,17 @@ sub cleanup_Math {
             push(@texts, $child); } } } }
     # and replace the whole Math with the pieces
     $document->replaceTree([undef, undef, @texts], $mathnode); }
+  # Single token PUNCTs can be taken out of math.
+  elsif (my $xmtok = $document->findnode('ltx:XMath[count(./*) = 1]/ltx:XMTok[@role="PUNCT" or @role="PERIOD"]', $mathnode)) {
+    my @texts = ();
+    for my $child ($xmtok->childNodes) {
+      if ($child->nodeType != XML_ELEMENT_NODE) {    # Make sure we've got an element
+        push(@texts, ['ltx:text', { class => 'ltx_markedasmath' }, $child]); }
+      else {
+        $document->addClass($child, 'ltx_markedasmath');
+        push(@texts, $child); } }
+    # and replace the whole Math with the pieces
+    $document->replaceTree([undef, undef, @texts], $mathnode); }
   else {    # Cleanup any remaining XMTexts
     cleanup_XMText_outer($document, $mathnode); }
   return; }


dginev:
I should mention the article in question is still timing out, this wasn't the underlying cause. 
There is a range of articles currently encountering post-processing timeouts at `Scan::addAsChild`, most of which are quite extensive in size. 
(This example has a 124 page PDF, and >8000 lines of TeX code as source).

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -3862,6 +3862,17 @@ sub cleanup_Math {
             push(@texts, $child); } } } }
     # and replace the whole Math with the pieces
     $document->replaceTree([undef, undef, @texts], $mathnode); }
+  # Single token PUNCTs can be taken out of math.
+  elsif (my $xmtok = $document->findnode('ltx:XMath[count(./*) = 1]/ltx:XMTok[@role="PUNCT" or @role="PERIOD"]', $mathnode)) {
```
brucemiller, line 3866: Hmm, I'm wondering if this whole else can be folded into the main if by including it into the xpath.  Something like
```
and not( local-name() = "XMTok" and (@role ="PUNCT" or @role="PERIOD") )
```
within the []. Eliminates a `findnode`, so possibly even faster, depending on the xpath compiler

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -3843,9 +3843,12 @@ Tag('ltx:Math', afterClose => \&cleanup_Math);
...
   my ($document, $mathnode) = @_;
   # If the Math ONLY contains XMath/XMText, it apparently isn't math at all!?!
+  # Single token PUNCTs can also be taken out of math.
   if (!$document->findnodes('ltx:XMath/ltx:*'
-        . '[local-name() != "XMText"'
-        . ' and local-name() != "XMHint"]', $mathnode)) {
+        . '[local-name() != "XMText" and local-name() != "XMHint" '
+        . 'and not('
+        . 'local-name() = "XMTok" and (@role="PUNCT" or @role="PERIOD") '
+        . 'and not(preceding-sibling::*) and not(following-sibling::*) )]', $mathnode)) {
```
brucemiller, line 3851: Does this really only apply to a single child of `XMath` ?  I'm thinking any sequence of only text, spacing and punctuation can be unwrapped; am I missing something?

lib/LaTeXML/Package/TeX.pool.ltxml
dginev: Sounds good, and it seemed a little faster when I benchmarked the article.

I had to also add the check that the PUNCT/PERIOD case is the *only* child in the XMath, since otherwise we match cases that we don't quite want to - in particular in t/math/sampler.tex, the expression
```
\mbox{\tiny $X$},\qquad\mbox{\scriptsize $X$},\qquad\mbox{\normalsize $X$}, %... [snip]
```

Parses as a list, but was wrongly getting its top-level unwrapped to text mode. Actually, I am not sure that is *wrong*, but is not my current intention - something to revisit in the future.

lib/LaTeXML/Package/TeX.pool.ltxml
dginev: That sounds like a stronger claim, that I hadn't thought about. Maybe? We have an example in the `t/math/sampler` test that actually tries that, and currently gets parsed to a list of items. 

Is that constructed list valuable to keep in MathML for search or accessibility? If not, we can unwrap.

dginev: I made an intent-enriched presentation MathML tree for the example, and will juxtapose that to the unwrapped HTML example:

MathML+Intent list:
```xml
<math>
  <mrow intent="list($x1,$x2,$x3,$x4,$x5,$x6,$x7,$x8)">
    <mi mathsize="50%" arg="x1">X</mi>
    <mo rspace="22.5pt">,</mo>
    <mi mathsize="70%" arg="x2">X</mi>
    <mo rspace="22.5pt">,</mo>
    <mi arg="x3">X</mi>
    <mo rspace="22.5pt">,</mo>
    <mi mathsize="120%" arg="x4">X</mi>
    <mo rspace="22.5pt">,</mo>
    <mi mathsize="144%" arg="x5">X</mi>
    <mo rspace="22.5pt">,</mo>
    <mi mathsize="173%" arg="x6">X</mi>
    <mo rspace="22.5pt">,</mo>
    <mi mathsize="207%" arg="x7">X</mi>
    <mo rspace="22.5pt">,</mo>
    <mi mathsize="298%" arg="x8">X</mi>
  </mrow>
</math>
```

unwrapped HTML+MathML
```html
<span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m1" class="ltx_Math" alttext="X" display="inline">
    <mi mathsize="50%">X</mi>
  </math>
</span>
<span class="ltx_text ltx_markedasmath">,</span>      <span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m2" class="ltx_Math" alttext="X" display="inline">
    <mi mathsize="70%">X</mi>
  </math>
</span>
<span class="ltx_text ltx_markedasmath">,</span>      <span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m3" class="ltx_Math" alttext="X" display="inline">
    <mi>X</mi>
  </math>
</span>
<span class="ltx_text ltx_markedasmath">,</span>      <span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m4" class="ltx_Math" alttext="X" display="inline">
    <mi mathsize="120%">X</mi>
  </math>
</span>
<span class="ltx_text ltx_markedasmath">,</span>      <span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m5" class="ltx_Math" alttext="X" display="inline">
    <mi mathsize="144%">X</mi>
  </math>
</span>
<span class="ltx_text ltx_markedasmath">,</span>      <span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m6" class="ltx_Math" alttext="X" display="inline">
    <mi mathsize="173%">X</mi>
  </math>
</span>
<span class="ltx_text ltx_markedasmath">,</span>      <span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m7" class="ltx_Math" alttext="X" display="inline">
    <mi mathsize="207%">X</mi>
  </math>
</span>
<span class="ltx_text ltx_markedasmath">,</span>      <span class="ltx_text ltx_markedasmath">
  <math id="p1.m1.m8" class="ltx_Math" alttext="X" display="inline">
    <mi mathsize="298%">X</mi>
  </math>
</span>
```

Arguments can be made either way... Almost tempted to ask Neil for the accessibility angle.

brucemiller: Hmm, interesting; since the expression is directly in an `ltx:equation`, there are some layout implications --- which we wanted to explore anyway (the padding columns of our table layout of equations squashes it and you get line breaks at the text).   And, of course there's something silly about `<span class='ltx_markedasmath'><math>...`
which we'd also want to collapse.

As for "List"; it's arguably an artefact of our insistence that the parse represents "A Thing".  Depending on what you think "List" means, it's more of a syntactic list than a semantic one; what we'd put in our hypothetical `indeterminate` CD.  Probably better to read off the content, rather than actually add an `@intent` to avoid being misleading.  Might be different if it were wrapped in delimiters (like an Open Interval :> ).

dginev: I agree there is more to explore... Also why I tried to defer it past this PR, since it feels a little open-ended :>

As to unfenced lists being syntactic - maybe they are (as in "simple enumeration of terms"), maybe they aren't?

Some examples of semantic uses of that syntax would be [geometric-progression](https://en.wikipedia.org/wiki/Geometric_progression), or [arithmetic-progression](https://en.wikipedia.org/wiki/Arithmetic_progression), or [harmonic-progression](https://en.wikipedia.org/wiki/Harmonic_progression_(mathematics)), the expansions of which are written as unfenced lists at least in Wikipedia. 

I guess the core observation is that if we auto-unwrap all such lists to HTML, we won't have an anchoring point for attaching the top-level `intent` (or Content MathML symbol, respectively). But if these kind of structures are out of scope for semantifying, we could indeed unwrap them... (and then compact down what HTML gets emitted)

brucemiller: Yeah, good point; best not to lose that potential, even if we can't exploit it yet.  But as we evolve the grammar, we'll want to make clearer distinctions between structure & meaning (as usual :>)

PR #1903, opened 2022-07-29
Fix unparsed

brucemiller:
I knew there was something peculiar there! When parsing the contents of an `XMWrap` fails, we don't need to add an extra wrapper, but can simply mark the container (the `XMWrap`) as unparsed; but *not* mark the children, since they are often quite legitimate and referred to by the content branch of an `XMDual`.


dginev (approved):
That makes sense!

PR #1906, opened 2022-07-30
providecommand for lxRequireResource

teepeemm:
lxRequireResource is defined in latexml.sty.ltxml, but not latexml.sty.  This defines it to do nothing.

Since I was editing the file, I also moved lxKeyWords to come before lxContextTOC, in order to match the order in latexml.sty.ltxml.


dginev (approved):
LGTM, leaving to Bruce to double-check

brucemiller:
Looks good to me to, thanks for fixing that oversight!

PR #1912, opened 2022-08-03
Avoid trailing spaces in vetoed filenames for ZIP inputs

dginev:
Fixes an edge cases reported in [ar5iv#43](https://github.com/dginev/ar5iv/issues/43), which lead to wrongly recognizing a dependent TeX file as the primary article source.

The arXiv "veto" algorithm works well, but there was a need to remove trailing spaces before appending the `.tex` extension.


dginev:
@brucemiller any objections to merging here? It's the 5th, so I'm about to convert the 2207 arXiv papers (just finished downloading).

dginev:
...ought to be safe+simple, so might as well try...

brucemiller:
yup, looks fine.

PR #1913, opened 2022-08-06
add \setuptodonotes stub

dginev:
Caught my eye in [arXiv:2205.10770](https://ar5iv.labs.arxiv.org/html/2205.10770) which is otherwise without any issues.

Looking at the [todonotes docs](http://tug.ctan.org/macros/latex/contrib/todonotes/todonotes.pdf), we are not emulating the full in-depth features yet, so just adding some harmless scaffolding to keep the conversions clean.


brucemiller:
Seems innocuous enough :> Thanks!

PR #1914, opened 2022-08-09
Blocks n frames

### LINKED ISSUE #1704: Missing left/right padding in framed.sty.ltxml
```latex
\documentclass{article}
\usepackage{framed}
\begin{document}
  \begin{framed}
    Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Ut purus elit, vestibulum ut, placerat ac, adipiscing vitae, felis. Curabitur dictum gravida mauris.
  \end{framed}
\end{document}
```
compiles as follows (PDF vs HTML)
PDF
HTML

Offending line (6 occurrences in the file):
https://github.com/brucemiller/LaTeXML/blob/48239014cf05ba2c72ad62b10f4105d592117cdc/lib/LaTeXML/Package/framed.sty.ltxml#L27
sets only `padding-top` and `padding-bottom`, even though `\framebox` applies the padding `\fboxsep` to all sides.

### LINKED ISSUE #1706: framed theorems with ntheorem.sty have no padding and other inconsistencies with LaTeX
Related to #1705
```latex
\documentclass{article}
\usepackage{framed}
\usepackage[framed]{ntheorem}
\newframedtheorem{thm}{Theorem}
\begin{document}
  \begin{thm}
    A theorem.
  \end{thm}
\end{document}
```
gives PDF and HTML
image
image

The frame is missing the `\FrameSep` padding and the vertical margins around the theorem. Note that `ntheorem.sty.ltxml` does not use `framed.sty.ltxml`, I guess because of the behaviour in #1705.

Plus there are other inconsistencies: no number, wrong font, an extra QED.

### LINKED ISSUE #1709: multirow escaping its table
The document
```tex
\documentclass{article}
\usepackage{multirow}
\begin{document}
stuff

%\noindent
\begin{minipage}{.5\linewidth}%
\begin{tabular}{ c c }
item 1 & \multirow{2}{*}{item 2 should be in the table} \\
item 3
\end{tabular}
\end{minipage}
\end{document}
```
renders as I expect, creating a `table` with a `td[rowspan="2"]`.  If I uncomment the `\noindent`, however, the `table` becomes a `div.ltx_noindent` and the `td` becomes a `span.ltx_rowspan`.  That last item is styled `position:absolute,top:0,bottom:0`, so that the rowspanned element moves to the top of the document, regardless of how much stuff comes before.

brucemiller:
Several seemingly related patches regarding frames and blocks, and side issues that cropped up in the process. This places more uniform spacing around frames; more consistent handling of some theorem features.  It makes slight improvements to `insertBlock`, but future work will need to generalize better.  Also slightly improves the validation & styling of rowspan and colspan on tables that are faked with span-soup due to HTML's schema.

Fixes #1704 
Fixes #1706
Fixes #1709

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3054,8 +3054,7 @@ sub defineNewTheorem {
       my $name = $_[1]->getArg(1);
       Digest('\the\thm@bodyfont\the\thm@styling'
           . '\def\lx@thistheorem{' . ($name ? UnTeX($name) : '') . '}'); },
-    beforeDigestEnd => sub { Digest('\thm@doendmark'); },
-    afterDigest     => sub { Digest('\the\thm@postwork'); },
+    beforeDigestEnd => sub { Digest('\thm@doendmark\the\thm@postwork'); },
     properties      => sub {
       my %ctr = ();
       if ($counter) {
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -2200,8 +2200,8 @@ sub insertBlock {
   my ($document, $contents, %blockattr) = @_;
   # Create something like:
   # "<ltx:inline-block vattach='$vattach' height='#height'>#2</ltx:inline-block>"
-  my $model   = $document->getModel;
-  my $context = $document->getElement;    # Where we originally start inserting.
+  my $model    = $document->getModel;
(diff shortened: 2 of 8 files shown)


dginev (approved):
The code looks solid on a first read.

I'll test it on a couple of arXiv articles, but it's good to merge irrespective.

dginev:
Tested 1011.3840 (framed theorems) and 1411.6363 (a framed block containing an algorithm) and they both work out nicely with the PR w.r.t the framed environment appearance. LGTM ✅

PR #1915, opened 2022-08-10
recognize \graphicspath{{dir1/}{dir2/}}

dginev:
A [social report](https://twitter.com/isomorphisms/status/1556444329811578881) spotted a particular lack of images in arxiv-vanity that also affects ar5iv, namely caused by:

```tex
\graphicspath{{figs/}{anc/}}
```

which our convenient `DirectoryList` parameter type wasn't parsing correctly. I added another level of indirection for the unbracing, and this should now work. The article in question is [1801.04486](https://ar5iv.labs.arxiv.org/html/1801.04486), in particular Figure 1.


brucemiller:
Ah, hmm, no commas? OK.

Issue #1919, opened 2022-08-15
support SOURCE_DATE_EPOCH


xworld21:
It would be nice to support [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/), which would allow for (mostly) reproducible outputs.

nschloe:
This is fixed by https://github.com/brucemiller/LaTeXML/pull/1920 and released in 0.8.7. Can be closed. Use as

```
export SOURCE_DATE_EPOCH=1641023584
```
(for Jan 1, 2022).

PR #1920, opened 2022-08-15
use SOURCE_DATE_EPOCH instead of current time

xworld21:
Quick example implementation of SOURCE_DATE_EPOCH (intended to address #1919). Not sure if it is complete. I just grepped "time" and changed modification times in Util::Pack.

The spec asks that LaTeXML "should" throw an error if SOURCE_DATE_EPOCH is set to an invalid value (i.e. not made of just digits) but I don't know where to do it.

I suppose one could refactor `localtime` into a function in some Util module.

### DIFF
--- a/bin/latexmlpost
+++ b/bin/latexmlpost
@@ -518,7 +518,7 @@ sub assemble_postprocessors {
       if ($icon) {
         $$parameters{ICON} = $icon; }
       if (!defined $timestamp) {
-        $timestamp = localtime(); }
+        $timestamp = defined $ENV{SOURCE_DATE_EPOCH} ? gmtime($ENV{SOURCE_DATE_EPOCH}) : localtime(); }
       if ($timestamp) {
         $$parameters{TIMESTAMP} = "'" . $timestamp . "'"; }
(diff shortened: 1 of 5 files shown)


xworld21:
What's missing is generating the EPUB uuid deterministically, but that would mean adding a dependency on Digest or Digest::SHA1. So I'll keep this PR to the minimum and just implement SOURCE_DATE_EPOCH.

brucemiller:
From a TeX perspective, this environment variable seems to come completely out of the blue, although I appreciate the sentiment. It also resonates with #2003, so perhaps there can be a common solution?

xworld21:
Actually pdfTeX and other engines have supported `SOURCE_DATE_EPOCH` for a while, although not always perfectly (the details escape me right now).

Other than that, I agree that a common solution with #2003 would be ideal, e.g. to have a single source of truth for 'current time' which can be overridden via `SOURCE_DATE_EPOCH`, `--timestamp` and/or similar.

brucemiller:
Oh, if it's supported by pdftex, we might as well follow suit for consistency.

Issue #1606, opened 2021-08-09
\tableofcontents within a section only generates TOC for that section if it's split out


xworld21:
If you run `latexmlc --splitat=section` on the following file, the desired table of contents contains only the entries for section B, rather than for the entire document. Not a huge bug, but slightly annoying.
```latex
\documentclass{amsart}
\begin{document}
  \tableofcontents
  \section{A}
  \subsection{AA}
  \subsection{AB}
  \subsection{AC}
  \section{B}
  \subsection{BA}
  \tableofcontents
\end{document}
```

dginev:
manually closing on behalf of PR #1923 ✅

PR #1923, opened 2022-08-17
generate global toc for \tableofcontents

xworld21:
Simple fix for #1606. I think `scope='global'` is safe here – `\tableofcontents` is supposed to generate a TOC for the entire document after all, just like for the list of figures, tables, etc.


brucemiller:
That makes sense; probably was an oversight as `ltx:TOC` evolved. While you're at it, could you also add `scope=>'global'` to `check_TOC` in `LaTeXML.pm`?  I'd guess ePub's nav TOC would want the same thing just to be safe.

xworld21:
> could you also add `scope=>'global'` to `check_TOC` in `LaTeXML.pm`?

Done!

brucemiller:
Great! Thanks!!

PR #1924, opened 2022-08-17
Img alt

### LINKED ISSUE #1452: accessibility: let users specify alt text for images

### LINKED ISSUE #1922: images with caption may have truncated alt text
When referring to images, the HTML stylesheet tries to lift the alternative text from the caption via:
https://github.com/brucemiller/LaTeXML/blob/58cbf1814f555e9741e9c2bea2cb5b35e87a4685/lib/LaTeXML/resources/XSLT/LaTeXML-misc-xhtml.xsl#L175
However selecting `text()` only picks the first text node, which is most often only a small piece of the caption. The stylesheet should rather extract the text content recursively (ideally following something like https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te).

brucemiller:
This PR adds support for the new `\includegraphics` keyword `alt` for the author to provide descriptive text for images.  It also simplifies the alt tag generation on html images, avoiding the redundant copying of caption (when present), saying "Refer to caption" instead. Images with no description or caption get "[Uncaptioned image]".

Fixes #1452
Fixes #1922 

Relates also to #1627, eventually...
depending on if you two (@dginev @xworld21 ) can come to agreement on if there is a proper working aria solution.
I'm kinda reluctant to add ids to `ltx:caption`; it seems unnatural, and will cause lot's of test case churn.  But if it's necessary and actually works, we can do that.

### DIFF
--- a/lib/LaTeXML/Package/graphicx.sty.ltxml
+++ b/lib/LaTeXML/Package/graphicx.sty.ltxml
@@ -47,10 +47,16 @@ DefMacro('\includegraphics OptionalMatch:* []',
   '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}', scope => 'global');
 
 DefConstructor('\@includegraphicx OptionalMatch:* OptionalKeyVals:Gin Semiverbatim',
-  "<ltx:graphics graphic='#path' candidates='#candidates' description='#alt' options='#options'/>",
-  alias      => '\includegraphics',
-  sizer      => \&image_graphicx_sizer,
-  scope      => 'global',
+  "<ltx:graphics graphic='#path' candidates='#candidates' options='#options'/>",
+  alias          => '\includegraphics',
+  sizer          => \&image_graphicx_sizer,
+  scope          => 'global',
+  afterConstruct => sub {
+    my ($document, $whatsit) = @_;
+    my $alt = $whatsit->getProperty('alt');
+    # Trickery to set @description EVEN IF empty string (constructor shorthand omits it)
+    if (defined $alt) {
+      $document->getNode->lastChild->setAttribute(description => ToString($alt)); } },
   properties => sub {
     my ($stomach, $starred, $kv, $graphic) = @_;
     my $options = graphicX_options($starred, $kv);
--- a/lib/LaTeXML/Package/graphicx.sty.ltxml
+++ b/lib/LaTeXML/Package/graphicx.sty.ltxml
@@ -32,6 +32,7 @@ DefKeyVal('Gin', 'keepaspectratio', '', 'true');
 DefKeyVal('Gin', 'clip',            '', 'true');
 DefKeyVal('Gin', 'scale',           '');
 DefKeyVal('Gin', 'angle',           '');
+DefKeyVal('Gin', 'alt',             '');
 # NOTE: graphicx defines @angle to actually carry out the rotation (on \box\z@) w/\Gin@erotate
 # rather than to simply record the angle for later use. (also origin redefines)
 # This is used by adjustbox.
@@ -46,7 +47,7 @@ DefMacro('\includegraphics OptionalMatch:* []',
   '\@ifnextchar[{\@includegraphics#1[#2]}{\@includegraphicx#1[#2]}', scope => 'global');
 
 DefConstructor('\@includegraphicx OptionalMatch:* OptionalKeyVals:Gin Semiverbatim',
-  "<ltx:graphics graphic='#path' candidates='#candidates' options='#options'/>",
+  "<ltx:graphics graphic='#path' candidates='#candidates' description='#alt' options='#options'/>",
   alias      => '\includegraphics',
   sizer      => \&image_graphicx_sizer,
   scope      => 'global',
@@ -54,9 +55,11 @@ DefConstructor('\@includegraphicx OptionalMatch:* OptionalKeyVals:Gin Semiverbat
     my ($stomach, $starred, $kv, $graphic) = @_;
     my $options = graphicX_options($starred, $kv);
     my ($path, @candidates) = image_candidates(ToString($graphic));
+    my $alt = $kv && $kv->getValue('alt');
     (path => $path,
       candidates => join(',', @candidates),
-      options    => $options); },
+      ($alt ? (alt => $alt) : ()),
+      options => $options); },
   mode => 'text');    # Hopefully only to affect reading of arguments
 
 #**********************************************************************
--- a/lib/LaTeXML/Package/graphicx.sty.ltxml
+++ b/lib/LaTeXML/Package/graphicx.sty.ltxml
@@ -58,7 +58,7 @@ DefConstructor('\@includegraphicx OptionalMatch:* OptionalKeyVals:Gin Semiverbat
     my $alt = $kv && $kv->getValue('alt');
     (path => $path,
       candidates => join(',', @candidates),
-      ($alt ? (alt => $alt) : ()),
+      (defined $alt ? (alt => $alt) : ()),
       options => $options); },
   mode => 'text');    # Hopefully only to affect reading of arguments
 
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-misc-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-misc-xhtml.xsl
@@ -171,12 +171,11 @@
           </xsl:attribute>
         </xsl:when>
         <xsl:when test="ancestor::ltx:figure/ltx:caption">
-          <xsl:attribute name='alt'>
-            <xsl:value-of select="ancestor::ltx:figure/ltx:caption/text()"/>
-          </xsl:attribute>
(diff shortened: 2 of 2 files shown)


xworld21 (commented):
That seems like a very good solution! It doesn't help with inline images (tikz and similar), but those are things that can be added separately. Same for the caption issue. My one concern is that there doesn't seem to be a way to set `alt=""` which is a valid use case.

lib/LaTeXML/Package/graphicx.sty.ltxml
```diff
@@ -46,17 +47,19 @@ DefMacro('\includegraphics OptionalMatch:* []',
...
   scope      => 'global',
   properties => sub {
     my ($stomach, $starred, $kv, $graphic) = @_;
     my $options = graphicX_options($starred, $kv);
     my ($path, @candidates) = image_candidates(ToString($graphic));
+    my $alt = $kv && $kv->getValue('alt');
     (path => $path,
       candidates => join(',', @candidates),
-      options    => $options); },
+      ($alt ? (alt => $alt) : ()),
```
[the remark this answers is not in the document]
dginev: Being able to force an empty `alt` seems useful. It appears to be recommended for [decorative images](https://www.w3.org/WAI/tutorials/images/decorative/).

The default for figures in scientific articles is likely better set to "refer to caption" / "uncaptioned image", as Bruce has done in this PR, but an author ought to be separately able to just force the silence via an empty `alt=""`. 

So indeed, mildly more involved code to detect the difference...

brucemiller:
As far as non-`img` objects, it seems the LaTeX3 folks are floating the idea of (eventually) anything that takes keyval arguments could end up with an `alt` keyword which would supply such documentation.

I adopted @xworld21 suggestion for empty alt.

And I'm not sure if I'm reading correctly between the lines; does this offer a sufficient solution to what #1627 was intended for? Or does it still need some aria something?  Or is it better to leave #1627 open? (although it will want to be rebased & rethought after this is merged).

dginev (changes requested):
Ok, so I tried using it... Needs more guarding I think.

```tex
\includegraphics[alt="",width=100pt]{fig2.pdf}
```
leads to `alt="””"`

if I instead use
```tex
\includegraphics[alt={},width=100pt]{fig2.pdf}
```

I get `alt="[Uncaptioned image]"`. Same for `alt=\relax`.

I would have assumed both of these should lead to a simple `alt=""` in the final HTML. Does that sound right?

Edit: using it with actual textual content propagates to the HTML just fine 👍🏻

lib/LaTeXML/Package/graphicx.sty.ltxml
```diff
@@ -32,6 +32,7 @@ DefKeyVal('Gin', 'keepaspectratio', '', 'true');
 DefKeyVal('Gin', 'clip',            '', 'true');
 DefKeyVal('Gin', 'scale',           '');
 DefKeyVal('Gin', 'angle',           '');
+DefKeyVal('Gin', 'alt',             '');
```
dginev, line 35: Adding an empty default via:
```perl
DefKeyVal('Gin', 'alt',             '', '');
```

would also allow the keyword-only empty use, as in:
```tex
\includegraphics[alt,width=100pt]{fig1.pdf}
```

lib/LaTeXML/Package/graphicx.sty.ltxml
```diff
@@ -47,16 +48,24 @@ DefMacro('\includegraphics OptionalMatch:* []',
...
-  scope      => 'global',
+  alias          => '\includegraphics',
+  sizer          => \&image_graphicx_sizer,
+  scope          => 'global',
+  afterConstruct => sub {
+    my ($document, $whatsit) = @_;
+    my $alt = $whatsit->getProperty('alt');
+    # Trickery to set @description EVEN IF empty string (constructor shorthand omits it)
+    if (defined $alt) {
+      $document->getNode->lastChild->setAttribute(description => ToString($alt)); } },
```
dginev, line 59: Cool, this works great for `alt={}` and `alt=\relax`.

I wonder if we can also default to empty for some expected uses of quotes. And just generally sanitize away unfortunate uses of outer-wrapping quotes that would make a screen reader speak unnecessary syntax.

I got some mileage with a couple of extra regexes:
```perl
    # Trickery to set @description EVEN IF empty string (constructor shorthand omits it)
    if (defined $alt) {
      my $description = ToString($alt);
      # sanitize away accidental quotations/fencing wrappers
      $description =~ s/^\W+//;
      $description =~ s/\W+$//;
      # discard values that do not contain traces of language
      $description = '' if $description !~ /\w/;
      $document->getNode->lastChild->setAttribute(description => $description); } },
```

which seems to produce nicely sanitized values for:
```tex
\includegraphics[alt="",width=100pt]{fig1.pdf}

\includegraphics[alt='',width=100pt]{fig1.pdf}

\includegraphics[alt={},width=100pt]{fig1.pdf}

\includegraphics[alt=first order approximation,width=100pt]{fig1.pdf}

\includegraphics[alt={second order approximation of the first kind},width=100pt]{fig1.pdf}

\includegraphics[alt="third order approximation of the second kind.",width=100pt]{fig1.pdf}

\includegraphics[alt=\relax,width=100pt]{fig1.pdf}

\includegraphics[alt,width=100pt]{fig1.pdf} % also needs the empty default in DefKeyVal
```

I realize that some of these may be considered "bad LaTeX" keyvals, but even so I expect people will end up writing them on occasion. So might as well guard some of the obvious cases.

lib/LaTeXML/Package/graphicx.sty.ltxml
brucemiller: value-less keywords are used for booleans, or assertions, like "draft"; an "alt" here without a value would seem to mean that this image is an alternative to something else. No, makes no sense.

lib/LaTeXML/Package/graphicx.sty.ltxml
brucemiller: Not so much "bad" TeX, as just not TeX, and very unexpected by a TeX author.  Besides, your regexps would strip away *all* leading/trailing non-alphanumerics.

xworld21: Further point (in support of no regexps), for forward compatibility: it is conceivable that graphicx will implement `alt` in the future, so the semantics implemented by LaTeXML should be as simple as possible.

dginev: @xworld21 oh, it is already implemented in graphicx, see [ltnews34, page 6](https://www.latex-project.org/news/latex2e-news/ltnews34.pdf) which also links to the github issue.

Maybe I am being overly cautious due to my experiences in dealing with arXiv sources, but I am indeed thinking about forward compatibility - the error we have if an author used a simple boolean `alt` are not informative, and if an author mistakenly used quotation marks (with or without content), they're likely to not notice until a non-sighted reader encounters the bumpy speech in their screen reader.

Whatever Bruce decides is fine by me here, just an extra consideration on my part.

brucemiller: If they don't give a value to alt, they'll get the same message they always get for missing keyval values (whether or not it is informative).  And besides, it just doesn't make any sense.

And if they intentionally used quotation marks (or [], (), math, or any other non alphanumeric), and we silently remove them? What if they used the wrong words and didn't proofread?  If the eventual LaTeX3 definition and usage suggests that the value should be stripped or massaged in some particular way, we can of course follow that usage.  But to proactively choose a very un-TeX-like interpretation? Nah.

PR #1926, opened 2022-08-17
Use unicode minus (for MathML Core in Chrome)

dginev:
This PR offers a simple way to address the current limitation of MathML Core in Chrome not rendering well the ascii minus when used as an operator.

By adding this check in the MathML post-processor (sub `stylizeContent`), we can keep the original character written by the author in the core XML, and can propagate it as-is to non-MathML outputs (e.g. it is still ascii in the TeX and lexeme annotations). 

Doing the replacement in Post instead of Core processing also avoids some internal reuse where a substituted unicode minus may end up in XML attributes that expect lengths, such as `xoffset` and `yoffset` (I've observed this in the test suite when I change the DefMath in TeX.pool).

So for the example call:
```
latexmlc 'literal:a-b=0'  --whatsin=math --whatsout=math --format=html5 \
  --pmml --cmml  --mathtex --preload=[mathlexemes]latexml.sty
```

the resulting fragment is:
```html
<math id="p1.m1" class="ltx_Math" alttext="a-b=0" display="inline">
  <semantics>
    <mrow id="p1.m1.7" xref="p1.m1.7.cmml">
      <mrow id="p1.m1.4" xref="p1.m1.4.cmml">
        <mi id="p1.m1.1" xref="p1.m1.1.cmml">a</mi>
        <mo id="p1.m1.2" xref="p1.m1.2.cmml">−</mo>
        <mi id="p1.m1.3" xref="p1.m1.3.cmml">b</mi>
      </mrow>
      <mo id="p1.m1.5" xref="p1.m1.5.cmml">=</mo>
      <mn id="p1.m1.6" xref="p1.m1.6.cmml">0</mn>
    </mrow>
    <annotation-xml encoding="MathML-Content">
      <apply id="p1.m1.7.cmml" xref="p1.m1.7">
        <eq id="p1.m1.5.cmml" xref="p1.m1.5"></eq>
        <apply id="p1.m1.4.cmml" xref="p1.m1.4">
          <minus id="p1.m1.2.cmml" xref="p1.m1.2"></minus>
          <ci id="p1.m1.1.cmml" xref="p1.m1.1">𝑎</ci>
          <ci id="p1.m1.3.cmml" xref="p1.m1.3">𝑏</ci>
        </apply>
        <cn type="integer" id="p1.m1.6.cmml" xref="p1.m1.6">0</cn>
      </apply>
    </annotation-xml>
    <annotation encoding="application/x-tex">a-b=0</annotation>
    <annotation encoding="application/x-llamapun">italic_a - italic_b = 0</annotation>
  </semantics>
</math>
```

Formulas using the U+2212 character seem to render quite nicely in today's Firefox and Chrome (+ experimental MathML Core), so there should be some direct benefit to taking care of this already on the latexml side.


brucemiller:
Hmm. Kinda inconsistent doing it there. It really should be done already at the digestion phase like all other conversions to Unicode.  Of course, since I was just whining about churn in the test suite...

brucemiller:
Yeah, `stylizeContent` is a sorta right place; possibly `DefMath('-'...` would be more correct, but as noted will cause lots of test case churn *and* we have to be much more careful about modes when digesting (many keywords have "-" in them!).

And as you say, Chrome looks *much* better, so we'll work around Chrome's limitations.

PR #1931, opened 2022-08-18
make csquotes commands inline

### LINKED ISSUE #1930: csquotes displayed as block
Also from LaTeX news:
```latex
\documentclass{article}
\usepackage{csquotes}
\begin{document}
Test: \enquote{should be inline}.
\end{document}
```
comes out as 
image

Most likely a simple CSS issue.

xworld21:
Fix #1930.

My theory is that the current meaning of `ltx_quote` (the class for `<ltx:quote>`) is not the same as it was in 2013. Since inline quotes in the sense of csquotes have little to do with `<ltx:quote>`, I changed the class name to `ltx_csquote`. I hope that doesn't break anything else.

### DIFF
--- a/lib/LaTeXML/Package/csquotes.sty.ltxml
+++ b/lib/LaTeXML/Package/csquotes.sty.ltxml
@@ -129,10 +129,10 @@ RawTeX(<<'EoTeX');
   \fi}
 EoTeX
 
-DefConstructor('\ltxml@oqmark@open', '<ltx:text class="ltx_quote ltx_outerquote" _noautoclose="1">');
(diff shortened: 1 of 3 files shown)


brucemiller:
You're right that over time the distinction between quote (assumed block-quote) and logical quote got conflated. But the distinction isn't unique to csquotes. I think the class name should be `ltx_inline-quote`, rather than `ltx_csquote`, and then it'll look goodl

xworld21:
> I think the class name should be `ltx_inline-quote`, rather than `ltx_csquote`, and then it'll look good

Done!

brucemiller:
I think you still need to run `tools/maketest` on `t/structure/csquotes`, as well.

PR #1932, opened 2022-08-18
do not create new dates in split documents

### LINKED ISSUE #1654: dates on split-off pages appear even if not in the document (before XSLT!)
Follow up on #1602: the postprocessor is adding `<ltx:date>` to the split files, even if no date is specified in the original document. Could this be the original cause of #1602?

Test:
```
latexmlpost --dest=mwe-post.xml --split mwe.xml
```
on any document without the date, e.g.
```latex
\documentclass{article}
\begin{document}
\section{A section}
\end{document}
```

xworld21:
Fix #1654. Assuming the code I removed is really safe to remove. I don't see the use case, so I am guessing it is an historical artefact (maybe `addDate` was originally run on the top document as well?).


brucemiller:
You may be right that it's an artefact; at least it doesn't seem to cause any harm in my current work.

Thanks!

Issue #1934, opened 2022-08-19
some CSS margins/paddings are not responsive


xworld21:
...so some blocks are hard to read on mobile, for instance
![image](https://user-images.githubusercontent.com/1962985/185610216-38b9e0a9-6ea4-444a-a4b2-614efdf44a25.png)
has hardcoded `4em` margins (this is an abstract on amsart).

I'll go through the styles and add `min(N em,N*1.5%)` where it seems uncontroversial. PR soon.

PR #1935, opened 2022-08-19
 Responsive margins

xworld21:
Proposed fix for #1934.

### DIFF
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -251,7 +251,7 @@ dl.ltx_description dl.ltx_description dd { margin-left:3em; }
 
 /* Bibliographies */
 .ltx_bibliography dt { margin-right:0.5em; float:left; }
-.ltx_bibliography dd { margin-left:3em; }
+.ltx_bibliography dd { margin-left: clamp(1em, 4.5%, 3em); }
 /*.ltx_biblist { list-style-type:none; }*/
 .ltx_bibitem { list-style-type:none; }
 .ltx_bibitem .ltx_tag { font-weight:bold; margin-left:-2em; width:3em; }
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -88,7 +88,7 @@ a.ltx_LaTeXML_logo { text-decoration: none; }
 /*======================================================================
   Para level */
 .ltx_float {
-    margin: 1ex 3em 1ex 3em; }
+    margin: 1ex clamp(1em, 4.5%, 3em) 1ex clamp(1em, 4.5%, 3em); }
 
 /* Multiple subfigures via flexbox */
 .ltx_flex_figure {
--- a/lib/LaTeXML/resources/CSS/ltx-amsart.css
+++ b/lib/LaTeXML/resources/CSS/ltx-amsart.css
@@ -44,4 +44,4 @@
 .ltx_itemize,
 .ltx_enumerate,
 .ltx_description
-    { margin-left:2em; }
+    { margin-left: clamp(1em, 3%, 2em); }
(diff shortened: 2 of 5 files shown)


brucemiller (approved):
Works for me.

@dginev does it fit in with your css approach?

dginev:
@brucemiller I think my approach is orthogonal to fine-tuning the margins in the official LaTeXML CSS for now.

Rather than specify margins for block content, most of the work is done via a `max-width: var(--main-width);` rule, where that variable is currently at `52rem`. 

I also have a top-level [margin rule](https://github.com/dginev/ar5iv-css/blob/main/css/ar5iv.css#L89) for the .ltx_page_content:
```css
ltx_page_content {
  margin: 4rem 1rem;
}
```

which combined lead to this current preview on the narrowest mobile option in the Firefox inspector:


That said, I'm keeping an eye on CSS-related PRs and will update ar5iv.css when appropriate (e.g. adding new class names in #1942 will need a couple of new rules for ar5iv too).

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -88,7 +88,8 @@ a.ltx_LaTeXML_logo { text-decoration: none; }
 /*======================================================================
   Para level */
 .ltx_float {
-    margin: 1ex 3em 1ex 3em; }
+    margin: 1ex 3em 1ex 3em;
+    margin: 1ex max(1em,min(3em,4.5%)) 1ex max(1em, min(3em, 4.5%)); }
```
dginev, line 92: I had to blink a few times but I think I'm seeing this right - shouldn't the first of each repeating rule be removed? as in:
```
margin: ...
margin: ...
```

and in a few rules further down:
```
margin-left: ...
margin-left: ...
```

I think you only need to keep the last of each of those, right?

xworld21: ...part of me still tries to support IE...

I have always wondered, does LaTeXML have a target in mind? E.g. anything that has been supported for at least 3 years in major browsers (just making one up)? Or something based on caniuse.com statistics?

brucemiller: Regarding margins versus width: If the goal is to keep from having unreadably wide paragraphs on wide displays, I'm inclined to think that we would end up with a more adaptable & tweakable system if we think of margins for things like abstracts, lists, tables, etc, that appear "within the flow".  But only restrict the overall width at higher level.   Then (perhaps), you can adjust overall page width or abstract "width" by making fewer changes, without having to track down as many surprises.

Obviously, there are many ways of getting the same look for any combination of page, browser, etc, but It can be simpler the fewer knobs there are that do the Right Thing.

xworld21: > Then (perhaps), you can adjust overall page width or abstract "width" by making fewer changes, without having to track down as many surprises.

Should I add a max-width to body, while I am here? Something like @dginev's `52em`?

dginev: @xworld21 I would add max-width in passing, since a variety of other bits will no longer work as intended (in particular I think it will harm the flexbox subfigures with the default CSS).

Best to do that separately, if at all.

dginev (approved):

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -88,7 +88,7 @@ a.ltx_LaTeXML_logo { text-decoration: none; }
 /*======================================================================
   Para level */
 .ltx_float {
-    margin: 1ex 3em 1ex 3em; }
+    margin: clamp(1em, 4.5%, 3em) 1ex clamp(1em, 4.5%, 3em); }
```
brucemiller, line 91: Did the leading `1ex` get lost, or are 3 args intentional?

xworld21: ...or how to mess up a trivial CSS change... fixed!

brucemiller:
Thanks!

PR #1936, opened 2022-08-19
stub \Gin@esetsize

dginev:
just a stub for `\Gin@esetsize`, that currently causes errors [in CorTeX](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/undefined?all=true) for 0.16% of arXiv articles (~3000 articles).

I quickly examined `1511.02120`, and I think the macro is invoked as a side-effect from the adjustbox.sty machinery (which we interpret raw), but likely we can't do much more than nod along for now.


brucemiller:
A quick glance suggests that adjustbox usually is expecting it's own `\adjustbox@esetsize` to be run, but maybe it occasionally leaks through to graphicx version?  Anyway, this looks like a plausible first fix.

Issue #1929, opened 2022-08-18
expansion of \title argument may happen too early


xworld21:
Discovered while trying to compile the LaTeX News https://www.latex-project.org/news/latex2e-news/: the argument of `\title` is expanded immediately, but it contains a macro that is only defined later, and so it fails.

MWE:
```latex
\documentclass{article}
\title{\mytitle}
\newcommand{\mytitle}{Title}
\begin{document}
\maketitle
\end{document}
```
causes
```
Error:undefined:\mytitle The token T_CS[\mytitle] is not defined. at String; line 2 col 0
```

dginev:
We may want to bundle the fix here with a fix for [ar5iv#174](https://github.com/dginev/ar5iv/issues/174) which has a sibling issue in how we handle `\title` currently.

dginev:
We had a follow-up discussion in PR #1937 which somewhat firmly established we will defer a proper fix for this issue till 0.8.8, refocusing the frontmatter implementation around the original LaTeX internals (such as `\@author`, `\@title`, `\@date`).

PR #1937, opened 2022-08-19
Handle frontmatter duplicates and digestion

tkw1536:
Previously when multiple '\title' commands were provided, LaTeXML would use all of them. 
This is different to how pdflatex handles it, where only the last title is handled.
See https://github.com/dginev/ar5iv/issues/174.

Furthermore, as mentioned in https://github.com/brucemiller/LaTeXML/issues/1929, native LaTeX does not expand
arguments to '\title' until it is inserted into the document.
LaTeXML on the other hand digests the argument as soon as '\title' is invoked. 
This causes differences in certain documents.

This PR addresses both issues.
First of all it adds a new '@frontmatter_autoreplace' array that causes specific frontmatter to only be used once.
Then this commit updates the behavior of '\title' to only digest the argument at a later point and store a Tokens() object in $STATE instead.


tkw1536:
As discussed last week, I've further tweaked this PR, splitting out separate 'digestion' and 'insertion' phases.
See https://github.com/brucemiller/LaTeXML/pull/1937/commits/1b3cf4e2da98d9eae191df8d78383b47b61a269f.

brucemiller:
Sorry for the late review.  I like parts of what you're doing, but sneaking in the mystery parameter into the array formatted XML will cause lots of confusion & problems.

I think a simpler, more LaTeX-like, solution would be better. Just have `\title` define '\@title`, as in LaTeX, which gets the deferred, trivially redefined effect (in stock LaTeX, author & date use the same approach).  Then the issue is before frontmatter is inserted, to have a hook (or list or ...?) that checks (eg) if `\@title` is defined and then runs what is currently the defn of `\title` to add it into the frontmatter.

dginev:
Could we get a variant of this before 0.8.7? 

I am working on something else but can take a look here tomorrow and make a branch incorporating Bruce's feedback, since I know Tom is busy with some intense coding hackathons at the moment.

dginev:
@brucemiller 
> the issue is before frontmatter is inserted

Right, getting the data in `\@title`, `\@date` and `\@author` was easy, but I remember you had a concern about LaTeX *only* performing the insertion on `\maketitle`. So if a document does not include `\maketitle`, the frontmatter would disappear.

I think with your latest refactor to frontmatter in #2015 the frontmatter can get inserted via four different triggers: 
- at `\maketitle` (via `\lx@frontmatterhere` macro)
- the `{abstract}` environment (afterConstruct)
- the `{titlepage}` environment (afterConstruct)
- at `ltx:document` (`afterOpen:late`)

each of which relying on the `insertFrontMatter` subroutine from TeX.pool.

The question I have is how to time the expansion+digestion of the frontmatter tokens with this setup in mind. Should I introduce a new subroutine `beforeInsertFrontMatter` and call it in any of the four cases where `insertFrontMatter` gets called? I suspect they need to be separated, since calling `Expand(Digest(T_CS('\@title'))` from e.g. `afterConstruct` isn't ideal.

brucemiller:
Starting 2 comments back:  For abstract and `\maketitle`, you should be able to use
```
afterDigest=> sub { return Digest(T_CS('\lx@frontmatterhere`)); }
```
(or add to existing afterDigest), instead of `afterConstruct`, with the same effect.

This allows you to stick something like your `\lx@assign@frontmatter` in `\lx@frontmatterhere` without lots of extra hooks and mods.  Then only the fallback Tag case needs to worry about digestion, and it is probably reasonable to have it start off by digesting frontmatter if it hasn't already.

This would seem to obviate any big sweeping changes;  And you really shouldn't move the frontmatter stuff to LaTeX.pool, since it in TeX to support non-LaTeX formats.

dginev:
> This would seem to obviate any big sweeping changes; 

Well, as one example, if I start consistently using `\@author` in LaTeX.pool, then in say the binding `inst_support.sty.ltxml` we have:

```perl
DefMacro('\author[]{}', sub {
    map { (T_CS('\lx@author'), T_BEGIN, @$_, T_END) } SplitTokens($_[2], T_CS('\and'), T_OTHER(',')); });
```

which will completely bypass the newly intended streamlined processing flow. So the expansion will happen at the time `\author` is invoked, rather than the time `\maketitle` gets invoked.

I was thinking that the move to `\@title`, `\@author` and `\@date` needs to be propagated to all bindings that deal with these metadata fields.

brucemiller:
> I was thinking that the move to ***@***.***|, ***@***.***| and ***@***.***| needs 
> to be propagated to all bindings that deal with these metadata fields.

Yep, that, for sure. But hopefully with a concise enough expression that 
we don't need major refactoring.

dginev:
Ugh, it breaks a bit too much. Say in `ams_support.sty.ltxml`, where we have the test amsarticle.tex:
```tex
\author{Joe Blow}
\address{Somewhere, USA}
\curraddr{same as above}
\email{joe@blow}
\thanks{My Mommy.}

\author{Frank Zappa}
\address{Baltimore}

\author{Someone Else}
\address{Somewhere Else}
```

Switching to carelessly using `\@author` leads to broken markup, as the underlying amsart.cls definition is doing dual bookkeeping between authors and addresses:
```tex
\renewcommand{\author}[2][]{%
  \ifx\@empty\authors
    \gdef\authors{#2}%
  \else
    \g@addto@macro\authors{\and#2}%
    \g@addto@macro\addresses{\author{}}%
  \fi
  \@ifnotempty{#1}{%
    \ifx\@empty\shortauthors
      \gdef\shortauthors{#1}%
    \else
      \g@addto@macro\shortauthors{\and#1}%
    \fi
  }%
}
```

At this point I am quite worried that we will break a variety of bindings if we rush this without adding enough tests for each case we had support for before...

brucemiller:
> Ugh, it breaks a bit too much.

Yeah, not something we can push through quickly.

tkw1536:
Is this still something that is planned? The PR is more than 2 years old at this point. 
If not, I'd suggest closing it.

dginev:
@tkw1536 The encouraging aspect I can share is the "continued relevance" of your work - I have triaged 27 arXiv user reports that require the correct timing of `\maketitle` and 18 reports that mention duplicated frontmatter. So the upgrade continues to be needed.

That said, this PR is now too stale to merge easily, as the LaTeXML engine has been refactored substantially since 2022. So let's close here. As you are aware, the kernel refactor and rewrite attempt have delayed all timelines in latexml land, so this kind of delay is more the norm than the exception.

PR #1939, opened 2022-08-19
make catcode ALIGN tokens always equal

### LINKED ISSUE #1938: Error when using non-& chars in "align" catcode
The following document is processed significantly differently by lualatex and (with an error) by latexml.

    \documentclass{article}

    \usepackage{amsmath}

    \catcode`\¶=4

    % or alternatively...
    % The following is a little special.
    %
    % Support using pilcrow (0xc2 0xb6) as alignment tab.
    % We do this by marking character "c2 as ignored,
    % and setting "b6 to have catcode 4.
    %\catcode"C2=9 % ignored
    %\catcode"B6=4

    \begin{document}
    Here is a display:
    \begin{align}
      E ¶= mc^2\\
      F ¶= ma
    \end{align}
    \end{document}

(those alignment characters are indeed pilcrows rather than ampersands).

LuaLaTeX is unperturbed by this, and I get an appropriately aligned display.

When I run this through latexml 0.8.6, however:

    % latexml --inputencoding=UTF-8 --dest=try-catcode.lxml try-catcode.tex
    latexml (LaTeXML version 0.8.6) processing try-catcode.tex
    Error:misdefined:¶ The token T_ALIGN[¶] should never reach Stomach! at try-catcode.tex; line 11 col 5
    Error:misdefined:¶ The token T_ALIGN[¶] should never reach Stomach! at try-catcode.tex; line 12 col 5
    Conversion complete 2 errors (See /Data/checkouts/me/lectures/A2-relativity/notes/try-catcode.latexml.log) (reqd. 0.76s)
    % latexml --VERSION
    latexml (LaTeXML version 0.8.6)

I notice that latexml does appear to be identifying the pilcrow as an alignment, but is clearly upset by the character itself.

If we replace the ``\catcode`\¶`` with the commented out `\catcode` lines in the source document, this is a hacky non-UTF8 way of achieving the same effect, which works somewhat better than you might imagine, with pdflatex.  However both versions produce the same error with latexml.  I also tried with `--inputencoding=iso-8859-1`, which produces different error messages in the two cases, but error messages nonetheless.

When I install LaTeXML 0.8.4, however, I get the hoped-for result, with both versions of the \catcode settings.

Is this a reversion from 0.8.4, or did this work by accident in that version?  I haven't tried with 0.8.5, but I _think_ this worked with that version, in the sense that it did produce the expected output, with alignment and no errors.

(Why am I doing such an odd thing, you ask?  The LaTeX I'm processing in fact originates from an XML file, and '&' is separately magic in that context, so it's convenient, when embedding LaTeX in that file, to have something else be the alignment character (and essentially everything below 0x7f is otherwise spoken for).)

dginev:
Fixes #1938 .

Even with all tests passing this kind of PR is quite scary.  
The reported use is working correctly again, but the PR change makes it impossible to use the main `Object::Equals` and `Token::equals` methods to determine if we are looking _exactly at_ an ampersand of the ALIGN catcode.

Maybe there is a more precise location to correct, specific to the amsmath align environment? How would you make that call @brucemiller ? The methods in question are latexml-specific, so it is in big part a design choice.

### DIFF
--- a/lib/LaTeXML/Core/Token.pm
+++ b/lib/LaTeXML/Core/Token.pm
@@ -222,7 +222,7 @@ our @CATCODE_SHORT_NAME =          #[CONSTANT]
   T_SUB T_IGNORE T_SPACE T_LETTER
   T_OTHER T_ACTIVE T_COMMENT T_INVALID
   T_CS T_MARKER T_ARG T_SMUGGLE_THE
-);
+  );
 
 our $SMUGGLE_THE_COMMANDS = {
   '\the'        => 1,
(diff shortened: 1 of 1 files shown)


dginev:
Closing in favor of the better targeted #1940

PR #1942, opened 2022-08-20
Indentation

### LINKED ISSUE #1928: \noindent not working?
I have a list of items I'm trying to typeset very plainly. I tried an enumeration (see #1927). I also tried a very plain list of paragraphs:

```
\noindent

item 1

item 2
```

When rendered in the PDF, these look normal, without an indent.

In the book.html from latexml, the first item is pulled to the left, as if the `\noindent` is ignored.

I fixed this by adding CSS to those particular items to change `text-indent` back to `2em`.

So, I have a workaround I'm okay with. Just thought you might like the report.

brucemiller:
This PR fixes several inconsistencies with LaTeX with respect to the indentation (or lack thereof) of the 1st paragraph within sections.  It is a little clunky and indirect; but I was trying to avoid ending up with all paragraphs explicitly marked as whether indented or not.  That would possibly be more accurate, but would probably require more code, testcase churn and be harder to customize the CSS?

Fixes #1928

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -663,7 +663,8 @@ DefConstructor('\@@numbered@section{} Undigested OptionalUndigested Undigested',
         Warn("malformed", "ltx:$type", "Tried to open an unknown tag ltx:$type for numbered section"); } }
     $document->openElement($tagname,
       'xml:id' => CleanID($id),
-      inlist   => ToString($inlist));
+      inlist   => ToString($inlist),
+      (LookupValue('INDENT_FIRST') ? (class => 'ltx_nofirstchild') : ()));
     if (my $tags = $props{tags}) {
       $document->absorb($tags); }
     $document->insertElement('ltx:title',    $props{title});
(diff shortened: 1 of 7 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -663,7 +663,8 @@ DefConstructor('\@@numbered@section{} Undigested OptionalUndigested Undigested',
         Warn("malformed", "ltx:$type", "Tried to open an unknown tag ltx:$type for numbered section"); } }
     $document->openElement($tagname,
       'xml:id' => CleanID($id),
-      inlist   => ToString($inlist));
+      inlist   => ToString($inlist),
+      (LookupValue('INDENT_FIRST') ? (class => 'ltx_nofirstchild') : ()));
```
dginev, line 667: I am rather confused by the "nofirstchild" name. Is there a better one?

dginev: For example, `ltx_indent_all_p` would be clearer (though I'm not sure that's the right intention - basing that on the new CSS rule).

brucemiller: Probably is a better name, but the way it ends up working is to override the firstchild behavior in css. Which of course has the effect (currently) of not indenting.

brucemiller: Actually worse: the use in `pruneEmpty` (which I wrote first) really means there is no first child, since it got removed; so what looks like the first child is actually the second. The use here does mean indent the first child.  I suppose it would be clearer intention to introduce another class name, even though there will be extra css rules.

dginev: From the perspective of someone reading classes to style with CSS, names that talk about `ltx_indent_first_p` and `ltx_noindent_first_p` are both concise and clear, so I think that is a very healthy renaming direction.

Encountering `ltx_nofirstchild` is quite confusing (to me) without a lot of extra context. If the emphasis is on the first paragraph getting pruned, maybe `ltx_pruned_first_p` is a first name idea?

dginev: Sure, it is "needlessly specific" if you can resolve the ambiguity that the first means the `p` and not the local title (e.g. section heading).

Good enough to merge in either case.

dginev (approved):

PR #1951, opened 2022-08-31
allow stylesheets to read and write files in the site directory

xworld21:
Small functionality & security improvement, while working on #1948: stylesheets can read (via `document()`) and write (via `<exsl:document>`) files. This PR ensures that read/write operations happen in the site directory.

For security, operations are also restricted to the site directory, and network access is disabled.


lib/LaTeXML/Post/XSLT.pm
```diff
@@ -76,7 +76,33 @@ sub process {
...
+  pathname_chdir($sitedir);
+
+  # ensure that the stylesheet can only read and write in the site directory
+  # and cannot access the network
+  my $cb = sub {
+    my $dest = pathname_canonical(pathname_absolute($_[1], $sitedir));
+    if (!pathname_is_contained($dest, $sitedir)) {
+      Error('unexpected', 'stylesheet', $doc, "'$dest' is outside of the site directory, denying access");
+      return 0; }
+    else { return 1; } };
```
xworld21, line 93: Thinking out loud: the security callback can save the list of files written by the stylesheet. Which suggests saving the list of resources within `$doc`, whether created by the stylesheet or via `copyResource`. Say with a new type `Post::Resource` containing basic info (source, parent document, destination, mime type).

This would allow Manifest modules to find all generated/copied content without running `find` and without hardcoding file names such as `content.opf` and `LaTeXML.cache`, as well as receive mime types from `\lxRequireResource`.

lib/LaTeXML/Post/XSLT.pm
```diff
@@ -76,7 +76,33 @@ sub process {
...
+  # move to the site directory when applying the stylesheet,
+  # since it can create additional files with <exsl:document>
+  my $sitedir  = $doc->getSiteDirectory;
+  my $orig_cwd = pathname_cwd();
+  pathname_chdir($sitedir);
+
+  # ensure that the stylesheet can only read and write in the site directory
+  # and cannot access the network
+  my $cb = sub {
+    my $dest = pathname_canonical(pathname_absolute($_[1], $sitedir));
```
xworld21, line 89: The following fits better with the behaviour of `copyResource`:
```suggestion
    my $dest = pathname_canonical(pathname_absolute($_[1], $destdir));
```
where `$destdir = $doc->getDestinationDirectory`.

lib/LaTeXML/Post/XSLT.pm
```diff
@@ -76,7 +76,33 @@ sub process {
...
+    if (!pathname_is_contained($dest, $sitedir)) {
+      Error('unexpected', 'stylesheet', $doc, "'$dest' is outside of the site directory, denying access");
```
xworld21, line 91:
```suggestion
      Fatal('unexpected', 'stylesheet', $doc, "'$dest' is outside of the site directory, access denied");
```
since libxslt will crash anyway after this.

lib/LaTeXML/Post/XSLT.pm
```diff
@@ -76,7 +76,33 @@ sub process {
   if (my $icon = $params{ICON}) {
     # Hmm.... what type? could be various image types
     $params{ICON} = '"' . $self->copyResource($doc, $icon, undef) . '"'; }
+
+  # move to the site directory when applying the stylesheet,
+  # since it can create additional files with <exsl:document>
+  my $sitedir  = $doc->getSiteDirectory;
+  my $orig_cwd = pathname_cwd();
+  pathname_chdir($sitedir);
```
xworld21, line 84: A sudden realisation: in principle, this is a breaking change. If somebody out there uses `document()` to read some data, this change of folder will break the workflow almost irreparably.

@brucemiller what do you think of using `exsl:document()` in the stylesheets (for manifests or other files the user might want to generate)? If that feels like a good strategy for the EPUB manifest, I can work on another solution that doesn't break the previous behaviour (most likely something like a function `f:dest(relative-path)` that resolves to the site directory – I'd have to think about it).

xworld21: Correction: `document(uri)` resolves `uri` relative to the stylesheet location, whereas `document(uri, node)` resolves `uri` relative to the path of the document containing `node`.

Hence, this PR is *not* a breaking change. It is appropriate to switch directory so that `exsl:document()` writes to the destination directory rather than other random places, and it won't change the behaviour of `document()`.

xworld21:
I have removed the libxslt security callback stuff, and kept this PR at a minimum: the point is, XSLT.pm is supposed to run in the site directory, because stylesheets can create additional files via `exsl:document()` (that could be very convenient to generate additional manifests, search indexes, custom EPUB tocs, etc). Simply changing directory is not controversial, I hope.

The security callbacks will reappear in the `--recorder` PR.

xworld21:
> should LaTeXML's XSLT framework use `exsl:document()` ?

Since this PR is under attention again, I should clarify the intention here, and split @dginev's question into two parts:
- Should LaTeXML use `exsl:document` for auxiliary files (e.g. EPUB manifest)? Tricky proposition, because the generated files go directly to disk, bypassing the postprocessing pipeline.
- Should LaTeXML support users that want to call `exsl:document`? This could be useful in e.g. BookML to generate search indices (not the best example, but you get my point I hope).

lib/LaTeXML/Post/XSLT.pm
```diff
@@ -76,7 +76,16 @@ sub process {
...
     $params{ICON} = '"' . $self->copyResource($doc, $icon, undef) . '"'; }
+
+  # change directory so that files created by <exsl:document> are relative to the destination
+  my $destdir  = $doc->getDestinationDirectory;
+  my $orig_cwd = pathname_cwd();
+  pathname_chdir($destdir);
+
   my $newdoc = $doc->new($$self{stylesheet}->transform($doc->getDocument, %params));
+
+  pathname_chdir($orig_cwd);
```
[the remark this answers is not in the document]
xworld21: I suppose we should also canonicalise the paths (especially if mixed path separators are involved) and emit an error/fatal if chdir fail.

PR #1954, opened 2022-10-03
Minor boilerplate / latex.ltx fixes

dginev:
I was quickly surveying mdframed.sty articles, and got side-tracked into adding some minor boilerplate that removes spurious PDF-related errors from arXiv:1306.2973

Mostly no-ops, list is:
 - add `\@thefnmark` and lock the footnote macros, as they will almost certainly produce broken markup upon redefinition.
 - add a handful of latex.ltx internal macros that may get exercised during raw interpretation, but have no real effect in latexml processing (PDF-related)
 - add `\@removeelement` with its real definition from latex.ltx
 - fix a typo of the macro `\pdfstringdef` which was written `\pdfstring` in hyperref
 - add stubs for the PDF bookmark macros in hyperref
 - sanitized `\MessageBreak` into a newline when it made its way into our custom generic messaging subroutine
 - even more minor: my editor ended up trimming some spaces in latexml.sty, which ended up here

Besides tweaking these minor bits, I could also state that `scrbook.cls`, `mdframed.sty` and `datetime.sty` will indeed need real bindings written in Perl - but of course we already knew that. The arXiv article in question exercises all three.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -530,10 +530,12 @@ DefConstructor('\lx@notetext[]{}[]{}',
     (role => $type . 'text', makeNoteTags($type, $_[1], $_[3] || Digest(T_CS('\the' . $type)))); },
   reversion => '');
 
-DefMacro('\footnote',      '\lx@note{footnote}');
-DefMacro('\footnotemark',  '\lx@notemark{footnote}');
-DefMacro('\footnotetext',  '\lx@notetext{footnote}');
-DefMacro('\@footnotetext', '\lx@notetext{footnote}');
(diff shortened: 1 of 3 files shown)


dginev:
Judging by the CI failure here, I think MikTeX has once again updated to a more recent texlive. 

We should likely file a test regression issue:
```
#   Failed test 't/babel/csquotes'
#   at D:\a\LaTeXML\LaTeXML\blib\lib/LaTeXML/Util/Test.pm line 113.
# Difference at line 58 for t/babel/csquotes
#       got : '<p>.0pt<text'
#  expected : '<p><text'
# Looks like you failed 1 test of 7.
t/81_babel.t .............. 
```

brucemiller:
Seems reasonable. Thanks!

PR #1955, opened 2022-10-05
[WiP] BibTeX  emulation

brucemiller:
This PR is an alternative refactoring of @tkw1536 's rather brilliant proposal in #1231 for emulating BibTeX within LaTeXML. It is much reorganized and (hopefully) simplified, and more integrated into LaTeXML's system of utilities, error handling, organization and coding style.

I've undoubtedly broken a few features that had been working, so more testing and development is needed (and probably add a few test cases) for completion, error recovery, corner cases.  There are also a couple of areas I'd still like to refactor (Runtime). And preservation of semantics for tei/jats.


tkw1536:
It's great to see some progress on this PR after several years, and this looks great in principle. 
Need to look at it in more detail though.

dginev:
Exciting PR!

We mentioned in the call today that this PR may be a very good occasion for adding a test exercising `chapterbib.sty.ltxml`, also as a just-in-case regression guard going forward.

PR #1957, opened 2022-10-08
fix binding typo - missing argument in elsart_support_core

dginev:
As it says - someone in a hurry must have forgotten the argument signature for a couple of frontmatter macros.

Spotted while debugging arXiv:math0603447


dginev:
I should probably also exclaim that we ought to have a hard binding compiler failure if an argument index gets used which goes beyond the length of the signature arguments.

PR #1946, opened 2022-08-23
Provision for data-* attributes in HTML5

dginev:
This PR contains a first approach that worked for me in providing `data-*` custom data attributes for HTML5. 
It is very much open to discussion and redesign.

* Uses a new `data` namespace prefix for the core LaTeXML XML output
   * (which needs a proper namespace URI, I "imagined" one)
* maps that into the flat `data-*` scheme in (X)HTML5 output formats.
* adds a new `\lxCustomData` macro in latexml.sty to demo the feature.

A (maybe whimsical) example use would be:
```bash
latexmlc --preload=latexml.sty --whatsin=fragment --format=xml \
  'literal:\section{Foo} hello \lxCustomData{mouse}{cookie} \
    \subsection{Bar} \lxCustomData{data-yes}{set} world.'
```

which produces the core processing XML (snipped):
```xml
  <section inlist="toc" xml:id="S1">
    <tags>
      <tag>1</tag>
      <tag role="refnum">1</tag>
      <tag role="typerefnum">§1</tag>
    </tags>
    <title><tag close=" ">1</tag>Foo</title>
    <para xml:id="S1.p1">
      <p xmlns:data="http://dlmf.nist.gov/LaTeXML/custom-data" data:mouse="cookie">hello</p>
    </para>
    <subsection xmlns:data="http://dlmf.nist.gov/LaTeXML/custom-data" data:yes="set"
                inlist="toc" xml:id="S1.SS1">
      <tags>
        <tag>1.1</tag>
        <tag role="refnum">1.1</tag>
        <tag role="typerefnum">§1.1</tag>
      </tags>
      <title><tag close=" ">1.1</tag>Bar</title>
      <para xml:id="S1.SS1.p1">
        <p>world.</p>
      </para>
    </subsection>
  </section>
```

with `--whatsout=fragment --format=html5` we get a desired HTML5 snippet:
```html
<article class="ltx_document">
  <section id="S1" class="ltx_section">
    <h2 class="ltx_title ltx_title_section">
      <span class="ltx_tag ltx_tag_section">1 </span>Foo</h2>

  <div id="S1.p1" class="ltx_para">
    <p class="ltx_p" data-mouse="cookie">hello</p>
  </div>
  <section id="S1.SS1" class="ltx_subsection" data-yes="set">
    <h3 class="ltx_title ltx_title_subsection">
      <span class="ltx_tag ltx_tag_subsection">1.1 </span>Bar</h3>

    <div id="S1.SS1.p1" class="ltx_para">
[...]

PR #1960, opened 2022-10-10
Foreign attributes

brucemiller:
This PR aims to gradually support the addition and propagation of any foreign namespaced attributes into the LaTeXML XML and ultimately into the resulting target format (html, etc). Likely options for conversion to `data-xxx` form for html5.

The following need to be verified
- [x] Foreign attributes explicitly allowed in schema
- [x] propagated through parsing
- [x] preserved in general postprocessing
- [x] preserved in Math conversion
- [x] optionally converted to `data-xxx`

Here's a simple tester that will show what attributes got added & lost from processing any text file. Use `./testall somearticle.tex`.
[foreignattr-testing.zip](https://github.com/brucemiller/LaTeXML/files/9746839/foreignattr-testing.zip)


brucemiller:
A more comprehensive tester 
[testall.zip](https://github.com/brucemiller/LaTeXML/files/9749864/testall.zip)

brucemiller:
I'm thinking this pretty much works as intended, and is done, other than:  Probably should add a namespace declaration for `data` to TeX.pool (since that prefix is slightly special), and possibly define in `latexml.sty` some convenience macro similar to what was proposed in #1946.

Jazzpirate:
...this sounds like it's exactly what I would need for sTeX! Thanks for letting me know, I will hopefully find some time very soon to try it out :)

brucemiller:
OK, I've added some bindings to `latexml.sty`, namely
```
\lxAddAnnotation{keyvalpairs}
\lxWithAnnotation{keyvalpairs}{something}
```
where the keyvals gives attribute names and values. 
If the attribute name has has a colon it is assumed to start with the namespace prefix; no colon, it goes into the `data` namespace. And if it starts with a colon, no namespace so you can add LaTeXL attributes, but should use care.  I have mixed feelings about defaulting to `data`, and likewise about *not* defaulting.

Anyway, should be a reasonable proposal, so removing the "WiP"

Jazzpirate:
> OK, I've added some bindings to `latexml.sty`

If I understand this correctly, that would mean we wouldn't even need any sepcial bindings for sTeX - those two bindings should subsume everything we need already <3

Just so you stay up-to-date: I'm currently very busy with completely cleaning up the mess that is the sTeX-code and improving lots of things in the process. Once that's done, I will check to what extent I can get sTeX itself and our sTeX libraries to work with LaTeXM. If the answer is "a decent amount", I will install LaTeXML on our active documents server and use that in the future, so that RusTeX can be degraded to a) a fallback whenever LaTeXML fails and b) mere "quick and dirty conversion" in the IDE. Either way, expect me to open new issues if anything should fail :)

Thanks a lot for the effort :)

brucemiller:
@Jazzpirate Indeed, it might be that you can accomplish a lot directly from TeX.  The kind of thing I was aiming for was something like
```
\begin{equation}\lxAddAnnotation{stex:purpose={Annotation demonstration}}
A = \lxWithAnnotation{stex:meaning={A meaningful function}}{f}(x)
\end{equation}
```
should add the obvious attributes to both the containing `ltx:equation`, and the `ltx:XMTok` for "f". Those (hopefully) would propagate into the XHTML (as attributes in stex namespace), or in html5 as `data-stex-purpose` and `data-stex-meaning`.

Hopefully, that can be adapted into something useful?  That said, I will be neither surprised, nor offended, if that leads you to new mysteries, questions, suggestions and requests! :>

Jazzpirate:
> ```
> \begin{equation}\lxAddAnnotation{stex:purpose={Annotation demonstration}}
> A = \lxWithAnnotation{stex:meaning={A meaningful function}}{f}(x)
> \end{equation}
> ```

Yes, that's what I thought :) Those two would perfectly subsume the primary macro functionalities I need, namely `\stex_annotate:nnn{key}{value}{content}` (which would be something like `\lxWithAnnotation{stex:key=value}{content}`) and `\begin{stex_annotate_env}{key}{value}...` (which would just be `\lxAddAnnotation{stex:key}{value}`) :)

I assume that for "invisible" annotations, introducing a `display=none`-style should be easily doable already, when in doubt via a new binding.

What I can easily imagine might cause *minor* issuess are 1. the mechanism I use for getting at environment variables, which works ultimately via something like `\@@input"|kpsewhich -var-value FOO"`, but that smells like the kind of situation where a perl binding is not just necessary but also very much desirable (and easy to implement even for me), and 2. I use very recent additions to LaTeX3 (I think the latest macro I use is `\prop_to_keyval:N`, which was added in April 2022), but I would guess those you would want to support anyway over time :)

...and of course whatever weird packages might show up in Michael's lecture notes and/or the smglom. But that will have to be determined :)

Again, thanks a lot, this looks very promising <3

brucemiller:
> I assume that for "invisible" annotations, introducing a 
> |display=none|-style should be easily doable already, when in doubt via 
> a new binding.

To clarify: this PR is only adding attributes, so no XML/HTML *markup*,
and thus invisible by default. They'll be visible only if javascript, or 
custom XSLT, uses them.


> I use very recent additions to LaTeX3

LaTeXML tries to work with whatever LaTeX3 code you have installed...
(and tries hard :>), but it's tricky and slow.

Jazzpirate:
> To clarify: this PR is only adding attributes, so no XML/HTML *markup*, and thus invisible by default. They'll be visible only if javascript, or custom XSLT, uses them.

Ah, sorry, I expressed myself badly. What I meant with "invisible annotation" was that I occasionally need to generate HTML that should not be rendered (i.e. normal nodes that I think of as "annotations" merely in the sense that they are invisible and only relevant for MMT, such as the types/definientia/notations of symbol declarations). But that can easily be done either way by introducing a CSS class and annotating the node with that, so I foresee no issues here.

dginev:
In the spirit of learning from sister projects, I was quickly skimming a [tralics article](https://hal.inria.fr/inria-00000198v2/document) from 2006, in particular pages 33 and 173, centered around `\XMLaddatt`.

The bit that struck me as particularly interesting is that they afford an optional argument where you can do long-distance attribute management by providing the explicit `xml:id` for the element on which the attribute is to be set. Maybe interesting, not necessarily for this PR, but possibly for developers such as @Jazzpirate who may end up needing more markup control, in some edge cases.

dginev (approved):
Pretty cool PR!

I tested around with the new macros, and had some attributes (in both a foreign namespace as well as an explicit/implicit `data:` namespace) preserved on 
- `ltx:section` → `<div>`
- `ltx:text` → `<span>`
- `<ltx:XMDual>` → `<mrow>`

I think it should be good to merge for public prototyping/testing in the upcoming release.

PR #1962, opened 2022-10-10
phase out *mathspace mspace values

### LINKED ISSUE #1801: Modernize mspace width values
There is a scary deprecation warning in Firefox:
```
“veryverythinmathspace”, “verythinmathspace”, “thinmathspace”, “mediummathspace”, “thickmathspace”,
  “verythickmathspace” and “veryverythickmathspace” are 
  deprecated values for MathML lengths and will be removed at a future date.
```

Changing these in latexml is technically easy -- the question is what should we change them to?

We might as well stay a step ahead of the deprecation status changing.

dginev:
Fixes #1801 .

Questions to @brucemiller : 
 - Should we use the `em` units MathML Core recommends, instead of `pt`, so that spaces scale along with font-size?
 - Is the PR correct so far?

I can claim the current commit phases out the old values, as a grep for "mathspace" doesn't find any other active code with them. But so far this is the kind of change that someone completely new to latexml would made, since I consider myself mostly novice when it comes to the post-processing unit considerations.

Thanks for any tips!

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -821,10 +821,10 @@ sub pmml_mo {
   my $islargeop = $role && ($role =~ /^(SUMOP|INTOP)$/);
   my $lpad      = $attr{lpadding}
     || ((ref $item) && $item->getAttribute('lpadding'))
-    || ($role && ($role eq 'MODIFIEROP') && 'mediummathspace');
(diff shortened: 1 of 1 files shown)


brucemiller:
looks plausible, and those few mathspace's were only used for internal spacing tweaks, so pretty innocuous allround.

dginev:
Alright, good to merge on my end if so. You can always revisit them in the other spacing work.

PR #1964, opened 2022-10-12
Strengthen auto-loads to global scope

dginev:
I was doing some diagnostics with arXiv:2009.06875 and arXiv:2112.11446 and there were a couple of easy wins by allowing an auto-load for bmatrix and tabularx.

The reason those were needed was that in the original papers there were guards such as (copied verbatim from the source):
```tex

%% These few lines make a distinction between latex and pdflatex calls and they
%% bring in essential packages for graphics and font handling.
%% Note that due to the \DeclareGraphicsExtensions{} call it is no longer necessary
%% to provide the the path and extension of a graphics file:
%% \includegraphics{diamondrule} is completely sufficient.
%%
\ifpdf%                                % if we use pdflatex
  \pdfoutput=1\relax                   % create PDFs from pdfLaTeX
  \pdfcompresslevel=9                  % PDF Compression
  \pdfoptionpdfminorversion=7          % create PDF 1.7
  \ExecuteOptions{pdftex}
  \usepackage{graphicx}                % allow us to embed graphics files
  \usepackage{gensymb}
  \usepackage{amsmath}
  \usepackage{tabularx}
  \usepackage{dblfloatfix}
  \usepackage{float}
  \DeclareGraphicsExtensions{.pdf,.png,.jpg,.jpeg} % for pdflatex we expect .pdf, .png, or .jpg files
\else%                                 % else we use pure latex
  \ExecuteOptions{dvips}
  \usepackage{graphicx}                % allow us to embed graphics files
  \DeclareGraphicsExtensions{.eps}     % for pure latex we expect eps files
\fi%
```

Since we intentially treat `\ifpdf` as false in latexml, such articles never load some of their packages - notably amsmath and tabularx in this particular example.

But then separately, if I make more provisions for auto-loading "on demand", it is often the case that the load event happens deep inside the article, often with several levels of open grouping. Which leads to the auto-loaded definitions very quickly "expiring" when the current group is closed, and again causing errors the next time the triggering macro is encountered.

My current experiments seem to lead to better outcomes if I explicitly set `\globaldefs` to true when auto-loading a _package_ in particular. That isn't quite correct, since I would only want to make `DefMacro` and similar binding definitions global, not *all* assignments to State's value table. So maybe I should introduce a separate switch that is checked in the Package.pm definitions of DefMacro, DefPrimitive, DefConstructor, DefEnvironment, DefConditional, DefExpandable, DefRegister, DefMath, DefColor, DefRewrite, DefMathRewrite, ... list goes on actually. Hm.

Proposing this for feedback, maybe there is a clearer way to solve the problem?


brucemiller:
Hmm... I wonder whether we should treat `\ifpdf` as true?  The original intention (I'd assume?) was to distinguish dvi from pdf, and we're certainly more pdf than dvi (or "as pdf").

dginev:
Right, though the symmetrical worry is that we may run into blocks that execute a lot of `\pdfxxx...` primitives if we openly claim we are emitting PDF.

A bit of an impossible choice, since we are "the option no one prepared for", really.

brucemiller:
but we have lots already; maybe most/all the built-in ones?

dginev:
Ok, I'll take the challenge.

Would you be open to me adding *all* of them (mostly stubbed in, naturally) and switching `\ifpdf` to true in an alternative PR?

Currently, I am seeing 112 primitives defined:
```
LaTeXML/lib/LaTeXML/Package$ rg -F "'\pdf"|wc -l
112
```

While the [Overleaf list of primitives](https://www.overleaf.com/learn/latex/TeX_primitives_listed_by_TeX_engine) lists 147.

So we have 35 to go and we're there. Doable?

brucemiller:
Interesting. There are 77 error undefined `\pdfxxxx` in CoreTeX :>  `\pdfbookmark` at 0.07%!  (or is that in the bookmark package, which may have accidentally not gotten loaded?)  All of which suggests many folks don't bother with `\ifpdf`; which suggests we should  pretend to make pdf.  

Are the Overleaf list actually primitives defined by pdftex itself?  If so I would assume some belong in `pdfTeX.pool.ltxml`, or where? Surprising there are so many missing; not documented, or we just haven't tracked it?

Probably a good thing way to go. I'm sure many are ignorable, but many will take arguments and be *interesting*.  Hopefully there's documentation for what they should do (if anything).

dginev:
> Interesting. There are 77 error undefined \pdfxxxx in CoreTeX :> \pdfbookmark at 0.07%! (or is that in the bookmark package, which may have accidentally not gotten loaded?) 

Curious, will have to debug a couple to be sure...

> All of which suggests many folks don't bother with \ifpdf; which suggests we should pretend to make pdf.

Alright, let's try.

> Are the Overleaf list actually primitives defined by pdftex itself? 

Almost all of them yes, pdftex 1.40.19 to be precise. There are just a couple that are exclusive to luatex and xetex. 

> If so I would assume some belong in pdfTeX.pool.ltxml, or where? Surprising there are so many missing; not documented, or we just haven't tracked it?

We don't have a process that goes and revisits the coverage of pdfTeX every time texlive gets bumped, and those folks seem to keep adding primitives :> I had a couple of PRs adding macros from other bindings we nominally claim to support, but haven't explicitly tracked all the way to texlive 2022. Version hell rares its ugly head...

> Probably a good thing way to go. I'm sure many are ignorable, but many will take arguments and be interesting. Hopefully there's documentation for what they should do (if anything).

Ok, I'll take it up and let you know how it goes. It will fix the original arXiv article I tested with, so I'll close here for simplicity's sake.

dginev:
Update on:
> >     \pdfbookmark at 0.07%! (or is that in the bookmark package, which may have accidentally not gotten loaded?)
> Curious, will have to debug a couple to be sure...

I tested 3 of the reported articles, and that error is no longer produced. I think our CorTeX reports are now sufficiently outdated that we need a new full rerun soon :> All 3 of the files I tested used an unsupported class, called `sigma.cls`, and still experienced other errors from macros it defines, though none start with `\pdf`.

So we can disregard the CorTeX reprots in this one instance, for now.

PR #1965, opened 2022-10-13
Add a couple of revtex internals

dginev:
I spotted that `\replace@command` is used in arXiv:1112.1218, and its lack was currently causing a (circumstantial) infinite loop in that document conversion.

Looks harmless enough to provide in the revtex support binding, hence this PR.
The document in question converts successfully, with only two minor errors remaining (missing support for `\sanitize@url` and `\@email`)


brucemiller:
looks plausible.

PR #1969, opened 2022-10-26
Address texlive 10.2022 regressions

### LINKED ISSUE #1963: Test suite fails w/ recent TeX Live snapshot
I recently uploaded a new snapshot of TeX Live into the Debian distribution. I had to notice that the test suite (partially) fails since then. The full log is visible [here](https://ci.debian.net/data/autopkgtest/testing/armhf/l/latexml/26934184/log.gz). I applied the patch from #1931 to the package, and even tried with the latest master checkout but this did not solve any issue.

Please be so kind to have a look at the issue. Many thanks!

dginev:
Fixes #1963 .

After the usual wild goose chases, I identified two causes for the test suite failing with the latest texlive 10.2022.
 1. We were missing support for `\the\font`, which I added via 
    - a special case in the `Register` parameter type
    - a special case in `\the`.
    - For now the PR returns a stub font CS, namely `\tenrm`
 2. The more sinister error was in the depths of expl3, namely loading of data `.txt` files, and in particular `SpecialCasing.txt` and `GraphemeBreakProperty.txt`
    - the fix here boiled down to respecting `\endlinechar` in the `\readline` reader in eTeX.
       - though that seems to slow down expl3 loading even further - maybe there is a faster way?

The font side is obvious, so I'll just add a bit of detail on the expl3 side:
After a multi-hour debugging session, I could reduce the exact cause to the empty lines in the 
data files, such as:
```
FB06; FB06; 0053 0074; 0053 0054; # LATIN SMALL LIGATURE ST

0587; 0587; 0535 0582; 0535 0552; # ARMENIAN SMALL LIGATURE ECH YIWN
```

Which in turn lead to finding the bit of code doing the reading, which relied on *both* `\endlinechar` and `\readline` and realizing we don't implement that correctly yet:
```tex
\cs_new_protected:Npn \__ior_str_get:NN #1#2
  {
    \exp_args:Nno \use:n
      {
        \int_set:Nn \tex_endlinechar:D { -1 }
        \tex_readline:D #1 to #2
        \int_set:Nn \tex_endlinechar:D
      }   { \int_use:N \tex_endlinechar:D }
  }
```

Curious to see if this PR passes CI on all texlives, but if it does - I think we can survive through yet another texlive update.

### DIFF
--- a/lib/LaTeXML/Package/csquotes.sty.ltxml
+++ b/lib/LaTeXML/Package/csquotes.sty.ltxml
@@ -27,11 +27,11 @@ InputDefinitions('csquotes', type => 'sty', noltxml => 1);
 ###########################################################################
 # Compatibility fixes
 
+# the unicode check does not work properly in LaTeXML
+# furthermore, package version <= 2018/04/06 v1.3b contain
+# a known bug, but might still be around.
+# therfore we skip the check here.
 RawTeX(<<'EoTeX');
-% the unicode check does not work properly in LaTeXML
-% furthermore, package version <= 2018/04/06 v1.3b contain
-% a known bug, but might still be around. 
-% therfore we skip the check here. 
 \def\csq@ifutfchar#1{%
   \ifundef\@inpenc@undefined
     {\@secondoftwo}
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Package/csquotes.sty.ltxml
```diff
@@ -27,11 +27,11 @@ InputDefinitions('csquotes', type => 'sty', noltxml => 1);
 ###########################################################################
 # Compatibility fixes
```
dginev, line 29: Taking the comments out from TeX into Perl is just a minor cosmetic change, since the performance implications were irking me while inspecting the binding.

dginev:
It seems MikTeX is determined to steal my CI thunder in fixing the TeX side of the problem. We now have a completely separate MikTeX-in-CI installation problem, as can be seen here (from a related PR):
https://github.com/brucemiller/LaTeXML/actions/runs/3275944409/jobs/5391413398#step:6:1328

Still, I can claim my local machine "Linux / TexLive 10.2022" has the tests passing, and seeing how all other TeXlives pass in CI, the PR should be good for review by @brucemiller .

brucemiller:
It passes all the texlive's I've got handy, so let's go with it; Thanks!!

PR #1970, opened 2022-10-26
Update MikTeX version for CI

dginev:
Trying to see if updating the MikTeX version to the one [currently listed in chocolatey](https://community.chocolatey.org/packages/miktex) fixes the installation in CI.


dginev:
I give up. The chocolatey miktex package is broken again, even with the newest miktex version.

Tempted to just yank that test case from the CI... Since otherwise we simply have to wait until it magically works again.

PR #1971, opened 2022-10-26
Yank miktex test from CI

dginev:
As it says.

Let's not waste the CPU cycles in Github Actions for a >20min test that we know will fail.

(for reasons completely independent of latexml)

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -5,11 +5,11 @@ name: Windows
 # significantly more difficult to implement.
 
 # This workflow caches data for both Perl and MiKTeX installations.
-# However GitHub does not provide a builtin way of evicting that cache. 
-# See https://github.com/actions/cache/issues/2. 
+# However GitHub does not provide a builtin way of evicting that cache.
+# See https://github.com/actions/cache/issues/2.
 #
 # Hence we assign the cache a 'hash' which is simply inserted into the internal cache key.
-# To manually evict the cache, simply change the hash to something else. 
+# To manually evict the cache, simply change the hash to something else.
 # Something like the output of `date | md5` usually works well.
 env:
   PERL_CACHE_HASH: 79d1fcba0f256b853d8daa7f3d973852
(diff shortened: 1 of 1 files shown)


dginev:
image

✅ What a sight for sore eyes. @brucemiller I will suggest skipping out on that test until the beginning of 2023, and the revisiting if the windows miktex package has become more robust.

brucemiller:
Having a preview of upcoming texlive problems is a good thing, but the way it's been working with MikTeX is too random, especially as a "required" test.  Don't we (or didn't we) have the option of allowed-to-fail tests?  What might be useful would be to run an allowed-to-fail test on the latest checkout of texlive, on a linux platform (which is otherwise quicker).

dginev:
@brucemiller indeed, the Windows+Miktex test is already allowed to fail, Tom took care of that. You can see the `allow_failure: true` line in the code I commented out from the YAML file in this PR.

And I wouldn't make the PR just for the texlive failure, I also find those early regression notifications useful. 

It's just that the chocolatey miltex package is currently failing to even install in CI, which guarantees a hard failure of the test before it even gets to anything latexml. But it only fails after a github CPU is actively spinning for 20 minutes.

I like your idea about trying to add another entry in the Linux matrix, where we try the "bleeding edge" texlive sources, and mark it as an "allowed failure".

brucemiller:
Maybe I'm confused as to what allowed failure means; (in other branches) it still seems to show up as a red X failure?

dginev:
Ah, you're right, it seems that is a gitlab idiom, not a github one. And it is a [missing feature](https://github.com/actions/toolkit/issues/399) in github actions.

There are all kinds of special-casing workarounds suggested in that open issue, but I kind of don't think they are worth the trouble, if the chocolatey package will have certain months where it is impossible to install. Much more reliable to replace that test with a "bleeding edge" Linux/texlive test.

dginev:
Feel free to add any comments on the topic @tkw1536 . Tagging you to attract attention.

brucemiller:
This calls for some semantic hairsplitting :>  "Replacing" the test loses the original intention of the test (to verify with MikTeX), but that has become too brittle to try to guarantee. But it maintains what has *become* the (more important) benefit of the test (verifying with upcoming texlive changes).  But then that latter test would be more useful as an advanced warning, rather than required.

So, I'm thinking the best solution would be to disable the test for now. And then add it, along with a new bleeding-edge texlive test, as allowed-failures, when that feature becomes available (which, from that thread, I expect will happen).

PR #1973, opened 2022-10-26
Default the html language to "en"

### LINKED ISSUE #1626: TeX default language is US English, LaTeXML has no default
Directly inspired by https://github.com/brucemiller/LaTeXML/issues/1440#issuecomment-897874048 (cc @nxg).

The TeX default language is deliberately US English. I think it makes sense for LaTeXML to set the language to US English by default, so that the output starts with the appropriate `<html lang="en-US">` even if the user does not import babel.

PS: I have tried sticking `MergeFont(language => "en-US")` in `TeX.pool.ltxml` and it seems to work correctly in like 90% of the situations. In some of the tests (especially the plain TeX ones), the language does not bubble up to `<ltx:document>`, and weird stuff happens.

dginev:
Fixes #1626 .

- I defaulted to "en", since it is allowed, and we have no real reason to assume US over UK/AU.
- Only in XSLT, so that the default only impacts (X)HTML formats, where we know there is benefit.

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
@@ -87,9 +87,14 @@
     <xsl:element name="html" namespace="{$html_ns}">
       <xsl:apply-templates select="." mode="begin"/>
       <xsl:call-template name="add_RDFa_prefix"/>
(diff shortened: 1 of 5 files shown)


brucemiller:
Perhaps in the future, we'll want to find ways (beyond `babel`) to determine or assert the language during generation. But if there's no language given by the end, assuming English seems reasonable to a native English speaker.

PR #1974, opened 2022-10-26
only run t/83_expl3.t in CI mode

dginev:
My patch in #1969 results in a more correct latexml interpretation of expl3, but it *also* results in a slower interpretation.

1. The `t/83_expl3.t` tests now take 7 minutes on my local PC, which has a decent CPU. 
    - In comparison, the second slowest test is `t/80_complex.t` which takes 3.5 minutes .

2. Separately, the expl3 test was a reason for raising a flag in the Debian "AutoPKG test suite", where the latexml package builds failed due to expl3 no longer loading error-free in interpretation.

3. I think it is a great goal to stay compatible with LaTeX 3, to keep that test, and to **always** run it in CI.

4. But at the same time I think regressions in our processing of LaTeX 3 should not be a reason for the latexml package builds to fail. Similarly to tikz, which may regress and we currently won't notice. Requiring every latexml user that builds from source to wait the extra time on every `make test` is considered too high a burden for tikz, and I think should now also be too high a burden for expl3.

I would be very happy to return the test to "always on" if we can get it back down to 3 minutes in runtime. Just my opinion, but I think the suggested change offers us a good trade-off.

So this PR switches the test to CI-only.

### DIFF
--- a/t/170_grammar_coverage.t
+++ b/t/170_grammar_coverage.t
@@ -13,7 +13,7 @@ use Data::Dumper;
 local $Data::Dumper::Sortkeys = 1;
 
 if (!$ENV{"CI"}) {
-  plan skip_all => "Only checked in continuous integration. (enable via: CI=true make test)";
+  plan skip_all => "Only checked in continuous integration. (set environment var CI=true and rerun tests)";
   done_testing();
   exit;
 }
--- a/t/170_grammar_coverage.t
+++ b/t/170_grammar_coverage.t
@@ -13,7 +13,7 @@ use Data::Dumper;
 local $Data::Dumper::Sortkeys = 1;
 
 if (!$ENV{"CI"}) {
-  plan skip_all => "Only checked in continuous integration. (set environment var CI=true and rerun tests)";
+  plan skip_all => "Only checked in continuous integration. (use make test CI=true)";
   done_testing();
   exit;
 }
--- a/t/170_grammar_coverage.t
+++ b/t/170_grammar_coverage.t
@@ -13,7 +13,7 @@ use Data::Dumper;
 local $Data::Dumper::Sortkeys = 1;
 
 if (!$ENV{"CI"}) {
-  plan skip_all => "Only checked in continuous integration.";
+  plan skip_all => "Only checked in continuous integration. (enable via: CI=true make test)";
   done_testing();
   exit;
 }
--- a/t/83_expl3.t
+++ b/t/83_expl3.t
@@ -7,7 +7,7 @@ use warnings;
 use LaTeXML::Util::Test;
 
 if (!$ENV{"CI"}) {
-  plan skip_all => "Only checked in continuous integration. (enable via: CI=true make test)";
+  plan skip_all => "Only checked in continuous integration. (set environment var CI=true and rerun tests)";
   done_testing();
   exit;
 }
--- a/t/83_expl3.t
+++ b/t/83_expl3.t
@@ -4,6 +4,12 @@
 #**********************************************************************
 use LaTeXML::Util::Test;
 
+if (!$ENV{"CI"}) {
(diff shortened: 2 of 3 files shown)


brucemiller:
The `expl3` test takes *twice* as long!? Wow.  Can you pinpoint exactly which change causes that? (eg using an older texlive, so you don't get errors).  Maybe there's a better way.

dginev:
Just a note that I went and checked the CI times from the last few PRs, only to realize that github actions have a *wild* variance in testing times (even upto 5 minutes), based on availability and load of their servers. So that was a very wrong avenue.

I only have one texlive on my machine, so going back to triage older versions will be easier for you than for me. Though the only real solution is making the current version faster - might as well benchmark the code in the current HEAD commit.

brucemiller:
I have lots of texlives, *except* the latest prerelease that causes the failure.  Maybe I should set up to track a bleeding-edge version?

At any rate, to the extent that I'm testing the right thing, there doesn't seem to be any particular slowdown with the patch from #1969; for t/83: 2020 52sec; 2021 66sec; 2022 61sec  (curious variation across texlive's though).

dginev:
According to nytprof, there were 188.6 million subroutine calls in converting `expl3/tilde_tricks.tex`. Impressive. Here is the exclusive time ranking:

image

Gullet seems to be a very heavy-weight actor. So here is its breakdown in turn:

image

So scanning for numbers is the most promising optimization spot? Curious.

dginev:
> At any rate, to the extent that I'm testing the right thing, there doesn't seem to be any particular slowdown with the patch from #1969; for t/83: 2020 52sec; 2021 66sec; 2022 61sec (curious variation across texlive's though).

Thank you for checking that. I would still suggest running the test only in CI, to keep local installations fast. We'll still catch regressions early, as we keep filing PRs on a regular basis.

brucemiller:
Well, ok, but how about adding a reminder like "(use CI=true make test ... )" to the "Only checked in continuous integration" line in testing?  It may be a bit much to *always* test expl3, but too often seemingly innocuous changes can break it.

dginev:
> Well, ok, but how about adding a reminder like "(use CI=true make test ... )" to the "Only checked in continuous integration" line in testing? It may be a bit much to _always_ test expl3, but too often seemingly innocuous changes can break it.

Good thinking, done.

brucemiller:
Not to be annoying, but it should clarify that it's an environment variable rather than a `make test` argument. (Eg. I always forget the underscore and plural in `TEST_FILES` :> )

dginev:
What text would you prefer instead? I wrote it the way it is clear to me, and it ought to be brief for that a 1-line message.

Maybe:
`"Only checked in continuous integration. (set environment var CI=true and rerun tests)";

brucemiller:
Sure, that works for me.  My only issue is that `make test` can use both 
environment variables *and* arguments, and both of them could look like 
`CI=true`.

dginev:
Got it, rephrased. And indeed, `make test` would also do what one would expect if the CI=true value was passed in as an argument. These two act identically:
```
$ make test CI=true TEST_FILES=t/83_expl3.t
$ CI=true make test TEST_FILES=t/83_expl3.t
```

But that doesn't seem to be true for `TEST_FILES` which **must** be passed in as an argument, and can't be an env var.

brucemiller:
Pfffttt! They *both* work? I shoulda just tried it instead of whining.  But if I'm going to (have to) be reminded (and as long as I'm torturing you with requests), I think I'd prefer the more concise & intuitive (imho) "(use 'make test CI=true...')"

brucemiller:
or just "(use make test CI=true)"  (the ellipsis were meant literal)

dginev:
Alright, reworded again. Let's see if this one sticks :>

brucemiller:
Well, actually, now that I look at it...

PR #1976, opened 2022-10-28
Tracking verb

### LINKED ISSUE #1848: Test t/tokenize/verb.tex inconsistent between texlives
I was seeing discrepancies in an unnamed porting effort, and when I tried to regenerate the PDF, I stumbled on:
```
! LaTeX Error: \verb ended by end of line.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.53 {\verb tricky} }
```

which was a hard error. This pointed out that latexml is using the space character as a `\verb` delimiter, which is not what my `latex` executable does in texlive 2021. Maybe this has changed, maybe it was always broken?

For the same reason, a different part of the test has the XML:
```
<para xml:id="S2.p9">
      <p><verbatim font="typewriter">|a}b#c^d_e$f|</verbatim>stuff</p>
    </para>
```

but in the PDF output, the two vertical bars are not displayed - as | is used as the fencing char.

brucemiller:
This PR updates the behaviour of `\verb` to match texlive after Apr, 2020.  Since `\verb` changes the catcode of space (for visible space), and since the space skipping after control sequences happens upon the next read, originally `\verb` would treat a following space as the delimiter. As this is unexpected (by most folks), this behaviour was changed to explicitly skip spaces. That change caused breakage in one of our test cases.  We are now implementing the current behaviour and have updated the test case to accomodate.

Fixes #1848


dginev (approved):
LGTM

PR #1980, opened 2022-10-31
jheppub.sty : Handle \notoc and fix \arxivnumber

j824h:
`\notoc` is missing. It should be handled as an empty macro, since TOC is not created at all, for now.

`\artxivnumber{}` seems to be a typo of `\arxivnumber{}`

Reference:
https://jhep.sissa.it/jhep/help/JHEP/TeXclass/DOCS/JHEP-author-manual.pdf


dginev:
@j824h I think we already have this covered - I added it in #1943 . Could you take a look that it meets expectations?

dginev:
I think this is also the reason for GitHub reporting your PR having a conflict - you need to rebase to the latest HEAD commit of LaTeXML, at which point you'll also obtain the changes I made in August.

PR #1981, opened 2022-11-03
declare Windows specific dependencies

xworld21:
This PR fixes something ugly I did pre-0.8.6 to make LaTeXML work in Windows consoles, which is to try importing `Win32::Console` rather than properly checking the platform, declaring the `Win32::Console` dependency in the Makefile, etc.

To declare the PR done, I want to make it behave properly under `$^O eq 'cygwin'`, but my Windows machine is taking its time. In the meanwhile, I sense an impending release, so please let me know if there is something obviously wrong here!

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -74,6 +74,11 @@ WriteMakefile(NAME => 'LaTeXML',
     'Time::HiRes'       => 0,
     'URI'               => 0,
     'version'           => 0,
+    # Windows terminal handling (see Common::Error)
+    ($^O eq 'MSWin32'
+      ? ('Win32::Console' => 0,
+        'Win32::Console::ANSI' => 0)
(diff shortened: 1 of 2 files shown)


dginev:
I've always dreamed that latexml gains a (part-time) Windows maintainer one day, and very much celebrate these kinds of PRs @xworld21 ! Thank you for looking into the fine-print here.

xworld21:
Oh well, LaTeXML on cygwin is broken for completely different reasons. I might try to understand why, but for now I am going to assume the intersection of cygwin users and LaTeXML users is currently empty. At any rate: this PR is good enough for normal Windows installations.

dginev (approved):
LGTM

brucemiller:
Cool! Thanks!!!

PR #1987, opened 2022-11-16
add {algorithm*} env; minor definedness guard

dginev:
Fixes an easy Fatal error in arXiv:2112.00827, which was a cascade of errors coming from the `{algorithm*}` environment not being defined in the algorithm2e binding.

Easy, as it is likely harmless to reuse the main `{algorithm}` environment in latexml: the star variant is targeting 2-column layout in PDF.

I also spotted a minor definedness warning in post-processing in this article, adding that in as well.

### DIFF
--- a/lib/LaTeXML/Package/algorithm2e.sty.ltxml
+++ b/lib/LaTeXML/Package/algorithm2e.sty.ltxml
@@ -59,36 +59,39 @@ DefMacro('\fnum@algorithm',      '\algorithmcfname\nobreakspace\thealgorithm');
 DefMacro('\fnum@font@algorithm', '\bf');
 DefMacro('\ext@algorithm',       'loa');
 
-DefEnvironment('{algorithm}[]',
-  "<ltx:float xml:id='#id'>"
-    . "#tags"
-    . "<ltx:listing class='ltx_lst_numbers_left'>"
-    . "<ltx:listingline>"
-    . "#body"
-    . "</ltx:listingline>"
-    . "</ltx:listing></ltx:float>",
-  properties   => { layout => 'vertical' },
-  beforeDigest => sub {
-    DigestIf(T_CS('\@ResetCounterIfNeeded'));
-    DigestIf(T_CS('\algocf@linesnumbered'));
-    Let('\par',    '\lx@algo@par');
-    Let('\parbox', '\lx@algo@parbox');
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Package/algorithm2e.sty.ltxml
```diff
@@ -59,36 +59,39 @@ DefMacro('\fnum@algorithm',      '\algorithmcfname\nobreakspace\thealgorithm');
...
+      DefMacro('\;', '\ifmmode\@mathsemicolon\else\@endalgoln\fi');
+      beforeFloat('algorithm');
+    },
+    afterDigest => sub {
+      if (ToString(DigestIf(T_CS('\algocf@style'))) =~ /box/) {
+        $_[1]->setProperty(frame => 'boxed'); }
+      afterFloat($_[1]); },
+    afterConstruct => sub { addFloatFrames($_[0], $_[1]->getProperty('frame')); }
+  );
+}
```
dginev, line 94: Same code as before here, just indented. Curious that the git diff is not recognizing it as a spacing change.

dginev:
PR Update:

A very timely report in [ar5iv#286](https://github.com/dginev/ar5iv/issues/286), with [this log report](https://ar5iv.labs.arxiv.org/log/1807.01774), led me to realize we also need to recognize an `{algorithm2e}` naming variant.

Separately, some lines in the algorithm output appear to get line-broken by pdflatex, which we are not doing in the latexml binding. I would like to provide a rough approximation of that via CSS, so I added a dedicated `ltx_algorithm` class next to the `ltx_float` class, so that I have an anchor for algorithm-specific styling.

PR continues to be ready for review+merge.

lib/LaTeXML/Post/MathML.pm
```diff
@@ -730,14 +730,14 @@ sub stylizeContent {
     else {
       $text = ($iselement ? $item->getAttribute('name') || $item->getAttribute('meaning') || $role : '?');
       $color = 'red'; } }
-  elsif ($role eq 'ADDOP' and $text eq '-') {    # MathML Core prefers unicode minus
+  elsif ($role and ($role eq 'ADDOP') and $text eq '-') {    # MathML Core prefers unicode minus
```
brucemiller, line 733: Probably a conflict with #1991, but will have to deal with it

PR #1988, opened 2022-11-16
glossaries.sty possible reliance on LaTeX 3

### LINKED ISSUE #1985: glossary.sty regression in texlive 11.2022
```
t/50_structure.t .......... 29/40 
#   Failed test 't/structure/glossary'
#   at /home/jazzpirate/work/Software/sTeX/LaTeXML/blib/lib/LaTeXML/Util/Test.pm line 113.
# Couldn't convert t/structure/glossary.tex: 0
```
LaTeXML version 0.8.6; revision e52efffc, pdfTeX 3.141592653-2.6-1.40.24 (TeX Live 2022)

dginev:
Fixes #1985 .

There are other possible approaches. This one tries to preserve the performance we have on pre-2022 texlives, trying to only load expl3 when needed.

Although I had to resort to explicitly loading expl3 in a new binding for mfirstuc.sty, as the several nested levels of loads appeared to make it impossible for the autoload to be visibile.

So overall, not really happy with this PR, but it keeps the test passing.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -586,6 +586,7 @@ lib/LaTeXML/Package/mathptm.sty.ltxml
 lib/LaTeXML/Package/mathptmx.sty.ltxml
 lib/LaTeXML/Package/mathrsfs.sty.ltxml
 lib/LaTeXML/Package/mathtools.sty.ltxml
+lib/LaTeXML/Package/mfirstuc.sty.ltxml
 lib/LaTeXML/Package/microtype.sty.ltxml
 lib/LaTeXML/Package/minimal.cls.ltxml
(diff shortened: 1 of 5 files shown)


dginev:
Good news: I correctly doubted my conclusion and double-checked: it looks like I had a subtle placement problem with the auto-load in my original try, rather than a scoping error.

So indeed, the rather nice-looking auto-loading directive in TeX.pool succeeds to recognize the case in glossary.sty and keeps the test passing on the latest texlive.

So no need for extra bindings or other advanced add-ons. Just minor polish that enables the new uppercase/lowercase macros from expl3 in all of LaTeX.

I am now happy with this PR, and think we can merge it right away. Ready for review @brucemiller .

PR #1989, opened 2022-11-17
refine parameter type for \href

### LINKED ISSUE #1986: Problems when a url has a percent sign
In the file below, I get an error from LaTeXML.  (I'll note that the input file has ten lines.)
```
jim@millstone:~/src/computing/tmp$ latexmlc test.tex --dest=test.html --includestyles
latexmlc (LaTeXML version 0.8.6; revision e52efffc)
processing started Sat Nov 12 08:00:40 2022
Error:expected:} Gullet->readBalanced ran out of input in an unbalanced state. at test.tex; line 11 col 0
Error:expected:{} Missing argument {} for Core::Definition::Expandable[\href Semiverbatim{}] at test.tex; line 13 col 0
Warning:unexpected:\end{document} Attempt to end document with open groups, environments or conditionals at test.tex; line 14 col 0
Error:unexpected:<closed> Mouth is unexpectedly already closed at String; line 0 col 0
Error:unexpected:<endgroup> Attempt to pop last locked stack frame at String; line 1 col 0
Error:expected:{} Missing argument {} for Core::Definition::Constructor[\@@Url Undigested{}{}Semiverbatim{}] at String; line 2 col 0
Conversion complete: 1 warning; 5 errors (See /home/jim/src/computing/tmp/test.latexml.log)
Post-processing complete: No obvious problems
Status:conversion:2
jim@millstone:~/src/computing/tmp$
```

If I run it through pdfLaTeX then I get no error and the link points to a URL with a percent sign (I have an up to date TeX Live).  If I comment out the first \href{}{} and uncomment the second then LaTeXML reports no error and the link goes to the URL in the uncommented line.

Just to have clearly stated the behavior I was hoping for: it would be great if it took the URL with the percent sign. 

Thank you,  Jim
+++++++++++++++++++++
```tex
\documentclass{article}
\usepackage{hyperref}

\begin{document}
Hello.
\href{http://scherk.pbworks.com/w/page/14864221/Quiz%3A%20Countable%20sets}{User scherk at \texttt{pbworks.com}}
% \href{http://scherk.pbworks.com/w/page/14864221/Quiz}{User scherk at \texttt{pbworks.com}}
Goodbye.

\end{document}
```

dginev:
Fixes #1986 .

One possible approach.


tkw1536:
Wouldn't it make sense to add a test with a URL that has a `%` sign in it? To make sure this keeps working in the future?

xworld21:
Just wondering, the catcodes used by hyperref are:
https://github.com/latex3/hyperref/blob/main/hyperref.dtx#L535-L558
```latex
\def\Hy@SetCatcodes{%
  \@makeother\`%
  \@makeother\=%
  \catcode`\$=3 %
  \catcode`\&=4 %
  \catcode`\^=7 %
  \catcode`\_=8 %
  \@makeother\|%
  \@makeother\:%
  \@makeother\(%
  \@makeother\)%
  \@makeother\[%
  \@makeother\]%
  \@makeother\/%
  \@makeother\!%
  \@makeother\<%
  \@makeother\>%
  \@makeother\.%
  \@makeother\;%
  \@makeother\+%
  \@makeother\-%
  \@makeother\"%
  \@makeother\'%
}
```
Do they match `Verbatim`?

Also, all of the above characters should be tested, if you add a test.

dginev:
@xworld21 close, for `\href` it's a similar but different macro, that is:
https://github.com/latex3/hyperref/blob/main/hyperref.dtx#L7392-L7416
```tex
  \gdef\hyper@normalise{^^A
    \begingroup
    \catcode`\^^M\active
    \def^^M{ }^^A
    \catcode`\%\active
    \let%\@percentchar
    \let\%\@percentchar
    \catcode`\#\active
    \def#{\hyper@hash}^^A
    \def\#{\hyper@hash}^^A
    \@makeother\&^^A
    \edef&{\string&}^^A
    \edef\&{\string&}^^A
    \edef\textunderscore{\string_}^^A
    \let\_\textunderscore
    \catcode`\_\active
    \let_\textunderscore
    \let~\hyper@tilde
    \let\~\hyper@tilde
    \let\textasciitilde\hyper@tilde
    \let\\\@backslashchar
    \edef${\string$}^^A
    \Hy@safe@activestrue
    \hyper@n@rmalise
  }^^A
```

I suppose we'll have to add a full test, now that we have so much expert attention on the issue :>

dginev:
I have now added a rather contrived test exercising the variety of character escapes in the macro above. See [t/complex/hyperchars.tex](https://github.com/brucemiller/LaTeXML/blob/775107aea24bc753866d2ab311265f109e025452/t/complex/hyperchars.tex).

It revealed another source of possible malformed URLs, as latexml's `Core::Document::serialize_attr` is very single-minded in always serializing using HTML entity escapes. So if say an ampersand `&` appeared inside an href attribute, that method would serialize it as `&amp;`, which would break the URL. Instead, for the very specific case of URI-holding attributes, we need to do URI escaping, where ampersand is mapped to `%26`.

I started extending the `f:url` XSLT function for this, as it seemed like a more well-positioned piece of code for doing such checks, which are specific to URL values.

But this should definitely get a look over by @brucemiller .

lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
```diff
@@ -151,7 +151,7 @@
   -->
   <func:function name="f:url">
     <xsl:param name="url"/>
-    <func:result><xsl:value-of select="$url"/></func:result>
+    <func:result><xsl:value-of select="f:subst($url,'&amp;','%26')"/></func:result>
```
xworld21, line 154: I think this breaks query strings: `&` should not be URL-encoded if it is separating different arguments after `?`. As far as I know, `&amp;` is fine inside an `@href` and will be correctly interpreted as a normal `&`. As long as it doesn't get *doubly* serialised, i.e. it doesn't become `&amp;amp;`. But that would be a different bug.

dginev: I think I managed to confuse myself here by testing the URLs in vacuum, rather than by clicking on `<a>` elements in HTML documents, where it would have become visible that the `&amp;` works just fine.

Thank you for catching that @xworld21 , I will undo the change.

brucemiller:
OK, fingers crossed :> Thanks for the patch & analysis

PR #1991, opened 2022-11-24
MathML spacing

### LINKED ISSUE #1804: Refactor away from relative mpadded widths
In the upcoming MathML Core recommendation, mpadded [width](https://www.w3.org/TR/2021/WD-mathml-core-20210816/#attribute-mpadded-width) is restricted to "a valid [length-percentage](https://www.w3.org/TR/2021/WD-mathml-core-20210816/#dfn-length-percentage)". This excludes the relative values such as `width="+5pt"`, `width="-2pt"`.

LaTeXML relies on these relative values in MathML.pm and Linebreaker.pm, and we should try to find a different formulation that achieves the same spacing effect.

Since the code seems to be dealing with left/right padding, maybe there is a way to translate that into a CSS-near "style" attribute with `padding-left` and `padding-right` values? For some standard cases we could also consider adding specialized spacing classes, which will be a little more configurable on the publisher end.

brucemiller:
This PR improves the spacing within math, and conforms to the restrictions in MathML Core specification.

It attempts to determine the desired spacing on the TeX side, by using the pair-wise spacing specified by TeX according to math atom type (See TeX Book, p.170), and by tracking explicit user spacing (as `XMHint`).  This is then compared to the spacing that MathML would use according to the OperatorDictionary. If there are differences, the spacing is adjusted, preferably by setting `lspace` or `rspace` on `m:mo`; in a pinch, using `m:mpadded`, but using the restricted form of the `width` attribute.

The results should conform to MathML Core.
Fixes #1804
Closes #1925


brucemiller:
@dginev OBTW: There's a failure with the CI test grammar_coverage, presumably due to the changes in 3 of the `t/parse` test cases? I'm not sure how to fix that.

dginev:
@brucemiller I have started studying the PR - very exciting!

Please rebase your local branch when you get a chance - I updated the `t/parse/nested_application` test so that it continues to exercise the necessary MathGrammar rules. The change of the role of `\partial` led to it no longer testing the same rule, so I reframed it to `\lxDeclare`. The PR should get a green checkmark in CI now.

Moving on to the studying the actual upgrades.

lib/LaTeXML/Post/MathML.pm
```diff
@@ -670,89 +645,56 @@ my %plane1hackable = (    # CONSTANT
...
+  my $cssstyle = join('; ', grep { $_; } $attr{ccsstyle}, ($iselement && $item->getAttribute('cssstyle')));
+  my $variant = ($font ? unicode_mathvariant($font) : '');
+  my $istoken = $tag =~ /^m:(?:mi|mo|mn)$/;                  # mrow? well, no....
+  my $href    = $istoken && ($attr{href}  || ($iselement && $item->getAttribute('href')));
+  my $title   = $istoken && ($attr{title} || ($iselement && $item->getAttribute('title')));
+
+  # Implied attributes, relevant for mo operators
+  my $stretchy = ((defined $attr{stretchy} ? $attr{stretchy} : ($iselement && $item->getAttribute('stretchy')))
+      || 'false') eq 'true';
+  my $isfence   = $role && ($role =~ /^(OPEN|CLOSE|MIDDLE)$/);
```
dginev, line 677: isn't the `VERTBAR` role also a potential fence?

lib/LaTeXML/Post/MathML.pm
```diff
@@ -670,89 +645,56 @@ my %plane1hackable = (    # CONSTANT
...
+  my $ispunct   = $role && ($role eq 'PUNCT');
+  my $islargeop = $role && ($role =~ /^(SUMOP|INTOP)$/);
```
dginev, line 679: the large operators according to MathGrammar are:
```
bigop   : BIGOP | SUMOP | INTOP | LIMITOP | DIFFOP
```

Is it worth including the missing `BIGOP`, `LIMITOP` and `DIFFOP` ?

lib/LaTeXML/Post/MathML.pm
```diff
@@ -670,89 +645,56 @@ my %plane1hackable = (    # CONSTANT
...
+  my $variant = ($font ? unicode_mathvariant($font) : '');
+  my $istoken = $tag =~ /^m:(?:mi|mo|mn)$/;                  # mrow? well, no....
+  my $href    = $istoken && ($attr{href}  || ($iselement && $item->getAttribute('href')));
+  my $title   = $istoken && ($attr{title} || ($iselement && $item->getAttribute('title')));
+
+  # Implied attributes, relevant for mo operators
+  my $stretchy = ((defined $attr{stretchy} ? $attr{stretchy} : ($iselement && $item->getAttribute('stretchy')))
+      || 'false') eq 'true';
+  my $isfence   = $role && ($role =~ /^(OPEN|CLOSE|MIDDLE)$/);
+  my $ispunct   = $role && ($role eq 'PUNCT');
```
dginev, line 678: Unsure if worth mentioning but `PERIOD` also acts as punctuation in the `endPunct` grammar rule. Would it be helpful to recognize here?

lib/LaTeXML/Post/MathML.pm
```diff
@@ -768,10 +710,48 @@ sub stylizeContent {
...
+  if ($text =~ /^[\x{2061}\x{2062}\x{2063}]*$/) {    # invisible get no size or stretchiness
+    $stretchy = $size = undef; }
+  if ($size) {
+    if ($size eq ($LaTeXML::MathML::SIZE || 'text')) {    # If default size, no need to mention.
+      $size = undef; }
+    # If requested relative size, and in script or scriptscript, we'll need to adjust the size
+    elsif (($size =~ /%$/) && ($LaTeXML::MathML::STYLE and $LaTeXML::MathML::STYLE =~ /script/)) {
+      my $req = $size;                               $req =~ s/%$//;
+      my $ex  = $stylesize{$LaTeXML::MathML::STYLE}; $ex  =~ s/%$//;
+      $size = int(100 * $req / $ex) . '%'; }
```
dginev, line 743: it would help to add a guard that ensures `$ex` is never zero or undefined here, or we may be at risk for a perl die.

lib/LaTeXML/Post/MathML.pm
```diff
@@ -781,69 +761,42 @@ sub stylizeContent {
...
-            # Just assume something between mediummathspace = 4/18em = 2.222pt
-            # and thickmathspace = 5/18em = 2.7777pt, so 2.5pt.
-        ($lpad ? (lspace => max(0, (2.5 + getXMHintSpacing($lpad))) . 'pt') : ()),
-        ($rpad ? (rspace => max(0, (2.5 + getXMHintSpacing($rpad))) . 'pt') : ()),
-        # If an operator has specifically located it's scripts,
-        # don't let mathml move them.
-        (($pos =~ /mid/) || $LaTeXML::MathML::NOMOVABLELIMITS
-          ? (movablelimits => 'false') : ()) },
-      $text]); }
+  return pmml_mayberesize($item, ['m:mo', { %mmlattr, }, $text]); }
```
dginev, line 799: `pmml_mayberesize` is a tiny bit less readable than `pmml_maybe_resize`

lib/LaTeXML/Post/MathML.pm
```diff
@@ -1071,6 +1024,210 @@ sub pmml_text_aux {
   else {
     return (); } }
 
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+# Adjust the spacing by comparing TeX & author specified spacing to MathML's defaults
+sub adjust_spacing {
+  my ($self, $mml) = @_;
+  space_walk($self, $mml);
```
dginev, line 1031: A name for the ages. 🛰️ 👨🏻‍🚀 🌎

dginev: An interesting observation -- the `adjust_spacing` routine does exactly a single call to `space_walk` and nothing else. So they're interchangeable. And possibly redundant? If one wanted to always use one of the names or the other, one could?

lib/LaTeXML/Post/MathML.pm
```diff
@@ -1071,6 +1024,210 @@ sub pmml_text_aux {
...
+  Debug("SPACEWALK: TeX $tex_space + $req_space vs OpDict $default : "
+      . ($needs_adjustment ? colorizeString("ADJUSTMENT $target", 'warning') : "  NO adjustment")
+      #        . ($invisop ? " [with invisible op]" : '')
+      . LaTeXML::Post::MathProcessor::shownode($prev) . " $prev_type "
+      . LaTeXML::Post::MathProcessor::shownode($next) . " $next_type "
+    ) if                                                         # $needs_adjustment &&
+    $LaTeXML::DEBUG{mathspacing};
+  return unless $needs_adjustment;
+  # Note that in MML Core, neither mspace nor mpadded can have negative width!
+  # It also does not support relative width using +/- previx!
```
dginev, line 1197: `previx` typo `prefix`

lib/LaTeXML/Post/MathML/OperatorDictionary.pm
```diff
@@ -0,0 +1,248 @@
...
+our $Operators_separator;
+our $Content_form;
+our $Category_data;
+# Spacing constants, defined at end
+our $thin;     #  = 0.167;    # 3 mu in ems
+our $med;      #   = 0.222;    # 4 mu
+our $thick;    # = 0.278;    # 5 mu
+
+# Rough assumption of what MathML form each role corresponds to
+# (without actually looking at its position within the expression!)
```
dginev, line 36: Indeed, that lack of contextual positioning is bound to crop up in the future -- great to have the comment here. 

Seeing `VERTBAR` as `postfix` is a notable example.

dginev (commented):
I made a pass reading through the code at length, and left some minor comments along the way.

Some of the detailed computations haven't clicked in my mind yet, but the changes to the test files, and the logical structure of the PR both look very healthy.

I'll focus on trying out some practical tests next and reviewing the DLMF draft.

lib/LaTeXML/Post/MathML.pm
brucemiller: It should be the case that MathParser has converted `VERTBAR` to one of those when it recognizes the usage.

lib/LaTeXML/Post/MathML.pm
brucemiller: Those are "big", not "large" :> There's a subtle (and not necessarily consistently respected) distinction between "big" operators in the notational sense of pre & post scripts, unparenthesized body arguments and such, versus "large" operators in the display sense (of MathML) which get larger font size under the right circumstances.

This should probably all be checked for consistency & documented, but I'm hesitant to get into it now.

lib/LaTeXML/Post/MathML.pm
brucemiller: Probably better to call it `$issep`, which is how it's used (upcoming commit)

lib/LaTeXML/Post/MathML.pm
brucemiller: Yeah, good idea.

lib/LaTeXML/Post/MathML.pm
brucemiller: It was initially proactive, in case weird arguments, `local` etc were needed. Then I kept it because the inner function name is too much fun, but too unclear at the "top" level.

lib/LaTeXML/Post/MathML.pm
brucemiller: Thanks;

lib/LaTeXML/Post/MathML/OperatorDictionary.pm
brucemiller: In the long run, I think a good strategy is having `role` indicate the (possibly ambiguous) grammatical roles going into parsing, and have them come out of parsing being tweaked to (notationally) unambiguous roles.  Hopefully, the unambiguous grammatical role is a sufficiently direct mapping to presentational display atom-types or forms.

dginev (approved):
Thank you for the comments, the co-debugging via email and the updated commit @brucemiller !

I have tested a range of content in DLMF and arXiv, and I believe the horizontal spacing upgrades from this PR behave quite nicely.

So, while I will continue to proofread some more, I'm marking the PR as good to merge - we can always provide more spacing tweaks later.

PR #1996, opened 2022-11-28
Better organization of copy-attribute templates 

brucemiller:
so they get the right priority.  This fixes a bug introduced in #1960.
Would be good to merge this quickly as it fixes a bug where `href` attributes don't get rewritten, breaking dlmf.

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
@@ -383,38 +383,23 @@
     </xsl:element>
   </xsl:template>
 
-  <!-- copy attribute? In general don't -->
-  <xsl:template match="@*" mode='copy-attribute'/>
+  <!-- Tricky to set up templates for attributes with & without namespaces that override correctly -->
 
-  <!-- but copy non-namespaced attributes -->
+  <!-- Copy NON-namespaced attributes -->
   <xsl:template match="@*[namespace-uri() = '']" mode='copy-attribute'>
     <xsl:attribute name="{local-name()}" namespace="{namespace-uri()}">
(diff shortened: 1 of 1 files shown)


dginev (approved):
Interesting, first time I encounter XSLT's priority. So
- it ranges between -9 and 9, 
- default is 0,
- rules with higher priority get executed before rules with lower priority

The PR looks good in light of that. Feel free to merge when the Linux CI passes (we are again having installation silliness with the Windows CI).

xworld21:
> * default is 0,

I can also be 0.5 according `@match`, which is probably what happens here https://www.w3.org/TR/xslt-10/#conflict. Presumably here `@*[namespace-uri() = '']` gets priority 0.5.

My unsolicited 2c, for what it's worth, `priority` should be avoided if possible, like `!important` in CSS, in this case replacing `@href | ...` with `@*[namespace-uri() = '' and (local-name() = 'href' ...)]`, or better yet exclude `@href` and similar from the blanket `@*`. This is because priority rules for imported stylesheets get even more confusing, plus it documents which template is causing the clash.

dginev:
Very interesting, thanks @xworld21 . I think Bruce may need the fix relatively quickly, but I agree we could/should improve the robustness as you suggest. Just like the first `!important` first `priority` looked innocuous, but I can imagine complete chaos if we had 20 of them.

brucemiller:
Indeed, I agree `priority` is nasty. OTOH, exceedingly complex matches are also kinda nasty.  There ought to be a simpler way, but it seems that namespaces being an afterthought make these templates more complicated (and more specific?) than they need to be. At least, if you want to override this particular template, you can override the definition of `f:url` :>

PR #2001, opened 2022-12-06
avoid infinite loop when XMRef does not contain idref

dginev:
Fixes an unfortunate Fatal error in arXiv:2210.07557, where an `XMRef` was created without an `idref` attribute, causing an unsuspecting check to infinitely recurse.

### DIFF
--- a/lib/LaTeXML/Package/physics.sty.ltxml
+++ b/lib/LaTeXML/Package/physics.sty.ltxml
@@ -188,7 +188,9 @@ DefMacro('\evaluated', sub {
           phys_revArg(I_arg(1), $open, $close),
           ($lower ? (T_SUB,   phys_revArg(I_arg(2))) : ()),
           ($upper ? (T_SUPER, phys_revArg(I_arg(3))) : ())) },
-      I_apply({}, I_symbol({ meaning => 'evaluated-at' }), I_arg(1), I_arg(2), I_arg(3)),
+      I_apply({}, I_symbol({ meaning => 'evaluated-at' }), I_arg(1),
+        ($lower ? (I_arg(2)) : ()),
+        ($upper ? (I_arg(3)) : ())),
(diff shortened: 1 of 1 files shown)


brucemiller:
Isn't this something that "shouldn't" happen?  Is an error reported before that no idref could be assigned or something? MWE?

dginev:
Took me a few minutes, but here is the minimal example that reproduces the Fatal:
```tex
\documentclass{article}
\usepackage{physics}
\begin{document}

\begin{equation}
\eval{ \pi }_{0}
\end{equation}

\end{document}
```

In master, that prints:
```
Warning:expected:id Missing idref on ltx:XMRef
Fatal:perl:deep_recursion Deep recursion on subroutine "LaTeXML::MathParser::is_genuinely_unparsed" at /home/deyan/perl5/lib/perl5/LaTeXML/MathParser.pm line 1009.
	at NanoSwimmer.tex; line 6 col 0 - line 6 col 17
```

brucemiller:
The guard is needed, since it is perhaps too easy to misapply `I_arg`, although it's not clear that counts as "unparsed".  Still would be better to catch the root cause.  Can you try whether this patch to `physics.sty.ltxml` takes care of it:
```
DefMacro('\evaluated', sub {
-      I_apply({}, I_symbol({ meaning => 'evaluated-at' }), I_arg(1), I_arg(2), I_arg(3)),
+      I_apply({}, I_symbol({ meaning => 'evaluated-at' }), I_arg(1),
+              ($lower ? (I_arg(2)):()),
+              ($upper ? (I_arg(3)):())),
 ```

dginev:
@brucemiller thanks, that defensive change to physics.sty improved the specific article as well - just tested it.

I agree we can have both the general guard in MathParser, and the specific binding fix. I've updated the PR.

brucemiller:
It's not really "defensive", just *correct*!

Issue #2000, opened 2022-12-01
longtable can't handle a pagebreak before a multicolumn


teepeemm:
The tex code
```tex
\documentclass{article}
\usepackage{longtable}
\begin{document}
\begin{longtable}{c}
\pagebreak
\multicolumn{1}{c}{text}
\end{longtable}
\end{document}
```
under `latexml` has the console output:
```
Error:unexpected:\omit \omit cannot be used here at script.tex; line 6 col 0
Error:unexpected:\endgroup Attempt to close non-boxing group at script.tex; line 7 col 0
Warning:unexpected:\end{document} Attempt to end document with open groups, environments or conditionals at script.tex; line 8 col 0
Error:unexpected:<endgroup> Attempt to pop last locked stack frame at String; line 0 col 0
Error:malformed:ltx:document Closing </ltx:document> whose open descendents do not auto-close at String; line 0 col 0
Error:malformed:ltx:table Attempt to close </ltx:table>, which isn't open
Conversion complete 1 warning; 5 errors
```
and in the xml, the page break has escaped to after the tabular, but within the table.  If a non-multicolumn comes after the page break, then the conversion has no errors or warnings, but the pagebreak escapes to after the tabular and after the table.  Since we don't do much with `\pagebreak` anyway, would it be better to have `\pagebreak` produce no xml output if it's in a longtable?

dginev:
Right, a really quick fix here would be to `Let('\pagebreak',`\relax');` inside the longtable environment.

And maybe in tabulars in general?

teepeemm:
Most tabular environments (I looked at booktabs, deluxetable, longtable, pgfplotstable, supertabular, tablefootnote, tabularx, tabulary, threeparttable) don't span multiple pages to begin with.  deluxetable and supertabular can span multiple pages, but they ignore `\pagebreak` in the general case, and report an error in the `\pagebreak\multicolumn` case.  Since that suggests people won't have them in TeX, I don't know that we need to worry about them.

But if we find ourselves wanting to make sure and ignore the pagebreak, the other thing I realized is that `\pagebreak` takes an optional argument that also needs to be ignored.  Consequently, I added `DefMacro('\@gobble@optional[]', Tokens());` to longtable.sty.ltxml.

PR #2002, opened 2022-12-06
ignore pagebreak within longtable

teepeemm:
This ignores `\pagebreak` within `{longtable}`, resolving #2000.  (I have checked that `\pagebreak` still works after the environment ends.)

This also removes the `{longtable*}` environment, which doesn't exist in TeXLive (and if it's supposed to mimic `{tabular*}{width}{colspec}`, then it's missing the `{width}` parameter).

### DIFF
--- a/lib/LaTeXML/Package/longtable.sty.ltxml
+++ b/lib/LaTeXML/Package/longtable.sty.ltxml
@@ -24,10 +24,15 @@ DefMacro('\longtable[]{}',
   '\@longtable@bindings{#2}\@@longtable[#1]{#2}\@start@alignment');
 DefMacro('\endlongtable',
   '\@finish@alignment\@end@tabular');
-DefMacro('\csname longtable*\endcsname []{}',
-  '\@longtable@bindings{#2}\@@longtable[#1]{#2}\@start@alignment');
-DefMacro('\csname endlongtable*\endcsname',
-  '\@finish@alignment\@end@tabular');
+# {longtable*} is not defined in longtable.sty
+# and if we're trying to mimic {tabular*},
+# then this needs \endcsname{Dimension}[]{}
+#DefMacro('\csname longtable*\endcsname []{}',
+#  '\@longtable@bindings{#2}\@@longtable[#1]{#2}\@start@alignment');
+#DefMacro('\csname endlongtable*\endcsname',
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/longtable.sty.ltxml
```diff
@@ -24,10 +24,15 @@ DefMacro('\longtable[]{}',
...
-  '\@longtable@bindings{#2}\@@longtable[#1]{#2}\@start@alignment');
-DefMacro('\csname endlongtable*\endcsname',
-  '\@finish@alignment\@end@tabular');
+# {longtable*} is not defined in longtable.sty
+# and if we're trying to mimic {tabular*},
+# then this needs \endcsname{Dimension}[]{}
+#DefMacro('\csname longtable*\endcsname []{}',
+#  '\@longtable@bindings{#2}\@@longtable[#1]{#2}\@start@alignment');
+#DefMacro('\csname endlongtable*\endcsname',
+#  '\@finish@alignment\@end@tabular');
```
dginev, line 33: Please don't remove `longtable*`, maybe just enhance it?

In texlive-proper, the variety of versioned revtex classes (such as `revtex4-1.cls`) defined `longtable*`, and in latexml that is realized via including the longtable package and adding the star variant in it.

Maybe moving the definition of `longtable*` to `revtex4_support.sty.ltxml` is "more correct" in respecting the texlive organization, but I don't see any harm with leaving it in longtable.sty.ltxml for now.

dginev (changes requested):
The `\pagebreak` change itself looks healthy, and if @teepeemm has tested in - I'm happy with merging it.

But we should avoid dropping `longtable*` here, I remember it getting used in the arXiv runs. Maybe a separate PR could tweak the definition to more closely match texlive's, but I am open to postponing that until we have a motivating example.

lib/LaTeXML/Package/longtable.sty.ltxml
teepeemm: I missed that it was defined in revtex4-1.  longtable* is not mimicking tabular*; instead, longtable* spans the whole page in a two column document.  So changing that can wait until we figure out how to deal with columns.

brucemiller:
Looks good! Thanks!!

PR #2003, opened 2022-12-06
Add "notoday" option for latexml.sty

dginev:
This PR adds a small option to latexml.sty, which allows to request disabling the `\today` macro.

This is motivated entirely by arXiv, where certain papers use the `\date{\today}` LaTeX idiom, which is only ever reasonable at the date when the article was originally submitted. 

Since we currently don't have a convenient handle on that submission metadata, a second-best workaround is to hide the date entirely in such cases, which I find simplest by disabling the `\today` macro in any arXiv conversion job.

It's opt-in only, and I expect only the arXiv builds will be using this in the near future.

### DIFF
--- a/lib/LaTeXML/Package/latexml.sty.ltxml
+++ b/lib/LaTeXML/Package/latexml.sty.ltxml
@@ -59,6 +59,9 @@ DeclareOption('rawclasses',     sub { AssignValue('INCLUDE_CLASSES' => 1,       
 DeclareOption('localrawclasses', sub { AssignValue('INCLUDE_CLASSES' => 'searchpaths', 'global'); });
 DeclareOption('norawclasses',    sub { AssignValue('INCLUDE_CLASSES' => 0, 'global'); });
 
+# No \today, useful when re-converting older archival articles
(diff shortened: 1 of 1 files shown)


brucemiller:
Closed meaning #1920 works for you, or are you planning to make suggestions there?

dginev:
@brucemiller I'll make a highly experimental `ar5iv.sty.ltxml` binding in the side repository at [ar5iv-bindings](https://github.com/dginev/ar5iv-bindings/), and will add the override there directly, to avoid polluting the main repository.

That way it doesn't even need to be an exposed option - just loading that binding file will set `\today` to `\relax` and lock it in place.

brucemiller:
Are file timestamps meaningful for arXiv? (I recall hearing rsync 
mentioned, so maybe...) Could be a useful strategy?

dginev:
File timestamps, or the official OAI metadata, could both provide the submission date when the file was written, yes. So long-term that would be better.

I won't have time to get that in place before the next rerun however (not important enough), so just disabling `\today` works for this pass.

PR #2009, opened 2022-12-08
Don't complicate mtext if no nested math

### LINKED ISSUE #1846: \emph{} in \text{} generates spurious mathvariant="italic"
It seems like `\emph` within `\text` occasionally generates an extra `@mathvariant` attribute. Not always, but definitely with the following:
```latex
\documentclass{article}
\usepackage{amsmath}
\begin{document}
  \[ x^2 \text{text in \emph{italic font}} \]
\end{document}
```
which becomes
```xml
<mtext mathvariant="italic"><em class="ltx_emph ltx_font_italic">italic font</em></mtext>
```
The `@mathvariant="italic"` looks pretty bad in Firefox.

brucemiller:
As it says; was unnecessarily scanning the children of `XMText`, resulting in multiple `mtext`.

Fixes #1846

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -492,7 +492,7 @@ sub pmml_internal {
   elsif ($tag eq 'ltx:XMText') {
     my @c = $node->childNodes;
     my $result;
-    if (!$$self{nestmath}) {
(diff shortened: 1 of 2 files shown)


dginev (approved):
Nice catch, LGTM

PR #2014, opened 2022-12-09
Refix mtext nesting

brucemiller:
This is an alternative fix for #1846 to avoid breaking MathJax.  It basically reverts #2009, but defines `\emph` so as to use `ltx:text` when within math, rather than `ltx:emph`.   We can better distill `ltx:text` down to essentials, but `ltx:emph` sticks around and finds its way inside the eventual `m:mtext`.  MathJax can't cope with this. So (arguably) the issue in #1846 is not the `mathvariant` on `m:mtext`, but rather the redundant html `<em>` inside of it.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -441,11 +441,16 @@ DefMacroI('\today', undef, sub { ExplodeText(today()); });
 
 # Use fonts (w/ special flag) to propogate emphasis as a font change,
 # but preserve it's "emph"-ness.
-DefConstructor('\emph{}',
-  "<ltx:emph _force_font='1'>#1",
-  mode           => 'text', bounded => 1, font => { emph => 1 }, alias => '\emph',
-  afterConstruct => sub { $_[0]->maybeCloseElement('ltx:emph'); },
-  beforeDigest   => sub {
+# Use ltx:emph in text, but use ltx:text within math, since it may collapse within XMText
+# (and minimize html within math, which causes MathJax problems)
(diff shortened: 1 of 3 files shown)


xworld21:
`mathvariant="italic"` looks quite ugly in Firefox, though
image

Further bit of info for the future, HTML in MathML will be supported in MathJax 4. Not sure what release schedule they have in mind.

dginev (approved):
Looks fine to merge on a surface read.

I have a slight skepticism about using an italic mathvariant on the `mtext` element (particularly if it ends up in a case where it has to contain arbitrary HTML, say `<a>` links), but the real question is whether the browsers do the "right thing" today.

If all 3 of Firefox, Chrome and MathJax are happy with this reformulation, fine by me.

brucemiller:
Yeah well, you've got to put just the right amount of "English" on it to have all your own bugs veer exactly between everybody else's.

DLMF having big red & yellow "Math input error" kinda ties my hands.

dginev:
Completely agree that's the right call, DLMF rightfully drives the latexml release cycle.

But maybe I can send an invitation to @xworld21 to let us know when MathJax 4 officially ships, or whichever version it may be that has "HTML in mtext" supported. I suspect the entire landscape of math-on-the-web rendering may have changed quite a bit in the next 6 months. We can always land a quicker-than-usual 0.8.8...

xworld21:
> But maybe I can send an invitation to @xworld21 to let us know when MathJax 4 officially ships, or whichever version it may be that has "HTML in mtext" supported.

Sure! I am pretty sure I will notice when MathJax 3.3/4/??? comes out. I'll let you know how things have changed when that happens.

PR #2015, opened 2022-12-10
Defer frontmatter

### LINKED ISSUE #1395: Order of titlepage and abstract
I'm trying to use the `titlepage` environment so that I can use `\title` to define a short title for LaTeXML to use in the HTML title and navigation links.  It appears strange to me that the `titlepage` content is generated after the `abstract`:
```tex
\documentclass{report}

\title{Title provided for LaTeXML}

\begin{document}

\begin{titlepage}
\huge
This is the titlepage.
\end{titlepage}

\begin{abstract}
Abstract goes here.
\end{abstract}

This follows the abstract.

\end{document}
```

I would have expected something more like the output when using `\maketitle`:
```tex
\documentclass{report}

\title{This is the titlepage.}

\begin{document}

\maketitle

\begin{abstract}
Abstract goes here.
\end{abstract}

This follows the abstract.

\end{document}
```

brucemiller:
Previously, all accumulated FrontMatter (author, title, etc) was placed at the beginning of the document. This PR notices when `\maketitle` or `{titlepage}` environments were used, and defers the placement of FrontMatter until immediately after the `\maketitle` or titlepage.

fixes #1395


t/complex/aliceblog.xml
```diff
@@ -8,12 +8,12 @@
 <document xmlns="http://dlmf.nist.gov/LaTeXML" class="ltx_authors_1line" prefix="cc: http://creativecommons.org/ns# dcterms: http://purl.org/dc/terms/ foaf: http://xmlns.com/foaf/0.1/ schema: http://schema.org/ xsd: http://www.w3.org/2001/XMLSchema#">
   <resource src="LaTeXML.css" type="text/css"/>
   <resource src="ltx-article.css" type="text/css"/>
+  <rdf about="" property="foaf:neighbor" resourceidref="fredblog"/>
   <title>Alice’s Blog</title>
   <creator role="author">
     <personname>Alice</personname>
   </creator>
   <rdf aboutlabelref="LABEL:blog:photos" property="cc:license" resourcelabelref="LABEL:license"/>
-  <rdf about="" property="foaf:neighbor" resourceidref="fredblog"/>
```
dginev, line 16: Interesting that this moves a single `rdf` element (rather than none or all).

And in that vein - do we want to have a small `{titlepage}` test, just to make sure the order remains as expected?

dginev (commented):
Looks reasonable.

I am not sure if `\maketitle` needs the `\lx@deferfrontmatter` given that it executes `lx@dofrontmatter` immediately ofter, which will always run the insertion and mark it as `done`, so that a further deferral isn't needed.

t/complex/aliceblog.xml
brucemiller: There needs to be both a digestion & construction component, but I was able to simplify to a single `\lx@frontmatterhere`.

But thanks for drawing my attention to the need to explain, which also had me realize that the semi-redundant abstract placement from (#1638) was not only redundant, but got broken.  So the next commits combine things much more cleanly.  However, you may want to recheck the article mentioned in #1638 (I can't get at it).

Also, added a couple of tests, thanks for mentioning.

dginev (approved):
All out of feedback, the latest code makes sense to me.

A parting minor comment is that `\lx@frontmatterhere` name is a little bit hard to understand intuitively. Maybe I would have gone for say `\lx@build@frontmatter@xml`, but I admit that is a different kind of clunky. Naming things is hard.

PR #2017, opened 2022-12-11
Citationkey case

### LINKED ISSUE #1674: \bibitem and \cite seems to be case-insensitive
In LateX, `\cite` is case-sensitive, but for LaTeXML, `\cite` seems to be case-insensitive.
A minimal example is [cite-test.zip](https://github.com/brucemiller/LaTeXML/files/7285455/cite-test.zip), which results in “[1] vs [2]” in LaTeX, but “[2] vs [2]” in LaTeXML.

brucemiller:
Try to emulate BibTeX's (curious) behavior: It uses case-insensitive lookup to find entries in the bib file, but then uses the key with the case as cited for the aux file, since LaTeX's labels & cites are case-sensitive.

fixes #1674


dginev (approved):
Fascinating evolutionary design...

If we can still process the DLMF bibliography, looks good to merge.

PR #2018, opened 2022-12-11
patch \bibitem buglet in natbib

dginev:
Feeling a little proud of spotting this one-liner PR.

There was a Fatal with `arXiv:1104.3156` due to a long chain of malformed ltx:bibitem of the form
```
Error:malformed:ltx:bibitem <ltx:bibitem> isn't allowed in <ltx:p> 
```

Upon inspection, this was due to a peculiar custom parsing algorithm in natbib.sty.ltxml, called `NAT_peel_arg`, which had a fascinating (and hard to hit) bug in using an array `pop` instead of `shift` in one of the recognized cases.

Changing that switches this Fatal to a Warning status.

### Debugging details

minimal TeX to reproduce:
```tex
\documentclass{article}
\usepackage{natbib}

\begin{document}

\begin{thebibliography}{1}
\bibitem[\protect\citeauthoryear{Chabrier} {2003}]{cha03} Test
\bibitem[Cooper et al.(2008)]{coo08} Test2
\end{thebibliography}
\end{document}
```


brucemiller:
Wow! good catch.

PR #2019, opened 2022-12-11
Guard against anonymous node names in math postprocessing

dginev:
There are certain cases in arXiv where we get a barrage of warnings due to math nodes with `undef` names.

As one example, I believe empty XMHint elements can get constructed in some tabular contexts (just holding a `role` attribute) in the XMath processing. Which then (sometimes?) get mapped into nodes with `undef` names in post-processing. I think. I haven't fully dug down to the bottom of the cause for the nodes. But they are present in `arXiv:hep-ph9607289`, and I could imagine other strange TeX contraptions to induce them as well.

So, this PR cleans up what are about 112 warnings in that article, by defensively expecting `undef` in node name checks, at least in the code locations where I saw warnings being raised.

### Debugging details

In `arXiv:hep-ph9607289` there was an unreasonable number of warnings stemming from nodes with `undef` names (in the latexml "array-ref triple" data structure formulation).

A minimal example isolating that is:
```tex
\def\GeV{{\rm GeV}}
\def\lsim{\mathrel{\mathpalette\versim<}}
\catcode`@=11
\def\versim#1#2{\lower0.2ex\vbox{\baselineskip\z@skip\lineskip\z@skip
       \lineskiplimit\z@\ialign{$\m@th#1\hfil##$\crcr#2\crcr\sim\crcr}}}

(2~$\GeV\lsim Q_0\lsim4$~GeV)

\end
```

That 1996 construct actually runs through pdftex and typesets a neat-looking formula. What happens to it in latexml is a much bigger topic, one that we may (or may not) want to examine in the mid-term future. The currently generated MathML is not ideal, but at least "reminds of" the expected outcome.

### DIFF
--- a/lib/LaTeXML/Post.pm
+++ b/lib/LaTeXML/Post.pm
@@ -947,7 +947,7 @@ sub checkDestination {
 
 sub stringify {
   my ($self) = @_;
-  return 'Post::Document[' . $self->siteRelativeDestination . ']'; }
+  return 'Post::Document[' . ($self->siteRelativeDestination || '') . ']'; }
 
 sub getLocator {
   my ($self) = @_;
@@ -1084,7 +1084,10 @@ sub addNodes {
   foreach my $child (@data) {
     if (ref $child eq 'ARRAY') {
       my ($tag, $attributes, @children) = @$child;
-      if ($tag eq '_Fragment_') {
+      if (!$tag) {
+        # don't add anonymous nodes? or?
+        next; }
+      elsif ($tag eq '_Fragment_') {
         my $indent;    # Derive indentation from indentation of $node
         if (my $pre = $node->previousSibling) {
           if (($pre->nodeType == XML_TEXT_NODE) && (($pre = $pre->textContent) =~ /^\s*$/)) {
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -511,6 +511,7 @@ sub needsMathstyle {
   my ($node) = @_;
   if (ref $node eq 'ARRAY') {
     my ($tag, $attr, @children) = @$node;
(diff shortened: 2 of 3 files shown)


dginev:
Also, @brucemiller there is a topical bit of markup that we may want to address here for MathML Core. It seems that such tabular snippets still generate mpadded with relative widths (which should have gone away with #1991 ).

Concretely, the minimal snippet adds a:
```xml
<mpadded depth="+0.9pt" height="-0.9pt" voffset="-0.9pt">
  <mtable> ... </mtable>
</mpadded>
```

I think we should consider phasing these out as well...

brucemiller:
Would be better to fix the problem than propagate it, even with guards. Alternative solution coming... Also, can't find the article; link would be better?

dginev:
I no longer have the easy download links, apologies, you can get the tarball from arXiv directly:
https://arxiv.org/format/hep-ph/9607289

dginev:
@brucemiller I see that with your PR we are still generating the relative width mpadded in the example here. Should I try to make a separate PR for that, or would you like to take a look?

brucemiller:
At this point, neither, actually. To do it "right" needs to have a more plausible measurement (the one I added really only approximates tokens, and the current case already has an `mtable`).  And improving the measuring really should be integrated with the linebreaker, which of course needs its own complete rethink/refactor.

Unless it can be achieved with a simple css addition (*without* touching `display` or sneaking in html, which recent Core discussions confirm are dangerous), I'd just as soon not pile too much more special casing which will be confusing later on.

dginev:
@brucemiller I'll see what I can come up with. I would like to claim that LaTeXML 0.8.7 "tries to" be compatible with MathML Core, and the relative mpadded widths would prevent me from saying that in good conscience.

PR #2020, opened 2022-12-11
allow more recursion on invokeToken

dginev:
Avoids another deep recursion Fatal in arXiv, useful for some large tikz diagrams.

### Debugging details

<details>

Motivating tikz example screenshot from `arXiv:2210.11132`:

onenode

That article recovers from Fatal to Error, in 8 minutes total conversion time.

</details>


dginev:
I also allowed a deeper recursion on the generic `->invoke` calls, which resolves a separate deep recursion Fatal of `arXiv:2210.11983`. Sadly that article remains a Fatal, but now with 100 Errors.

PR #2023, opened 2022-12-12
Deal with stray/unexpected XHMints in MathML

brucemiller:
As it says...

closes #2019


brucemiller:
That's why I find the actual xml to be so useful; you can see the path rather than only a mysterious end result.

dginev:
The XML is part of it, but certainly not all of it - I did generate the core XML and spotted the hints (which is how I knew to mention XMHint in my PR), but couldn't quite put my finger on how they ended up in those XMCell elements and whether they were intended to serve a useful purpose.

PR #2024, opened 2022-12-12
allow to globally toggle line-breaking of UnTeX

### LINKED ISSUE #1393: Line break deactivation on long equations
I observe that LaTeXML breaks line on very long equation and add `%` at the end of the line before the break. Is there any possibility to deactivate this feature? I do not find the solution in the documentation.

dginev:
Fixes #1393 .

This PR keeps the current behavior of latexml as-is, but it exposes a new global toggle through latexml.sty, which can set/unset the line-breaking **default** of the `UnTeX` helper subroutine. That can still be overridden using the optional second argument of `UnTeX`.

The expected use case (which I will test in the next ar5iv run) is to disable the line-breaks globally, via:

```bash
latexmlc  --preload=[nobreakuntex]latexml.sty \
  'literal:$111111111+2222222222+333333333+444444444+555555555+
            66666666+777777777+888888888+99999999$'
```

which will generate the tex attribute:
```
tex="111111111+2222222222+333333333+444444444+555555555+66666666+777777777+888888888+99999999"
```

in contrast the default would generate an extra line-break via `%\n`, as in:
```
tex="111111111+2222222222+333333333+444444444+555555555+66666666+777777777+88888888%&#10;8+99999999"
```

which is identical to the explicit `--preload=[breakuntex]latexml.sty`.

I have not allowed customizing the line length for the breaks, which is currently always 78, but I switched it to an `our` declaration, just in case. If accessing a variable length through State is also desired change, I can update the PR.

### DIFF
--- a/lib/LaTeXML/Core/Token.pm
+++ b/lib/LaTeXML/Core/Token.pm
@@ -131,11 +131,20 @@ sub ExplodeText {
       split('', $string)
     : ()); }
 
-my $UNTEX_LINELENGTH = 78;    # [CONSTANT]
+our $UNTEX_LINELENGTH = 78;    # [CONSTANT]
 
 sub UnTeX {
   my ($thing, $suppress_linebreak) = @_;
   return unless defined $thing;
+  # Linebreak suppression could be a bit misleading for third-party users:
+  # even if the $suppress_linebreak argument flag is passed in,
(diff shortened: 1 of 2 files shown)


dginev:
And (obviously) suggestions for better naming are quite welcome.

brucemiller:
Not pretty, but I guess it's consistent with the mangle that evolved up to now :>

Issue #2025, opened 2022-12-12
MikTeX kpsewhich "maintenance" output breaks pathname_kpsewhich


xworld21:
It seems like kpsewhich from MIkTeX prints stuff on the *standard output* when installing a package on the fly:
```

======[not sure how many!]
starting package maintenance...
installation directory: "C:\..."
...
======[...]
C:\....
```
which of course breaks `pathname_kpsewhich`.

If anybody reading this knows the 'official' behaviour of MikTeX, please let me know, because I can't find any documentation about it! In the meanwhile, I'll send a mildly unsatisfying PR that ignores the lines between `/^=+$/`, as well as empty lines.

brucemiller:
Fixed upstream (?)

PR #2026, opened 2022-12-12
ignore MikTeX maintenance info

xworld21:
Not clean, but not dirty either, fix for #2025

### DIFF
--- a/lib/LaTeXML/Util/Pathname.pm
+++ b/lib/LaTeXML/Util/Pathname.pm
@@ -406,6 +406,8 @@ sub pathname_kpsewhich {
     close($resfh);             # ignore exit status (only one of @candidates exists, usually)
     if ($result) {
       chomp $result;
+      my ($vol, $dir, $name) = File::Spec->splitpath($result);
+      LaTeXML::Common::Error::Warn('unexpected', "$name", 'pathname_kpsewhich', "not among the requested files (@candidates)") if !grep { $_ eq $name } @candidates;
       return $result; } }
   return; }
 


brucemiller:
I haven't dug into what's happening, but the test failures aren't reassuring.

xworld21:
Just tried with nixpkgs (perl 5.34.1, Tex Live 2021 final) and it passes!

Is there any way to extract more detailed logs from the CI tests?

dginev:
I have cancelled and marked for rerun the two failing CI jobs, on the off chance it is some transient error.

But the code is indeed scarier than ideal... 

I am almost tempted to instead suggest a "documentation fix" by recommending Windows users of latexml to first build to PDF, letting MikTeX install everything, and only then rebuild with latexml.

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -410,16 +410,24 @@ sub pathname_kpsewhich {
...
-    my $result = <$resfh>;     # we only need the first line
+    my $ignore = 0;
+    my $result;
+    while (<$resfh>) {
+      chomp;
+      if (/^=+$/) {    # ignore MikTeX maintenance info between ==== lines
+        $ignore = !$ignore;
+        next; }
+      if (!$ignore && $_) {
+        $result = $_;    # we only need the first non-empty result
```
dginev, line 421: General note: the less you use the implicit `$_` (or other Perl implicits) in latexml code, the better.

It's very easy to have strange behavior slip through unnoticed in such cases.

dginev:
I could reproduce the failure locally from your branch:

<details>

```
$ perl -Iblib/lib blib/script/latexml t/complex/si.tex

latexml (LaTeXML version 0.8.6; revision d587043c) processing t/complex/si.tex
Package: expl3
Error:misdefined:# The token T_PARAM[#] should never reach Stomach! at expl3-code.tex; line 12585 col 9
Error:unexpected:\fi: Didn't expect a T_CS[\fi:] since we seem not to be in a conditional at expl3-code.tex; line 12588 col 0
Warning:expected:<number> Missing number, treated as zero at expl3-code.tex; line 12602 col 3
Error:undefined:\s__prg_mark The token T_CS[\s__prg_mark] is not defined. at expl3-code.tex; line 13240 col 0
Error:undefined:\q__prg_recursion_tail The token T_CS[\q__prg_recursion_tail] is not defined. at expl3-code.tex; line 13240 col 0
Error:undefined:\q__prg_recursion_stop The token T_CS[\q__prg_recursion_stop] is not defined. at expl3-code.tex; line 13240 col 0
Warning:expected:<number> Missing number, treated as zero at expl3-code.tex; line 13240 col 0
Error:misdefined:# The token T_PARAM[#] should never reach Stomach! at expl3-code.tex; line 15740 col 7
Error:misdefined:# The token T_PARAM[#] should never reach Stomach! at expl3-code.tex; line 15740 col 10
Error:misdefined:# The token T_PARAM[#] should never reach Stomach! at expl3-code.tex; line 15740 col 13
Error:undefined:\__fp_parse_one_register_dim:ww The token T_CS[\__fp_parse_one_register_dim:ww] is not defined. at expl3-code.tex; line 15744 col 37
Error:misdefined:# The token T_PARAM[#] should never reach Stomach! at expl3-code.tex; line 15744 col 39
Warning:expected:<number> Missing number, treated as zero at expl3-code.tex; line 24033 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 31919 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 31919 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 31930 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 31930 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 31974 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 31974 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 32013 col 0
Error:unexpected:} Attempt to close boxing group at expl3-code.tex; line 32013 col 0
```

</details>

Then I added explicit variable names, and exited the loop unconditionally on the first non-ignored line, which got tests to pass. I can't commit to your branch, but here is the snippet where I made changes:
```perl
    while (my $resline = <$resfh>) {
      chomp($resline);
      if ($resline =~ /^=+$/) {    # ignore MikTeX maintenance info between ==== lines
        $ignore = !$ignore;
        next; }
      if (!$ignore) {
        $result = $resline;    # we only need the first non-empty result
        last; } }
```

I can't tell what I improved with 100% certainty, except to guess that some of the implicit operations on `$_` had more subtle side-effects than anticipated.

lib/LaTeXML/Util/Pathname.pm
xworld21: > General note: the less you use the implicit `$_` (or other Perl implicits) in latexml code, the better.

All very true. I kind of feel embarrassed now. My money is on that `/^=+$/` test – I don't see what else could interfere with `$_`.

dginev:
A general thought I had initially here was:
"We are going to run a regex every time we call kpsewhich, really?"

Thinking that this will hurt performance, given how often we look for assets in the tex installation (e.g. for tikz and expl3).

But my next thought was - calling out to kpsewhich is already much slower than the regex cost, so it won't be noticeable.

In the end, I feel neutral about the change. My ideal fix would be to work with the MikTeX maintainers towards them providing a friendly programmatic interface for tools such as latexml. That is currently lacking, so we are put into the position of retrofitting a user-oriented command-line interface.

That said, my previous interactions with them were largely met with disinterest (maybe I was reading too much into the curt responses), so that's not a likely avenue forward.

That's my two cents, leaving the hard decision-making to Bruce.

xworld21:
> In the end, I feel neutral about the change. My ideal fix would be to work with the MikTeX maintainers towards them providing a friendly programmatic interface for tools such as latexml. That is currently lacking, so we are put into the position of retrofitting a user-oriented command-line interface.

I have just tried reporting the issue to MikTeX, asking to confirm if `===` checking is reliable, also if the output can be switched to standard error.

xworld21:
It seems like MikTeX is not affected by this (the issue https://github.com/MiKTeX/miktex/issues/1246#event-8844733139 was closed by a bot, but I confirm it's been resolved). So possible alternative, should LaTeXML print a warning when it sees a line of `=====`? Just so that users have a chance to know that something went wrong, and may need to rerun.

dginev:
Great to hear that the MikTeX setup has improved!

> should LaTeXML print a warning when it sees a line of =====? Just so that users have a chance to know that something went wrong, and may need to rerun.

I think we are generally fond of printing simple warning messages, especially with good cause. So I'd be in favor of the warning.

xworld21:
> I think we are generally fond of printing simple warning messages, especially with good cause. So I'd be in favor of the warning.

...and actually, we could print a warning whenever the file name returned by kpsewhich does not match the one requested. That should just never happen as far as I know. I'll update the PR.

brucemiller:
I may have overlooked some commentary, but as near as I can tell, the patch that the MikTeX folks applied simply removes the extra messages entirely.  If that's so, there're no "===" lines to detect and warn about.

xworld21:
What about the warning I put to replace the previous PR? Still helpful to detect if kpsewhich is returning nonsense, for whatever reason (most likely a non-recent MikTeX).

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -406,6 +406,8 @@ sub pathname_kpsewhich {
     close($resfh);             # ignore exit status (only one of @candidates exists, usually)
     if ($result) {
       chomp $result;
+      my ($vol, $dir, $name) = File::Spec->splitpath($result);
+      LaTeXML::Common::Error::Warn('unexpected', "$name", 'pathname_kpsewhich', "not among the requested files (@candidates)") if !grep { $_ eq $name } @candidates;
```
xworld21, line 410: I am not sure why `use LaTeXML::Common::Error` does not work, by the way. I don't always understand perl.

xworld21: > Presumably there will end up some error or warning higher up that the file couldn't be found?

I think the answer must be yes, or I wouldn't have found the source, but it was hard to debug. You only get an error about a file not existing and no explanation. Very puzzling when it happens.

brucemiller:
Sorry, this patch is pretty uncomfortable; there's got to be a better way (and apparently the root cause has been fixed upstream anyway, right?)

PR #2028, opened 2022-12-13
Mark mpadded with relative sizes as explicitly MathML Full

dginev:
I had an idea for a compromise solution for the mpadded piece of PR #2019 .

Namely, in the cases where we continue to emit an mpadded with relative sizes for width, height or depth, we can also add a class attribute stating "mathml-full".

This way we can leave the decision to the specific deployment how to handle these cases (if at all). I have tested that there is a CSS rule which can avoid the current Chrome "collisions" on relative widths, such as:

```css
mpadded.mathml-full {
  width: initial;
  height: initial;
}
```

which I am happy to add for the possible remaining cases in the millions of yet-to-be-vetted arXiv pages.

Does this sound reasonable?

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -546,11 +546,12 @@ sub pmml_maybe_resize {
       if (!$depth) {
         if ($yoff =~ /^-/) { $depth = $yoff; $depth =~ s/^-/+/; }
         else               { $depth = "-" . $yoff; } } }
-    $$attr{width}   = $width  if $width;
-    $$attr{height}  = $height if $height;
-    $$attr{depth}   = $depth  if $depth;
-    $$attr{lspace}  = $xoff   if $xoff;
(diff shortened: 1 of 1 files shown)


brucemiller:
This seems a bit desperate and doesn't quite meet the goal of Core.  A couple of points: the full-style attributes only come in when `$yoff` is set, not width, height or depth.  Secondly, in the specific case you're looking at, the shift is less than 1 point which is arguably ignorable in html.  And finally, I'm wondering (late in the game) whether adjusting height & depth is even needed/desired when setting voffset --- would help to have a more realistic case to  test on.

Lacking that test, I'd be tempted to just comment out the whole branch of `if($yoff){....}`, but not remove until we're sure.

dginev:
> Lacking that test, I'd be tempted to just comment out the whole branch of if($yoff){....}, but not remove until we're sure.

Certainly, that alternative is now in a new PR #2029

PR #2029, opened 2022-12-13
deprecate MathML Full use of relative height/depth attributes for yoff

dginev:
Alternative to #2028 per Bruce's suggestion.


brucemiller:
Jeez, didn't take you long to jump onto that.  I did find 1 case in DLMF that used this Feature:  1.12.E1, but it would appear that the adjustments to `height` & `depth` don't have any effect, so perhaps this is a safe approach after all!  Cool!

dginev:
Looks like I have good velocity when it comes to deprecations and removals.

If I could only get up to that speed for writing acceptable new code...

PR #2032, opened 2022-12-15
apply CSS to parts in article

teepeemm:
`\documentclass{article}` defines `\part` (just not `\chapter`).  We currently do not style the resulting html.  This PR applies the same styling to parts in articles that we apply to parts in books.


dginev (approved):
Curious, it does indeed!

And latexml appears to create the right numbering already, the only missing piece is this CSS change.

After experimenting with a small sample article, this PR looks like an easy merge to me.

Issue #2044, opened 2023-01-25
Test fails again w/ latest CTAN snapshot


hpreusse:
Unfortunately I have to tell you that the LaTeXML test suite again fails with an CTAN snapshot as of 2023/01/22. The test logs are [here](https://ci.debian.net/data/autopkgtest/testing/amd64/l/latexml/30688493/log.gz) and [here](https://qa.debian.org/excuses.php?package=texlive-base). Please be so kind to have a look at the issue.

dginev:
The `glossary` and `si` tests fail.

If you need an immediate resolution, we can move them to CI-only. What is the preferred time for packages like ours to get build failures resolved?

I must say that having *every* new texlive snapshot break the latexml package is very distracting.

dginev:
Thanks for the report @hpreusse ! I'ts actually very much appreciated.

Apologies if I was a bit curt, I'm just a bit disillusioned with the lack of resilience we have against texlive moving.

dginev:
Another suggestion would be to add a "max" version parameter and avoid having the installation fail in bleeding-edge builds.

e.g. add to the current condition 50_structure.t
```perl
    glossary   => {
      texlive_min => 2016,
      packages    => 'glossaries.sty' }
```

another line:
```perl
      texlive_max => 2022,
```

dginev:
As an example of a well-defended production system, consider arXiv:
https://arxiv.org/help/faq/texlive

who have moved from texlive 2009, to 2011, to 2016, to 2020. And texlive 2020 is what they use in 2023.

I am more than excited to have latexml issues targeting the latest bleeding edge of texlive (the date mentioned was 3 days ago, 2023/01/22), but it would be quite reassuring if the actual OS packaging of latexml was insulated against this level of update frequency.

These thoughts have made me like the `texlive_max` idea even more, as we can sit on a specific version for one or two years (with latexml packages building smoothly in that time), until we have had the time to resolve any issues with newer texlive variants. Given we are external to texlive, it would make sense for us to codify that we should be lagging behind their releases, rather than expect to stay ahead of them.

At least that's one view.

brucemiller:
But LaTeXML is a component tool, rather than a production system. It 
might be a production system only in combination with a TeX-live.
So if you want such a production system, you'll want to bundle latexml
with a texlive that you know cooperates.

A texlive_max limit on tests would only make the tests pass
(or, rather, be skipped). But the test would no longer be an indicator
[...]

PR #2045, opened 2023-01-27
only run glossary and SI tests in CI

dginev:
This PR has the quick consensus resolution for getting OS packages to build after the #2044 report.

I added a `CI_only => 1` flag in the known test requirement fields, since we want to skip individual files, rather than the full `50_structure.t` or `80_complex.t` bundles.

Hopefully this looks good.

### DIFF
--- a/lib/LaTeXML/Util/Test.pm
+++ b/lib/LaTeXML/Util/Test.pm
@@ -86,13 +86,15 @@ sub check_requirements {
     next unless $reqmts;
     my @required_packages = ();
     my $texlive_min       = 0;
+    my $CI_only           = 0;
     if (!(ref $reqmts)) {
       @required_packages = ($reqmts); }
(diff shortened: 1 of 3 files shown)


dginev:
Any comments before merging?

brucemiller:
Well, I hate that we need it, but I guess we do. And, the tests you've blocked are specific to a particular Issue, rather than marking all things likely to fail when texlive changes.

That said, we should probably switch the expl3 & grammar_coverage tests to use the same mechanism (two ways to do the same thing will lead to confusion).

And it would be preferable (if possible) to have an allowed-to-fail concept; we've got two quite different reasons for avoiding tests. Tests that take too long (like grammar, expl3?) make sense to defer for CI; but the tests in this PR are different (and likely other tests may want to be added).

dginev:
This PR has outlived its usefulness as well I believe.

PR #2048, opened 2023-02-01
transitional state of columnspan + colspan synonym attributes

dginev:
It is very easy to make the PR, although the outcome looks rather silly.

Tracking https://github.com/w3c/mathml-core/issues/180 , we *could* emit both variants of the `columnspan` attribute, until the newer `colspan` is available cross-browser.

Currently even with `colspan` set, the Chrome 109 rendering will *not* use the attribute, so this PR is likely to only become useful in Q2 of 2023 (or later)


dginev:
Chrome intends to ship the original MathML names now, so we can avoid the due diligence on our end (given that we haven't acted on this yet):
https://groups.google.com/a/chromium.org/g/blink-dev/c/SqIQf6nKxxk/m/Gc-9fBVAAQAJ

PR #2056, opened 2023-03-08
Make ltx:text->span add class=ltx_inline-block if given a width

brucemiller:
When an `ltx:text` is given a `width` attribute (eg from `\makebox`), when it is converted to html's `span`, the width gets ignored since CSS doesn't allow width on inline elements. 

It would be tempting to have such constructors to create an `ltx:inline-block` instead of an `ltx:text`, but that also has a different content model and so introduces seemingly extraneous `ltx:p` elements, which seem unnatural.

This PR shifts the burden to the XSLT, since it is really CSS that has the "problem": When converting such an `ltx:text` to a `span`, if it has a `@width`, also add the class `ltx_inline-block`.   This seems to solve the immediate problem.

It also reveals a different one: that LaTeXML's sizes, which appear to be accurate within TeX's context, when re-expressed in HTML pts, with different, unknown, fonts, tends to become wrong.  I suspect that we really should be mapping to `rem` (not `em`, since that varies with context).  But `rem` is unnatural outside of a CSS context (it seems to me).  So this will need some experimentation.

It could also be (or not) that `rem` is a better solution to the SVG measurement mismatch?

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-inline-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-inline-xhtml.xsl
@@ -36,7 +36,9 @@
     <xsl:element name="span" namespace="{$html_ns}">
       <xsl:variable name="innercontext" select="'inline'"/><!-- override -->
       <xsl:call-template name="add_id"/>
-      <xsl:call-template name="add_attributes"/>
+      <xsl:call-template name="add_attributes">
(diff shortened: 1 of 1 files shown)


dginev (approved):
Both the general direction and the code in the PR sound highly reasonable to me.

Switching to an internal model that is centered on `rem`, and then converting back out of `rem` to say `pt` or `px` in contexts that have no CSS capability seems reasonable to me long-term.

brucemiller:
I guess I kept this open as a reminder to think about `rem`, but that'll be something for later.

Issue #1995, opened 2022-11-28
\rule should be on the baseline


teepeemm:
The document
```tex
\documentclass{article}
\begin{document}
A\rule{1cm}{1pt}A
\end{document}
```
produces the pdf
<img width="91" alt="Screen Shot 2022-11-28 at 9 55 31 AM" src="https://user-images.githubusercontent.com/13021169/204322749-bd19177e-1648-4762-bc2a-24d6d9700c7e.png">
but the html
<img width="71" alt="Screen Shot 2022-11-28 at 9 52 03 AM" src="https://user-images.githubusercontent.com/13021169/204322025-305dcee5-3ffc-4fcb-927e-962eb847e115.png">

I think this is a matter of changing the styling of ltx:rule within `lib/LaTeXML/resources/XSLT/LaTeXML-inline-xhtml.xsl`

teepeemm:
The PR resolved this issue.

I was wondering why my PRs weren't linking to the issues I was claiming they were resolving, and found that the magic words are "fix", "resolve", and "close", along with their "-es", and "-ed" variations.  But I tended to use "-ing" in my descriptions, which doesn't count: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

PR #2061, opened 2023-03-31
align rules to baseline

teepeemm:
This aligns rules to the baseline, resolving #1995.  It also provides default dimensions (otherwise, vrules end up with a width of text-indent).  Some other notes:
* hrule should be in vertical mode but isn't.  (I think that's the correct terminology: TeX put the hrule on its own line and subsequent text on another line; LaTeXML put everything on the same line.)
* I don't appear to have html.sty, so I'm not sure what html.sty.ltxml is trying to emulate.
* TeX.pool.ltxml claims that hrule and vrule can appear [as rules within an alignment/table](https://github.com/brucemiller/LaTeXML/blob/c3aea809d9b677b8237103f9166dfb43328e5d0f/lib/LaTeXML/Package/TeX.pool.ltxml#L2353).  I was unable to see this behavior in LaTeXML: all tabular rules were rendered via a ltx:td/@ltx:border attribute, not a ltx:rule element.  Did I overlook some construction that would have resulted in ltx:rule?

The tex code I was using to test this out:

```tex
\documentclass{article}
\usepackage[lined]{algorithm2e}
\usepackage{xcolor}
\usepackage{latexml}
\iflatexml\usepackage{html}\fi
\begin{document}
A\rule{1cm}{.4pt}A
B\vrule height 1cm\relax B

C\hrule width 5cm\relax C

D\boxframe{5cm}{.1pt}{1cm}D

\begin{algorithm}[H]
  \If{understand}{go to next section\;}
\end{algorithm}

\iflatexml
E\htmlrule E
F\htmlrule*F
G\HTMLrule G
H\HTMLrule*H
\fi

\end{document}
```

### DIFF
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -377,6 +377,8 @@ span.ltx_rowspan { position:absolute; top:0; bottom:0; }
 .ltx_framed_leftright { border-left-style:solid; border-left-width:1px;
                         border-right-style:solid; border-right-width:1px; }
 
+.ltx_rule { vertical-align: bottom; height: 0.4pt; width: 0.4pt; }
+
 /*======================================================================
  Misc */
 /* .ltx_verbatim*/


brucemiller:
Works a charm; Thanks!

BTW: Yes, the `\hrule` & '\vrule` in tables are handled specially (and are complicated); basically they disappear into attributes on the table cells, eventually ending up as specific CSS classes.

Issue #2041, opened 2023-01-13
too close marginpars overlap


teepeemm:
LaTeX takes care that consecutive margin pars are adjusted so that they don't overlap:
```tex
\documentclass{article}
\begin{document}
\hfill Some text\marginpar{first}\marginpar{second} some more text
\end{document}
```
compiles to:
<img width="197" alt="Screen Shot 2023-01-12 at 7 43 27 PM" src="https://user-images.githubusercontent.com/13021169/212218232-6c74a365-c2da-4551-a861-363a111e75a8.png">
and even provides a warning: `LaTeX Warning: Marginpar on page 1 moved.`

`latexmlc --css=LaTeXML-marginpar.css` does not do any adjusting:
<img width="457" alt="Screen Shot 2023-01-12 at 7 43 43 PM" src="https://user-images.githubusercontent.com/13021169/212218257-a72100bc-29f0-47d7-9a52-a29826265d3f.png">

I'm not quite sure how we could fix this.  Javascript could space them out easily enough, but that requires Javascript.  It would be possible to adjust the DOM and CSS to place consecutive margin pars into a single margin par that includes line breaks.  But a long margin par followed by intervening text (whose height changes as the window resizes) followed by another margin par would be harder to deal with.

teepeemm:
The PR resolved this issue.

I was wondering why my PRs weren't linking to the issues I was claiming they were resolving, and found that the magic words are "fix", "resolve", and "close", along with their "-es", and "-ed" variations. But I tended to use "-ing" in my descriptions, which doesn't count: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

PR #2068, opened 2023-04-13
floating marginpar

teepeemm:
This allows marginpars to float (in the CSS sense), so that they display in order instead of overlapping, fixing #2041.  LaTeXML-marginpar.css doesn't appear to mesh well with LaTeXML-navbar-left.css (for example `.ltx_page_main { width }` is determined by which is loaded last), so I've not tried to resolve that.  On the other hand, if someone is loading both, then they will probably want their own css file anyway.

The tex code I was using to test this out:

```tex
\documentclass{article}
\usepackage{latexml}
\begin{lxNavbar}
\lxContextTOC
\end{lxNavbar}
\begin{document}
\section{Section Title}
A small amount of text.\marginpar{first}\marginpar{second second second second}
A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text. A large amount of text.
A large amount of text.\marginpar{third}\marginpar{fourth}
A small amount of text.
\end{document}
```

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3558,12 +3558,16 @@ DefMacro('\@dblfloatplacement', '');
 # C.9.2 Marginal Notes
 #======================================================================
 
-AssignValue(marginparclass => undef);
+DefConditional('\if@reversemargin');
+Let('\reversemarginpar', '\@reversemargintrue');
+Let('\normalmarginpar',  '\@reversemarginfalse');
 DefConstructor('\marginpar[]{}',
-  "<ltx:note role='margin' class='#class'><ltx:inline-para>#2</ltx:inline-para></ltx:note>",
-  properties => sub { (class => LookupValue('marginparclass')); });
-DefPrimitiveI('\reversemarginpar', undef, sub { AssignValue(marginparclass => 'ltx_marginpar_reverse'); });
-DefPrimitiveI('\normalmarginpar', undef, sub { AssignValue(marginparclass => undef); });
+  "<ltx:note role='margin' class='#class'><ltx:inline-para>#content</ltx:inline-para></ltx:note>",
+  properties => sub {
+    my $isReversed = IfCondition('\if@reversemargin');
+    my $class      = $isReversed             ? 'ltx_marginpar_reverse' : 'ltx_marginpar_normal';
+    my $content    = ($isReversed and $_[1]) ? $_[1]->toString         : $_[2]->toString;
+    return (class => $class, content => $content); });
 DefRegister('\marginparpush', Dimension(0));
 
 #**********************************************************************
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3562,12 +3562,9 @@ DefConditional('\if@reversemargin');
 Let('\reversemarginpar', '\@reversemargintrue');
 Let('\normalmarginpar',  '\@reversemarginfalse');
 DefConstructor('\marginpar[]{}',
-  "<ltx:note role='margin' class='#class'><ltx:inline-para>#content</ltx:inline-para></ltx:note>",
-  properties => sub {
-    my $isReversed = IfCondition('\if@reversemargin');
-    my $class      = 'ltx_marginpar_' . ($isReversed ? 'reverse' : 'normal');
-    my $content    = ($isReversed and $_[1]) ? $_[1] : $_[2];
-    return (class => $class, content => $content); });
+"?#1(<ltx:note role='margin' class='ltx_marginpar_left'><ltx:inline-para>#1</ltx:inline-para></ltx:note>"
+    . "?#2(<ltx:note role='margin' class='ltx_marginpar_right'><ltx:inline-para>#2</ltx:inline-para></ltx:note>))"
+    . "(<ltx:note role='margin' class='ltx_marginpar'><ltx:inline-para>#2</ltx:inline-para></ltx:note>)");
 DefRegister('\marginparpush', Dimension(0));
 
 #**********************************************************************
@@ -4708,11 +4705,11 @@ DefConstructor('\@framebox[Dimension][]{}',
     $stomach->beginMode('text');
     AssignValue(FRAME_IN_MATH => $wasmath); },
   properties => sub {
-    my $sep= LookupRegister('\fboxsep')->toAttribute;
+    my $sep = LookupRegister('\fboxsep')->toAttribute;
     (($_[2] ? (align => $makebox_alignment{ ToString($_[2]) }) : ()),
       framecolor => LookupValue('font')->getColor,
-     ($sep ne '3.0pt' ? (cssstyle   => 'padding:' . $sep):()),
-      ($_[1] ? (width => $_[1]) : ())); },
+      ($sep ne '3.0pt' ? (cssstyle => 'padding:' . $sep) : ()),
+      ($_[1]           ? (width    => $_[1])             : ())); },
   afterDigest => sub {
     my ($stomach, $whatsit) = @_;
     my $wasmath = LookupValue('FRAME_IN_MATH');
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -3565,8 +3565,8 @@ DefConstructor('\marginpar[]{}',
   "<ltx:note role='margin' class='#class'><ltx:inline-para>#content</ltx:inline-para></ltx:note>",
   properties => sub {
     my $isReversed = IfCondition('\if@reversemargin');
-    my $class      = $isReversed             ? 'ltx_marginpar_reverse' : 'ltx_marginpar_normal';
-    my $content    = ($isReversed and $_[1]) ? $_[1]->toString         : $_[2]->toString;
+    my $class      = 'ltx_marginpar_' . ($isReversed ? 'reverse' : 'normal');
+    my $content    = ($isReversed and $_[1]) ? $_[1] : $_[2];
     return (class => $class, content => $content); });
 DefRegister('\marginparpush', Dimension(0));
 
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
@@ -403,13 +403,7 @@
   <xsl:template match="/" mode="body-main">
     <xsl:text>&#x0A;</xsl:text>
     <xsl:element name="div" namespace="{$html_ns}">
-      <xsl:attribute name="class">
-          <xsl:text>ltx_page_main</xsl:text>
-          <xsl:if test="//ltx:note[@role='margin'][@class='ltx_marginpar_reverse']"
-              > ltx_has_marginpar_reverse</xsl:if>
-          <xsl:if test="//ltx:note[@role='margin'][@class='ltx_marginpar_normal']"
-              > ltx_has_marginpar_normal</xsl:if>
-      </xsl:attribute>
+      <xsl:attribute name="class">ltx_page_main</xsl:attribute>
       <xsl:apply-templates select="." mode="body-main-begin"/>
       <xsl:apply-templates select="." mode="header"/>
       <xsl:apply-templates select="." mode="body-content"/>
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
@@ -403,7 +403,13 @@
   <xsl:template match="/" mode="body-main">
     <xsl:text>&#x0A;</xsl:text>
     <xsl:element name="div" namespace="{$html_ns}">
-      <xsl:attribute name="class">ltx_page_main</xsl:attribute>
+      <xsl:attribute name="class">
+          <xsl:text>ltx_page_main</xsl:text>
+          <xsl:if test="//ltx:note[@role='margin'][@class='ltx_marginpar_reverse']"
+              > ltx_has_marginpar_reverse</xsl:if>
+          <xsl:if test="//ltx:note[@role='margin'][@class='ltx_marginpar_normal']"
+              > ltx_has_marginpar_normal</xsl:if>
+      </xsl:attribute>
       <xsl:apply-templates select="." mode="body-main-begin"/>
       <xsl:apply-templates select="." mode="header"/>
       <xsl:apply-templates select="." mode="body-content"/>
--- a/lib/LaTeXML/resources/CSS/LaTeXML-marginpar.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML-marginpar.css
@@ -2,10 +2,13 @@
   (more sophisticated layout probably requires xslt) */
 .ltx_note.ltx_role_margin .ltx_note_mark { display:none; }
 .ltx_note.ltx_role_margin .ltx_note_type { display:none; }
+.ltx_note.ltx_role_margin {
+    margin-right: -15%; width:15%;
+    float: right; clear: right; }
 .ltx_note.ltx_role_margin .ltx_note_content {
     display:block;
-    position:absolute; left:83%; width:15%;
+    position:relative; right:-15%;
     background-color: transparent; border:0pt; }
 
 /* Narrower, to make room for marginpar */
-.ltx_page_main { width:80%; }
+.ltx_page_main { width:75%; }
--- a/lib/LaTeXML/resources/CSS/LaTeXML-marginpar.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML-marginpar.css
@@ -2,13 +2,23 @@
   (more sophisticated layout probably requires xslt) */
 .ltx_note.ltx_role_margin .ltx_note_mark { display:none; }
 .ltx_note.ltx_role_margin .ltx_note_type { display:none; }
-.ltx_note.ltx_role_margin {
-    margin-right: -15%; width:15%;
-    float: right; clear: right; }
+.ltx_note.ltx_role_margin { width:15%; }
 .ltx_note.ltx_role_margin .ltx_note_content {
-    display:block;
-    position:relative; right:-15%;
-    background-color: transparent; border:0pt; }
+    display:block; position:relative;
+    background-color:transparent; border:0pt; }
+.ltx_note.ltx_role_margin.ltx_marginpar_reverse {
+    margin-left:-15%; float:left; clear:left; }
+.ltx_note.ltx_role_margin.ltx_marginpar_normal {
+    margin-right:-15%; float:right; clear:right; }
+.ltx_note.ltx_role_margin.ltx_marginpar_reverse .ltx_note_content { left:-15%; }
+.ltx_note.ltx_role_margin.ltx_marginpar_normal .ltx_note_content { right:-15%; }
 
-/* Narrower, to make room for marginpar */
-.ltx_page_main { width:75%; }
+/* Narrower, to make room for marginpar.
+ * We don't want to change .ltx_page_main, because navbar chnages that and
+ * these would interfere. */
+.ltx_page_main.ltx_has_marginpar_reverse .ltx_page_content {
+    left:15%; position:relative; }
+.ltx_page_main.ltx_has_marginpar_normal .ltx_page_content,
+.ltx_page_main.ltx_has_marginpar_reverse .ltx_page_content { width:85%; }
+.ltx_page_main.ltx_has_marginpar_normal.ltx_has_marginpar_reverse .ltx_page_content {
+    width:70%; }
--- a/lib/LaTeXML/resources/CSS/LaTeXML-marginpar.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML-marginpar.css
@@ -6,21 +6,19 @@
 .ltx_note.ltx_role_margin .ltx_note_content {
     display:block; position:relative;
     background-color:transparent; border:0pt; }
-.ltx_note.ltx_role_margin.ltx_marginpar_reverse {
-    margin-left:-15%; float:left; clear:left; }
-.ltx_note.ltx_role_margin.ltx_marginpar_normal {
-    margin-right:-15%; float:right; clear:right; }
-.ltx_note.ltx_role_margin.ltx_marginpar_reverse .ltx_note_content { left:-15%; }
-.ltx_note.ltx_role_margin.ltx_marginpar_normal .ltx_note_content { right:-15%; }
 
 /* Narrower, to make room for marginpar.
  * We don't want to change .ltx_page_main, because navbar changes that and
- * these would interfere. */
-.ltx_page_main:has(.ltx_marginpar_reverse) .ltx_page_content {
-    left:15%; position:relative; }
-/* if there's either one */
-.ltx_page_main:has(.ltx_marginpar_normal,.ltx_marginpar_reverse) .ltx_page_content {
-    width:85%; }
-/* if there's both */
-.ltx_page_main:has(.ltx_marginpar_normal):has(.ltx_marginpar_reverse) .ltx_page_content {
-    width:70%; }
+ * this would interfere. */
+.ltx_page_main .ltx_page_content { width:85%; }
+
+.ltx_note.ltx_role_margin.ltx_marginpar,
+.ltx_note.ltx_role_margin.ltx_marginpar_right {
+    margin-right:-15%; float:right; clear:right; }
+.ltx_note.ltx_role_margin.ltx_marginpar .ltx_note_content,
+.ltx_note.ltx_role_margin.ltx_marginpar_right .ltx_note_content { right:-15%; }
+.ltx_note.ltx_role_margin.ltx_marginpar_left { display:none; }
+
+/* For left marginpar:
+ * swap all occurences of 'left' and 'right' and add
+ * .ltx_page_main .ltx_page_content { left:15%; position:relative; } */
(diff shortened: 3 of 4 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -3547,12 +3547,16 @@ DefMacro('\@dblfloatplacement', '');
 # C.9.2 Marginal Notes
 #======================================================================
 
-AssignValue(marginparclass => undef);
+DefConditional('\if@reversemargin');
+Let('\reversemarginpar', '\@reversemargintrue');
+Let('\normalmarginpar',  '\@reversemarginfalse');
```
brucemiller, line 3552: nice!

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -3547,12 +3547,16 @@ DefMacro('\@dblfloatplacement', '');
...
 DefConstructor('\marginpar[]{}',
-  "<ltx:note role='margin' class='#class'><ltx:inline-para>#2</ltx:inline-para></ltx:note>",
-  properties => sub { (class => LookupValue('marginparclass')); });
-DefPrimitiveI('\reversemarginpar', undef, sub { AssignValue(marginparclass => 'ltx_marginpar_reverse'); });
-DefPrimitiveI('\normalmarginpar', undef, sub { AssignValue(marginparclass => undef); });
+  "<ltx:note role='margin' class='#class'><ltx:inline-para>#content</ltx:inline-para></ltx:note>",
+  properties => sub {
+    my $isReversed = IfCondition('\if@reversemargin');
+    my $class      = $isReversed             ? 'ltx_marginpar_reverse' : 'ltx_marginpar_normal';
+    my $content    = ($isReversed and $_[1]) ? $_[1]->toString         : $_[2]->toString;
```
brucemiller, line 3558: You'll want to omit those `->toString` conversions, and just store whichever argument directly in `$content`, so that it will be converted correctly to XML.

lib/LaTeXML/resources/XSLT/LaTeXML-webpage-xhtml.xsl
```diff
@@ -403,7 +403,13 @@
   <xsl:template match="/" mode="body-main">
     <xsl:text>&#x0A;</xsl:text>
     <xsl:element name="div" namespace="{$html_ns}">
-      <xsl:attribute name="class">ltx_page_main</xsl:attribute>
+      <xsl:attribute name="class">
+          <xsl:text>ltx_page_main</xsl:text>
+          <xsl:if test="//ltx:note[@role='margin'][@class='ltx_marginpar_reverse']"
+              > ltx_has_marginpar_reverse</xsl:if>
```
brucemiller, line 409: Isn't there a `has()` CSS directive now that could be used?

Maybe @dginev has some comments on the CSS.

dginev: [:has](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) is still missing from Firefox, so the recommendation I've heard is to use it for "progressive enhancement" only until it arrives. There is a single experiment in ar5iv with it, but I am mostly waiting for Firefox to ship it...

brucemiller: Well, maybe we can avoid this `ltx_has_marginpar` complication with the assumption that the author will setup page dimensions to accommodate margins, whether or not there are marginpars? (you do in ar5iv).

teepeemm: I'd like to have some styling so that there would be a possibility that a user could use the default style file.  It does appear that [Firefox will implement :has soon](https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idi-p/23794), although that thread starts with "first half of 2023".  Maybe go with :has anyway?  That should be fully functional soon enough, and even if not, it would show the user what their style file could have.

teepeemm:
Grr.

The real kicker is that I can't commit the actual changes because latexmllint says
```
lib/LaTeXML/Package/LaTeX.pool.ltxml  has syntax issues
  Bareword "CC_ESCAPE" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_BEGIN" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_END" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_MATH" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ALIGN" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_EOL" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_PARAM" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SUPER" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SUB" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SPACE" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SPACE" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_COMMENT" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ACTIVE" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ESCAPE" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ACTIVE" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_LETTER" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 114.
  Bareword "CC_LETTER" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 115.
  Bareword "CC_LETTER" not allowed while "strict subs" in use at LaTeXML/blib/lib/LaTeXML/Core/State.pm line 118.
```
but I've not touched State.pm, so I'm not sure why that happened.  Any idea what's going on?  Any idea how I can fix the git commits to be more reasonable?

dginev:
@teepeemm Ah this is unfortunate, this branch has diverged. I generally never use git `merge` exactly since it can be quite treacherous when it misfires - I get better mileage out of`rebase`.

You can branch off from master and cherry-pick individual commits from here, making a new PR, or use this branch by rewinding back to before your first git merge. (git checkout to the right commit, then a `git push --force` would yank out the merges).

I always do a `git fetch upstream; git rebase upstream/master` myself, resolving any potential conflicts and then `git push --force origin/mybranch` to update the PR.

dginev:
Btw @brucemiller , it looks like latexmllint is indeed failing on the master branch, as Tim reports here.

I am not sure how to convince perltidy to recognize constants, currently it complains:

```
$ perl tools/latexmllint lib/LaTeXML/Core/State.pm

lib/LaTeXML/Core/State.pm  has syntax issues
  Bareword "CC_ESCAPE" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_BEGIN" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_END" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_MATH" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ALIGN" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_EOL" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_PARAM" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SUPER" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SUB" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SPACE" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_SPACE" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_COMMENT" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ACTIVE" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ESCAPE" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_ACTIVE" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 108.
  Bareword "CC_LETTER" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 114.
  Bareword "CC_LETTER" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 115.
  Bareword "CC_LETTER" not allowed while "strict subs" in use at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/State.pm line 118.
  Compilation failed in require at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/Token.pm line 22.
  BEGIN failed--compilation aborted at /home/deyan/git/my-LaTeXML/tools/../blib/lib/LaTeXML/Core/Token.pm line 22.
  Compilation failed in require at lib/LaTeXML/Core/State.pm line 17.
  BEGIN failed--compilation aborted at lib/LaTeXML/Core/State.pm line 17.
lib/LaTeXML/Core/State.pm fails commit criterion
```

dginev:
As a qucik experiment, I moved the catcode constants to a new `LaTeXML::Core::Catcode` which avoids the circular dependencies, and perltidy seemed happier. Something to consider, @brucemiller ?

<details>

say in `LaTeXML/Core/Catcode.pm`

```perl
use base qw(Exporter);
our @EXPORT = (
  # Catcode constants
  qw( CC_ESCAPE  CC_BEGIN  CC_END     CC_MATH
    CC_ALIGN   CC_EOL    CC_PARAM   CC_SUPER
    CC_SUB     CC_IGNORE CC_SPACE   CC_LETTER
    CC_OTHER   CC_ACTIVE CC_COMMENT CC_INVALID
    CC_CS      CC_MARKER CC_ARG)
);

use constant CC_ESCAPE  => 0;
use constant CC_BEGIN   => 1;
use constant CC_END     => 2;
use constant CC_MATH    => 3;
use constant CC_ALIGN   => 4;
use constant CC_EOL     => 5;
use constant CC_PARAM   => 6;
use constant CC_SUPER   => 7;
use constant CC_SUB     => 8;
use constant CC_IGNORE  => 9;
use constant CC_SPACE   => 10;
use constant CC_LETTER  => 11;
use constant CC_OTHER   => 12;
use constant CC_ACTIVE  => 13;
use constant CC_COMMENT => 14;
use constant CC_INVALID => 15;
# Extended Catcodes for expanded output.
use constant CC_CS     => 16;
use constant CC_MARKER => 17;    # non TeX extension!
use constant CC_ARG    => 18;    # "out_param" in B Book

# [The documentation for constant is a bit confusing about subs,
# but these apparently DO generate constants; you always get the same one]
# These are immutable
use constant T_BEGIN => bless ['{',  CC_BEGIN], 'LaTeXML::Core::Token';
use constant T_END   => bless ['}',  CC_END],   'LaTeXML::Core::Token';
use constant T_MATH  => bless ['$',  CC_MATH],  'LaTeXML::Core::Token';
use constant T_ALIGN => bless ['&',  CC_ALIGN], 'LaTeXML::Core::Token';
use constant T_PARAM => bless ['#',  CC_PARAM], 'LaTeXML::Core::Token';
use constant T_SUPER => bless ['^',  CC_SUPER], 'LaTeXML::Core::Token';
use constant T_SUB   => bless ['_',  CC_SUB],   'LaTeXML::Core::Token';
use constant T_SPACE => bless [' ',  CC_SPACE], 'LaTeXML::Core::Token';
use constant T_CR    => bless ["\n", CC_SPACE], 'LaTeXML::Core::Token';
```

</details>

teepeemm:
@dginev I think I managed to resolve the conversations and get the commit history back to looking reasonable, so that this is ready to review and merge.  Let me know if I need to do anything else.

dginev:
Nice, thank you @teepeemm .

Hm, so I made a quick exploration into how the optional argument of \marginpar is used, and I wonder if we aren't creating a mildly misleading feature by respecting it. Generally, latexml does not keep page counts, and does not have a reliable distinction between being on an "even" or "odd" page (unless I am missing something new?). As an example, in the PDF you would expect this simple \marginpar call to use its different arguments on even/odd pages:

```tex
\documentclass{book}
\def\greeting{\marginpar[even greetings]{odd greetings}}
\begin{document}

\greeting
\clearpage
\greeting

\end{document}
```

I wonder if we the Perl binding should only ever use the mandatory (odd page) argument for marginpar, to avoid ever preserving a note expected for the left margin in the XML. 
As my usual example, a site such as ar5iv only ever uses the right margin (of a sufficiently wide viewport) for showing notes, so in the \marginpar sense all pages appear to be "odd" there.

What do you think?

teepeemm:
Why can't we alternate left and right marginpar reliably?  Is there an arXiv document that uses `\reversemarginpar` that we could experiment with?  My guess is that `\reversemarginpar` isn't used all that often, so deactivation to `\relax` may not be necessary.

dginev:
> Why can't we alternate left and right marginpar reliably? 

We don't even have the concept of pages in HTML, but we kind of do in ePub. Seen differently - say we emit an ePub and show it on a fixed page view of an eink reader. The concrete page numbers will depend on the resolution and size of the display. So a margin note which appears at the top of page `2k+1` on a 10 inch screen may be at the bottom of page `2q` on a 7 inch screen.

> Is there an arXiv document that uses \reversemarginpar that we could experiment with?

I can try to find one, but the server hosting the documents is down for maintenance this week. Will dig through my local copy...

brucemiller:
I don't think we've quite captured the intention of `\marginpar`, and are maybe drifting away from it.  The basic idea is to put a note in the "outside" margin, opposite the usually narrower gutter where the pages are sewn or between columns.   No counter or dagger marker is needed, since the note should appear directly alongside the place the note was introduced. The optional argument allows you to provide an alternate text for when the note appears on the left (for whatever reason that it appears there), in case there is some directionality in the text (eg. an arrow pointing to the text?).  The `\reversemarginpar` says to place the note on the "inner" margin (except double column); it doesn't (directly) choose which note to use.  I'd expect both the optional argument and reversing to be relatively rare, used only for production page design & layout.

Understanding that, we can now ask how that should be carried over to XML/HTML, where we don't (necessarily) have pages, columns, or even margins --- but we might.  The lowest level would be latexml's original behavior to synthesize a marker (dagger) and have the note appear as a popup.  The next level would be to build-in a wide margin, either on left or right and place the notes in that margin, without marker, alongside its source (choosing the optional argument as appropriate).  The `\reversemarginpar` would be relevant at most for choosing which side you're creating the wide margin.

For a truly responsive solution, however, you may have variable margins, may have multiple columns, and you won't know till you detect the device & the user resizes the window.  So the XML really should convert & to store *both* versions of the note,  with appropriate attributes/classes, and defer the marker, display & positioning as late as possible to CSS (& javascript if needed).  In this scenario the optional argument would be used, but `\reversemarginpar` is not really meaningful.

teepeemm:
I'm not seeing a purely CSS way to determine if a marginpar has appeared in the left or right margin, so I think  it would need to involve Javascript for the truly responsive solution.

I think this PR does accomplish the "next level" rendering.

`doc.sty` (often used for typesetting tex documentation) does use `\reversemarginpar` and `\marginpar{term}` (along with a much wider left margin).  But I do agree that reversing is in general rare, and switching back and forth within a document especially so.

brucemiller:
I'm just not so clear that this is actually an improvement (yet); `\reversemarginpar` seems incorrect; it's not used to select the right or left form of the note, it's used to select whether notes go on the inside or outside. That may seem like the same thing if the default is to put the note in the right margin, but it isn't quite.  Consider the margin note:
```
\marginnote[TaDah $\rightarrow$]{$leftarrow$ TaDah}
```
The first arg should *never* appear in the right margin, the 2nd arg should *never* appear in the left margin.

So, I'm thinking if it's worth handling the optional arg at all, we should be storing two notes, with appropriate roles or classes and using CSS; for purpose of argument `role=margin-left|margin-right`.  I think the point with CSS is not to determine if a specific note appeared in the left or right, but rather to use CSS (and media queries, etc) to decide on the margin layout, and then to decide *which* versions of the notes will appear at all, and where.  Thus if you're doing notes in the left margin, you'd be activating the margin-left and generic margin, while the margin-right would get display none.

teepeemm:
But in html, we don't have pages, so everything acts like page 1, and outside <=> right and inside <=> left.  An ePub will have left and right pages, but then we need to figure out if we have a left page or a right page, which doesn't appear to be doable within CSS.

Without ePub and some reliable way to determine if it's a left or right page, then we can assume we have a right page.  And if we know that we have a right page in the output, then when converting to XML, we know which argument we want to use, and don't need to keep the other.

dginev:
> But in html, we don't have pages, so everything acts like page 1

That's if you are doing a right-margin CSS design. If you do a left-margin CSS design, everything acts like page 2 :>

A problem that hasn't gotten enough attention is legacy authored documents. In arXiv, an author may have *known for certain* that a specific `\marginpar` will appear on an odd page, and written it as `\marginpar[Important remark $\rightarrow$]{}`. For that case, irrespective of which CSS theme is used, the only usable argument is the optional `#1`. Similarly in the inverse case of `\marginpar[]{$\leftarrow$ Important remark}`.

With that limited application in mind, I'd be happy with a solution that uses arg `#2` if non-empty, otherwise `#1` if non-empty, or do nothing if both are left empty. And to ignore all other subtleties needed for "two-sided layout" in PDF, leaving us with a single note to style in the HTML. Of course we *could* open a large project working on "two-sided layout" in HTML+CSS, but I'm not sure it's the right time to start on such an adventure...

brucemiller:
The point is to create a passably nice document by default, while preserving enough information and semantics to give people the flexibility to adapt the styling and layout for whatever medium they're using; webpages, cellphones, epub, etc.  So while a default assumption to make the entire document look like a right-hand page is perfectly reasonable, to build in the assumptions that that's the only thing you'll ever be able to do is not ok.

Folks ought to be able to easily create a CSS layout with the margin notes on the right *or* the left. As you point out left notes likely will interfere with a left navbar; so let people choose how it should be laid out.  On that basis, I'd be more inclined to think we should ignore `\reversemarginpar` entirely, since it really doesn't apply in most cases, or should be left to the top-level design choices.

In any case, the class `ltx_marginpar_reverse` is misnamed & confusing as you've interpreted it; you're using it to mark left marginpars.

While you're probably right that you can't determine whether the note is on the left or right from CSS, you've got the process backwards: it's the CSS that will be applying the styling and layout that makes it *be* a left or right hand page (eg. unsymmetric margins, or whatever); so the CSS can easily display or hide whichever notes it wants based on the same choices.

brucemiller:
@dginev tells me I wasn't very clear about what I was hoping for; it's pretty simple actually, just awkward to explain.
I'm thinking that `\marginpar` with no optional argument creates an `<ltx:note role="margin" class="ltx_marginpar"...`.  But if both arguments are given, it should create *two* `ltx:note`,  one with `class="ltx_marginpar_left"`, using the optional argument, the other with `class="ltx_marginpar_right"` using the required argument.

Then, if you want notes in the right margin, you'd use CSS to add some space on the right, and make both `ltx_marginpar` and `ltx_marginpar_right` be visible, positioned to the right, and make `ltx_marginpar_left` have `display:none`.  For left margins, you'd add the space on the left and position `ltx_marginpar` and `ltx_marginpar_left` notes to the left and make `ltx_marginpar_right` invisible.  And there's sufficient data there to do fancy things if you want to play more and add javascript.

The `\reversemarginpar` would basically be ignored as it (probably) doesn't correspond to anything you'd want to do in html.

Does that make more sense?

dginev:
I really like that proposal actually!

teepeemm:
I think that ends up as one of
```perl
DefConstructor('\marginpar[]{}',
  "?#1(?#2("
      ."<ltx:note role='margin' class='ltx_marginpar_left'><ltx:inline-para>#1</ltx:inline-para></ltx:note>"
      ."<ltx:note role='margin' class='ltx_marginpar_right'><ltx:inline-para>#2</ltx:inline-para></ltx:note>)"
    ."(<ltx:note role='margin' class='ltx_marginpar'><ltx:inline-para>#1</ltx:inline-para></ltx:note>))"
  ."(<ltx:note role='margin' class='ltx_marginpar'><ltx:inline-para>#2</ltx:inline-para></ltx:note>)");
# or
DefConstructor('\marginpar[]{}', sub {
    my ($document, $left, $right) = @_;
    if ($left) {
      $document->openElement('ltx:note', role => 'margin', 'class' => 'ltx_marginpar' . ($right ? '_left' : ''));
      $document->insertElement('ltx:inline-para', $left);
      $document->closeElement('ltx:note'); }
    if ($right) {
      $document->openElement('ltx:note', role => 'margin', 'class' => 'ltx_marginpar' . ($left ? '_right' : ''));
      $document->insertElement('ltx:inline-para', $right);
      $document->closeElement('ltx:note'); } });
```
Do you have a preference on which one?  I think using the `sub` is more readable, but it is 4 lines longer.

Somewhat related, but I'm having some sort of trouble compiling.  No matter what I do to `LaTeX.pool.ltxml`, it is somehow using the file from the prior commit, so I'm not able to test this out.  Do you know if this is a git thing, or a LaTeXML thing, or a Perl thing, or some strange thing that you can't quite tell from my vague description?  I've tried running `make`, as well as `git add` the file to be committed.  (Could it be relevant that `git status` keeps saying that the file has changes to be committed as well as changes not staged for commit?)

brucemiller:
I tend to prefer the xml pattern forms, when they're not too complex; plus you probably can assume that there will be an arg 2, so you should be able to use:
```
DefConstructor('\marginpar[]{}',
  "?#1("
      ."<ltx:note role='margin' class='ltx_marginpar_left'><ltx:inline-para>#1</ltx:inline-para></ltx:note>"
      ."<ltx:note role='margin' class='ltx_marginpar_right'><ltx:inline-para>#2</ltx:inline-para></ltx:note>)"
    ."(<ltx:note role='margin' class='ltx_marginpar'><ltx:inline-para>#2</ltx:inline-para></ltx:note>)");
  ```
make and running latexml normally shouldn't be affected by git; just what's in your working directory; Or the local directory where you're testing from.  Maybe there's a stray copy of `LaTeXML.pool.ltxml` sitting somewhere that LaTeXML is picking up instead of the one from git directory?  The log of your test should show where things were fetched from, for example
```
...
(Loading /home/bruce/LaTeXML/blib/lib/LaTeXML/Package/LaTeX.pool.ltxml...
...
```

brucemiller:
Oh, if a file has staged changes *and* local changes, that can make things like `git diff` confusing, but still it will be version in the working directory that latexml uses when it runs.

teepeemm:
Ah.  I'd copied LaTeX.pool.ltxml into my working directory so that my git fumblings wouldn't overwrite what I was trying to do.

Deyan did have a possible example where there would only be the optional argument: if the author was sure that the marginpar was going to be on an even page.  So I think we do need to watch out for that possibility.

brucemiller:
Well there has to be the required argument, although it may be empty. 
Was the author sure that the marginpar would be on the left? Or were 
they saying they didn't want any text, if it appeared on the right? Who 
knows; I'd saying such abuses deserve what they get (or don't get :> )

brucemiller:
Moreover, whatever the author had in mind, the text provided in the optional arg should *not* be used on the right margin and should still have `ltx_marginpar_left`, not `ltx_marginpar`.

teepeemm:
I believe this incorporates everything we've discussed.  If there's `#1`, it gets `.ltx_marginpar_left` and `#2` (if present) gets `.ltx_marginpar_right`.  Otherwise, there's only `#2`, and we just use `.ltx_marginpar` (this assumes no one will do a completely empty `\marginpar{}`).  By default, the CSS hides the `_left` version.

dginev (approved):
Cool! I haven't tested it on a real-world example, but the code looks solid.

brucemiller:
Nice! I think this will work well by default, as well as with the optional `LaTeXML-marginpar.css`, and still provide users the flexibility to do left marginpars or other fancy stuff if they wish.  Thanks!!

Issue #2027, opened 2022-12-12
ignored characters should not prevent a paragraph break


teepeemm:
The TeX:
```tex
\documentclass{article}
\begin{document}
one

two

{ % catcode 9 is ``ignored''
\catcode`\P=9
P three
P
P four
}
\end{document}
```
creates four paragraphs.  `latexml --nocomments` yields:
```xml
  <para xml:id="p1">
    <p>one</p>
  </para>
  <para xml:id="p2">
    <p>two</p>
  </para>
  <para xml:id="p3">
    <p>three

 four</p>
  </para>
```
and the latter two paragraphs have merged into one.  I suspect this means that when we're looking to see if we should have a paragraph break, we're only checking if the line matches `/^\s*$/`, and not ignoring characters that should be ignored.  But I'm not sure where in the code we make that determination.

teepeemm:
I've done some investigating. The following TeX results in 7 (counting the X) paragraphs:
```tex
\catcode`\I=9
\catcode`\S=10
3 initial line^^M nothing past the end
I
5 line after ignore catcoded character
S
7 line after space catcoded character
  
9 line after line with space characters
		
11\vadjust{X} line after line with tab characters

13 Another paragraph.

\bye
```
latexml only recognizes the paragraph break at lines 8 and 12 and puts the X in front of the 13 (the resulting whitespace for the other paragraph breaks goes into the xml source).  I think a fix would be to change Mouth.pm around line 320 from
```perl
      # In state N, skip spaces
      while (($$self{colno} < $$self{nchars})
        # DIRECT ACCESS to $STATE's catcode table!!!
        && ( (($$STATE{catcode}{ $$self{chars}[$$self{colno}] }[0] || CC_OTHER) == CC_SPACE) )) {
        $$self{colno}++; }
      # If upcoming line is empty, and there is no recognizable EOL, fake one
      return T_MARKER('EOL') if $read_mode
        && ($$self{colno} >= $$self{nchars}) && ((!defined $eolch) || ($eolch ne "\r"));
      # Sneak a comment out, every so often.
```
to
```perl
      # In state N, skip spaces
      while (($$self{colno} < $$self{nchars})
        # DIRECT ACCESS to $STATE's catcode table!!!
        && ( (($$STATE{catcode}{ $$self{chars}[$$self{colno}] }[0] || CC_OTHER) == CC_SPACE)
[...]

PR #2071, opened 2023-04-21
handle ignored catcode and blank lines

teepeemm:
Lines containing only ignored characters or whitespace should trigger a `\par`.  The adjustment to the Mouth causes this to happen, fixing #2027.  This caused center, flushleft, and flushright to have extra `ltx:break` tags when they have a `\par`, because LaTeX.pool made it so.  Consequently, LaTeX.pool no longer makes it happen.  I also introduced a new test for those environments.  (And a new test for the ignored and whitespace characters.)

A few test files have been changed as a result.  The resulting html files are generally closer to the pdfs, primarily because of better linespacing and indentation.


brucemiller:
Interesting; I'll have to dig into this. I suspect you're changing a bit too much, and some test cases are changing in ways that don't seem right.  But clearly you've found some bugs.

brucemiller:
See #2184 for a more focused patch to the `\par` problem, and dealing with corner cases (like `^^` chars, etc) that turned up while exploring --- thanks for the nice test samples!

Your patch is also changing `centering` behavior (but introducing too many `\par` and breaks, it seems to me); and I'm not quite sure what we're after with `\vadjust`.  If #2184 doesn't take care of those, it would be best to open a separate issue making clear what we're trying to do.

Thanks for the report & suggestions, though!

PR #2094, opened 2023-04-28
latexmlc: Only unlink logfile when it is a regular file

### LINKED ISSUE #2069: `make test` as root moving a file over `/dev/null`
How to reproduce:

```
sudo make test
```

This command is doing something that is moving a file over /dev/null, which breaks all of the system. Running tests as root is a part of some package building systems, and is unavoidable sometimes.

UPD: the actual failing test is:
```
sudo perl -Mblib t/91_latexmlc_api.t
```

UPD:`t/92_profiles.t` has the same problem
UPD :`t/93_formats.t` has the same problem
UPD: `t/95_complex_config.t` has the same problem
UPD:`t/96_fatal.t` has the same problem

I don't think this code is okay:

```
push @$opts, (['destination', "$localname.test.xml"],
    ['log',                "/dev/null"],
    ['quiet',              ''],
    ['quiet',              ''],
    ['quiet',              ''],
    ['quiet',              ''],
    ['autoflush',          1],
    ['timestamp',          '0'],
    ['nodefaultresources', ''],
    ['xsltparameter',      'LATEXML_VERSION:TEST'],
    ['nocomments',         '']);
```

I suspect that this setting doesn't actually do what you expect it to do. I suspect that it doesn't actually "pipe" output to `/dev/null`, it probably unlinks the file if it is present, and then makes a new one. It works as a user, because a user cannot delete `/dev/null`.

tkw1536:
As noted in #2069, `latexmlc` and friends were unlinking and re-creating logfiles. This was not a problem in the regular case, however on certain system when `make test` is run with root rights, this caused `/dev/null` to be deleted and re-created as a regular file.

This PR updates the behavior of `latexmlc` to only unlink the logfile if it is a regular file.

Fixes #2069.

### DIFF
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -117,7 +117,7 @@ if (!$logfile) {
   $opts->set('log', $logfile); }
 # A bit special: Since LaTeXML.pm will always try to append to the log file
 # make sure the file is new before we start.
-unlink($logfile) if (-e $logfile);
+unlink($logfile) if (-f $logfile);
 
 #***************************************************************************
 # Prepare output variables:


dginev:
Quick check on Ubuntu, works as advertised:
```
$ perl -e 'print -f "/dev/null" ? "exists\n" : "none\n"'
none
$ perl -e 'print -e "/dev/null" ? "exists\n" : "none\n"'
exists
```

LGTM

brucemiller:
Seems reasonable; Thanks!

PR #2098, opened 2023-05-10
Patch: attribute typo in "invisible" flag

dginev:
As it says. Spotted while auditing the alignment code (and small enough to fix fast).


dginev:
@brucemiller it's a typo either way, but there's a chance this causes differences in some SVG exports. Since you're stewarding the extended SVG test suite, might be worth checking before/after merging.

PR #2106, opened 2023-05-16
Fix profiling

brucemiller:
This PR gets the profiling option (to `latexml.sty`) working again.

@dginev I've added a call to `LaTeXML.pm` so that `latexmlc` also provides reports, but I'm not sure I added in the ideal place.


dginev (approved):
Looks good!

The addition to LaTeXML.pm seems to be in a reasonable place (before the daemon frame gets popped, and after the fatal recovery attempt).

PR #2108, opened 2023-05-24
also detect .ltx source files in ZIP input

dginev:
An easy problem with an easy fix I stumbled on for [arXiv:2303.15130](https://arxiv.org/abs/2303.15130), which uses a `.ltx` format for its main TeX source.

Currently latexml's ZIP input only checks the extensions `.tex` and `.txt`. This PR also adds the `.ltx` variant, which recognizes that paper's source.


brucemiller:
Looks good; Thanks!

Issue #373, opened 2013-03-26
rudimentary biblatex support


kohlhase:
[Originally Ticket 1697]

It seems that more and more people are moving over to biblatex, so we should have some rudimentary support. I started with my private stub `biblatex.sty.ltxml` by copying from `LaTeX.pool.ltxml` and got stumped. 

I think we should support at least the three main commands 
- `\addbibresource` which specifies a bibTeX (or other) database, the bibTeX case seems the most prevalent to support  
- `\bibliography` does the same, but allows multiple bibTeX files. Note that this is deprecated in favor of `\addbibresource` but is still commonly used. 
- `\printbibliography` which prints the bibliography
  So `\bibliography`/`\addbibresource` do the same as the old `\bibliography` from LaTeX, so I thought I would just copy and split the code; but this here does not quite work. 

```
DefConstructor('\printbibliography []',
           "<ltx:bibliography files='local'>"# files='#1' xml:id='#id'>"
           . "<ltx:title>References</ltx:title>"# font='#titlefont' _force_font='1'>#title</ltx:title>"
           ."</ltx:bibliography>");
DefConstructor('\bibliography Semiverbatim','',
           afterDigest=>sub { $_[0]->begingroup; # wrapped so redefns don't take effect!
                  beginBibliography($_[1]); 
                  $_[0]->endgroup; });

DefConstructor('\addbibresource Semiverbatim','',
           afterDigest=>sub { $_[0]->begingroup; # wrapped so redefns don't take effect!
                  beginBibliography($_[1]); 
                  $_[0]->endgroup; });
```

I would be grateful, if I could get my hands on a functional replacement. All I really want is a functional 

```
<bibliography files='foo bar'>
  <title>References</title>
</bibliography>
```

from 

```
\addbibresource{foo}
\addbibresource{bar}
...
\printbibliography
```

More coverage of biblatex would be nice, but this would be sufficient for the majority (all of mine) cases.

dginev:
Giving this issue a :+1: for higher priority, now that I am looking at the latest arXiv rerun. biblatex is quite common, and it could potentially be easy to add rudimentary support as far as an arXiv document is concerned (as it uses .bbl directly). 

As an example, one of our very own articles would be upgraded from error status to no_problem, with just the most basic of biblatex support:

https://corpora.mathweb.org/preview/arxmliv/tex_to_html/1404.6549

kohlhase:
[...]

PR #2109, opened 2023-05-26
fixing expl3

### LINKED ISSUE #2064: LaTeXML with current Fedora 38 beta LaTeX causes infinite (?) loop
I process LaTeX on Fedora 38 beta using LaTeXML. Recently, I started observing that some latexml runs seem to enter in infinite loop. I spent some time narrowing this down until I arrived at the following minimum example:
```
\documentclass{article}
\usepackage{siunitx}
\begin{document}
\end{document}
```
When I run `latexml foo.tex`, the conversion hangs with:
```
latexml (LaTeXML version 0.8.7) processing foo.tex
Package: expl3
 Digesting > Processing > Loading > Loading > Loading > Processing > Processing definitions /usr/share/texlive/texmf-dist/tex/latex/l3kernel/expl3-code.tex
```
Removing in import of `siunitx` fixes the problem. My Fedora 38 beta install provides LaTeXML-0.8.7-1.fc38.noarch and texlive-l3kernel-svn65299-65.fc38.noarch.

teepeemm:
I've poked and pushed this as much as I can, and now need a bit of help.  After a deep dive into the weeds of expl3, this PR will (1) eliminate the bug that was crashing expl3 and (2) cut the time to load expl3 by around 90%.

(1) The 100 errors were caused by TeX trying to access some intarrays.  LaTeX3 implements intarrays as a `\font` at size `++$nthNewIntArray` scaled points.  The length of the array is `\hyphenchar\font`, and the (1-indexed) $nth element is `\fontdimen{$n}\font` as scaled points.  The main part of fixing this was making the getters and setters for those two registers use private hashes with a key of the font name.  The other part of the fix is to look for when a font is involve in `\let`.  Otherwise, we fall back to the default values for the two registers, which corresponds to the 45 element intarray
```perl
[0,327680,0,0,282168,655361,0,0,0,0,
 0,0,0,0,0,0,0,0,0,0,
 0,0,0,0,0,0,0,0,0,0,
 0,0,0,0,0,0,0,0,0,0,
 0,0,0,0,0]
```
This does point to a failing of this current approach: if you try to work with `\nonExistantIntArry`, TeX will throw an error, whereas this may (also or instead) act as if you worked with the default array.  To fix this would require getting the default values only for the default fonts, whatever those are.

The solution does feel a bit heavy handed to me.  `lib/LaTeXML/Package/TeX.pool.ltxml` defines `\hyphenchar` and `\fontdimen`, so it's responsible for setting and getting.  But `lib/LaTeXML/Package.pm` defines `Let`, so it needs access to the hashes as well.  I don't think it's possible to have the module call out the ltxml file, so instead I do the reverse and have Pacakge.pm save the hash references.  I assume you'll also want to adjust how the hashes are defined and used.

(2) Debugging expl3 shows that my laptop takes about 20 seconds to get to the point where it opens UnicodeData.txt, over 20 minutes working through the file, and 30 seconds afterward.  There are two broad classes of what it's doing in UnicodeData.txt:
(a) Defining `\c__codepoint_[nfd|titlecase]_[someUTFchar]_tl` about 2000 or 60 times (respectively)
(b) Creating the four intarrays `\c__codepoint_[upper|lower]case_[blocks|index]_intarray`.
Since we have now defined how intarrays work, we can rewrite the ~200 lines of TeX into ~110 lines of Perl.  That takes the 20 minutes down to 30 seconds.

Item (1) works correctly.  If you dump out of `Package::AdjustExplThree` without doing anything, this will (after 80 minutes) pass all of the expl3 tests, including the newly included `t/expl3/explIntarray` making sure that we've defined the intarrays and the variables correctly.  Unfortunately, something is not quite right with how I'm handling the items in (a), and is the primary reason this needs to be a WIP.  When I try to use a variable, I get an error that I think indicates the variable isn't defined.  If we're not able to resolve the error, another approach would be to Storable::store the four intarrays during installation (a 288k file), and then Storable::retrieve them at runtime (and leave defining the variables to the existing LaTeXML).  That would still take about 30 seconds for UnicodeData, at the cost of storing a 288k file somewhere.

One complication in all of this is there is a narrow window where the necessary functions have been defined but not yet used.  I opted to look for UnicodeData.txt being opened, and act at that time.  From an object oriented perspective, I think it would make more sense to have `&AdjustExplThree` in one of the `lib/LaTeXML/Package/expl3.*` instead of `lib/LaTeXML/Package.pm`.  But again, I don't think it's possible to have `lib/LaTeXML/Core/Mouth/file.pm` call a ltxml file's function.

So:
* How should the variables be defined so that they can later be used?
* If we can't figure that out, should we Storable::store the intarrays into a file?  Where should that file go?
* How should my use of the font hashes be adjusted to better match the rest of LaTeXML?

Post Script:
To save for reference how to edit and debug expl3: In an empty directory:
* `cp $(kpsewhich expl.sty) myexpl.sty`
* `cp $(kpsewhich expl-code.tex) myexpl-code.tex`
* `cp $(kpsewhich expl.ltx) myexpl.ltx`
* Open them up, and change references of expl to myexpl.
* Then `\RequirePackage{myexpl}` and `--includestyles`.
Debugging from within LaTeX is harder, because expl is already loaded and checks to not load itself again.  In addition to the above, you also need to:
* `cp $(kpsewhich latex.ltx) mylatex.ltx`
* Open it up, and change references of expl to myexpl.
* Use the preamble from https://tex.stackexchange.com/a/91050/107497, inputting mylatex.ltx
* Compile with xetex

This PR resolves #2064, and is related to others.


dginev:
@teepeemm Ouch, I see you have sunk quite some time here.

Bruce has been working quite actively on patching this recent expl3 regression, and we just discussed his progress internally in our weekly meeting yesterday. It's quite a big bite to chew on.

I wouldn't recommend to external contributors to jump this far into the deep end of latexml debugging, as expl3 is one of the highest difficulty pieces (and one of the slowest ones to run), coming close to tikz and pgf work.

Thanks for sharing your progress in any case, maybe @brucemiller will spot a couple of helpful crumbs while reviewing...

brucemiller:
Wow, impressive debugging (which I can say, having done much of the same!)!  Let me put a different spin than @dginev : It's gratifying and reassuring that you came to many of the same conclusions & pointers that I did; and very helpful that you've found some things that I haven't (yet?).  So, you shouldn't feel like your work is wasted, even if we end up going with my WiP patch --- which will be that much better because of your observations.  That said, I should follow @dginev 's example and be more responsive on this forum about what I'm doing, or trying to do, so that contributors have a better sense of where to put their effort. 

You're right that the errors are coming from  `\font` metrics being used as arbitrary array storage.  I've fleshed out `\font,\fontdimen,\hyphenchar,\skewchar` much as you've done. However, I think the question of where to store the data that respects `\let` font symbols can be finessed with a devious bit of indirection: store the data much like is currently done, but use the definition's CS instead of the fonttoken that was given.
There're still a couple of inconsistencies in my code w.r.t. TeX, and `\fontdimen` should probably be better tied to the FontMetrics, some of which we have actually read in.

It is interesting that you managed to find a way to speed things up with Perl code. I've tracked a lot of that down to some inefficiencies forced on us by dealing with `\the`, but I'll have to look at how you've done things.  It may line up with something else I noticed: The LaTeX3 code will use several  `__intarray_XXX` routines if they've been predefined; the idea seems to be to provide lua versions for luatex (see `expl3.lua`). We can probably do perl versions and that may help.

vigoux:
Hi, I just wanted to add that I tested this PR in the context of another bug I encountered.

In this context, I wanted to use `biblatex`, and it in turns required `expl3` which chaused the problem.

With this PR, the `expl3` problem seems to disappear (but still takes an awfully long time to load), but after that file is loaded I get a `biblatex.sty file not found` which is weird.

I can open an issue for that specific problem, but nevertheless, this PR seems to fix the `expl3` issue !

I can help with the testing if needed further down the line.

dginev:
@vigoux there is no official biblatex support in LaTeXML today, see the open issue at #373 . There is a very partial experimental binding that works in the context of arXiv, but that's about all we have for now. expl3 is a bit of a red herring in that regard.

PR #2111, opened 2023-06-01
add support for marginnote.sty

teepeemm:
marginnotes are marginpars that (1) do not float and (2) have an optional argument to move them vertically.  (1) is the responsibility of CSS.  And as long as CSS is going to start moving them around, (2) probably doesn't make much sense.  The TeX package also provides a few commands to hook into styling a marginnote.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -577,6 +577,7 @@ lib/LaTeXML/Package/lxRDFa.sty.ltxml
 lib/LaTeXML/Package/ly1.fontmap.ltxml
 lib/LaTeXML/Package/makecell.sty.ltxml
 lib/LaTeXML/Package/makeidx.sty.ltxml
+lib/LaTeXML/Package/marginnote.sty.ltxml
 lib/LaTeXML/Package/marvosym.sty.ltxml
 lib/LaTeXML/Package/mathabx.sty.ltxml
 lib/LaTeXML/Package/mathbbol.sty.ltxml
--- a/lib/LaTeXML/Package/marginnote.sty.ltxml
+++ b/lib/LaTeXML/Package/marginnote.sty.ltxml
@@ -1,0 +1,31 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# |  marginnote                                                         | #
+# | Implementation for LaTeXML                                          | #
+# |=====================================================================| #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
+# |  United States Government & not subject to copyright in the US.     | #
+# |---------------------------------------------------------------------| #
+# | Tim Prescott <teepeemm@gmail.com>                           #_#     | #
+# | http://dlmf.nist.gov/LaTeXML/                              (o o)    | #
+# \=========================================================ooo==U==ooo=/ #
+package LaTeXML::Package::Pool;
+use strict;
+use warnings;
+use LaTeXML::Package;
+
+DefRegister('\marginnotevadjust'   => Dimension(0));
+DefRegister('\marginnotetextwidth' => LookupRegister('\textwidth'));
+
+DefMacro('\marginfont',            '\normalcolor');
+DefMacro('\raggedleftmarginnote',  '\raggedleft');
+DefMacro('\raggedrightmarginnote', '\raggedright');
+
+# marginnote is marginpar shifted vertically by #3 (which we ignore)
+DefMacro('\marginnote[]{}[]', sub {
+    Tokenize('\marginpar'
+        . ($_[1] ? '[\marginfont\raggedleftmarginnote ' . $_[1]->toString . ']' : '')
+        . '{\marginfont\raggedrightmarginnote ' . $_[2]->toString . '}'); });
+
+1;
--- a/lib/LaTeXML/Package/marginnote.sty.ltxml
+++ b/lib/LaTeXML/Package/marginnote.sty.ltxml
@@ -15,8 +15,19 @@ use strict;
 use warnings;
 use LaTeXML::Package;
 
-DefRegister('\marginnotevadjust'   => Dimension(0));
-DefRegister('\marginnotetextwidth' => LookupRegister('\textwidth'));
+DefConditional('\if@mn@verbose');
+
+# not documented, but in the implementation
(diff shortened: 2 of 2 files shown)


dginev:
This may be a good time to mention that there is a stub of marginnote.sty currently used in ar5iv, which can be found [here](https://github.com/dginev/ar5iv-bindings/blob/main/bindings/marginnote.sty.ltxml).

You may want to check it for completeness.

teepeemm:
The differences that I see are that the stub file
1. Defines `\marginnote[]{}[]` as `\footnote{#2}` instead of `\marginpar[#1]{#2}`.  I prefer my approach: if we change how we handle `\marginpar`, we would have to remember to change this as well.  (And I'm wanting to better handle `[#1]` if `\reversemarginpar` is issued, but that can be a separate item.)
2. Defines `\newmarginnote` analagously to `\newlabel`.  I suppose I should do this.
3. Processes (but ignores) the package options.  Is there a reason to `DeclareOption('option',undef)`?  The only thing I can think of is to avoid an `Unused global option(s)` warning.
4. Defines the internal commands.  Do we generally want to do this?

teepeemm:
I've used the stub file to make sure I've implemented everything in the style file.  I did stick with defining `\marginnote` in terms of `\marginpar`, instead of `\footnote`.

dginev:
Sure, the PR here is certainly an improvement over my stub file. 

We should keep in mind the narrative markup we are aiming for at the different stages of conversion. A `\marginpar` is internally represented by a `ltx:note` with `role="margin"`, while a `\footnote` is represented by a `ltx:note` with `role="footnote"` at the XML stage. So you're certainly right to choose the variation closer to the original macro intention.

That said, in the final HTML, each of those constructs (i.e. their `<span>`-soup mapping) could appear in a screen's footer or a margin, depending on viewport size and the CSS theme used by the final deployment which serves the pages. 

To take ar5iv as an example, footnotes are displayed in the side margins on wide screens, but as on-hover notes on narrow screens (which is the only behavior in the default latexml CSS). So whether a note is in a "margin" or a "footer" becomes a more liquid distinction on the web. At least if that responsive behavior is a design goal.

I need to look into styling margin notes, and how that compares to footnotes in ar5iv. As one random example from our reports, the first 6 notes in [ar5iv:1511.00350](https://ar5iv.labs.arxiv.org/html/1511.00350) are using `\marginnote`, and currently mapped to footnotes. I'll write back once I test convert with this branch.

brucemiller:
This looks good; Thanks!

dginev:
Looking into the styling here, I still think we've regressed visually a little compared to the footnote style (from the ar5iv side). Here is the old stub:

Screenshot 2023-07-20 at 09-50-42 Beyond Degree Choosability

and here is the state with this PR merged (and me hiding the `margin:` note type with CSS):
image

It is a lot harder to discern which mark belongs to which note, as we are not maintaining the same linebreaking structure as the PDF.

teepeemm:
But this does end up being closer to the pdf, which doesn't even have a footnotemark within the text.

I'm having some Git difficulties that are making me unsure of my results, but I think I'm getting a different output with this PR (and some errors that would need to be fixed).  Are you using this merge for your second picture, or have you also used some Arxiv styling?

dginev:
@teepeemm my screenshots are using [ar5iv-css](https://github.com/dginev/ar5iv-css) over [ar5iv:1511.00350](https://ar5iv.labs.arxiv.org/html/1511.00350) that I mentioned above.

Being "closer to the pdf" is a red herring here - we need an understandable visual connection between the notes and their associated mainmatter content in HTML, and if we can't guarantee the line level matches the margin level of the note (which we can't), the numbered footnote marks are a reasonable approach.

I could alternatively also hide the marks entirely in CSS, which makes the display a little cleaner and matches the PDF approach, but that would make it really hard to understand which notes associate with which pieces of content, because our lines reflow with screen size.

So this is neither here nor there...

brucemiller:
I'm open to further patches. Certainly the repeated dagger is much less clear what each note relates to.  Perhaps the connection should be more footnote-like, while still making the `ltx:note` have a class related to margin pars?

teepeemm:
Or should we leave this so that marginnotes become marginpars, and instead reconsider whether marginpars should behave more like footnotes?  With the old stub, wouldn't mixing marginnotes and marginpars leave some numbered and some with daggers?

brucemiller:
I'm not sure how much it matters whether marginnotes *are* marginpars so long as there are a rational set of `@role`s and classes and that these roles are recognized when they need to be; they all come down to an `ltx:note` with role and class attributes. 

n retrospect, perhaps the 1st arxiv example above only looks "better" because both are so wrong :>  The notes are spread out in a way that looks pretty, but loses the alignment with the thing being annotated (eg note 6  f-choosable is next to the wrong paragraph). I think marginpars (and notes) are used more to draw attention to (or explain, illustrate, etc) a sentence or paragraph, rather than a specific word. Thus you don't normally use numbers or daggers if you're displaying the notes (unless you're hiding them until hover).

Perhaps it's the default treatment of `role=margin`, or all `ltx:note`, that needs improvement:  The word "margin:" is added to distinguish different kinds of notes simply because margin is not an explicitly recognized role. And the dagger is added since the note doesn't display by default, but only when you hover over the dagger --- we don't assume there *is* a margin by default :>  Both of those additions is happening in the XSLT, so harder to patch up later in CSS. A better division between the XSLT & CSS might help?

I

dginev:
> The notes are spread out in a way that looks pretty, but loses the alignment with the thing being annotated (eg note 6 f-choosable is next to the wrong paragraph). 

Exactly, we can't guarantee the adjacency that PDF can, becase we reflow (and may choose to stack or fold-in the notes).

> I think marginpars (and notes) are used more to draw attention to (or explain, illustrate, etc) a sentence or paragraph, rather than a specific word.

Ah, how they are *actually* used is kind of an open question. An author could add anything they find sensible in those margin spaces. The screenshot of this one article used them as an in-margin index kind of thing, where you can quickly navigate to a paragraph mentioning a term.

But they can be actual notes as well - the Overleaf [margin notes tutorial](https://www.overleaf.com/learn/latex/Margin_notes) provides a demonstration of that as well.

We can spend some dedicated HTML design time to improve how we represent notes, but in the absence of volunteers, I think I will go ahead and add a new PR that anchors them using the footnote mark numbered counters for now.

Issue #2035, opened 2022-12-18
Problem with cmap.sty


henrikt-ma:
This doesn't work unless the use of _cmap.sty_ is removed:
```
\documentclass{article}
\usepackage{cmap}
\usepackage[utf8]{inputenc}
\usepackage{natbib}
\begin{document}
Broken citation: \cite{Hello}
\bibliography{minimal}
\end{document}
```

Symptom: Citation just shows bibliography key, like this:
> Broken citation: Hello

and the bibliography is empty.

This is the corresponding _minimal.bib_:
```
@InProceedings{Hello,
  author = {Lastname, Firstname},
  title = {Hello},
}
```

Build rule:
```
minimal.html: minimal.tex
	$(LATEXMLPREFIX)latexml minimal.tex --includestyles --dest minimal.xml
	$(LATEXMLPREFIX)latexmlpost minimal.xml -format html -pmml --navigationtoc=context --dest $@
```

Build log:
```
➜  minimal-LaTeXML touch minimal.tex;make minimal.html
latexml minimal.tex --includestyles --dest minimal.xml
latexml (LaTeXML version 0.8.6) processing minimal.tex
Error:expected:{} Missing argument {} for Core::Definition::Expandable[\@gobble {}] at cmap.sty; line 43 col 0
Warning:latex:(cmap) Package cmap Warning: pdftex not detected - exiting at cmap.sty; line 43 col 0
Conversion complete 1 warning; 1 error (See /Users/henrikt/Documents/minimal-LaTeXML/minimal.latexml.log) (reqd. 0.53s)
latexmlpost minimal.xml -format html -pmml --navigationtoc=context --dest minimal.html
latexmlpost (LaTeXML version 0.8.6) paginating /Users/henrikt/Documents/minimal-LaTeXML/minimal.xml
Error:expected:{} Missing argument {} for Core::Definition::Expandable[\@gobble {}] at cmap.sty; line 43 col 0
Warning:latex:(cmap) Package cmap Warning: pdftex not detected - exiting at cmap.sty; line 43 col 0
Initialization complete: 1 warning; 1 error (See /Users/henrikt/Documents/minimal-LaTeXML/minimal.latexmlpost.log). Aborting.
Error:expected:{} Missing argument {} for Core::Definition::Expandable[\@gobble {}] at cmap.sty; line 43 col 0
Warning:latex:(cmap) Package cmap Warning: pdftex not detected - exiting at cmap.sty; line 43 col 0
[...]

PR #2116, opened 2023-06-07
"implement" cmap.sty

teepeemm:
With `--includestyles`, the package cmap causes a LaTeXML `Warning:latex:(cmap) Package cmap Warning: pdftex not detected - exiting`.  But the purpose of cmap is 'to make the PDF files generated by pdflatex "searchable and copyable"'.  Therefore, the only thing to do is check off any global options and exit successfully.

This is one possible fix for #2035.  It appears that we've already fixed the problem with `\PackageWarningNoLine{}{}` leading to an Error, so that issue is completely resolved.


dginev (approved):
Looks good to me, thanks @teepeemm

brucemiller:
Looks good; Thanks!

PR #2117, opened 2023-06-07
Fix allocated registers

brucemiller:
This PR improves TeX fidelity by: allocating registers from `\count,\dimen` etc; fixing `\meaning` results for registers (distinguishing allocated from primitive registers); and more using `(Lookup|Allocate)Register` API rather than the `Value` ones (making risky assumptions about storage).


lib/LaTeXML/Package.pm
```diff
@@ -817,9 +838,9 @@ sub RefStepID {
 
 sub ResetCounter {
   my ($ctr) = @_;
-  AssignValue('\c@' . $ctr   => Number(0), 'global');
-  AssignValue('\c@UN' . $ctr => Number(0), 'global');    # ?????
-  DefMacroI(T_CS("\\\@$ctr\@ID"), undef, Tokens(Explode(LookupValue('\c@' . $ctr)->valueOf)),
+  AssignRegister('\c@' . $ctr   => Number(0), 'global');
+  AssignRegister('\c@UN' . $ctr => Number(0), 'global') unless $ctr =~ /^UN/;    # ?????
```
dginev, line 842: I see there was an `# ?????` comment before, but now it has a new `unless $ctr=~/^UN/` added... Could I request a more verbose comment explaining the intention?

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -907,19 +908,7 @@ DefMacro('\meaning Token', sub {
...
-          if ($register_type =~ /glue/) {
-            $prefix = '\skip'; }
-          elsif ($register_type =~ /dimension/) {
-            $prefix = '\dimen'; }
-          elsif ($register_type =~ /tokens/) {
-            $prefix = '\toks'; } }
-        my $literal_value = $register_type && $value->valueOf;
-        # Should we be more careful to distinguish between latex and tex counters?
-        $meaning = $prefix . $literal_value; }
+        $meaning = $$definition{address} || $definition->getCSName; }
```
dginev, line 911: Could one add a new `Register::getCSName` which returns the `{address}` field? That may make porting this a little easier in well-typed languages as just `$meaning = $definition->getCSName` on any `Definition`.

lib/LaTeXML/Package/thmtools.sty.ltxml
```diff
@@ -62,7 +62,7 @@ sub thmtools_style {
   if (my $bodyfont = $kv && $kv->getValue('bodyfont')) {
     $parameters{'\thm@bodyfont'} = $bodyfont; }
   if (my $headformat = ToString(($kv && $kv->getValue('headformat')) || '')) {
-    $parameters{'\thm@swap'} = $headformat eq 'swapnumber'; }
+    $parameters{'thm@swap'} = $headformat eq 'swapnumber'; }
```
dginev, line 65: I'm not closely familiar with the code here, but if `thm@swap` loses the backslash, shouldn't all other `thm@` parameters lose it as well? As in `\thm@bodyfont` right above? 

It's a little unclear if they're subtly different.

dginev (approved):
The PR reads really nicely and the changes look robust. It's great that there is a new set of tests as well.

I guess the harder question will be if any definitions got overlooked for the AssignRegister refactor. Is there a way to craft a new error that spots such situations?

dginev:
We have lost the MikTeX test suite again (for reasons unrelated to the PR), but CI passes on all working images. So that's ✅

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: It probably needs an accessor (but not getCSName), even though this probably should be the only place that should know that it has an address stored internally.

lib/LaTeXML/Package/thmtools.sty.ltxml
brucemiller: Yeah, way too subtly different. The others are registers, and really should be. The swap is effectively boolean, though it could be forced to be a register.

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: Or even possibly a `getMeaning` (non-mathematical), and shift the burden do the Definitions?

brucemiller:
Spotting errors turns out kinda tricky by eye. I've caught most, but...  I did some semi-automated detection over the test-suite, but  each one has to be examined to see which way is "correct".

lib/LaTeXML/Package.pm
brucemiller: ok, done

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: ok, done

PR #2126, opened 2023-06-20
Scan toks

brucemiller:
This PR defines a `gullet->readTokens($macrodef,$expand)` which acts much like TeX's `scan_toks` (but doesn't read macro parameter lists). It ends up being a synthesis of `readXToken` and `readBalanced` . Since we're collecting the results of the expansions, we can easily defer further expansion of `\the`-like commands without cumbersome wrapping of tokens in a "smuggle".  Thus all the various `CC_SMUGGLE_THE` code, along with several additional utilities (`T_SMUGGLE_THE`, `without_dont_expand`, `setup_scan`, etc) are no longer needed and have been removed..

The Token smuggling caused a lot of overhead in intensive expansion computations, such as with recent LaTeX 3.

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -117,12 +117,11 @@ DefParameterType('Plain', sub {
 
 DefParameterType('DefPlain', sub {
     my ($gullet, $inner) = @_;
-    my $value = $gullet->readArg;
+    my $value = $gullet->readTokens(1, 0);
     if ($inner) {
       ($value) = $inner->reparseArgument($gullet, $value); }
     return $value; },
-  packParameters => 1,
-  reversion      => sub {
+  reversion => sub {
     my ($arg, $inner) = @_;
     (T_BEGIN,
       ($inner ? $inner->revertArguments($arg) : Revert($arg)),
@@ -152,13 +151,8 @@ DefParameterType('Optional', sub {
 # which IS required in contrast to a general argument; ie a single token is not correct.
 DefParameterType('GeneralText', sub {
     my ($gullet) = @_;
-    my $open = $gullet->readXToken;
-    if ($open && ($open->getCatcode == CC_BEGIN)) {
-      return $gullet->readBalanced; }
-    else {
-      Error('expected', '{', $gullet,
-        "Expected <general text> here");
-      return $open; } });
+    $gullet->unread($gullet->readXToken);    # Force expansion to skip <filler> before required {
+    return $gullet->readTokens(0, 0); });
 
 DefParameterType('Until', sub {
     my ($gullet, $until) = @_;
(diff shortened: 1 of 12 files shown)


dginev (approved):
Really nice!

Curious side additions too, changing the second argument of `\accent` and switching UndigestedDefKey to an explicit read+pack sub, rather than `packParameters => 1`.

Also interesting that you kept gullet's `readArg` as-is, (using readBalanced), but switched some of the parameter types that relied on readArg to now use the new readTokens. I need to ponder the implications there, but I'll probably need to first understand what TeX is actually doing in the regular argument cases...

brucemiller:
There's potential to unify things a bit more regarding `readArg` and `readBalanced`.  `readArg` accepts a single token, or a balanced list if "{" is encountered. Some of it is adapting to history: `readBalanced` expects the opening "{" to already have been read, whereas `readTokens` (aka `scan_toks`) requires to read it.  So, for the moment, `readBalanced` simply pushes a `T_BEGIN` before calling `readTokens`.

brucemiller:
Most of the parameter changes were to achieve the "correct" `\the` expansions and `macro_def` behavior (packParameters); this is particularly true for things that will end up as macro bodies.

dginev:
@brucemiller rebasing this on master will also help get the fresh CI checks.

brucemiller:
OK, this should wrap this up; a proper `scan_toks` equivalent, along with simpler and more TeX-like `\noexpand`, and deferred expansion of `\the`.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -2472,21 +2484,17 @@ DefPrimitive('\lowercase GeneralText', sub {
...
   @tokens = map {
-    my $t  = ($$_[2] || $_);
-    my $cc = $$t[1];
-    if    ($cc == CC_CS)    { ($t, T_SPACE); }
+    my $cc = $$_[1];
+    if    ($cc == CC_CS)    { ($_, T_SPACE); }
     elsif ($cc == CC_SPACE) { (T_SPACE); }
-    elsif ($cc == CC_PARAM) { ($t, $t); }
-    elsif ($cc == CC_ARG) {
-      # B Book, 294. Reduce to param+integer
```
dginev, line 2521:
```
# B Book, 294. Reduce to param+integer
```

For the record: I liked my comment!

dginev (approved):
This PR is now an absolutely fantastic upgrade with far-reaching performance and correctness implications for the low-level TeX interpretation.

I've read through the changes, they look really nice. CI passes too, so you've really solved it! Good to merge.

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: Yeahhhhhh, I figured you did, but you also gave a better reference to B book in the comment above.

PR #2133, opened 2023-06-26
implement verbatim*

### LINKED ISSUE #2085: `\begin{verbatim*}` doesn't use ␣
The inline macro `\verbatim*` works as expected.

MWE:
```latex
\documentclass{article}

\begin{document}
\verb*+12 3+

\begin{verbatim*}
ab c
\end{verbatim*}
\end{document}
```

LaTeX output:

> `12␣3`
>
> ```
> ab␣c
> ```

LaTeXML output:
```xml
  <para xml:id="p1">
    <p><verbatim font="typewriter">12␣3</verbatim></p>
  </para>
  <para xml:id="p2">
    <verbatim>
ab c
</verbatim>
  </para>
```
(Note the missing `␣` in `ab c`.)

teepeemm:
This implements the `verbatim*` environment, which is like `verbatim`, but uses `␣` instead of the space character.  This resolves #2085.

This also fixes a bug where `\verb *T*T` should use `*` as the delimiter, because of the space after `\verb`.

These two cases have also been added to `t/tokenize/verb`.


brucemiller (changes requested):

brucemiller (approved):
Yeah, I think that does it! Thanks!!!

PR #2137, opened 2023-07-02
Dont defer tokenization of macro body

brucemiller:
As it says;  We had been deferring tokenization of macro bodies when given (in perl) as strings. That gives a small performance gain for multiple small conversions, but the extra checks are a penalty for conversions involving heavy expansion.


dginev (approved):
Code looks good, optimization sounds sensible.

PR #2138, opened 2023-07-02
Tracing optimization

brucemiller:
This combines the tracing and profiling flags into a single bitmask, which reduces the number of lookups required (and we're usually not tracing or profiling).  This is a performance boost since it is happening frequently and at lower level.

Yes, magic numbers; but context should make them clear & the hassle (in Perl) of faking constants isn't worth it.

### DIFF
--- a/lib/LaTeXML/Package/latexml.sty.ltxml
+++ b/lib/LaTeXML/Package/latexml.sty.ltxml
@@ -34,10 +34,14 @@ DeclareOption('noids', sub { AssignValue('GENERATE_IDS' => 0, 'global'); });
 DeclareOption('comments',   sub { AssignValue('INCLUDE_COMMENTS' => 1, 'global'); });
 DeclareOption('nocomments', sub { AssignValue('INCLUDE_COMMENTS' => 0, 'global'); });
 
-DeclareOption('tracing',     sub { AssignValue(TRACING => (LookupValue('TRACING') || 0) | 3); });
-DeclareOption('notracing',   sub { AssignValue(TRACING => (LookupValue('TRACING') || 0) & ~3); });
-DeclareOption('profiling',   sub { AssignValue(TRACING => (LookupValue('TRACING') || 0) | 4); });
-DeclareOption('noprofiling', sub { AssignValue(TRACING => (LookupValue('TRACING') || 0) & ~4); });
+DeclareOption('tracing', sub {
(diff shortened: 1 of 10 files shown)


lib/LaTeXML/Package/latexml.sty.ltxml
```diff
@@ -34,8 +34,10 @@ DeclareOption('noids', sub { AssignValue('GENERATE_IDS' => 0, 'global'); });
 DeclareOption('comments',   sub { AssignValue('INCLUDE_COMMENTS' => 1, 'global'); });
 DeclareOption('nocomments', sub { AssignValue('INCLUDE_COMMENTS' => 0, 'global'); });
 
-DeclareOption('profiling',   sub { AssignValue('PROFILING' => 1, 'global'); });
-DeclareOption('noprofiling', sub { AssignValue('PROFILING' => 0, 'global'); });
+DeclareOption('tracing',     sub { AssignValue(TRACING => (LookupValue('TRACING') || 0) | 3); });
```
dginev, line 37: Could we document the bitmask details here (in latexml.sty)? Especially since it is likely to be ported to an `enum` type, it would be nice to match the implied intention.

dginev:
@brucemiller this branch now has a conflict with master, likely can use a rebase.

dginev:
Ah. That last merge commit is problematic - it goes in the wrong flow.

I think you should either rebase onto master from the previous commit 49da94953cb043c50a9bb6756813dc24205481e5 as in
```
git checkout 49da94953cb043c50a9bb6756813dc24205481e5
git reset HEAD --hard
git rebase master
git push origin tracing-optimization --force
```

which will update this branch with force, but avoid a merge commit.

Or alternatively (and in my opinion worse) you could merge this branch directly onto master. I think that goes as in:

```
git checkout tracing-optimization
git revert -m 1 01e254ce0b30638821c873a62acadc23384cd6b2
git checkout master
git merge tracing-optimization
git push -u origin master
```

In either case the last merge commit is going to be an issue and needs to be undone. At least that's how I'd go about it, git probably has a handful of other options available.

brucemiller:
Cant it justbe a suash?

dginev:
@brucemiller right, a squash would erase the entire commit history here. So - sure, that works.

brucemiller:
OK, that should be better...

PR #2139, opened 2023-07-02
$ctr is never used in \@startsection

dginev:
A little compiler birdie told me that a `$ctr` very similar to the one we have in `\@startsection` is never used.

I double-checked and indeed, some wasted compute cycles can be eliminated. Enjoy!


dginev:
Bonus: I recovered the Windows CI back into a working state.

The chocolatey short commands `cinst` and `cuninst` have (recently?) been removed in favor of the main `choco` command.

brucemiller:
Oh! Cool! and Cool!! Thanks

PR #2140, opened 2023-07-03
Comments welcome

brucemiller:
This PR deals more consistently with comments, whether Tokens, digested or XML nodes, appearing in places we previously didn't expect them.

The function `IsEmpty` tests whether a tokens or list  is empty, ignoring comments, and is generally preferred over `scalar($x->unlist )`; `element_next($node), element_prev($node)` are usually preferred over `$node->previousSibling`.   Also, avoid splitting xml text nodes with comments, pushing them to the front.

Also, remove an obsolete form of `DefMathLigature`.

### DIFF
--- a/lib/LaTeXML/Common/XML.pm
+++ b/lib/LaTeXML/Common/XML.pm
@@ -75,7 +75,7 @@ our @EXPORT = (
     decodeFromUTF8  ),
   @XML::LibXML::EXPORT,
   # Possibly (later) export these utility functions
-  qw(&element_nodes &text_in_node &new_node
+  qw(&element_nodes &text_in_node &new_node &element_next &element_prev
     &append_nodes &clear_node &maybe_clone
     &valid_attributes &copy_attributes &rename_attribute &remove_attr
     &get_attr &isTextNode &isElementNode &isChild &isDescendant &isDescendantOrSelf
@@ -96,6 +96,20 @@ sub element_nodes {
 sub text_in_node {
   my ($node) = @_;
   return ($node ? join("\n", map { $_->data } grep { $_->nodeType == XML_TEXT_NODE } $node->childNodes) : ''); }
+
+sub element_next {
+  my ($node) = @_;
+  my $next;
+  while (($next = $node->nextSibling) && ($next->nodeType != XML_ELEMENT_NODE)) {
+    $node = $next; }
+  return $next; }
+
+sub element_prev {
+  my ($node) = @_;
+  my $prev;
+  while (($prev = $node->previousSibling) && ($prev->nodeType != XML_ELEMENT_NODE)) {
+    $node = $prev; }
+  return $prev; }
 
 sub isTextNode {
   my ($node) = @_;
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -2859,45 +2859,38 @@ sub DefLigature {
       %options });
   return; }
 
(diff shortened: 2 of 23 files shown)


lib/LaTeXML/Common/XML.pm
```diff
@@ -97,6 +97,20 @@ sub text_in_node {
...
+  while (($next = $node->nextSibling) && ($next->nodeType != XML_ELEMENT_NODE)) {
+    $node = $next; }
+  return $next; }
+
+sub element_prev {
+  my ($node) = @_;
+  my $prev;
+  while (($prev = $node->previousSibling) && ($prev->nodeType != XML_ELEMENT_NODE)) {
+    $node = $prev; }
+  return $prev; }
```
dginev, line 112: Naming question: Given `nextSibling` and `previousSibling` have the naming scheme with the modifier in front and the kind of object second, I'd be tempted to suggest `next_element` and `previous_element` (which are in a way abbreviated forms of `get_next_element` and `get_previous_element`).

That's roughly how I've internalized the already existing `element_nodes` which seems to align with `childNodes` in the libxml wrapper.

lib/LaTeXML/Package.pm
```diff
@@ -974,6 +975,32 @@ sub TokenizeInternal {
...
+  my (@things) = @_;
+  foreach my $thing (@things) {
+    my $ref = ref $thing;
+    if    (!$thing)                          { }
+    elsif ($ref eq 'LaTeXML::Core::Comment') { }
+    elsif ($ref eq 'LaTeXML::Core::Tokens') {
+      return 0 unless IsEmpty($thing->unlist); }
+    elsif ($ref eq 'LaTeXML::Core::Token') {
+      my $cc = $$thing[1];
+      return 0 if ($cc == CC_LETTER) || ($cc == CC_OTHER) || ($cc == CC_ACTIVE) || ($cc == CC_CS); }
```
dginev, line 991: Hm, this seems incomplete? A T_BEGIN, T_END, T_MATH, T_ALIGN, T_PARAM aren't empty either. Maybe it's somewhat briefer to invert the condition:
```perl
return 0 unless ($cc == CC_COMMENT) || ($cc == CC_MARKER) || ($cc == CC_IGNORE);
```

dginev: though if this is meant to be used for a single Token, I'd also wonder if we shouldn't treat all tokens that have empty text as empty as well.
```perl
return 0 unless ($cc == CC_COMMENT) || ($cc == CC_MARKER) || ($cc == CC_IGNORE) 
  || (!length($$thing[0]));
```

dginev (approved):
Read through the code, the `IsEmpty` consistency looks nice to have. Just the couple of minor comments.

lib/LaTeXML/Common/XML.pm
brucemiller: I guess at the time I was writing, I was emphasizing "element".

lib/LaTeXML/Package.pm
brucemiller: I think you have the logic flipped; this line should short-circuit immediately returning "false" if we encounter a token that  is *not* empty.
OTOH, your point is good: should we consider `T_LETTER("")` to be empty?
OTOOH, I expect this issue may have to be revisited to distinguish between truly empty, and just spacey things? Hopefully I can defer that concern?

brucemiller:
You raise good questions, but with the refactors in this PR, it should be easy to find & revisit later if we want to.

PR #2141, opened 2023-07-04
Additional Comment accommodations

brucemiller:
Additional Comment accommodations that should have been in previous batch; already had been tested locally, but accidentally got left out.


dginev (approved):
Looks good!

This kind of situation is a good reminder that we don't test the handling of comment tokens in CI. We could benefit from a special test mode that exercises comments, with a new test. 

If/when it fits our development schedule...

brucemiller:
Good thing you mentioned it; I had at least one test case I could incorporate, and when I fleshed it out it turned up another corner case!  Thanks!

PR #2149, opened 2023-07-10
More optimization

brucemiller:
Given a bit more wisdom about gullet, after #2126, I couldn't resist another session with NYTProf for some optimizing... with some promising improvements.  A bit more to come.

### DIFF
--- a/lib/LaTeXML/Core/Token.pm
+++ b/lib/LaTeXML/Core/Token.pm
@@ -19,6 +19,7 @@ package LaTeXML::Core::Token;
 use strict;
 use warnings;
 use LaTeXML::Global;
+use LaTeXML::Core::State;
 use LaTeXML::Common::Error;
 use LaTeXML::Common::Object;
 use base qw(LaTeXML::Common::Object);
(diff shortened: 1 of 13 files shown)


brucemiller:
OK, I think this is sufficient for this PR. (though there's still more that warrants a separate PR)

lib/LaTeXML/Core/Token.pm
```diff
@@ -325,6 +325,21 @@ sub equals {
     && ($$a[1] == $$b[1])
     && (($$a[1] == CC_SPACE) || ($$a[0] eq $$b[0])); }
 
+# Check whether $token is "equivalent" to $self,
+# that is, equal to $self, or \let to $self.
+# $self is is presumed to be some "constant", explicit token,
+# such as  T_SPACE, T_CS('\endcsname').
+sub equivalent {
```
dginev, line 332: My one remaining suggestion for this PR is to avoid the overloaded name [equivalent](https://en.wikipedia.org/wiki/Equivalence_relation) here and instead come up with a name that is a little closer to the low-level details of what the method is trying to achieve.

It seems the method is interested in the way a token is defined (technically, its stored Definition), if at all, and in recognizing the special case of a space token.

I think one would also like to capture the asymmetry implied by the documentation string, where one of the arguments (currently `$self`) is already a "terminal" token, which won't get further lookups into State, while the other is a "content" Token, which needs the `lookupMeaning` check. An "equivalence" doesn't really capture that.

The best suggestion I have is `defined_as`, where we switch the argument order and provide the constant token as the argument.

Namely:
```perl
$token->defined_as(T_endcsname);
```

instead of 
```perl
T_endcsname->equivalent($token);
```

To my taste this is a "must have" improvement, but you're welcome to overrule otherwise.

dginev (changes requested):

lib/LaTeXML/Core/Token.pm
brucemiller: Hmm, yeah; `defined_as` is much clearer, w/o becoming unreadable.  And it captures the asymmetry which `equivalent` doesn't. Thanks!

dginev (approved):
Everything looks encouraging now, thanks.

PR #2151, opened 2023-07-14
Intarray

brucemiller:
This PR provides the `intarray` infrastructure needed by the latest (2023) TeXLive versions of `expl3`.  It no longer fails with errors, but still takes way too long to run (working on that).

This partially, but not completely, addresses #2044 and #2064.

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -1338,7 +1338,7 @@ DefPrimitive('\font Token SkipMatch:= SkipSpaces TeXFileName', sub {
     my $f = ($at ? $at->divide(Dimension('1em'))->valueOf
       : ($scaled ? $scaled->valueOf / 1000
         : 1));
-    my $fontinfo = {%props};
+    my $fontinfo = \%props;
     $$fontinfo{data} = [map { $_->multiply($f); }
         Dimension(0), Dimension('0.5em'), Dimension(0),
       Dimension(0), Dimension('1ex'), Dimension('1em')];
@@ -1346,7 +1346,7 @@ DefPrimitive('\font Token SkipMatch:= SkipSpaces TeXFileName', sub {
     # Store the font info & metrics
     AssignValue('fontinfo_' . ToString($cs) => $fontinfo);
     # The font $cs should select the font
-    DefPrimitiveI($cs, undef, undef, font => {%props});
+    DefPrimitiveI($cs, undef, undef, font => $fontinfo);
     return; });
 
 # Not sure what this should be...
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -1319,26 +1319,35 @@ sub lookupFontinfo {
   #  return LookupValue(($defn ? ToString($defn) : ToString($token)) . '_fontinfo'); }
   return LookupValue('fontinfo_' . ($defn ? $defn->getCSName : ToString($token))); }
 
-# Need to handle "at" too!!!
+# This should eventually actually load the font metrics,
+# and tie-in to the FontMetrics data used by Font.
 DefPrimitive('\font Token SkipMatch:= SkipSpaces TeXFileName', sub {
     my ($stomach, $cs, $name) = @_;
     my $gullet = $stomach->getGullet;
     $name = ToString($name);
+    my ($at, $scaled);
+    if ($gullet->readKeyword('at'))     { $at     = $gullet->readDimension; }
+    if ($gullet->readKeyword('scaled')) { $scaled = $gullet->readNumber; }
     my %props = LaTeXML::Common::Font::decodeFontname($name,
-      ($gullet->readKeyword('at')
-        ? $gullet->readDimension->ptValue : undef),
-      ($gullet->readKeyword('scaled')
-        ? $gullet->readNumber->valueOf / 1000 : undef));
+      $at && $at->ptValue, $scaled && $scaled->valueOf / 1000);
     if (!keys %props) {    # Failed?
       Info('unexpected', $name, $stomach, "Unrecognized font name '$name'",
         "Font switch macro " . ToString($cs) . " will have no effect"); }
     else {
       $props{fontname} = $name; }
+    my $f = ($at ? $at->divide(Dimension('1em'))->valueOf
+      : ($scaled ? $scaled->valueOf / 1000
+        : 1));
+    my $fontinfo = {%props};
+    $$fontinfo{data} = [map { $_->multiply($f); }
+        Dimension(0), Dimension('0.5em'), Dimension(0),
+      Dimension(0), Dimension('1ex'), Dimension('1em')];
     $gullet->skipSpaces;
-    AssignValue('fontinfo_' . ToString($cs) => {%props});
+    # Store the font info & metrics
+    AssignValue('fontinfo_' . ToString($cs) => $fontinfo);
+    # The font $cs should select the font
     DefPrimitiveI($cs, undef, undef, font => {%props});
-
-});
+    return; });
 
 # Not sure what this should be...
 DefPrimitiveI('\nullfont', undef, undef, font => { family => 'nullfont' });
@@ -1499,9 +1508,6 @@ foreach my $letter (ord('A') .. ord('Z')) {
   $STATE->assignUCcode(chr($letter + 0x20), $letter, 'global'); }
 
 # Stub definitions ???
-DefRegister('\hyphenchar{}', Number(ord('-')));
-DefRegister('\skewchar{}',   Number(0));          # no idea what the default is here
-
 DefMacro('\hyphenation GeneralText', Tokens());
 DefMacro('\patterns{}',              Tokens());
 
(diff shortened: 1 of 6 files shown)


dginev:
I would also like to invite @teepeemm to offer a review here (if he's willing and interested), given that he did very relevant overlapping work in #2109 . No obligation to do so, but he's welcome to!

lib/LaTeXML/Package/expl3.lua.ltxml
```diff
@@ -0,0 +1,124 @@
...
+
+DefMacroI('\intarray_count:N', 'Intarray', sub {
+    my ($gullet, $table) = @_;
+    return Explode(scalar(@$table)); });
+
+DefMacroI('\__intarray_gset:wF', 'Number Intarray Number {}', sub {
+    my ($stomach, $pos, $table, $value, $ifmissing) = @_;
+    my $length = scalar(@$table);
+    $pos = $pos->valueOf;
+    if (($pos > 0) && ($pos < $length)) {
```
dginev, line 64: Is there an off-by-one error here, or am I misreading the bounds?

Say, if `$table = [SomeSingleElement];` and `$pos` is `1`, then `$length` is `1` as well, but `$pos == $length`. I think we won't be able to assign an element in that slot?

So maybe `$pos < $length` should be `$pos <= $length` ?

There are a number of other subs with this `$$table[$pos - 1]` kind of indexing, so posing the question in general for all of them.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -1319,26 +1319,35 @@ sub lookupFontinfo {
...
+      $at && $at->ptValue, $scaled && $scaled->valueOf / 1000);
     if (!keys %props) {    # Failed?
       Info('unexpected', $name, $stomach, "Unrecognized font name '$name'",
         "Font switch macro " . ToString($cs) . " will have no effect"); }
     else {
       $props{fontname} = $name; }
+    my $f = ($at ? $at->divide(Dimension('1em'))->valueOf
+      : ($scaled ? $scaled->valueOf / 1000
+        : 1));
+    my $fontinfo = {%props};
```
dginev, line 1341: Out of curiosity, why not `my $fontinfo = \%props;` ?
I think that avoids cloning the hash, and if I'm reading the code right `%props` has no further uses that may cause data race conflicts.

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -1319,26 +1319,35 @@ sub lookupFontinfo {
...
+    $$fontinfo{data} = [map { $_->multiply($f); }
+        Dimension(0), Dimension('0.5em'), Dimension(0),
+      Dimension(0), Dimension('1ex'), Dimension('1em')];
     $gullet->skipSpaces;
-    AssignValue('fontinfo_' . ToString($cs) => {%props});
+    # Store the font info & metrics
+    AssignValue('fontinfo_' . ToString($cs) => $fontinfo);
+    # The font $cs should select the font
     DefPrimitiveI($cs, undef, undef, font => {%props});
```
dginev, line 1349: Ah, my comment from line 1341 applies better here at line 1349, where we truly have the last use of `%props`.

lib/LaTeXML/Package/expl3.lua.ltxml
```diff
@@ -0,0 +1,124 @@
...
+# Translation of the intarray portion of expl3.lua
+# lua arrays are typically 1-based
+# Apparently an expl3 intarray is encoded at the TeX level by \__intarray:w <Number>
+# We store actual ints in the array, seems more efficient (could be Number or Tokens)
+DefParameterType('Intarray', sub {
+    my ($gullet) = @_;
+    my $tok = $gullet->readXToken;
+    if (my $defn = LookupDefinition($tok)) {    # token may be \let to the identifier!!!
+      $tok = $defn->getCS; }
+    if ($tok && $tok->equals(T_CS('\__intarray:w'))) {
```
dginev, line 27: Interesting, the 3 lines here seem to be a good possible use of the new `$tok->defined_as(T_CS('\__intarray:w'))`, right?

The current code looks correct too, just marking the opportunity.

lib/LaTeXML/Package/expl3.lua.ltxml
```diff
@@ -0,0 +1,124 @@
...
+    return; });
+
+DefMacroI('\__intarray_to_clist:Nn', 'Intarray', sub {
+    my ($stomach, $table) = @_;
+    return Tokenize(join(', ', @$table)); });
+
+DefMacroI('\__intarray_range_to_clist:w', 'Intarray Number Number', sub {
+    my ($stomach, $table, $from, $to) = @_;
+    $from = $from->valueOf - 1;
+    $to   = $to->valueOf - 1;
```
dginev, line 112: This is a rare edge case, but if either of the input numbers happens to be (erroneously?) zero `0`, the `$from` or `$to` index will be `-1`, which will lead to a rather peculiar indexing in perl. It may be worth guarding them to be within the allowable range of `$table` here as well.

dginev (commented):
Looks good on a surface read! I left some low-level perl comments.

lib/LaTeXML/Package/expl3.lua.ltxml
teepeemm: It seems like the expl3 code usually has the public functions do the input validation, and then the `__private` functions don't need to worry about that.  The corresponding lua code doesn't validate the inputs.  (I've also not noticed a public version of this function, so it might all be moot.)

lib/LaTeXML/Package/expl3.lua.ltxml
```diff
@@ -0,0 +1,124 @@
...
+DefMacroI('\__intarray_gset:w', 'Number Intarray Number', sub {
+    my ($stomach, $pos, $table, $value) = @_;
+    my $length = scalar(@$table);
+    $pos = $pos->valueOf;
+    if ($pos > 0) {
+      $$table[$pos - 1] = $value->valueOf; }
+    return; },
+  protected => 1);
+
+DefPrimitiveI('\intarray_gzero:w', 'Intarray', sub {
```
teepeemm, line 80: My dtx file is defining `\intarray_gzero:N` at this point, and no `\intarray_gzero:w`.

lib/LaTeXML/Package/expl3.lua.ltxml
```diff
@@ -0,0 +1,124 @@
...
+    my ($stomach, $table) = @_;
+    return Tokenize(join(', ', @$table)); });
+
+DefMacroI('\__intarray_range_to_clist:w', 'Intarray Number Number', sub {
+    my ($stomach, $table, $from, $to) = @_;
+    $from = $from->valueOf - 1;
+    $to   = $to->valueOf - 1;
+    return Tokenize(join(', ', @$table[$from .. $to])); });
+
+DefPrimitiveI('\__intarray_gset_range:w', 'Number Intarray', sub {
```
teepeemm, line 115: Since `\__kernel_intarray_gset_range_from_clist:Nnn` calls `\__intarray_gset_range:w \int_eval:w #2 #1 #3 , , \scan_stop:`, I think this will leave `#3 , , \scan_stop:` in the token stream.  Should this be something like
```perl
DefPrimitiveI('\__intarray_gset_range:w',
  LaTeXML::Core::Parameters->new(
    LaTeXML::Core::Parameter->new('Number'),
    LaTeXML::Core::Parameter->new('Intarray'),
    LaTeXML::Core::Parameter->new('Until','Until:\scan_stop:',extra=>[T_CS('\scan_stop:')]) ),
  sub {
    ... } );
```
On the other hand, there doesn't appear to be a public version of the function, so I'm not sure how often this will get called (also, being unimplemented will throw an error anyway).

lib/LaTeXML/Package/expl3.lua.ltxml
```diff
@@ -0,0 +1,124 @@
...
+# Apparently an expl3 intarray is encoded at the TeX level by \__intarray:w <Number>
+# We store actual ints in the array, seems more efficient (could be Number or Tokens)
+DefParameterType('Intarray', sub {
+    my ($gullet) = @_;
+    my $tok = $gullet->readXToken;
+    if (my $defn = LookupDefinition($tok)) {    # token may be \let to the identifier!!!
+      $tok = $defn->getCS; }
+    if ($tok && $tok->equals(T_CS('\__intarray:w'))) {
+      my $n       = $gullet->readNumber;
+      my $address = '__intarray' . '_' . ToString($tok) . '_' . $n->valueOf;
```
teepeemm, line 29: Not really a bug, but having every `$address` begin with `__intarray_\__intarray:w_` seems a bit redundant.  (Unless there's some other way to get an intarray that I wasn't finding.)

lib/LaTeXML/Package/TeX.pool.ltxml
brucemiller: Indeed, I'm just not in the `\%` habit. Patch coming.

lib/LaTeXML/Package/expl3.lua.ltxml
brucemiller: But all that *wasted* space!?!?!  I think I'll be better off keeping the Perl arrays 0-based, but do need to be careful about the limits. Patch coming.

lib/LaTeXML/Package/expl3.lua.ltxml
brucemiller: I wasn't completely clear on the lua interface, so I just implemented (or didn't) everything related to `intarray`.

lib/LaTeXML/Package/expl3.lua.ltxml
brucemiller: ditto.

dginev (approved):
Looks good!

brucemiller:
I think this one's ready, if you guys agree.

teepeemm:
We do still need `\intarray_gzero:N`, which can use the `:w` definition.  I'm moderately certain that we don't need `\intarray_gzero:w`.

I'm still trying to dig into why it takes so long to go through UnicodeData, but that shouldn't stop this from merging.

brucemiller:
Thanks for the reviews. I still would like to come back later to #2109 to see if we can speed things up, but given the evolving nature of latex3, I'd want to avoid building in too much expl3 code (even translated into perl); it'll likely be different tomorrow.

PR #2160, opened 2023-07-24
Be more careful translating ReLaxNG wildcard attributes

### LINKED ISSUE #2148: rawhtml no longer working?
I'm pretty sure that this
```perl
use strict;
use warnings;
use LaTeXML::Package;
DefConstructor('\makeiframe',
           '<ltx:rawhtml>'.
           '<xhtml:iframe src="https://math.nist.gov/~BMiller/LaTeXML/" />'.
           '</ltx:rawhtml>');
```
```tex
\documentclass{article}
\providecommand{\makeiframe}{make iframe}
\begin{document}
\makeiframe
\end{document}
```
used to work (for example, back when you helped me with #1007).  But now I'm getting the xml `<xhtml:iframe/>` and losing any attributes.  If I understood git bisect correctly, I think this started with #1960 `2313ec53706e5ac3c6cb41147ffdd024fac20985` on 2022-10-20, and looking through the committed files, it does seem plausible.  Can I adjust my function call in some way to make this work?

brucemiller:
Be more careful translating ReLaXNG wildcard attributes. In the internal model, * stands for any NON-namespaced name and *:* stands for any namespaced name.

Fixes #2148


dginev (approved):
Interesting. PR itself seems clearcut!

XML namespaces are notorious for a reason... Having empty, optional and default namespaces be three potentially different settings can quite dangerous for unsuspecting developers.

Issue #2165, opened 2023-07-24
image descriptions for SVGs (\tikzpictures)


rzach:
Issue  #1452 added support for `alt` attributes for `img` tags produced by `\includegraphics`. It looks like no similar mechanism exists for images produced with `\tikzpicture` (which are turned into inline SVG).

dginev:
Interesting observation, @rzach . I believe we followed the lead of the LaTeX team for the `alt` support, who added the key to the `\includegraphics` macro. Is there a similar development for the native `\tikzpicture`? 

I don't know the answer, and maybe this extra information can be requested from the [pgf-tikz](https://github.com/pgf-tikz) project? 

I am finding a similar request for [htlatex's {tikzpicture}](https://tex.stackexchange.com/questions/641335/htlatex-tikzpicture-with-alt-tag-and-custom-class), but that doesn't really indicate if this has become native since...

rzach:
If I understand the spec correctly, `<svg>` doesn't have an `alt` attribute, but rather you put title and/or description inside the `<svg>` between `<title>...</title>` and `<desc>...</desc>` tags. Since you don't know if `latexml` will generate inline SVGs or write to an image file or whatnot, I suppose it's harder to decide how to implement.

Perhaps the easiest way to do this is to provide something in `latexml.sty` so you can insert `<title>` and `<desc>` into the SVG code (`\lxSvgTitle`, `\lxSvgDesc`, or something more general that adds raw XML?)

[The way I get around it now is to use features provided by [https://github.com/vlmantova/bookml](BookML): Surround the `tikzpicture` with `\begin{bmlimage}...\end{bmlimage}`. BookML then converts the picture to SVG by running it through LaTeX and converting the DVI to an SVG file which is then included using `<img>`. And you can set the `alt` tag on that image using `\bmlDescription{...}` right after `\end{bmlimage}`. That works well enough but the images generated aren't optimal--for one, everything is in Computer Modern. But if LaTeXML had an option to write the SVG to a file instead of inlining it (which I think is issue #2144?), and a way to provide the ALT tag on the resulting `<img>`, ...]

dginev:
Right, sure. I am not that concerned about passing the information into the SVG, various nice [accessibility guides](https://www.tpgi.com/using-aria-enhance-svg-accessibility/) exist to that end, and it is relatively easy to accommodate the different approaches in latexml.

[...]

PR #2166, opened 2023-07-25
preserve xhtml:attrs in XSLT

dginev:
This is a PR-shaped question to @brucemiller, while brainstorming on #2165 .

It is generally quite simple to write a small binding that deposits ARIA attributes for latexml, except that those attributes are not declared in the schema, and aren't allowed to pass through in the XSLT post-processing.

We could allow each and every one of the list of ARIA attributes in the `ltx` namespace, but - in purely pragmatic terms - such a requirement limits experimentation for lay binding authors a bit. I myself find it somewhat uncomfortable to be buffered through the latexml schema, before being able to experiment with emitting the ultimate HTML markup. Going back to enhance the schema after a stable HTML markup pattern emerges feels slightly more natural to exploratory coding.

The PR contains the first quick upgrade while wearing my HTML hat - allow any `xhtml:attribute` to pass through the XSLT as a plain `attribute` in the final (X)HTML, in the same conditional check that currently emits `data-` attributes as a fallback.

This allowed to write the binding:
```perl
DefEnvironment("{namedblock}{}{}", "<ltx:block xhtml:aria-label='#1' xhtml:aria-description='#2'>#body</ltx:block>");
```
author:
```tex
\begin{namedblock}{sample-value}{describe it}
  Hello World!
\end{namedblock}
```
and emit to HTML5:
```html
<div class="ltx_block" aria-description="describe it" aria-label="sample-value">
<p class="ltx_p">Hello World!</p>
</div>
```

More generally, passing arbitrary `xhtml:names` along allows direct experimentation with (X)HTML attributes of any variety, bypassing that aspect of latexml's schema.

---

In addition, maybe in a next PR (or next commit), if one was to be Principled in the XML sense, we'd follow the [WAI-ARIA host languages](https://www.w3.org/TR/wai-aria-1.0/host_languages#host_general_attrs) instructions and declare an `aria` namespace at the official URI `http://www.w3.org/ns/wai-aria`. Then I'd need a dedicated XSLT rule that maps `aria:` namespaced values into `aria-` values in HTML5, as we do for `data-`.

I believe this simplest PR can be useful for general quick experimentation, and I can bring along the namespaced aria approach in addition to that.

How much of that aligns with your perspective here @brucemiller ?

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-common.xsl
@@ -413,6 +413,11 @@
           <xsl:value-of select="."/>
         </xsl:attribute>
       </xsl:when>
+      <xsl:when test="namespace-uri() = 'http://www.w3.org/1999/xhtml'">
+        <xsl:attribute name="{local-name()}">
+          <xsl:value-of select="."/>
+        </xsl:attribute>
+      </xsl:when>
       <xsl:when test="$prefix = 'data'">
         <xsl:attribute name="{concat('data-',local-name())}">
           <xsl:value-of select="."/>


dginev:
I should also mention that the master branch currently uses the `data-` check for the xhtml namespaced attributes, creating the rather amusing variation on my example:
```html
<div class="ltx_block" data-xhtml-aria-label="sample-value">...</div>
```

dginev:
A technical note I noticed this morning, this kind of experimentation is already possible if it is done *only* as HTML, without inter-operating with the native schema. Namely this:

```perl
DefEnvironment("{ariadiv}{}{}", "<ltx:rawhtml><xhtml:div aria-label='#1' aria-description='#2'>#body</xhtml:div></ltx:rawhtml>");
```

gets successfully carried into (X)HTML, thanks to the recently fixed #2148.

dginev:
Most of this sounds like a code smell to me. Of course courtesy of XML's design which makes the simple and intuitive way of annotating something adjacent to "rude surprises", rather than helpful intuitive behavior.

Irrespective of this PR, LaTeXML should be able to pass along any `xhtml:` namespaced attribute into HTML5, which it currently mangles, as [I showed above](https://github.com/brucemiller/LaTeXML/pull/2166#issuecomment-1649957930).

But I think there is more important work than this for now, closing here.

brucemiller:
> Irrespective of this PR, LaTeXML should be able to pass along any 
> |xhtml:| namespaced attribute into HTML5, which it currently mangles,

No, since you have `USE_DATA_ATTRIBUTES` true, it did exactly what you 
told it to do.

If, on the other hand, you want the ability to say that some attributes 
(or namespaces) get turned into data attributes, some keep their 
namespace, some drop their namespace, etc, etc, then you've got to have 
a mechanism to specify what you want, without too many short signed 
assumptions.

PR #2170, opened 2023-07-27
Not a function

### LINKED ISSUE #2163: \lnot ¬ should be a prefix operator
`$\lnot A$` will generate `<mi mathvariant="normal">¬</mi><mo>⁢</mo><mi>A</mi>` but the ¬ is a prefix operator not an identifier.

In general, is there a way to help LaTeXML generate the right MathML for prefix operators? E.g., if I want to take a relation (say, `\sim`) and turn it into a prefix operator (i.e., avoid the spacing around relations in LaTeX) I'd say `\mathord{sim}` but that turns it into an `<mi>`. My screen reader, e.g., will read out `$\mathord{sim}A$ as "tilde times upper A".

brucemiller:
make `\lnot` and `\neg` be OPERATOR rather than FUNCTION, which leads to better parsing & MathML.

Fixes #2163

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -5906,10 +5906,10 @@ DefMathI('\angle', undef, "\x{2220}");
 
 # NOTE: This is probably the wrong role.
 # Also, should probably carry info about Binding for OpenMath
-DefMathI('\forall',    undef, "\x{2200}", role => 'BIGOP',    meaning => 'for-all');
-DefMathI('\exists',    undef, "\x{2203}", role => 'BIGOP',    meaning => 'exists');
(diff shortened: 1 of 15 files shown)


dginev (approved):
Right, we do have OPERATOR. That unfortunately makes `\neg P(x)` parse as `(\neg P)(x)`. But it emits an `<mo>` and succeeds fine on single letter arguments.

Maybe that's fine for now...

brucemiller:
`BIGOP` works too; it has lower precedence, but has some implications for scripts which probably doesn't matter. Maybe we'll want a little more nuanced grammatical roles...

dginev:
Right, `BIGOP` is a decent fit as well. 

My mention in the issue of a simple `PREFIX` is something that you've already implemented for unary uses of `ADDOP` (`-f(x)`, `+f(x)` etc.), but would need to be factored out and reused.  I'll play around with the idea in a different grammar realm...

PR #2172, opened 2023-07-28
Fix \ion macro

### LINKED ISSUE #2090: aastex `\ion`: Numeral not capitalized
MWE:

```latex
\documentclass{aastex}
\begin{document}
\ion{Fe}{2}
\end{document}
```

LaTeX output:

> Fe II

LaTeXML output:

```xml
  <para xml:id="p1">
    <p>Fe ii</p>
  </para>
```

(`ii` not capitalized.)

Perhaps it'd also be worth thinking about applying a no-break space (U+00A0) or narrow no-break space (U+202F) -- even though aastex doesn't do that. (The fundamental question: Should LaTeXML "fix" the LaTeX output, or should it be as faithful as possible?)

brucemiller:
needed some `\expandafter` magic.

Fixes #2090


dginev (approved):
sure

PR #2173, opened 2023-07-29
set URI to destination for postprocessed documents

### LINKED ISSUE #2093: mismatch between $doc URI and working folder
Tricky issue to explain. During postprocessing, all URLs to resources (CSS, images, etc) are relative to the working folder. So if I call e.g. `latexmlpost folder/file.xml` and `file.xml` includes the image `images/one.svg`, latexmlpost will look for `images/one.svg` in the current directory. However, if I call `document(@href)`, where `@href="images/one.svg"`, latexmlpost will look for `folder/images/one.svg`. ^

Now `document()` is not exactly supported by LaTeXML, but it has use cases (e.g. if you remember, I inspect SVGs in postprocessing to decide whether to use img or object; very effective in practice). So it would be nice if `document()` and latexmlpost treated relative URLs in a consistent way.

~~The backward compatible solution is to change the internal URI of `$doc` so that it's in the working folder; it doesn't change the behaviour of latexmlpost and fixes the discrepancy. It should be an easy change too.~~

Edit: `document()` receives URIs that are relative to the *destination* folder, so the URI must be set to the destination.

^ The semantics of `document()` is delicate – if called on a pure string, it's relative to the _stylesheet_. When called on a node, such as `document(@href)`, it will resolve relative to the URI of the document containing that node.

xworld21:
Fixes #2093 by setting the URI of postprocessing documents to their destination. This guarantees that calling e.g. `document(@imagesrc)` resolves relative URLs correctly, i.e., relative to the destination directory.

### DIFF
--- a/lib/LaTeXML/Post.pm
+++ b/lib/LaTeXML/Post.pm
@@ -664,6 +664,7 @@ use DB_File;
 use Unicode::Normalize;
 use LaTeXML::Post;    # to import error handling...
 use LaTeXML::Common::Error;
+use URI::file;
 use base qw(LaTeXML::Common::Object);
 our $NSURI = "http://dlmf.nist.gov/LaTeXML";
 our $XPATH = LaTeXML::Common::XML::XPath->new(ltx => $NSURI);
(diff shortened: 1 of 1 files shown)


brucemiller:
Seems harmless and maybe even useful! :>  Thanks!!!

PR #2176, opened 2023-07-30
use normal spaces with CSS white-space:pre instead of non-breaking spaces in listings

### LINKED ISSUE #2067: make listings output copy-pasteable
It looks like LaTeXML translates spaces in listings to non-breaking spaces, which has the side effect of breaking copy & paste of the code presented on screen. Minimal example:
```latex
\documentclass{article}
\usepackage{listings}
\begin{document}
\begin{lstlisting}
  Three spaces   then the rest.
\end{lstlisting}
\end{document}
```
will produce non-breaking spaces in the XML output.

I figure LaTeXML should output normal spaces and adjust the CSS instead (and maybe also use `<pre>` and `<code>` tags rather than plain `<div>` and `<span>`).

xworld21:
Fixes #2067 in so far as non-breaking spaces are concerned.

### DIFF
--- a/lib/LaTeXML/Package/listings.sty.ltxml
+++ b/lib/LaTeXML/Package/listings.sty.ltxml
@@ -1277,7 +1277,7 @@ sub lstProcess {
   local $LaTeXML::ESCAPE_RE = join('|', map { $$delimiters{$_}{open} }
       grep { $$delimiters{$_}{escape} } sort keys %$delimiters);
   local $LaTeXML::QUOTED_RE = "\\\\\\\\";    # start w/ backslashed backslash?
-  local $LaTeXML::SPACE     = (lstGetBoolean('showspaces') ? T_CS('\@lst@visible@space') : T_CS("~"));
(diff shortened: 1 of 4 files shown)


brucemiller:
Maybe I've missed something, but if I grab your `listing.xml` and css, the html doesn't really look right; there's a lot of line-breaks and vertical space. Your patch is turning the former non-break spaces to regular spaces and using `white-space:pre` to make them not break. But there is already a lot of regular whitespace in the listing markup, so that CSS is applying to that whitespace as well.  I think it'd be hard to exclude that whitespace(?)

xworld21:
> there is already a lot of regular whitespace in the listing markup

Truth is, I was playing with this on a file with no multiline example... I changed the CSS so that it applies to each `.ltx_text` separately, which achieves the desired effect without interfering with the surrounding white space.

brucemiller:
Cool! That seems to focus the `pre` where it's needed.  Thanks!!

PR #2180, opened 2023-08-05
use value of name in \lstinputlisting

### LINKED ISSUE #2179: \lstinputlisting ignores name for caption
```latex
\documentclass{article}
\usepackage{listings}
\begin{document}
\lstinputlisting[name=test-context]{t/000_test_context.t}
\lstlistoflistings
\end{document}
```
produces the same output as in #2178 (i.e. use the filename in the TOC) instead of using the value of `name`.

xworld21:
Fix #2179


brucemiller:
Cool; that should work. Thanks!

PR #2181, opened 2023-08-05
preserve filename in listings

### LINKED ISSUE #2177: add file name to listing download button
I mentioned ages ago in #1450, when talking about the `<a download>` button used in listings:

> I suppose it would be even better to use the actual file name when it is available (i.e. when using \lstinputlisting).

When clicking on a `<a download>` link, browsers use the `@download` value as file name, if there is one, but a default name if `@download` is empty. For data URLs, the file name can be `download.txt` or a random string depending on the browser.

However, if the file was read using `\lstinputlisting`, we should set `@download` to the original file name.

This is not difficult, I am preparing a PR.

xworld21:
Fix #2177. Note that this uses the listing `name` as download name, so its effect not limited to `\lstinputlisting`.

I couldn't run `tools/compileschema`, could you please verify that my manual schema changes are correct?


dginev:
FYI I wouldn't trust a change I myself make to `LaTeXML.model` if it wasn't automatically generated. There are some indirect relationships that get made explicit in it, which are easy to miss. Maybe not relevant for the specific PR, but I can highly recommend auto-generating the model file.

What is the issue with `tools/compileschema`? If I recall, it mostly needs a working `trang` installation.

brucemiller:
This is a nice enhancement!  And, as you say, it's not restricted to listings, but any raw data annotation. So, how about we add `dataname` to the `Data.attributes` collection (`name` is a little too generic, and is already *way* too overloaded)?

`compileschema` requires `trang` to work, but it keeps things consistent.  Should I make the schema change & push that change back to your PR/branch? Or would you prefer to battle it out yourself :>

xworld21:
> Or would you prefer to battle it out yourself :>

It's a Nixpkgs packaging issue, so I do need to battle it out myself... working on it!

xworld21:
> What is the issue with `tools/compileschema`? If I recall, it mostly needs a working `trang` installation.

Problem solved (it was Nixpkgs). I changed the PR to use a generic `@dataname`.

lib/LaTeXML/resources/XSLT/LaTeXML-block-xhtml.xsl
```diff
@@ -155,7 +155,9 @@
           <xsl:with-param name="name" select="'href'"/>
         </xsl:call-template>
         <xsl:if test="$USE_HTML5='true'">
-          <xsl:attribute name="download"/>
+          <xsl:attribute name="download">
+            <xsl:value-of select="@name"/>
```
brucemiller, line 159: This wants to be `@dataname`, right?

xworld21: ...forgot to run the final `git commit --amend`...

brucemiller (approved):
Nice! Thanks!!

PR #2182, opened 2023-08-06
Malformed xmtok

brucemiller:
This PR fixes a couple of `malformed ltx:XMTok` problems in the xy arXiv sandbox.


dginev:
Would this be a good time to suggest we include a CI-only test that exercises an xy diagram? 

The xmtok detail sounds like a nice place to guard for sneaky regressions.

brucemiller:
A good time, but bad place, as it turns out that neither of these patches actually has anything to do with `xy`, other than coincidentally showing up in the xy sandbox.  I did enhance a testcase for the tricky `&` chars, though.

dginev (approved):
LGTM

PR #2184, opened 2023-08-08
Fix par

### LINKED ISSUE #2027: ignored characters should not prevent a paragraph break

brucemiller:
This PR deals more correctly with spaces and ignored chars which shouldn't affect a "blank" line which produces a `\par`; this should be the case whether they are the normal chars, or chars with catcode changed, or even `^^` encoded chars.

Fixes #2027
Alternative to #2071


dginev (approved):
LGTM

PR #2185, opened 2023-08-09
proof-of-concept implementation of -recorder

### LINKED ISSUE #2183: implement (or document alternative of) -recorder option for latexml/latexmlpost
Most (all?) tex commands support the `-recorder` option, which writes each file opened for input or output in a `.fls` file. It would be nice to have something comparable for latexml(post), e.g. to produce Makefile dependencies.

As of now, latexml writes `(Loading [filename]...` in the log for many files being read but not all of them, e.g. `\lstinputlisting` does not produce any output, and latexmlpost says very little about the files being read, not to mention resources being copied.

xworld21:
It turns out that implementing `-recorder` is a small task (fix #2183 in principle). Since it is essentially a logging mechanism, I have attached it to Common:Error.

This is not complete (I couldn't make sense of the log handling in LaTeXML.pm!), but does it go in the right direction?


lib/LaTeXML/Common/Error.pm
```diff
@@ -134,11 +136,49 @@ sub UseLog {
     $log_count++;
     return if $LOG or not($path);                 # already opened?
     pathname_mkdir(pathname_directory($path));    # and hopefully no errors! :>
+    RecordOutput($path) if ! $append && ! $log_count;
```
xworld21, line 139:
```suggestion
    RecordOutput($path) if ! $append;
```
I believe `$log_count` is redundant, and `RecordOutput($logfile)` when the log is in append mode should be responsibility of `latexmlc` (I did say I am struggling to follow the logging code in LaTeXML.pm...).

lib/LaTeXML/Common/XML/Parser.pm
```diff
@@ -12,12 +12,18 @@
...
+use URI;
+use URI::Escape;
 use XML::LibXML;
 
 sub new {
   my ($class) = @_;
   my $parser = XML::LibXML->new();
   $parser->validation(0);
+  my $input_callbacks = XML::LibXML::InputCallback->new();
+  $input_callbacks->register_callbacks([sub { RecordInput(uri_unescape(URI->new($_[0])->path)); return 0; }, undef, undef, undef]);
```
xworld21, line 25: TODO: fix Windows problems with mixed `\` and '/' path separators.

dginev:
To vocalize the implied silence on the PR, I am still unsure if we want to add this level of boilerplate to the logging mechanisms in latexml. As you also mentioned, we'll have to make sure we add the recorder lines at the right places and then keep maintaining them each time we introduce new file I/O.

My personal preference (also based on how I've implemented the log-based tracking over arXiv) is to make the main log entries emitted by latexml as machine-readable as possible, so that the single log file can be parsed with relative ease. If there is a subtlety where messages aren't emitted at present but should, we could also leverage latexml's `--debug` mechanism, to add recorder-only messages (`--debug=recorder`). A brand new command-line option is usually something I feel very conservative about, because each addition makes the CLI tool harder to understand for newcomers.

A different suggestion I have heard recently by the arXiv team is that there is a generic way to track this for any process using `strace`, as with

```
strace -e trace=open latexml [arguments]
```

xworld21:
> My personal preference (also based on how I've implemented the log-based tracking over arXiv) is to make the main log entries emitted by latexml as machine-readable as possible

I quite agree with this option. Not sure why I didn't make this a draft PR as it's named "proof-of-concept". The exercise was about picking up all the relevant open events, including hidden ones such as schemas, but omit irrelevant ones such as temporary files or system libraries. (Intercepting I/O is necessary anyway to properly deal with e.g. utf8 filename encoding, so that's not a high cost once I get around proposing new pathname_open and the likes.)

We could simply drop lines like LXINPUT, LXOUTPUT, or something else in the log to get the same result and that's certainly less invasive than the approach here.

(strace is good, but not very portable, and it adds lots of noise.)

Issue #341, opened 2013-01-04
qtree.sty support


physikerwelt:
[Originally Ticket 1665]

I'd like to make Tree drawings with qtree unsing LaTeXML.
As Michael mentioned there is also a tikz-qtree version.

dginev:
qtree.sty, tikz-qtree.sty, and today I have seen some great diagrams produced in a very recent paper via [forest.sty](http://mirrors.ibiblio.org/CTAN/info/forest-quickstart/ForestQuickstart.pdf)

As far s the output SVG is concerned, it may be best to keep these three in mind together as a single project.

dginev:
Some free lunch from the latest master (from Bruce's latest tikz upgrades) -- we are now getting some proof-of-concept successful trees via the `tikz-qtree` route. 

First the (unchanged) bad news:
1. I also tried forest.sty -- which sadly has an infinite loop when interpreting raw, and that after loading expl3, making it somewhat scary to debug. 
2. Using the original `qtree.sty` actually gets all the way to depositing specials in the final HTML, e.g. 
```html
<span id="p1.p1.pic1" class="ltx_picture" style="width:79.1pt;height:19.8pt;">
\begin{picture}(2.0,0.5)\put(0.0,0.0){\special{ps: save /setgray { pop } def}%
\special{" \x 0 0 moveto 39.562114 19.781057 lineto\space\y}\special{ps: %
restore}}
\put(2.0,0.0){\special{ps: save /setgray { pop } def}\special{" \x 0 0 moveto %
-39.562114 19.781057 lineto\space\y}\special{ps: restore}}
\end{picture}
</span>
```
but we'd need some binding work to map those to SVG.

Now for the good news.

Here is a minimal example [from the manual](https://mirrors.concertpass.com/tex-archive/graphics/pgf/contrib/tikz-qtree/tikz-qtree-manual.pdf):
```tex
\documentclass{article}
\usepackage{tikz}
\usepackage{tikz-qtree}
\begin{document}

\Tree [.S [.NP [.Det the ] [.N cat ] ]
              [.VP [.V sat ]
                      [.PP [.P on ]
                              [.NP [.Det the ] [.N mat ] ] ] ] ]

\end{document}
```
generating the following HTML rendering with the latest master and using `--includestyles`:

![image](https://user-images.githubusercontent.com/348975/115124830-60064f80-9f92-11eb-991f-4017b8073501.png)

and here is the PDF for comparison:
[...]

PR #2186, opened 2023-08-09
allow deeper recursion for tikz diagrams

dginev:
I was double-checking if our raw support for `tikz-qtree.sty` has improved (based on [a small example](https://github.com/brucemiller/LaTeXML/issues/341#issuecomment-821877053) from #341 ), and indeed it has!

The only catch was that due to the deeper interpretation, the conversion was running into `Fatal:perl:deep_recursion` fatal errors. I allowed more recursion in 3 locations, which then allowed the conversion to succeed.

Tested with a simple:
```
$ latexmlc test.tex --dest=test.html --includestyles
```

That minimal example tree now renders exactly as in the PDF - with the only exception of having the entire outer frame translated too far down. Tested with today's latexml master at 6143eb15bbbb435bcb0cf67544acac1c55e6b3e2

TeX example:
<details>

```tex
\documentclass{article}
\usepackage{tikz}
\usepackage{tikz-qtree}
\begin{document}

\Tree [.S [.NP [.Det the ] [.N cat ] ]
              [.VP [.V sat ]
                      [.PP [.P on ]
                              [.NP [.Det the ] [.N mat ] ] ] ] ]

\end{document}
```

</details>

### DIFF
--- a/lib/LaTeXML/Core/List.pm
+++ b/lib/LaTeXML/Core/List.pm
@@ -63,6 +63,7 @@ sub unlist {
   return @{ $$self{boxes} }; }
 
 sub revert {
+  no warnings 'recursion';
   my ($self) = @_;
   return map { $_->revert } $self->unlist; }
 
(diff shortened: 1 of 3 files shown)


brucemiller:
Seems reasonable. Thanks!

PR #2190, opened 2023-08-10
Fix token register

### LINKED ISSUE #2097: weird font setting incorrectly make the entire document italic
Here's a weird LaTeX font setting that I don't understand. For native LaTeX, it doesn't seem to do anything, in LaTeXML the entire document is typeset in italic.

```latex
\documentclass{article}

\newfam\bifam
\font\tenbi=cmmib10 scaled \magstep1 \font\sevenbi=cmmib10 at 11pt
\font\fivebi=cmmib10 at 6pt \textfont\bifam = \tenbi

\begin{document}
x
\end{document}
```

LaTeXML output:
```xml
<p><text font="italic" fontsize="120%">x</text></p>
```

brucemiller:
This PR skips spaces before reading the value of (pseudo) registers that take a token as value, such as `\textfont`.

Fixes #2097

Also removes an unneeded circular dependency which broke perl-critic.


dginev:
Nice, looks good!

PR #2191, opened 2023-08-11
detect raw PDF inclusion documents and mark them as invalid inputs

dginev:
This PR marks another class of arXiv inputs as invalid, following a user report from today at [ar5iv#376](https://github.com/dginev/ar5iv/issues/376).

There is a range of special case documents that are erroneously included in the arXiv source bundles, which precompile a local PDF and re-include it via `\includepdf` to satisfy the arXiv submission requirement.

I think it is an honest "invalid" marker that a document which only has a single use of that macro isn't meant for latexml conversion. Suggestions welcome if there is a better test than the one I tried - I simply checked at the end of the document construction whether we have more than one content node under `ltx:document`, when `\includepdf` has been used.


brucemiller:
Was this intended to be a PR for the arXiv setup?  It seems quite reasonable to consider such files invalid for coretex, and fairly pointless to convert with latexml in general, but throwing a fatal error seems kinda gratuitous.

dginev:
The PR was filed in the intended repository.

PR #2193, opened 2023-08-14
Several longtable fixes

### LINKED ISSUE #2004: using a better caption for longtable
longtable allows specifying table captions in the firsthead, head, foot, and lastfoot.  By my testing, any caption that gets typeset in TeX can contribute to the list of tables, so most of the captions should have an empty optional argument to avoid that.  In LaTeXML, we use whatever caption came last in the code, and always put it in the head.  I think it would be better to emulate what happens when a longtable ends up only taking one page: we use the firsthead and lastfoot, with the appropriate captions.  The following TeX:
```tex
\documentclass{article}
\usepackage{longtable}
\begin{document}
\listoftables
\section{The section}
one page longtable:
\begin{longtable}{c}
 \caption[first head optional caption]{first head caption}\\first head\endfirsthead
 \caption[]{other head caption}\\other head\endhead
\caption[]{other foot caption}\\other foot\endfoot
\caption[]{last foot caption}\\last foot\endlastfoot
 data
\end{longtable}
\end{document}
```
has the pdf:
Screen Shot 2022-12-07 at 12 27 47 PM
The html renders as:
Screen Shot 2022-12-07 at 12 28 19 PM

### LINKED ISSUE #2006: killing a line in longtable
One more table issue.

It appears that longtable is not implementing `\kill` correctly.  We define `\@longtable@kill` and `\@longtable@kill@marker`, but only the latter actually kills a line.  And we don't define `\kill`.  Or am I not understanding how we're coding it?  In the code
```tex
\documentclass{article}
\usepackage{longtable}
\usepackage{latexml}
\makeatletter
\newcommand{\atLongTableAtKill}{\@longtable@kill}
\newcommand{\atLongTableAtKillAtMarker}{\@longtable@kill@marker}
\makeatother
\begin{document}
\begin{longtable}{c}
not killed\\
not killed 1\kill\\
\iflatexml not killed 2\atLongTableAtKill\\\fi
\iflatexml killed\atLongTableAtKillAtMarker\\\fi
\end{longtable}
\end{document}
```
"not killed 1" should be killed but isn't.  Only "killed" is actually killed.

brucemiller:
Fix longtable `\kill`. Fix treatment of repeated `\caption` when used with \endfirsthead, etc, to prefer firsthead or lastfoot. Captions in firsthead goes before table, last foot goes after. Use \lx@ prefix for LaTeXML specific internal macros

Fixes #2006 
Fixes #2004

### DIFF
--- a/lib/LaTeXML/Package/lxRDFa.sty.ltxml
+++ b/lib/LaTeXML/Package/lxRDFa.sty.ltxml
@@ -18,11 +18,11 @@ use LaTeXML::Package;
 #======================================================================
 # Package Options
 DeclareOption('labels', sub {
-    Let(T_CS('\lxRDF@original@label'),           T_CS('\label'));
-    Let(T_CS('\lxRDF@original@longtable@label'), T_CS('\@longtable@label'));
+    Let(T_CS('\lxRDF@original@label'),             T_CS('\label'));
+    Let(T_CS('\lxRDF@originallx@longtable@label'), T_CS('\lx@longtable@label'));
     DefMacro('\label Semiverbatim', '\lxRDF@original@label{#1}\lxRDFa{property=dcterms:alternative,content=#1}');
-    DefMacro('\@longtable@label Semiverbatim',
-      '\lxRDF@original@longtable@label{#1}\lxRDFa{property=dcterms:alternative,content=#1}');
+    DefMacro('\lx@longtable@label Semiverbatim',
+      '\lxRDF@originallx@longtable@label{#1}\lxRDFa{property=dcterms:alternative,content=#1}');
 });
 
 ProcessOptions();
--- a/lib/LaTeXML/Package/longtable.sty.ltxml
+++ b/lib/LaTeXML/Package/longtable.sty.ltxml
@@ -21,12 +21,12 @@ use LaTeXML::Package;
 #======================================================================
 # Environment \begin{longtable}[align]{pattern} ... \end{longtable}
 DefMacro('\longtable[]{}',
-  '\@longtable@bindings{#2}\@@longtable[#1]{#2}\@start@alignment');
+  '\lx@longtable@bindings{#2}\@@longtable[#1]{#2}\@start@alignment');
 DefMacro('\endlongtable',
   '\@finish@alignment\@end@tabular');
 # {longtable*} is defined in revtex4-1 to be able to span a two column document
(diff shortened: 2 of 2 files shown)


lib/LaTeXML/Package/lxRDFa.sty.ltxml
```diff
@@ -18,11 +18,11 @@ use LaTeXML::Package;
 #======================================================================
 # Package Options
 DeclareOption('labels', sub {
-    Let(T_CS('\lxRDF@original@label'),           T_CS('\label'));
-    Let(T_CS('\lxRDF@original@longtable@label'), T_CS('\@longtable@label'));
+    Let(T_CS('\lxRDF@original@label'),             T_CS('\label'));
+    Let(T_CS('\lxRDF@originallx@longtable@label'), T_CS('\lx@longtable@label'));
```
dginev, line 22: Maybe the inner `originallx` renaming here was accidental? That macro already started with `\lx`.

dginev:
Interesting. So I modified the reported example to the new `\lx` naming as:
```tex
\documentclass{article}
\usepackage{longtable}
\usepackage{latexml}
\makeatletter
\newcommand{\atLongTableAtKill}{\lx@longtable@kill}
\newcommand{\atLongTableAtKillAtMarker}{\lx@longtable@kill@marker}
\makeatother
\begin{document}
\begin{longtable}{c}
not killed\\
not killed 1\kill\\
\iflatexml not killed 2\atLongTableAtKill\\\fi
\iflatexml killed\atLongTableAtKillAtMarker\\\fi
\end{longtable}
\end{document}
```

And a pdflatex run over that source outputs a single centered line with content "not killed".

This branch also outputs that line, but also adds underneath the "not killed 2".
Not sure what to make of that (I'm a bit new to `\kill`), so I thought I'd just raise the flag.

dginev (approved):
Looks good to merge (left a minor naming comment).

I once again got fooled by a reported test case and left a comment not realizing that a difference I was seeing was due to `\iflatexml` only being true in latexml and not pdflatex. Please ignore that if you see it fly by in an email notification.

lib/LaTeXML/Package/lxRDFa.sty.ltxml
brucemiller: yeah, weird renaming, but isn't actually wrong, so...

PR #2196, opened 2023-08-15
Respect listings-style semiverbatim name/file arguments

### LINKED ISSUE #2178: broken handling of underscore in file names read by `\lstinputlisting`

dginev:
Fixes #2178 .

Successor to the discussion in #2189 .

This PR now matches the original listings behavior of replacing `_`, `-` and `$` with `\textunderscore`, `\textendash` and `\textdollar`.

That keeps the XML values close to the source, as desired, avoiding any font questions.
But it also opens two new edge cases, namely filenames containing a dollar sign, or an endash `–`. The underscore case doesn't have complications, so I won't discuss it.

I modified the example to use a file named `00$–test_context.t`, which continues to fail, as it is probably wise for us to keep treating `$` as a "nasty" character in a filename and back out.

But I decided to allow `00–test_context.t`, even if it is quite strange - one would need to use the simple dash in the TeX source, but the Unicode endash `–` in the actual filename, if they are to match up.

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -2130,7 +2130,7 @@ sub FindFile_fallback {
 
 sub pathname_is_nasty {
   my ($pathname) = @_;
-  return $pathname =~ /[^\w\-_\+\=\/\\\.~\:\s]/; }
+  return $pathname =~ /[^\w\-_\+\=\/\\\.~\:\s\x{2013}]/; }
 
 sub maybeReportSearchPaths {
   if (LookupValue('SEARCHPATHS_REPORTED')) {
--- a/lib/LaTeXML/Package/listings.sty.ltxml
+++ b/lib/LaTeXML/Package/listings.sty.ltxml
@@ -118,16 +118,32 @@ DefMacroI(T_CS('\begin{lstlisting}'), 'OptionalKeyVals:LST', sub {
     lstActivate($keyvals);
     return lstProcessDisplay(lstGetTokens('name'), $text); });
 
-DefMacro('\lstinputlisting OptionalKeyVals:LST Semiverbatim', sub {
+# Read a semiverbatim name, neutralizing $, _ and -
+DefParameterType('LstNameVerbatim', sub {
+    my ($gullet) = @_;
+    $gullet->readUntil(T_BEGIN);
+    StartSemiverbatim();
+    AssignCatcode('$', CC_ACTIVE);
+    AssignCatcode('-', CC_ACTIVE);
+    AssignCatcode('_', CC_ACTIVE);
+    DefMacroI(T_ACTIVE('$'), undef, T_CS('\textdollar'),     scope => 'local');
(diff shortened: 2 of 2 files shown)


brucemiller:
yuck, this is getting worse.  I almost think we were closer originally; using verbatim for the file access, but formatting with tt (which it kinda should have been anyway, imho).

xworld21:
> using verbatim for the file access, but formatting with tt (which it kinda should have been anyway, imho).

But that won't expand macros, will it?

And to be hypercorrect, `LstNameVerbatim` should really replace the tokens `$#_`, not activate them. Otherwise the command won't be robust.

dginev:
@brucemiller I think I was gradually heading towards being `hypercorrect` as prodded by @xworld21 .

Should I reverse course or triple-down?

I'm not really enjoying doing this four different times, each of which slightly differently perverse than the other. But we seem forced into it and I can still hold my nose (or did I also duct-tape that by accident...)

brucemiller:
Maybe I've misunderstood the comments, but does this patch fail to accept & input files that `listings` would accept?  Failure to accept the same input files is a more serious problem; failure to format the filenames in a caption is much less so.

Earlier in this adventure, when we had much simpler code, but the filename in typewriter font, I was wondering if we should have a mechanism to use the typewriter font-mapping, but without the CSS to use typewriter font?  Maybe not so TeX-y, but... (See `DeclareFontMap` in `TeX.pool`: typewriter maps to different unicode than roman!).

dginev:
> does this patch fail to accept & input files that listings would accept?

Both this PR and the mainline latexml will fail to input `00$–test_context.t`, because `$` is treated as a "nasty" character and the pathname is refused. `pdflatex` will accept the file.

This patch has the additional incorrect side-effect of not recognizing the dashed case (while the mainline latexml can accept `000–test_context.t`) as it gets encoded to the unicode endash too early and expects that unicode char in the filename.

I think I will end my participation here (for now), having tried twice and only gotten into deeper weeds.

brucemiller:
FWIW, the `$` in the pathname is failing in `FindFile`, so has nothing to do with how we're playing with tokens, catcodes or arguments.

xworld21:
Let me take over this one! There is no easy answer apparently:
- *all* of the arguments of `\lstinputlisting` are read with a different set of cat codes, because internally it switches to its own 'SemiVerbatim' before even reading the optional arguments; 
- the hyphen is tricky: `@dataname` should get the original file name, but to replicate listings.sty 100% it should be rendered as en dash – in this particular case the pragmatic solution is probably to let the hyphen be.

Modulo figuring out what to do with the above, I think I know how to please everybody.

PR #2205, opened 2023-08-25
spurious characters in css file removed

evoludolab:
Some numbers crept into the ltx-book.css file that shouldn't be there and make validation to fail.


brucemiller:
Good catch! Thanks!! (Samantha did it)

Issue #2169, opened 2023-07-26
title attribute on links does not honor cleveref types


rzach:
This is extremely minor but since I noticed it I figured I'd report it: If you change the type of an item and make a link using `cleveref`, the `title` attribute (tooltip) does not reflect the change:
```
\documentclass{book}
\usepackage{latexml}
\usepackage{hyperref}
\usepackage{cleveref}
\crefname{enumi}{sentence}{sentences}
\begin{document}
\begin{enumerate}
  \item \label{one} Item
\end{enumerate}
Test: \cref{one}
\end{document}
```
results in
```
<a href="#Ch0.S0.I1.i1" title="item 1" class="ltx_ref"
```
but should result in
```
<a href="#Ch0.S0.I1.i1" title="sentence 1" class="ltx_ref"
```

brucemiller:
Interesting point; To the extent that `cref` type names are defined for an object, and to the extent that the title attribute (generated in postprocessing) wants to use a type name, I'm thinking the title should use any `cref` names available, whether or not the link was specifically one of the `cref` link commands.

OTOH, in light of #2162, I'm wondering what the title attribute for a link like the reported one should be.  The suggested one is exactly a duplicate of the link's content. So I'm guessing that in this case it would be best to omit the title completely, right?

Issue #2162, opened 2023-07-24
Redundant text in <title> and title attribute in links


rzach:
LaTeXML generates very verbose document titles (`<title>`) tags and `title=` attributes in navigation links. The `<title>` tag of a chapter is 
```
Chapter X <chapter title> ‣ <document title> ‣ <document title>
```
and if there are parts: 
```
Chapter X <chapter title> ‣ Part X <part title> ‣ <document title> ‣ Part X <part title> ‣ <document title>
```
(That's definitely a bug since the document title should at least not repeat).

A navigation link to a section, say, will have 
```
<a href=... title="Chapter X <chapter title> ‣ Part X <part title> ‣ <document title>">X Chapter title</a>
```
A screen reader reads out the document title when the page loads, and the `title` attribute in addition to the link text, so navigating a table of contents or navigation menu with a screen reader is not a good experience.

rzach:
PS Just noticed that every `\ref` will generate a link with such a verbose `title` attribute, e.g., "item X > Y.Z Section title > Chapter Y > Part U Part title > Document title".

dginev:
This is quite important, thank you for flagging it:

> A screen reader reads out the document title when the page loads, and the title attribute in addition to the link text, so navigating a table of contents or navigation menu with a screen reader is not a good experience.

rzach:
I meant the title of the page in the `<head>` section, e.g., on https://forallx.openlogicproject.org/html/Ch4.html you have
```
<!DOCTYPE html><html lang="en">
<head>
<meta http-equiv="content-type" content="text/html; charset=UTF-8">
<title>Chapter 4 First steps to symbolization ‣ Part II Truth-functional logic ‣ forall x: Calgary. An Introduction to Formal Logic‣ Part II Truth-functional logic ‣ forall x: Calgary. An Introduction to Formal Logic</title>
<!--Generated on Wed Jul  5 22:21:48 2023 by LaTeXML (version 0.8.7) http://dlmf.nist.gov/LaTeXML/.-->
```
Chapter, part, book title, part title repeated, book title repeated. Or on https://math.nist.gov/~BMiller/LaTeXML/manual/intro/ you have 
```
<title>Chapter 1 Introduction ‣ LATExml The Manual‣ LATExml The Manual</title>
```

And yes, that's what the screen reader is doing: it assumes that what's in the `title` attribute is important and so reads it out, but of course also reads out the link text.

[...]

PR #2207, opened 2023-08-30
Redundant titles

brucemiller:
This PR avoids the accidental duplication in the `<title>` element;  keeps the `@title` on html anchors avoid repeating the contents, to improve accessibility. But also provides for `link` elements to get the full, out-of-context, text.

Fixes: #2162
Fixes: #2169


lib/LaTeXML/resources/RelaxNG/LaTeXML-inline.rnc
```diff
@@ -175,10 +175,15 @@ ref_attributes =
...
   attribute show { text }?,
 
+  ## gives a description of the target, not repeating the content,
+  ## used for accessibility or a tooltip in HTML.
+  ## Typically filled in by postprocessor.
+  attribute title { text }?,
+
   ## gives a longer form description of the target,
-  ## this would typically appear as a tooltip in HTML.
+  ## usedful when the link appears outside its original context, eg in navigation.
```
dginev, line 184: `usedful` typo

dginev (approved):
Looks good!

brucemiller:
I meant to also ping @rzach to see if this worked out for him.  And also a question about appropriate style for combining title components.

In a hierarchically structured website like DLMF it's *very* helpful for `@title` to show the context of each link; so for https://dlmf.nist.gov/1.7.ii you get  "\S1.7(ii) Integrals IN \S1.7 Inequalities IN Ch. 1 Algebraic and Analytic Methods".  It seemed to me that inserting the english word "IN" was potentially confusing (and weird, if the site is not english!), so I arbitrarily chose a triangle (U+2023) as the separator.  It looks plausible to my eye, but maybe not so accessible?

That question is even more acute since we're now omitting the "\S1.7(ii) Integrals" when it's already present in the link's content. Having `@title` start with "\S1.7 Inequalities" is misleading; starting with the triangle is very weird; So, I fell back to prefixing these cases with "In ".  So, the above link would get `@title` starting "In \S1.7 Inequalities...".

So, feedback or suggestions are welcome.

brucemiller:
I don't quite understand what you're suggesting; both the spec (somewhere) and the WAVE tool are saying that `@title` is intended for accessibility (as well as tooltip elsewhere in the same spec).

dginev:
> I don't quite understand what you're suggesting; both the spec (somewhere) and the WAVE tool are saying that `@title` is intended for accessibility (as well as tooltip elsewhere in the same spec).

WAI-ARIA explicitly calls out `@title` and `@placeholder` as fallback attributes and recommends against reliance on them. They recommend the `aria-label` for the text which is always read:
https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#naming_rule_avoid_fallback

So my suggestion was to optimize `@title` for tooltip use, and set `aria-label` for accessible vocalization, when that is needed. And I am OK with holding off on that for now, as `<a>` links have decent vocalization in our current HTML output.

dginev (approved):
Looks good!

rzach:
I haven't been able to check what an actual screen reader (JAWS or NVDA) does but given what @dginev says in https://github.com/brucemiller/LaTeXML/pull/2207#issuecomment-1700242563 I fully expect that they don't exhibit the weird behavior I saw in ChromeVox. So I'm happy with it.

dginev:
Btw @rzach when you mentioned in the issue:
> that's what the screen reader (ChromeVox) is doing: it assumes that what's in the title attribute is important and so reads it out, but of course also reads out the link text. 

That felt like some sort of strange bug on ChromeVox's part. I don't have it available so can't verify (although Chromebooks are cheap, so maybe I should grab one). When I tried Chrome's Screen Reader extension (available on any OS), it didn't read out the `@title` attribute on your site or the DLMF. Which is reasonable. I would have expected a choice of *either* reading the `<a>` content, *or* reading the `@title` attribute. Reading both could be more frustrating than helpful.

rzach:
Yes several people who know about a11y and SRs subsequently to me filing the bug told me that the ChromeVox extension is not good... but I'm on Linux so I haven't yet been able to test it with a SR that is actually good. Even the built-in SR on Chromebooks might be much better.

brucemiller:
I was also going by the warnings that the WAVE extension gives (and trying to minimize them).

PR #2209, opened 2023-09-01
Listings fix

### LINKED ISSUE #2007: Implementing pipe (call command) in indices
What would it take to implement the `|` notation for indices?  It appears that `\index{term|mycommand{arg}}` should create the entry "term, \mycommand{arg}{page}" in TeX.  On the other hand, we only handle a few font changing commands (see [& process_index_phrases in LaTeX.pool.ltxml](https://github.com/brucemiller/LaTeXML/blob/0f8c2e389fe528bc7d20f838daef1110a7d7bf2f/lib/LaTeXML/Package/LaTeX.pool.ltxml#L4343)), and put the result in the styling for the indexmark.  For example:
```tex
\documentclass{article}
\usepackage{makeidx}
\makeatletter
\let\mycommand=\@firstoftwo
\makeatother
\makeindex
\begin{document}
\section{Introduction}
one\index{isbold|textbf} two\index{isemph|emph} three\index{iscustom|mycommand{15}} four\index{isplain}
\printindex
\end{document}
```
has the index:
Screen Shot 2022-12-08 at 9 48 13 AM
but we create the xml:
```xml
      <p>one<indexmark style="bold">
          <indexphrase key="isbold">isbold</indexphrase>
        </indexmark> two<indexmark style="italic">
          <indexphrase key="isemph">isemph</indexphrase>
        </indexmark> three<indexmark style="mycommand–15˝">
          <indexphrase key="iscustom">iscustom</indexphrase>
        </indexmark> four<indexmark>
          <indexphrase key="isplain">isplain</indexphrase>
        </indexmark></p>
```
and the output:
Screen Shot 2022-12-08 at 9 49 08 AM

Why am I doing this?  I'm wanting to index the line of code where keywords occur.  I've figured out how to create a custom environment that mimics an lstlisting and spots the keywords so that I end up tacking `\index{keyword|mycommand{lineno}}` onto the end of the environment.  I think it would be possible to hijack the xslt so that `style="mycommand–1234˝"` became `<a href="#lstnumberx1234">1234</a>`, but that seems to be a roundabout way to accomplish things (and I'm not good at xslt under the best of circumstances, and the index is not the best of circumstances).

### LINKED ISSUE #2178: broken handling of underscore in file names read by `\lstinputlisting`

### LINKED ISSUE #2179: \lstinputlisting ignores name for caption

brucemiller:
This PR cherry-picks the best bits of @dginev and @xworld21 patches.  Apologies for all the round-about, but thanks for the patience & ideas.

This PR also adds full numbering tags to the lines in a listing, and gives them a type-name ("line"), so that they can be reasonably referenced (also by index), so hopefully provides the  tools to answer #2007.

Closes #2179, closes #2189, closes #2196, closes #2200.

Fixes #2178
Fixes #2007

### DIFF
--- a/lib/LaTeXML/Package/listings.sty.ltxml
+++ b/lib/LaTeXML/Package/listings.sty.ltxml
@@ -118,25 +118,13 @@ DefMacroI(T_CS('\begin{lstlisting}'), 'OptionalKeyVals:LST', sub {
     lstActivate($keyvals);
     return lstProcessDisplay(lstGetTokens('name'), $text); });
 
-DefParameterType('LstNameVerbatim', sub {
-    my ($gullet) = @_;
-    # Closer to listings approch, but do NOT make - => \textendash! (ends up in file name)
-    AssignCatcode('$', CC_ACTIVE);
-    AssignCatcode('_', CC_ACTIVE);
-    DefMacroI(T_ACTIVE('$'), undef, T_CS('\textdollar'),     scope => 'local');
-    DefMacroI(T_ACTIVE('_'), undef, T_CS('\textunderscore'), scope => 'local');
-    return Tokens(Expand($gullet->readArg)); },
-  semiverbatim => 1,
-  reversion    => sub { (T_BEGIN, Revert($_[0]), T_END); });
-
-DefMacro('\lstinputlisting OptionalKeyVals:LST LstNameVerbatim', sub {
+DefMacro('\lstinputlisting OptionalKeyVals:LST Semiverbatim', sub {
     my ($gullet, $keyvals, $file) = @_;
     my $text = listingsReadRawFile($gullet, $file);
     $STATE->getStomach->bgroup;
     lstActivate($keyvals);
     my $name = lstGetTokens('name');
     $name = $file if IsEmpty($name);
-    AssignValue('LST@toctitle', $name);    # so it shows up in list of..
     return lstProcessDisplay($name, $text); });
 
 NewCounter('lstlisting', 'document', idprefix => 'LST');
@@ -176,9 +164,15 @@ sub lstProcessBlock {
     [@{ LookupValue('LISTINGS_POSTAMBLE') },
       T_END]); }
 
+our %lstFilenameRPL = (
+  '_' => T_CS('\textunderscore'), '$' => T_CS('\textdollar'), '-' => T_CS('\textendash'));
+
 sub lstProcessDisplay {
   my ($name, $text)    = @_;
   my ($body, $trailer) = lstProcessBlock($name, $text);
+  # reencode certain chars, so typewriter-like, but not typewriter font.
+  $name = Tokens(map { $lstFilenameRPL{ ToString($_) } || $_; } $name->unlist) if $name;
+  AssignValue('LST@toctitle', $name);    # so it shows up in list of..
   my @body = @$body;
   # Figure out whether the display is numbered, or has a caption or titles.
   my @caption = ();
@@ -321,7 +315,7 @@ sub listingsReadRawLines {
 
 sub listingsReadRawFile {
   my ($gullet, $file) = @_;
-  my $filename = ToString(Digest($file));
+  my $filename = ToString(Expand($file));
   my $path     = FindFile($filename);
   my $text;
   my $LST_FH;
@@ -874,7 +868,7 @@ DefKeyVal('LST', 'numberstyle',      '');
 DefKeyVal('LST', 'numbersep',        'Dimension');
 DefKeyVal('LST', 'numberblanklines', '', 'true');
 DefKeyVal('LST', 'firstnumber',      '');
-DefKeyVal('LST', 'name',             'LstNameVerbatim');
+DefKeyVal('LST', 'name',             'Semiverbatim');
 NewCounter('lstnumber');
 DefMacro('\thelstnumber', '\arabic{lstnumber}');
 
--- a/lib/LaTeXML/Package/listings.sty.ltxml
+++ b/lib/LaTeXML/Package/listings.sty.ltxml
@@ -165,13 +165,15 @@ sub lstProcessBlock {
       T_END]); }
 
 our %lstFilenameRPL = (
-  '_' => T_CS('\textunderscore'), '$' => T_CS('\textdollar'), '-' => T_CS('\textendash'));
+  '_' => T_CS('\textunderscore'), '$' => T_CS('\textdollar'),
+  ## '-' => T_CS('\textendash')  # Probably unwise, since it ends up in dataname
+);
 
 sub lstProcessDisplay {
-  my ($name, $text)    = @_;
-  my ($body, $trailer) = lstProcessBlock($name, $text);
+  my ($name, $text) = @_;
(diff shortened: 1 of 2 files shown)


dginev (approved):
Code looks very familiar :> And should be a healthy upgrade, now that the endash cases is avoided (thanks). No objections from me.

If @xworld21 feels this is useful, good to merge by me.

lib/LaTeXML/Package/listings.sty.ltxml
```diff
@@ -118,16 +118,26 @@ DefMacroI(T_CS('\begin{lstlisting}'), 'OptionalKeyVals:LST', sub {
...
     return lstProcessDisplay(lstGetTokens('name'), $text); });
 
-DefMacro('\lstinputlisting OptionalKeyVals:LST Semiverbatim', sub {
+DefParameterType('LstNameVerbatim', sub {
+    my ($gullet) = @_;
+    # Closer to listings approch, but do NOT make - => \textendash! (ends up in file name)
+    AssignCatcode('$', CC_ACTIVE);
+    AssignCatcode('_', CC_ACTIVE);
+    DefMacroI(T_ACTIVE('$'), undef, T_CS('\textdollar'),     scope => 'local');
+    DefMacroI(T_ACTIVE('_'), undef, T_CS('\textunderscore'), scope => 'local');
```
xworld21, line 127: Does this work with
```latex
\lxWithClass{class}{\lstinputlisting{a_b}}
```
? I gather `\lxWithClass` will prevent `LstNameVerbatim` from affecting the catcode and so you get the wrong name again. In this respect, `\lstinputlisting` is a robust command.

brucemiller: Yeah, right, but then *all* the proposed solutions suffer this, don't they?  `\lstinputlisting` being robust (which it apparently isn't) doesn't really affect this either, since it's the argument which already will have been read, rather than delaying the expansion of `\lstinputlisting`.

Digging in the `listings` code, I see that we've been barking up the wrong tree; Listings isn't changing catcodes (at least not for `$_-`) before reading the filename, but rather it's substituting for them before using it in the caption!  Perversely, it's much like using `\texttt` (or a non-typewriter `\texttt`)!

xworld21: > _all_ the proposed solutions suffer this, don't they?

Actually my PR does exactly what `\lstinputlisting` does, and should survive `\lxWithClass` (I admit I haven't checked). Well, I left out the `&ndash;` because it would end up in `@download`, and a minor visual discrepancy felt ok compared to saving the original name in a different variable (which would be `\lst@intname` if we follow the original code).

(My PR also does *some* catcode handling related to `=` due to babel activating `=`, which I find totally mysterious. Also a `\makeatletter` before reading the keyvals, because... uhm, so that the key names can contain `@` as letter? The listings package does some weird stuff.)

brucemiller: Indeed, the substitution was the effect; but burying it in `lstActivate` confused me.

xworld21: > burying it in `lstActivate` confused me.

I suspect I wanted to avoid the thing I mention below: listings reads the file whose name is the string of characters received from the mouth, *then* edits and saves the filename. So doing this directly in KeyVals is too early, especially if you decide you actually want the en dashes after all.

xworld21: > whose name is the string of characters received from the mouth

What did I just say... it's the _expansion_ of the string received from the mouth. The point being that listings has two distinct logical names: the physical filename, and the TOC filename, and using a KeyVals type forces them to coincide.

brucemiller:
Thanks; updated to only change the special characters for the displayed version of the filename.

I haven't done anything special for keyval's "=" (yet?); wonder if that should actually be built into keyvals in the first place (or if it's ignorable).

lib/LaTeXML/Package/listings.sty.ltxml
```diff
@@ -166,9 +164,15 @@ sub lstProcessBlock {
...
+our %lstFilenameRPL = (
+  '_' => T_CS('\textunderscore'), '$' => T_CS('\textdollar'), '-' => T_CS('\textendash'));
+
 sub lstProcessDisplay {
-  my ($name, $text) = @_;
+  my ($name, $text)    = @_;
   my ($body, $trailer) = lstProcessBlock($name, $text);
+  # reencode certain chars, so typewriter-like, but not typewriter font.
+  $name = Tokens(map { $lstFilenameRPL{ ToString($_) } || $_; } $name->unlist) if $name;
+  AssignValue('LST@toctitle', $name);    # so it shows up in list of..
```
xworld21, line 175: Let's see if I can break this one: does
```latex
\lstset{caption=\lstname}
```
behave as intended? It seems to be like `\lstname` is still incorrect. (Example from https://texdoc.org/serve/listings/0.)

lib/LaTeXML/Package/listings.sty.ltxml
```diff
@@ -864,7 +868,7 @@ DefKeyVal('LST', 'numberstyle',      '');
 DefKeyVal('LST', 'numbersep',        'Dimension');
 DefKeyVal('LST', 'numberblanklines', '', 'true');
 DefKeyVal('LST', 'firstnumber',      '');
-DefKeyVal('LST', 'name',             '');
+DefKeyVal('LST', 'name',             'Semiverbatim');
```
xworld21, line 871: This should be reverted I think.

lib/LaTeXML/Package/listings.sty.ltxml
xworld21: And it should be done before `lstProcessBlock` to fix `@dataname` as well.

brucemiller: What did you try, and what did you expect? It seems to work for me.

lib/LaTeXML/Package/listings.sty.ltxml
brucemiller: should be harmless, though, and is maybe a good hint :>

lib/LaTeXML/Package/listings.sty.ltxml
xworld21: > What did you try for `\lstname` and what did you expect? It seems to work for me.

Oh, `\lstname` is set some time in `lstProcessDisplay`, not in `lstActivate`... so it does work. I assume nobody else sets `\lstname` and/or uses `\LST@name`?

brucemiller:
OK, let's call this close enough for the moment. Thanks for all the help!

PR #2210, opened 2023-09-08
Image glitches

### LINKED ISSUE #2114: complex keyvals in \includegraphics
I was diagnozing a strange timeout during graphics post-processing for arXiv:2305.15717, which I tracked down to the single line:

```tex
\includegraphics[trim={0cm, 6cm, 0cm, 6cm}, width=1\linewidth]{figures/mturk_rating.pdf}
```

The timeout is a combination of that PDF asset having some peculiar properties, but also encounters a limitation in the current keyval option serialization (and re-parsing for graphics).

Namely, the trim component is a complex value (held in a braced group), but upon reserializing from `KeyVals::toString` it loses the braces. The `options` attribute of `ltx:graphics` that reaches post-processing is concretely the string:
```
trim=0cm, 6cm, 0cm, 6cm,width=433.62pt
```

Which is already malformed. So I am filing an issue that we should try to preserve the braces in arguments that contain commas, and we should also enhance `Post::Graphics::image_graphicx_parse` to be able to parse those arguments. Ideally then the PDF will convert normally (or we can focus on what is specifically difficult about that asset in 2305.15717).

brucemiller:
This PR fixes a couple of glitches processing `graphicx` images: the incorrect (but accepted) use of commas to separate viewport values; inconsistent up/downscaling of vector graphics images.

Fixes #2114

### DIFF
--- a/lib/LaTeXML/Package/graphics.sty.ltxml
+++ b/lib/LaTeXML/Package/graphics.sty.ltxml
@@ -22,6 +22,7 @@ use LaTeXML::Util::Image;
 # (See  LaTeXML::Post::Graphics for suggested postprocessing)
 # Package options: draft, final, hiderotate, hidescale, hiresbb
 
+# Dimensions, but \calc syntax SHOULD be allowed (if calc loaded)
 DefParameterType('GraphixDimension', sub {
     my ($gullet) = @_;
     $gullet->skipSpaces;
@@ -32,6 +33,27 @@ DefParameterType('GraphixDimension', sub {
     else {
       $gullet->unread($next);
       $gullet->readDimension; } },
+  optional => 1);
+
+# For trim, viewport: a sequence of 4 dimensions
+# These must be space separated, but apparently trailing commas are ignored.
+DefParameterType('GraphixDimensions', sub {
+    my ($gullet) = @_;
+    $gullet->skipSpaces;
+    my @dims = ();
+    while (scalar(@dims) < 4) {
+      if (scalar(@dims)) {
+        my $t = $gullet->readToken;
+        if ($t && !$t->equals(T_OTHER(','))) {
+          $gullet->unread($t); } }
+      my $s = $gullet->readOptionalSigns();
+      if (defined(my $d = $gullet->readRegisterValue('Dimension', $s, 1))) {
+        push(@dims, $d); }
+      elsif (defined($d = $gullet->readFactor())) {
+        my $unit = $gullet->readUnit() || $STATE->convertUnit('bp');
+        push(@dims, Dimension(fixpoint($s * $d, $unit))); }
+      else { last; } }
+    return LaTeXML::Core::Array->new(values => [@dims], separator => T_SPACE); },
   optional => 1);
 
 # Should probably be more clever about whether to create an ltx:inline-block vs just ltx:text
@@ -178,6 +200,10 @@ sub rotatedProperties {
     ytranslate  => $ysh,
   ); }
 
+#======================================================================
+# See here and graphicx.sty.ltxml for recognized graphicx options
+# See LaTeXML::Util::Image for internal implementation
+
 # NOTE: Need to implement the origin of rotation!
 # [code so far only reads them]
 DefKeyVal('Grot', 'origin', '');            # c,l,r,t,b,B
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Package/graphicx.sty.ltxml
```diff
@@ -34,6 +34,8 @@ DefKeyVal('Gin', 'clip',            '', 'true');
 DefKeyVal('Gin', 'scale',           '');
 DefKeyVal('Gin', 'angle',           '');
 DefKeyVal('Gin', 'alt',             '');
+DefKeyVal('Gin', 'trim',            'GraphixDimensions');
+DefKeyVal('Gin', 'vieport',         'GraphixDimensions');
```
dginev, line 38: vie-w-port typo?

lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -34,6 +35,27 @@ DefParameterType('GraphixDimension', sub {
...
+# These must be space separated, but apparently trailing commas are ignored.
+DefParameterType('GraphixDimensions', sub {
+    my ($gullet) = @_;
+    $gullet->skipSpaces;
+    my @dims = ();
+    while (scalar(@dims) < 4) {
+      if (scalar(@dims)) {
+        my $t = $gullet->readToken;
+        if ($t && !$t->equals(T_OTHER(','))) {
+          $gullet->unread($t); } }
```
dginev, line 48: minor aside: do we need a `$gullet->skipComma` method?

lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -34,6 +35,27 @@ DefParameterType('GraphixDimension', sub {
...
+      my $s = $gullet->readOptionalSigns();
+      if (defined(my $d = $gullet->readRegisterValue('Dimension', $s, 1))) {
+        push(@dims, $d); }
+      elsif (defined($d = $gullet->readFactor())) {
+        my $unit = $gullet->readUnit() || $STATE->convertUnit('bp');
+        push(@dims, Dimension(fixpoint($s * $d, $unit))); }
+      else { last; } }
+    return LaTeXML::Core::Array->new(values => [@dims], separator => T_SPACE); },
```
dginev, line 56: minor: I think `[@dims]` does a copy, while `\@dims` would pass the current ones by reference.

lib/LaTeXML/Package/graphics.sty.ltxml
```diff
@@ -244,17 +270,23 @@ sub graphicS_options {
...
+  #  my $keepaspect;
+  my ($saw_w, $saw_h);
+  while (@kv) {
+    my ($key, $value) = (shift(@kv), shift(@kv));
+    $saw_w = 1 if $key =~ /width$/;
+    $saw_h = 1 if $key =~ /height$/;
+    $value = ToString($value);
+    $value =~ s/,/\\,/g;    # Slashify any "," within the value
+    push(@options, $key . '=' . $value); }
+  push(@options, 'keepaspectratio=true') if ($saw_w xor $saw_h) && !$kv->hasKey('keepaspectratio');
```
dginev, line 288: Minor: this `$saw_w xor $saw_h` startled me, so I reread and noticed you could switch to say:
```perl
my $saw_w_or_h = ($key =~ /(?:width|height)$/);
```
and then use that directly (as the individual cases are never used on their own).

lib/LaTeXML/Util/Image.pm
```diff
@@ -336,7 +338,7 @@ sub image_graphicx_complex {
   # Get some defaults from the read-in image.
   my ($imagedpi) = image_getvalue($image, 'x-resolution');
   # image_setvalue($image,debug=>'exception');
-  my ($bg) = image_getvalue($image, 'transparent-color');
+  my ($bg) = image_getvalue($image, 'mattecolor');
```
dginev, line 341: What is the story of `mattecolor`? Just out of curiosity.

dginev (approved):
Seems great on a quick read! 

Caught a minor typo, and left a range of comments for ignorable nits :>

lib/LaTeXML/Package/graphics.sty.ltxml
brucemiller: but that's `or`, not `xor`... or did some bug creep in?

lib/LaTeXML/Package/graphicx.sty.ltxml
brucemiller: samantha did it!

lib/LaTeXML/Package/graphics.sty.ltxml
brucemiller: possibly, but seems to be used slightly differently every time?

lib/LaTeXML/Package/graphics.sty.ltxml
brucemiller: ok

lib/LaTeXML/Util/Image.pm
brucemiller: I don't know the story, but it seems that `mattecolor` is the name used by ImageMagick; not sure if `transparent-color` actually ever worked?

lib/LaTeXML/Package/graphics.sty.ltxml
dginev: It's not a bug, it's a feature :> since both are matching for an exact substring at the end of the same string, it is not theoretically possible for both conditions to be true at the same time.

PR #2211, opened 2023-09-12
Itemize fixes

### LINKED ISSUE #1645: lists are misaligned due to extra whitespace
Small styling issue:
```latex
\documentclass{article}
\usepackage[T1]{fontenc}
\begin{document}
\begin{itemize}
\item A long sentence that is not aligned properly because of additional whitespace between \texttt{<div>} and the rest.
\item To better see the misalignment.
\end{itemize}
\end{document}
```
compiles to
image

This is kind of tricky to fix properly. The only satisfying solution I have found: replace (the first) `&#x0A;` in the XSLT with a zero-width space, or word joiner (`toStringHTML`/`toString(1)` can still add newlines); add some `padding-right` to the tag to compensate for the now missing whitespace.

### LINKED ISSUE #1927: enumitem package parameters not working ?
I wrote

```
\usepackage{enumitem}

% ...

\begin{enumerate} [label=,wide=0pt,nosep]
\item item 1
\item item 2
\end{enumerate}
```

In the PDF I generate with LaTeX, it looks like:

> item 1
> item 2

However, in the book.html it looks like:

> [label=,wide=0pt,nosep]
> 1. item 1
> 2. item 2

Perhaps the `enumitem` isn't being used?

The command I used to generate:

```
latexml --dest ../book.xml main.tex
latexmlpost  -dest=../book.html  ../book.xml
```

I'm using `latexml` 0.8.5.

### LINKED ISSUE #1961: errors with nested list's
The following code
```latex
\newcounter{question}
\newcounter{subquestion}
\begin{list}{\textbf{\thequestion.}}{\usecounter{question}}
  \item One.
  \item Two.
  \begin{list}{\emph{\thesubquestion.}}{\usecounter{subquestion}}
    \item a.
    \item b.
  \end{list}
\end{list}
```
triggers something odd with the counters:
```
Error:undefined:\thequestioni@ID The token T_CS[\thequestioni@ID] is not defined. at test.tex; line 56 col 0
```
Of course the output is not quite right.

I think the culprit are `\usecounter` and `beginItemize` (LaTeXML.pool.ltxml), but I can't completely make sense of it.

### LINKED ISSUE #2155: enumitem lists don't honor label= formatting
A cloned enumerate list should format labels as "1." (with trailing dot) and references to items without the dot. They don't, even if label and ref format are set explicitly:
```
\documentclass{article}
\usepackage{enumitem}
\newlist{test1}{enumerate}{2}
\newlist{test2}{enumerate}{2}
\setlist[test2]{label=\arabic*.,ref=\arabic*}
\begin{document}

\begin{enumerate}
	\item\label{enum} Mary is in Barcelona.
\end{enumerate}
Reference: \ref{enum}
\begin{test1}
	\item\label{test1} Mary is in Barcelona.
\end{test1}
Reference: \ref{test1}
\begin{test2}
	\item\label{test2} Mary is in Barcelona.
\end{test2}
Reference: \ref{test2}
\end{document}
```
produces output:
Screenshot from 2023-07-21 15-39-48
I can get the item labels to format by saying `ref=arabic*.` but then the cross references also include the dot.

brucemiller:
This PR fixes several itemization issues

Fixes #1645
Fixes #1927
Fixes #1961
Fixes #2155

### DIFF
--- a/lib/LaTeXML/Package.pm
+++ b/lib/LaTeXML/Package.pm
@@ -1816,10 +1816,18 @@ sub DefEnvironmentI {
   my $mode = $options{mode};
   $name      = ToString($name)                    if ref $name;
   $paramlist = parseParameters($paramlist, $name) if defined $paramlist && !ref $paramlist;
+  # Magic form: CS with name \begin{env} bypasses some LaTeX
+  # However, in pure LaTeX would usually have expanded to \env
+  # and would have skipped spaces before parsing args, if any.
+  my $paramlist_skips;
+  if ($paramlist && $paramlist->getNumArgs) {
+    $paramlist_skips = LaTeXML::Core::Parameters->new(
+      LaTeXML::Core::Parameter->new('SkipSpaces', 'SkipSpaces'),
+      $paramlist->getParameters); }
   # This is for the common case where the environment is opened by \begin{env}
   my $sizer = inferSizer($options{sizer}, $options{reversion});
   $STATE->installDefinition(LaTeXML::Core::Definition::Constructor
-      ->new(T_CS("\\begin{$name}"), $paramlist, $replacement,
+      ->new(T_CS("\\begin{$name}"), $paramlist_skips, $replacement,
       beforeDigest => flatten(($options{requireMath} ? (sub { requireMath($name); }) : ()),
         ($options{forbidMath} ? (sub { forbidMath($name); }) : ()),
         sub { $_[0]->bgroup; },
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -231,14 +231,16 @@ table.ltx_eqn_eqnarray tr.ltx_eqn_lefteqn + tr td.ltx_align_right { min-width:2e
     display:block; }
 .ltx_itemize .ltx_item,
 .ltx_enumerate .ltx_item {
-    display: list-item; }
+    display: list-item; white-space:nowrap; }
 
 /* Position the tag to look like a normal item bullet. */
 li.ltx_item > .ltx_tag {
     display:inline-block; margin-left:-1.5em; min-width:1.5em;
     text-align:right; }
-.ltx_item .ltx_tag + .ltx_para,
-.ltx_item .ltx_tag + .ltx_para .ltx_p  { display:inline; }
+.ltx_item .ltx_tag + .ltx_para {
+    display:inline-block; vertical-align:top;}
+.ltx_item .ltx_tag + .ltx_para .ltx_p  {
+    display:inline; }
 
 /* NOTE: Need to try harder to get runin appearance? */
 dl.ltx_description dt { margin-right:0.5em; float:left;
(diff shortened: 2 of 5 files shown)


lib/LaTeXML/Package.pm
```diff
@@ -1816,10 +1816,18 @@ sub DefEnvironmentI {
...
   $name      = ToString($name)                    if ref $name;
   $paramlist = parseParameters($paramlist, $name) if defined $paramlist && !ref $paramlist;
+  # Magic form: CS with name \begin{env} bypasses some LaTeX
+  # However, in pure LaTeX would usually have expanded to \env
+  # and would have skipped spaces before parsing args, if any.
+  my $paramlist_skips;
+  if ($paramlist && $paramlist->getNumArgs) {
+    $paramlist_skips = LaTeXML::Core::Parameters->new(
+      LaTeXML::Core::Parameter->new('SkipSpaces', 'SkipSpaces'),
+      $paramlist->getParameters); }
```
dginev, line 1826: Hm, this is a little subtle. So `->getNumArgs` counts all the arguments which take a parametric value. But it wouldn't count a `Match:` parameter, or a `RequireBrace`, or some of the `Skip` cases.

What is the downside to instead *always* skipping the spaces here?

dginev (approved):
Looks good! 

I left a clarifying question.

lib/LaTeXML/Package.pm
brucemiller: It is subtle and a bit heuristic; trying to mimic that LaTeX would have turned `\begin{env}` into `\env` which normally would have skipped spaces, unless they've made other arrangements. If it skips too far, or in the wrong cases, we end up on the next line which throws some environments off (eg. listings).

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -231,14 +231,16 @@ table.ltx_eqn_eqnarray tr.ltx_eqn_lefteqn + tr td.ltx_align_right { min-width:2e
     display:block; }
 .ltx_itemize .ltx_item,
 .ltx_enumerate .ltx_item {
-    display: list-item; }
+    display: list-item; white-space:nowrap; }
```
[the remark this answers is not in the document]
brucemiller: Apparently doesn't. The idea is that LaTeXML's items will always have `para` and such in them, so any line breaking should be handled within them, and not by the item itself. I was concerned about those para inheriting the nowrap, but it doesn't seem to be a problem. Mysterious css.

dfrankow:
I applaud your energy to maintain a valuable package that requires so much work.

I am a little surprised to see no test cases associated with this PR. Maybe I'm missing something?

It seems really hard (I'd say impossible) to preserve such subtle behavior without test cases.

Anyway, best wishes.

dginev:
> It seems really hard (I'd say impossible) to preserve such subtle behavior without test cases.

That is a fair point in general @dfrankow , and I often tend to harp on that tune in PR reviews. Here I took the update to `t/keyval/keyvalstyle.xml` as a sufficient minimal proof that we have at least one test tracking the changes - but it is clearly partial.

Some of the original snippets in the reported issues could be converted into small tests, if we had a subtlety we thought was particularly fragile - does one come to mind here?

PR #2212, opened 2023-09-12
Enhance the itemize/enumerate tests

brucemiller:
OK, more tests! Thanks for the prodding @dfrankow


dginev (approved):
Ok, read through the PDF and XML, elements match up, square bracket refnums match up, bold and italicized number tags match up.

Yay for more tests!

dfrankow:
> OK, more tests! Thanks for the prodding @dfrankow

Ha, you're welcome.

Ironically I think I didn't see tests because I didn't know what they looked like. You must have some framework that takes the xml and makes it into tex and pdf and checks it.

Anyway, well done.

brucemiller:
Basically Perl's testing framework, with some tooling & conventions 
added. The `t` directory contains test programs (with "t" extension) 
that run all tests in a subdirectory. Most of those subtests consist of 
a .tex file and the expected .xml file (the .pdf is for informative 
comparison).

PR #2215, opened 2023-09-19
extra guards for empty lines for Mouth::readToken

### LINKED ISSUE #2175: Test fails w/ latest CTAN snapshot (babel/greek)
I checked out TeX Live another time at end of June and noticed that LaTeXML again fails to run a test using that TL snapshot.

```
t/81_babel.t ..............
1..7
ok 1 - use LaTeXML::Core;
ok 2 - t/babel/csquotes
ok 3 - t/babel/french
ok 4 - t/babel/german
not ok 5 - t/babel/greek

#   Failed test 't/babel/greek'
#   at /home/hille/latexml-0.8.8-pre/blib/lib/LaTeXML/Util/Test.pm line 120.
Wide character in print at /usr/share/perl/5.36/Test2/Formatter/TAP.pm line 125.
# Difference at line 41 for t/babel/greek
#       got : '<p>Ηερε´ς'
#  expected : '<p>Here’s'

Test Summary Report
-------------------
t/81_babel.t            (Wstat: 256 (exited 1) Tests: 7 Failed: 1)
  Failed test:  5
  Non-zero exit status: 1
Files=34, Tests=443, 524 wallclock secs ( 0.24 usr  0.12 sys + 492.41 cusr 29.91 csys = 522.68 CPU)
Result: FAIL
Failed 1/34 test programs. 1/443 subtests failed.
make: *** [Makefile:1848: test_dynamic] Error 255
```
This is reproducible with a git clone of LaTeXML I made this morning. The full build log can be seen [here](https://ci.debian.net/data/autopkgtest/unstable/amd64/l/latexml/36138938/log.gz).

Sorry for the bad news! Thanks for help.

dginev:
Fixes #2175

The regression is really subtle, so this PR definitely deserves a careful look-over. As Tim helpfully diagnosed in the issue, the underlying cause were the improvements from commit https://github.com/brucemiller/LaTeXML/commit/acaa51d9bdf4ae582c57f527fe34580290efac34 .

The remaining subtlety seemed to be revealed when babel.sty loads its english `.ini` file on a recent texlive (`babel-en.ini`). Every time that file read encounters a completely empty line, it seems to terminate the read early. Explaining *how* that happens is not something I can do with sufficient clarity at the moment, but it has to do with the interplay of `Gullet::readUntil`, `Mouth::readToken` and `Mouth::getNextChar`, where the `undef` from the empty line needs to be handled at just the right context point - or the assembled token list for the argument is incorrect, raising the higher level Errors of the regression.

I arrived at the PR contents by experiment - trying to tweak in a variety of contexts and settling on the first reasonable combination that worked out. The key change was adding a `$$self{nchars} == 0` check to the branch of `Mouth::readToken` that creates an EOL marker if a line is completely empty - tests continue to pass, and the regression is resolved.

I also changed a couple of checks relying on `$ch` to instead rely on `defined $ch` along the way, as I kept worrying about `0` chars.

### DIFF
--- a/lib/LaTeXML/Core/Mouth.pm
+++ b/lib/LaTeXML/Core/Mouth.pm
@@ -20,7 +20,7 @@ use LaTeXML::Core::Token;
 use LaTeXML::Core::Tokens;
 use LaTeXML::Util::Pathname;
 use Encode qw(decode);
-use base   qw(LaTeXML::Common::Object);
+use base qw(LaTeXML::Common::Object);
 
 our $READLINE_PROGRESS_QUANTUM = 25;
 
@@ -214,7 +214,7 @@ sub handle_escape {    # Read control sequence
   # Bit I believe that he does NOT mean within control sequences
   my $cs = "\\" . $ch;    # I need this standardized to be able to lookup tokens (A better way???)
   if ((defined $cc) && ($cc == CC_LETTER)) {    # For letter, read more letters for csname.
-    while ((($ch, $cc) = getNextChar($self)) && $ch && ($cc == CC_LETTER)) {
+    while ((($ch, $cc) = getNextChar($self)) && (length($ch) > 0) && ($cc == CC_LETTER)) {
       $cs .= $ch; }
     # We WILL skip spaces, but not till next token is read (in case catcode changes!!!!)
     $$self{skipping_spaces} = 1;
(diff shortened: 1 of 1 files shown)


dginev:
I should also mention that I have tested this on the way the regression manifested itself on `texlive 2022` (as I described in the issue) - but I haven't checked a texlive 2023 installation, where the test suite was also reported to be failing. I guess I should install that and double-check the fix is also appropriate there...

lib/LaTeXML/Core/Mouth.pm
```diff
@@ -214,7 +214,7 @@ sub handle_escape {    # Read control sequence
   # Bit I believe that he does NOT mean within control sequences
   my $cs = "\\" . $ch;    # I need this standardized to be able to lookup tokens (A better way???)
   if ((defined $cc) && ($cc == CC_LETTER)) {    # For letter, read more letters for csname.
-    while ((($ch, $cc) = getNextChar($self)) && $ch && ($cc == CC_LETTER)) {
+    while ((($ch, $cc) = getNextChar($self)) && (length($ch) > 0) && ($cc == CC_LETTER)) {
```
brucemiller, line 217: Is this really supposed to be positive length, rather than `defined`?

dginev: True, the length check is extra-paranoid (I believe I was triple-checking I am not misunderstanding the return values).

My fear was that the empty string `""` is defined, but empty. But studying `getNextChar`, there is no code path that could return `""`. So I can switch this to a "defined" test, thanks.

dginev: Done. Also rebased to master - the English babel load continues to succeed.

brucemiller:
Slightly scary, but seems right, and doesn't cause any problems that I can detect. Thanks!

PR #2216, opened 2023-09-20
allow ltx:ERROR in ltx:biblist schema

### LINKED ISSUE #2099: bibliography + undefined macro → nested bibitems
When putting the meaningless `\xyz` (or any other undefined macro) in a bibliography, the following bibitems will be _nested_.

MWE:

```latex
\documentclass{article}
\begin{document}

\begin{thebibliography}{99}
\xyz  % !
\bibitem {A} Lorem
\bibitem {B} ipsum
\end{thebibliography}

\end{document}
```

LaTeXML output:

```xml
  <bibliography xml:id="bib">
    <title>References</title>
    <biblist/>
    <ERROR class="undefined">\xyz</ERROR>
    <bibitem key="A" xml:id="bib.bib1">
      <tags>
        <tag>[1]</tag>
        <tag role="refnum">1</tag>
      </tags>
      <bibblock> Lorem
<bibitem key="B" xml:id="bib.bib2">
          <tags>
            <tag>[2]</tag>
            <tag role="refnum">2</tag>
          </tags>
          <bibblock> ipsum
</bibblock>
        </bibitem></bibblock>
    </bibitem>
  </bibliography>
```

dginev:
Fixes #2099 .

This PR allows ltx:ERROR in an ltx:biblist, allowing for more graceful recovery for bibliography content that follows an encountered error (such as an undefined macro name, as reported in the issue). 

I was wondering if I should be allowing all of Meta.class for consistency, but spotted that only the single ERROR is allowed in XMath, so there was a precedent I could follow.


dginev:
One drawback of just allowing the element is that the current stylesheets will just create a `<span>` for the error under the `<ul>`, which is technically invalid, as in

```html
<ul class="ltx_biblist">
  <span class="ltx_ERROR undefined">\xyz</span>
  <li id="bib.bib1" class="ltx_bibitem">...
```

So the code here may also need a special case rule in the XSLT that adds an `<li>` for this error case?

dginev:
Ok, some curious progress here. I was studying the different approaches to passing parametric information into XSLT templates (of which we also use several), and I think I found a convenient vehicle for creating better `ltx:ERROR` wrapping elements in HTML.

Namely, a quick test via the XPath `parent::` selector, for now recognizing the common list item scenarios, and setting the respective HTML item as the wrapping element. I also tied that together with the ability to force the choice of the element via a named `xsl:param` input.

I then also noticed we don't really handle `ltx:ERROR` explicitly when it is the first item in a `ltx:description`, and also that we auto-open a variety of unpleasant wrappers during Core conversion when an error is the first child of a `ltx:enumerate` or `ltx:itemize`. So I also allowed errors in the schema models for those list-like elements.

That allows the following larger example TeX:
```tex
\documentclass{article}
\begin{document}

\begin{itemize}
  \iyz
  \item A
  \item B
\end{itemize}

\begin{enumerate}
  \eyz
  \item A
  \item B
\end{enumerate}

\begin{description}
  \dyz
  \item[Second] The second item
  \item[Third] The third item
\end{description}


\begin{thebibliography}{3}
\xyz  % !
\bibitem {A} Lorem
\bibitem {B} ipsum
\end{thebibliography}

\end{document}
```

to now produce a mildly neater HTML structure (which passes validation) and renders as follows:
image

Should be good enough for an initial discussion I think - always open to better XSLT idioms.

brucemiller:
Well, I've seen (& written) worse XSLT, but what concerns me is the amount of complication for what seems a relatively narrow case. I wonder if there might be a better solution earlier which would "autoopen" a `bibitem` in such cases, so you'd end up with `<bibitem><ERROR.../></bibitem>`.

Normally, when `Document` is inserting an element, it will auto-open any needed elements, provided they have that `TAG_PROPERTY`.  Perhaps in the case of `ltx:ERROR` we relax that condition? (could be cans of worms in there, though)

PR #2219, opened 2023-09-26
Schema docs

brucemiller:
This PR continues on from #2214 to better organize the schema documentation. It is now hopefully clearer and more concise.  It leverages the naming conventions used in the RelaxNG schema, handles wildcards and external schema more cleanly (mostly by avoiding).

To review, you'll probably want to use
```
doc/makemanual --force --nopdf --webdir=<htmldir>
```
to put the documentation in `htmldir` for comparison to the current manual.


dginev (approved):
This seems great - I could build the full manual with the command above. Indeed the wildcards and foreign namespaces look nicer, and I am not spotting any artifacts after browsing a number of the pages.

It's worth mentioning that a couple of general changes to `bin/latexml` and `LaTeXML.css` also made their way in here, which will also be noticed in regular conversion jobs.

PR #2220, opened 2023-09-26
Better errors in bibliographies

### LINKED ISSUE #2099: bibliography + undefined macro → nested bibitems

dginev:
Fixes #2099 (using the feedback from #2216). This PR:

1. Moves `ltx:ERROR` from `Meta.class` to `Misc.class` in the schema, as errors are typically visible in the final formats, which fits better with the purpose of Misc.
2. Allows bibitem and bibblock to auto-open, as suggested by Bruce. 
   -  I tried to double-check for side-effects, and it looks like as long as `ltx:biblist` is never allowed to auto-open, this change will have limited impact, as bibitem is only allowed in a biblist, and bibblock is only allowed in its respective item.

This approach has the nice side-effect that it will group together any trailing content from an unknown macro as in `\myitem{Some content}` will now produce a self-contained:
```xml
<bibitem>
  <bibblock><ERROR class="undefined">\myitem</ERROR>Some content
  </bibblock>
</bibitem>
```

Lastly, I think it is worthwhile to quickly open the diff for `LaTeXML.model` here and check the implied cases where ERROR (dis)appeared, as the change from Meta to Misc led to some interesting differences. As just one example - `ltx:chapter` used to allow error, but now doesn't, so it auto-opens an `ltx:para` to host the error instead (for latex such as `\chapter{Test}\badmacro`).

Here is this PRs visual HTML behavior on the test from #2216 :

Aside: I also tried investigating the `Document::makeError` method that Bruce mentioned in our recent meeting, but it looks like the `TAG_PROPERTIES` are consulted quite early - already in `computeIndirectModel`, which gets cached after it is computed once. That seems a little tricky to interfere with and make exceptions for `ltx:ERROR`? Could be possible of course, but this solution did the trick as well.


brucemiller:
This looks promising. Your case for `misc` vs `meta` is fairly convincing (though `Error` is still kinda meta too). Now that that the schema is easier to read, might be a further improvement available? I worry a bit about `autoOpen` for misbehaved arXiv papers, but perhaps they'll sort themselves out no worse than before.

dginev:
> might be a further improvement available?

Do you mean as part of this PR, or as a follow-up improvement? The `autoOpen` approach is consistent with the other list environments insofar as `ltx:item` and `ltx:para` also auto-open for this kind of error in `{itemize}` and `{enumerate}`.

Aside: actually `ltx:para` isn't declared as `autoOpen`, but Document.pm auto-opens it as a special case. I just realized that while reviewing code for this comment.

But I got curious and tried one broken test with a bibliography followed by a paragraph/section/chapter. The PR approach leads to some good recovery where it won't slurp in a following paragraph as part of the bibliography, but on the downside it auto-opens empty bibitems for each empty line until the auto-close takes place:

<details>

```tex
\documentclass{article}
\begin{document}
\begin{thebibliography}{3}
  \xyz

\paragraph Trailing
\end{document}
```

```xml
  <bibliography xml:id="bib">
    <title>References</title>
    <biblist>
      <bibitem>
        <bibblock><ERROR class="undefined">\xyz</ERROR></bibblock>
      </bibitem>
      <bibitem xml:id="bib.bib1">
        <tags>
          <tag>[1]</tag>
          <tag role="refnum">1</tag>
        </tags>
        <bibblock/>
      </bibitem>
      <bibitem xml:id="bib.bib2">
        <tags>
          <tag>[2]</tag>
          <tag role="refnum">2</tag>
        </tags>
        <bibblock/>
      </bibitem>
    </biblist>
  </bibliography>
  <paragraph inlist="toc" xml:id="S0.SS0.SSS0.Px1">
    <title>T</title>
    <para xml:id="S0.SS0.SSS0.Px1.p1">
      <p>railing</p>
    </para>
  </paragraph>
```

</details>

brucemiller:
All in all, this seems pretty reasonable and doesn't cause problems that I can see in normal cases. We'll have to watch for anything pathological in arXiv bibliographies, tho it may turn out to be the best solution overall.

PR #2223, opened 2023-10-05
Improved \color parsing and scaling for custom RGB directives

dginev:
Fixes a Fatal error in [ar5iv#388](https://github.com/dginev/ar5iv/issues/388), as well as a subtle precision bug in color conversion. 

I have added the offending examples to the `t/graphics/xcolor.tex` test, to avoid regressions.

There were 3 conceptual changes here:
 - allow spaces after the comma separators in custom rgb specifications used in `\color` 
 - switch to a generalized conversion to hex via `sprintf`
 - and fix a subtle precision bug: ensure we switch to the target color scheme *before* a color is scaled. There was an interesting example with "yellow" scaled down by 0.25, which is represented via CMYK `(0,0,1,0)`, resulting in `(0,0,0.75,0)`. The issue is that if only then a conversion to the RGB model is performed, we lose the scaling information on all but the yellow axis (due to the zero components). If we instead first switch from CMYK to RGB and then scale down by 0.25, we get the exact right results as in the PDF.

This PR feels complete to me, welcoming a review.


brucemiller:
Cool! Thanks!!!

PR #2226, opened 2023-10-07
replicate framed padding and border width

### LINKED ISSUE #2224: styling inconsistencies with framed.sty
I have noticed that the framed.sty.ltxml binding differs from framed.sty in a few places, some with ugly results. At a minimum
- the default value of `\FrameSep` is three times `\fboxsep` (at package load time), not `12pt`
- the titled-frame environment has no padding at all (can be fixed using negative margins on the title span, instead of `width:100%`)
- padding inside the title span seems to be hardcoded to `8pt`
- similar considerations for `\FrameRule`

This might be all. If you don't mind, I'll just go through the code of both binding and package line by line, and write a PR with CSS only fixes.

xworld21:
Fix #2224, mostly. Remaining discrepancies:
- vertical spacing is different, but this seems unavoidable: as far as I can tell, CSS padding behaves differently than `\fboxsep`;
- LaTeXML does not indent the content, unless the box is at the start of an `<ltx:para>`; I presume this requires starting a new `<ltx:para>` for each box, but I promised a CSS only PR, and I am afraid of breaking something else.

### DIFF
--- a/lib/LaTeXML/Package/framed.sty.ltxml
+++ b/lib/LaTeXML/Package/framed.sty.ltxml
@@ -107,7 +107,8 @@ DefEnvironment('{snugshade*}', sub {
 DefEnvironment('{leftbar}', sub {
     my ($document, %props) = @_;
     $_[0]->maybeCloseElement('ltx:p');    # this starts a new vertical block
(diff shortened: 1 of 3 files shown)


brucemiller:
This looks pretty plausible (I haven't tried out much of it)

brucemiller:
Thanks!

PR #2227, opened 2023-10-07
amsthm.sty.ltxml: ignore \thmnote, \thmnumber, \thmnote if arguments are empty

### LINKED ISSUE #2225: \newtheoremstyle does not deal correctly with \thmnote in amsthm
The `amsthm` package tweaks `\newtheoremstyle` so that `\thmnote` is ignored if there is no optional argument to a theorem. LaTeXML does not deal with `\thmnote`, resulting in the following

```latex
\documentclass{amsart}
\newtheoremstyle{note}{3pt}{3pt}{}{}{\itshape}{:}{.5em}
{\thmname{#1}\thmnumber{ #2}\thmnote{ (#3)}}
\theoremstyle{note}
\newtheorem{note}{Note}
\begin{document}
\begin{note}[Title]
    A note.
\end{note}
\begin{note}
    A note without title.
\end{note}
\end{document}
```
image

Each theorem environment should instead redefine `\thmnote` to either copy or discard the input depending on whether the optional argument is empty. To be precise, `\thmname` and `\thmnumber` should also copy or discard according to the content of the corresponding argument.

Source: search for `\thmnote` in https://mirror.apps.cam.ac.uk/pub/tex-archive/macros/latex/required/amscls/amsclass.dtx

xworld21:
Fix #2225. I'll update it if tests fail (tests are too slow on my machine!).

### DIFF
--- a/lib/LaTeXML/Package/amsthm.sty.ltxml
+++ b/lib/LaTeXML/Package/amsthm.sty.ltxml
@@ -21,8 +21,7 @@ use warnings;
 use LaTeXML::Package;
 
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-Let('\@xp', '\expandafter');
-Let('\@nx', '\noexpand');
+RequirePackage('amsgen');
 
 # Core is defined in LaTeX.ltxml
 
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/amsthm.sty.ltxml
```diff
@@ -21,8 +21,7 @@ use warnings;
 use LaTeXML::Package;
 
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-Let('\@xp', '\expandafter');
-Let('\@nx', '\noexpand');
+RequirePackage('amsgen');
```
xworld21, line 24: Incidentally, amsgen.sty.ltxml defines `\@ifempty` *twice*.

brucemiller:
Cool! Thanks!!!

PR #2229, opened 2023-10-08
Handle some amsmath options (reqno,leqno,fleqn)

### LINKED ISSUE #2121: Flush equation left
I would like to have flushed left equations in HTML output. The following minimal example
```
\documentclass{article}

\usepackage[fleqn]{amsmath}

\begin{document}
\begin{equation}
  x^2 = 1
\end{equation}

\begin{align}
  x^2 &= 1\\
  x &= \pm 1
\end{align}
\end{document}

```
When compiled with
`latexml main.tex | latexmlpost --format=html -dest=main.html -`
produce horizontal center equations.

Is there any workaround?

brucemiller:
Fixes #2121

### DIFF
--- a/lib/LaTeXML/Package/amsmath.sty.ltxml
+++ b/lib/LaTeXML/Package/amsmath.sty.ltxml
@@ -41,6 +41,21 @@ RequirePackage('amsopn');
 DefMacroI('\AmSfont', undef, Tokens());
 DefMacroI('\AmS',     undef, "AmS");
 
+# Unhandled options (for now)
+foreach my $option (qw(
+  centertags tbtags
+  sumlimits nosumlimits
+  intlimits nointlimits
+  namelimits nonamelimits
+  alignedleftspaceyes alignedleftspaceno alignedleftspaceyesifneg)) {
+  DeclareOption($option, undef); }
+
+DeclareOption('reqno', sub { AssignMapping('DOCUMENT_CLASSES', ltx_leqno => undef); });
+DeclareOption('leqno', sub { AssignMapping('DOCUMENT_CLASSES', ltx_leqno => 1); });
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/amsmath.sty.ltxml
```diff
@@ -41,6 +41,21 @@ RequirePackage('amsopn');
...
+# Unhandled options (for now)
+foreach my $option (qw(
+  centertags tbtags
+  sumlimits nosumlimits
+  intlimits nointlimits
+  namelimits nonamelimits
+  alignedleftspaceyes alignedleftspaceno alignedleftspaceyesifneg)) {
+  DeclareOption($option, undef); }
+
+DeclareOption('reqno', sub { AssignMapping('DOCUMENT_CLASSES', ltx_leqno => undef); });
```
dginev, line 53: Shouldn't the false toggle here have value `0` instead of `undef` since it was explicitly set? 

And relatedly, since we have classes `ltx_leqno` and `ltx_fleqn`, one could imagine a (no-op for now) class `ltx_reqno`.

dginev (approved):
Looks good, with one comment.

lib/LaTeXML/Package/amsmath.sty.ltxml
brucemiller: Equation numbers on the right (and the associated CSS) are the default, as are centered equations, so there's (currently) no `ltx_reqno` or `ltx_ceqn`(?).  If there was such classes, you'd kind end up *requiring* one of each class to be in your document, else the math & numbering position would be unspecified.

You could indeed set the toggle to 0, but you'd want to do it in all places where it's set, since it somewhat implies there's a distinction (which you have to be careful about in Perl).

dginev: Alright, thanks for the explanation.

PR #2232, opened 2023-10-12
Patch restricted horizontal again

### LINKED ISSUE #1994: tcolorbox and double dollar signs
I think this ends up being a minor error, but I've come across a strange interaction between `$$` and `tcolorbox`:
```tex
\documentclass{article}
\usepackage{tcolorbox}
\begin{document}
$$a^b$$
\begin{tcolorbox}
\[a^b\]
\end{tcolorbox}
\begin{tcolorbox}
$$a^b$$
\end{tcolorbox}
\end{document}
```
Not only does the last one give "Error:unexpected:^ Script ^ can only appear in math mode", but the equation disappears from within the box.  (I am aware that `$$` isn't proper LaTeX, but there seems to be something else going on here as well.)

brucemiller:
minipage (and `\parbox`) act in vertical mode, which (should have) overrides the (fake) restricted horizontal mode of `\hbox`.   This fixes that.
Fixes #1994

The underlying issue is that `$` should look for a second `$` only when *not* in restricted-horizontal mode.  We really need to upgrade to the proper 6 processing modes, but that should be in the next release.


dginev (approved):
Curious - looks good!

I only wish the dollar.tex tests used `\mathrm{displaymath}` to minimize the XML size.

brucemiller:
Funny; I found it annoying while testing, but needed prodding to change it.

Issue #2204, opened 2023-08-23
lipsum having problems


teepeemm:
The following:
```tex
\documentclass{article}
%\usepackage{expl3}
\usepackage{lipsum}
\begin{document}.\end{document}
```
fails after 10 minutes with 100 errors (that are similar to the errors we were having before we got expl3 working).  More strangely, if I `\usepackage{expl3}` before lipsum, then LaTeXML gets stuck in an infinite loop after 10 minutes, with the first error being that `\ExplSyntaxOn` is not defined.  In both cases, it seems like LaTeXML has forgotten some aspect of expl3.  Unfortunately, I'm having trouble figuring out how that happened.

I've boiled the infinite loop down to
```perl
RequirePackage('expl3');
LoadPool('expl3');
```
The infinite loop happens while trying to `InputDefinitions('expl3', type => 'ltx', noltxml => 1);`, but if I call that instead of `LoadPool`, then it works fine.

Without `\usepackage{expl3}`, lipsum seems to get through loading expl3 alright, but then starts having trouble with intarrays as soon as it tries to do something more than defining new functions.

dginev:
@teepeemm just in case, could you also add the output of `tex --version` here, as well as the exact snippet that fails?

The one above succeeds for me with the latest latexml master, on texlive 2022:
```
$ tex --version
TeX 3.141592653 (TeX Live 2022)
kpathsea version 6.3.4
```

teepeemm:
```
% tex --version  
TeX 3.141592653 (TeX Live 2023)
kpathsea version 6.3.5
```

```tex
\documentclass{article}
\usepackage{expl3}
\usepackage{lipsum}
\begin{document}.\end{document}
```

LaTeXML starts with
```
latexml (LaTeXML version 0.8.7) processing lipsumtest
Package: expl3
```
then pauses for 10 minutes, then
```
Error:undefined:\ExplSyntaxOn The token T_CS[\ExplSyntaxOn] is not defined. at expl3.ltx; line 75 col 0
Error:undefined:\bool The token T_CS[\bool] is not defined. at expl3.ltx; line 76 col 7
Error:unexpected:_ Script _ can only appear in math mode at expl3.ltx; line 76 col 9
Error:unexpected:_ Script _ can only appear in math mode at expl3.ltx; line 76 col 14
Error:undefined:\g The token T_CS[\g] is not defined. at expl3.ltx; line 76 col 22
[...]

PR #2234, opened 2023-10-13
only run expl3-dependent tests in continuous integration

dginev:
This PR is essentially a repeat on my first attempt in #2045, and is one way to interpret the sentiment I recently expressed in #2175 and #2204.

Namely, I suggest we temporarily only run expl3-related tests in continuous integration, allowing systems where latexml's expl3 emulation is broken to still have a clean/successful installation. Those are currently two tests - glossary and si. Also, this is specifically targeted to the `0.8.8` release of latexml.

For the next, `0.8.9` release, I think Bruce has shown some convincing prototypes where both the performance in loading expl3 can be improved, as well as the robustness in interpretation, but that will require some sweeping changes to the pool-related internals of Core processing.

P.S. Attached with taking this route would be a suggestion to defer resolving #2204 until after the more ambitious approach to expl3 loading is in place.

### DIFF
--- a/lib/LaTeXML/Util/Test.pm
+++ b/lib/LaTeXML/Util/Test.pm
@@ -86,12 +86,14 @@ sub check_requirements {
     next unless $reqmts;
     my @required_packages = ();
     my $texlive_min       = 0;
+    my $required_env;
     if (!(ref $reqmts)) {
       @required_packages = ($reqmts); }
     elsif (ref $reqmts eq 'ARRAY') {
       @required_packages = @$reqmts; }
     elsif (ref $reqmts eq 'HASH') {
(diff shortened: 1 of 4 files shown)


brucemiller:
Yeah, sadly it seems like it's time to do this. Hopefully we'll be able to back off soonish.

But since you've "formalized" the `CI` switch, shouldn't we tweak the existing `CI` tests to use the same tagging (to avoid surprises later)?  Those seem to be `t/83_expl3.t, t/170_grammar_coverage.t, t/97_manifest.t`

dginev:
Sure. I rebased to master and switched `t/83_expl3.t` to the new idiom.

But the other two cases - `t/170_grammar_coverage.t` and `t/97_manifest.t` are not testing individual `.tex` files, and thus do not rely on `LaTeXML::Util::Test`, so they can't really be unified more? At least I am not sure how that would look.

brucemiller:
Cool! Thanks!

PR #2236, opened 2023-10-15
Color tweaks

### LINKED ISSUE #1945: `\color` in formulas when tikz is loaded
I convert the following with LaTeXML 0.8.6; revision ffaa301 and TeXLive 2021:
```
\documentclass{article}
\usepackage{tikz}
\begin{document}
${\color{red}a}$
\end{document}
```
what results in:
```
...<Math mode="inline" tex="{\color[rgb]{1,0,0}\definecolor[named]{pgfstrokecolor}{rgb}{1,0,0}%&#10;\pgfsys@color@rgb@stroke{1}{0}{0}\pgfsys@color@rgb@fill{1}{0}{0}a}" text="a" xml:id="p1.m1">...
```
Now, I cannot process this with latexmlpost to get mathimages because `\color` is expanded by tikz to @-commands.
Can you add something to the bindings that avoids this expansion?

brucemiller:
The actual issue fix was a simple one liner in the pgf driver, but it turned up some inconsistency in color defaulting. The most interesting bit is the test for black in `TeX.pool` which affects how color adapts to a box being moved, copied, etc.  The rest is just consistency that `black` becomes `Black()`, and thus `#000000`.

Fixes #1945

### DIFF
--- a/lib/LaTeXML/Common/Font.pm
+++ b/lib/LaTeXML/Common/Font.pm
@@ -20,8 +20,9 @@ use LaTeXML::Common::Number;
 use LaTeXML::Common::Dimension;
 use LaTeXML::Common::Font::Metric;
 use LaTeXML::Common::Font::StandardMetrics;
+use LaTeXML::Common::Color;
 use List::Util qw(min max sum);
-use base qw(LaTeXML::Common::Object);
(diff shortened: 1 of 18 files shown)


dginev (approved):
Seems reasonable!

PR #2237, opened 2023-10-16
pgfmath and pgfplots enhancements

dginev:
This is a draft PR, which may already benefit from some visibility/discussion.

I started with a motivating issue [ar5iv#390](https://github.com/dginev/ar5iv/issues/390), which caught an important oversight with the perl-eval optimization for pgfmath. Namely, the `^` is a valid char in both pgf and perl, but it means binary-xor in perl, and a classic power in pgf.

I also spotted the reports of the precision errors, where a `$T_\x$` was getting typeset as $T_{0.000000}$ instead of $T_0$, where the `\x` was computed via pgfmathparse. Which lead me down a deeper rabbit hole, and added the [falling dominoes](https://texample.net/tikz/examples/dominoes/) test to my tests. That example also revealed that expressions such as `3*sin \x` are validly parse-able - here as `3*(sin(\x))` - so I added a grammar case for a function call on a bare argument.

The PR now includes a first pgfmath unit test, which is gradually growing in coverage of the math operators, in order to try and get closer to the precision and eval-accuracy of the native pgf driver.

I have some more work in increasing that test's coverage, and then I'd like to try some more diverse tests, to ensure I didn't cause some unexpected regressions. But the original issue in the ar5iv report should already be patched here.


dginev:
The dominoes test also renders perfectly now. Which allows the distraction of noticing text size (= domino numbers) is a just a little large. That is also the case for the axes labels in the pgfplots examples. 

In ar5iv, I currently have a CSS directive anchored at `foreignObject .ltx_text` and I seem to get really good outcomes for the handful of tests here if I add a property `font-size: 0.9em`.

But I am recording that only as an informal comment for now, since I am unsure if there is a general sizing distinction, or this is limited to a few special cases of examples.

dginev:
The last commit [ccd423f](https://github.com/brucemiller/LaTeXML/pull/2237/commits/ccd423f5fcb5d2c35ad2abdc37fa8d2eaa7669ce) fixes https://github.com/dginev/ar5iv/issues/318 and https://github.com/dginev/ar5iv/issues/402

t/tikz/ac-drive-voltage.xml
```diff
@@ -0,0 +1,119 @@
...
...0.8pt}\pgfsys@invoke{ }\pgfsys@setdash{3.0pt,3.0pt}{0.0pt%&#10;}\pgfsys@invoke{ }{}\pgfsys@moveto{0.0pt}{85.35828pt}\pgfsys@lineto{56.90552pt%&#10;}{85.35828pt}\pgfsys@curveto{69.7096pt}{72.5542pt}{113.81104pt}{28.45276pt}{17%&#10;0.71655pt}{28.45276pt}\pgfsys@stroke\pgfsys@invoke{ }&#10;\pgfsys@invoke{\lxSVG@closescope }\pgfsys@endscope{}{{}}{}{{}}{}\pgfsys@moveto%&#10;{71.1319pt}{85.35828pt}\pgfsys@stroke\pgfsys@invoke{ }\hbox{\hbox{{%&#10;\pgfsys@beginscope\pgfsys@invoke{ }{{}{}{{&#10;{}{}}}{&#10;{}{}}&#10;{{}{{}}}{{}{}}{}{{}{}}&#10;{&#10;}{{{{}}\pgfsys@beginscope\pgfsys@invoke{ }\pgfsys@transformcm{1.0}{0.0}{0.0}{1%&#10;.0}{65.9305pt}{82.54439pt}\pgfsys@invoke{ }\hbox{{\definecolor{pgfstrokecolor}%&#10;{rgb}{0,0,0}\pgfsys@color@rgb@stroke{0}{0}{0}\pgfsys@invoke{ }%&#10;\pgfsys@color@rgb@fill{0}{0}{0}\pgfsys@invoke{ }\hbox{{$\varPsi_{s}$}}&#10;}}\pgfsys@invoke{\lxSVG@closescope }\pgfsys@endscope}}}&#10;\pgfsys@invoke{\lxSVG@closescope }\pgfsys@endscope}}}&#10;\par&#10;\pgfsys@invoke{\lxSVG@closescope }\pgfsys@endscope{}{}{}\hss}%&#10;\pgfsys@discardpath\pgfsys@invoke{\lxSVG@closescope }\pgfsys@endscope\hss}}%&#10;\lxSVG@closescope\endpgfpicture}}" width="303.8" xml:id="p1.pic1">
```
dginev, line 12: While studying the XML in the tests here, I have been wondering if we should add a switch to disable the `tex` attribute in the `<picture>` element, especially when they come from some pgf/tikz source.

The expanded low-level pgf instructions get overwhelming quite quickly, and contain LaTeXML-only macros (such as `\lxSVG@close`) so they won't be reusable in other, non-latexml, contexts.

But even if we keep them in general, I can imagine various spurious diffs in these tests later on, e.g. if we refactor internal macro names.

dginev: Ah, interesting: all the more so since the exact source snippets differ between texlives. I generated the tests on texlive 2022, and our 2018-2021 CI versions all failed on the source line.

dginev:
@brucemiller the newly added pgf and tikz tests are now part of the `extratests` target and only run with:
```
CI=true make extratest
```

That should allow some freedom to move until we have figured out how to make the XML easy to regenerate when texlives switch. The current XML was generated under texlive 2022.

Also, here is a shortlist of notes that are still not fully polished in the Silviu unit test, which some of us may return to when we re-focus on Tikz again:
- `t/tikz/unit_tests_by_silviu.t`
  - 4 Paths, pgfclosepath, left triangle should have a sharp top end 
  - 5 Decorations, first example should be fully wavy
  - 8 Nodes, Use Case, width of "Use Case" should match circle width (too narrow)
  - 8 Nodes, 2nd example, "qe", "qb" etc text should be centered in their circles (currently close to top)
  - 8 Nodes, "More: [integral]" width should match red rectangle width (other nodes have same problem)
  - 11 Patterns, missing stars
  - 14 Shadings, 3 square "applications" examples, inner shapes should be centered.
  - 16 Transparency - red rectangle background shouldn't be visible
  - 18 Fun Stacking - the black line isn't inside its square?

I have rebased to master, and am largely out of time to keep fine-tuning here, so please consider this ready for review.

brucemiller (approved):
Scary, but seems right. Thanks!!

PR #2249, opened 2023-10-26
respect \@sanitize guard for \index 

dginev:
Fixes [ar5iv#393](https://github.com/dginev/ar5iv/issues/393), 

Minimal motivating example:
<details>

```tex
\documentclass{article}
\usepackage{makeidx}
\makeindex
\begin{document}
\def\w{w}

\begin{itemize}
\item A\index{Choice Principle!({{\sf AC$^\{<\w}_\w$}})}

\item B\index{Choice Principle!({{\sf BC$^{<\w}_\w$}})}
\end{itemize}

\printindex

\end{document}
```

</details>

---

In a way this is an edge case problem in LaTeX documents which have an actual error in the argument of the `\index` command. The arXiv example is a particularly egregious case, where a big portion of a 70 page manuscript goes haywire due to a single malformed index argument.

pdflatex is largely immune to such problems, as the argument to `\index` is neutralized via `\@sanitize` and written in the auxiliary `.idx` file for second-stage processing. The `makeidx` binary can then (relatively quietly) veto malformed arguments, avoiding any errors in the main pdflatex workflow.

To this end, this PR adds the sanitization guard via a new parameter type - which then also retokenizes back to Plain catcodes. I then add the usual balanced-argument check, offering a warning in cases where the `\index` argument was ill-formed + discarding the entry. This matches pdflatex+makeidx "in spirit".

One tricky detail that was revealed by the tests is that if we have `DefMacro` bindings that expand into `\index`, there is some care needed for the tokens not to get mangled. There appear to be some subtle details around re-tokenizing spaces that I am not too certain about (they have to do with space skipping after a command sequence is completed). I wonder if I can implement the parameter type in a way that is more compatible for binding reuse.

Feedback welcome.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -4422,9 +4422,22 @@ our %index_style = (textbf => 'bold', bf => 'bold', textrm => '', rm => '',
 sub process_index_phrases {
   my ($gullet, $phrases, $inlist) = @_;
   my @expansion = ();
+  my @tokens    = $phrases->unlist;
+  # check we have a well-formed argument
+  return unless @tokens;
+  my $group_level = 0;
+  for my $t (@tokens) {
+    my $cc = $t->getCatcode;
+    if    ($cc == CC_BEGIN) { $group_level++; }
+    elsif ($cc == CC_END) {
+      $group_level--;
+      # discard if unbalanced close group;
+      last if ($group_level < 0); } }
+  if ($group_level != 0) {    # if a group is still open by the end, ill-formed, discard;
+    Warn("malformed", "indexentry", $gullet,
+      'index entry has unbalanced groups, discarding: "' . ToString($phrases) . '"');
+    return; }
   # Split the text into phrases, separated by "!"
-  my @tokens = $phrases->unlist;
-  return                      unless @tokens;
   push(@tokens, T_OTHER('!')) unless $tokens[-1]->getString eq '!';    # Add terminal !
   my @phrase = ();
   my @sortas = ();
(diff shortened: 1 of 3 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -4418,9 +4418,22 @@ our %index_style = (textbf => 'bold', bf => 'bold', textrm => '', rm => '',
...
+    my $cc = $t->getCatcode;
+    if    ($cc == CC_BEGIN) { $group_level++; }
+    elsif ($cc == CC_END) {
+      $group_level--;
+      # discard if unbalanced close group;
+      last if ($group_level < 0); } }
+  if ($group_level != 0) {    # if a group is still open by the end, ill-formed, discard;
+    Warn("malformed", "indexentry", $gullet,
+      'index entry has unbalanced groups, discarding: "' . ToString($phrases) . '"');
+    return; }
```
brucemiller, line 4435: Have you seen `$tokens->isBalanced` ?

lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -4458,7 +4471,37 @@ sub process_index_phrases {
...
+    # where the \index argument was malformed for one reason or another.
+    #
+    # the strangeness comes from the original TeX workflow requiring multiple conversion calls,
+    # alongside a call to the `makeidx` binary, which we don't do in latexml. This parameter type
+    # emulates one important aspect implied by those steps.
+    $arg = Tokenize(ToString($arg));
+    return $arg; },
+  beforeDigest => sub {
+    $_[0]->bgroup;
+    MergeFont(family => 'typewriter'); },
```
brucemiller, line 4496: this will only have effect if the parameter is used in a Constructor; are you anticipating this?

lib/LaTeXML/Package/listings.sty.ltxml
```diff
@@ -1019,7 +1019,9 @@ DefKeyVal('LST', 'indexstyle', '');
 DefMacro('\lst@@indexstyle [Number] Until:\end', sub {
     lstSetClassStyle(lstClassName('index', $_[1]), $_[2]); });
 
-DefMacro('\lstindexmacro{}', '\index{{\ttfamily #1}}');
+# use the simple plain-argument index,
+# otherwise we need to double-check any relevant Semiverbatim cases
+DefMacro('\lstindexmacro{}', '\lx@simple@index{{\ttfamily #1}}');
```
brucemiller, line 1025: Does your safe index not actually work in the listings case? That would seem troubling.

lib/LaTeXML/Package/LaTeX.pool.ltxml
dginev: Excellent reminder, and it even allowed me to improve that method a tiny bit.

lib/LaTeXML/Package/LaTeX.pool.ltxml
dginev: I was not, thank you for raising awareness. But it also was cargo-culted without good reason (from `HyperVerbatim`), so I could erase it safely.

lib/LaTeXML/Package/listings.sty.ltxml
dginev: It was quite troubling indeed. Fixed now!

dginev:
Alright, I managed to catch all the subtleties I got wrong I believe.

Double-checking arXiv:2006.01613 I am again seeing a healthy index, and all tests pass. Ready for review.

brucemiller (approved):
Looks good! Thanks!

PR #2269, opened 2023-11-22
pass package options in the right order

dginev:
Hello! Well I sure found an interesting example of what happens when argument lists get shuffled unintentionally.

Apparently `\PassOptionsToPackage` and `\PassOptionsToClass` have never really worked, as their arguments had been swapped.

I landed on this discovery when diagnozing why [ar5iv:2310.13021](https://ar5iv.labs.arxiv.org/html/2310.13021) used AuthorYear citation styles rather than the Number style used in the PDF. The source revealed the lines:

```tex
% if you need to pass options to natbib, use, e.g.:
%     \PassOptionsToPackage{numbers, compress}{natbib}
% before loading neurips_2023
\PassOptionsToPackage{numbers, compress}{natbib}
% ready for submission
% \usepackage{neurips_2023}

\usepackage[preprint]{neurips_2023}
```

which appears to have been supplied by the neurips conference organizers. Straightening up the argument order in the binding definition fixes the problem.

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -933,13 +933,13 @@ DefPrimitive('\DeclareOption{}{}', sub {
         DeclareOption(ToString($option), $code)); });
 
 DefPrimitive('\PassOptionsToPackage{}{}', sub {
-    my ($stomach, $name, $options) = @_;
+    my ($stomach, $options, $name) = @_;
     $name = ToString($name);
     $name =~ s/\s+//g;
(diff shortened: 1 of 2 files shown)


dginev:
Also snuck in a fix for the same kind of problem due to a mildly incomplete binding for elsarticle.cls. The last commit [a7fcfa6](https://github.com/brucemiller/LaTeXML/pull/2269/commits/a7fcfa686eae567af4f41e8b5ab7db0491d7ef6e) should default to the numbers style, but still allow using the `authoryear` option to set an alternative.

brucemiller (approved):
Cool! Thanks!

PR #2273, opened 2023-11-29
add \ion for mnras.cls and friends

dginev:
I got pinged that this was missing and it seemed faster to make the PR than an issue.

mnras.cls has a recent addition of `\ion`, which I transferred as-is to the binding.

Will fix arXiv:2311.17062


brucemiller (approved):
OK, thanks!

PR #2281, opened 2023-12-08
Digest some keys in thmtools; avoid error cascades in \hypertarget 

dginev:
By the time I had checked the latest reported problem in [ar5iv#412](https://github.com/dginev/ar5iv/issues/412) I had it fixed, so filing the two extra `Digest` calls in this PR.

Separately, I am also filing a small softening of the `\hypertarget` error of inserting an `ltx:anchor` to a Warning, since it turns out when it occurs it can be really frequent (100+ times). We stumbled onto that when converting the [Linear Algebra textbook by Hefferon](https://hefferon.net/linearalgebra/) this week. I will probably file an issue to bookkeep some of the more subtle problems there, but this one seems clearcut.


brucemiller (approved):
Thanks!

PR #2285, opened 2023-12-18
Fix relaxng

brucemiller:
This PR patches a small oversight #2219 generating the simplified schema model when additions to a symbol are combined with a symbol not initially declared as a combination.

[A separate PR since it belongs to #2219, but needed for the next PR]


dginev (approved):
LGTM

PR #2287, opened 2023-12-21
Fix centering

### LINKED ISSUE #1712: \centering escaping its group
In the tex document
```tex
\documentclass{article}
\begin{document}
This should not be centered.

{\centering
This is the only thing that should be centered.\\
}

This should not be centered.
\end{document}
```
The `{}` aren't forming a group in the xml (nor does `\bgroup/\egroup` nor `\begingroup/\endgroup`), so that the `.ltx_align_center` escapes and applies to everything on the page.

brucemiller:
This PR fixes the effects of `\centering` (and related) leaking outside their groups.

Fixes #1712

### DIFF
--- a/lib/LaTeXML/Package/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Package/LaTeX.pool.ltxml
@@ -1377,11 +1377,11 @@ sub applyAligningContext {
   my ($document, $align, $class) = @_;
   if (my $container = LookupValue('ALIGNING_NODE')) {
     my ($node, $previous) = @$container;
-    my @children = $node->childNodes($node);
+    my @children = $node->childNodes;
     while (my $skip = shift(@children)) {
       last if !$previous || $previous->isSameNode($skip); }
     while (my $child = shift(@children)) {
-      setAlignOrClass($document, $child, $align, $class); } }
+      setAlignOrClass($document, $child, $align, $class) if $child->nodeType == XML_ELEMENT_NODE; } }
   return; }
 
 DefConstructorI('\centering', undef, \&setupAligningContext,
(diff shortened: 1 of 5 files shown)


lib/LaTeXML/Package/LaTeX.pool.ltxml
```diff
@@ -1367,42 +1367,41 @@ DefEnvironment('{flushright}', sub {
...
+  my ($document) = @_;
+  my $node = $document->getElement;
+  AssignValue(ALIGNING_NODE => [$node, $node->lastChild]);
+  return; }
+
+sub applyAligningContext {
+  my ($document, $align, $class) = @_;
+  if (my $container = LookupValue('ALIGNING_NODE')) {
+    my ($node, $previous) = @$container;
+    my @children = $node->childNodes($node);
```
dginev, line 1380: `childNodes()` doesn't take arguments right? Likely has no effect, but threw me off.

dginev: A related detail is that the previous code used `$document->getChildElements($node)` instead of `childNodes`, the difference being whether text nodes are excluded/included.

`setAlignOrClass` executes `setAttribute` on the input node, so I think it only expects element nodes.

dginev (approved):
Looks healthy! Left a note on element nodes.

lib/LaTeXML/Package/LaTeX.pool.ltxml
brucemiller: Extra arg has no effect for now, but will be confusing later.
We do want text nodes here, since that might be the marker node where we start marking. But you're right that we need to protect `setAlignOrClass`.

PR #2292, opened 2023-12-31
Mathvariant fixes

### LINKED ISSUE #2021: Unneeded mathsize in munder for large operators?
I noticed a visual glitch which may be something to examine for the next release. In tabulated range arguments for large operators, such as the minimal TeX:

```tex
$$ \sum_{\begin{array}[]{c}i_{1}\ldots i_{n}\geq 0\\ i_{1}+\cdots+i_{n}=n\end{array}} 1 $$
```

We seem to annotate the final MathML token elements with an explicit "mathsize" that scales them up, as in (generated via latexmlmath):

<details>

```xml
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="\sum_{\begin{array}[]{c}i_{1}\ldots i_{n}\geq 0\\&#10;i_{1}+\cdots+i_{n}=n\end{array}}1" display="block">
  <mrow>
    <munder>
      <mo movablelimits="false">∑</mo>
      <mtable rowspacing="0pt">
        <mtr>
          <mtd>
            <mrow>
              <mrow>
                <msub>
                  <mi mathsize="142%">i</mi>
                  <mn mathsize="140%">1</mn>
                </msub>
                <mo>⁢</mo>
                <mi mathsize="142%" mathvariant="normal">…</mi>
                <mo>⁢</mo>
                <msub>
                  <mi mathsize="142%">i</mi>
                  <mi mathsize="140%">n</mi>
                </msub>
              </mrow>
              <mo mathsize="142%">≥</mo>
              <mn mathsize="142%">0</mn>
            </mrow>
          </mtd>
        </mtr>
        <mtr>
          <mtd>
            <mrow>
              <mrow>
                <msub>
                  <mi mathsize="142%">i</mi>
                  <mn mathsize="140%">1</mn>
                </msub>
                <mo mathsize="142%">+</mo>
                <mi mathsize="142%" mathvariant="normal">⋯</mi>
                <mo mathsize="142%">+</mo>
                <msub>
                  <mi mathsize="142%">i</mi>
                  <mi mathsize="140%">n</mi>
                </msub>
              </mrow>
              <mo mathsize="142%">=</mo>
              <mi mathsize="142%">n</mi>
            </mrow>
          </mtd>
        </mtr>
      </mtable>
    </munder>
    <mn>1</mn>
  </mrow>
</math>
```

</details>

If the `{array}` construct appeared on the baseline instead of a `\sum` subscript, the mathsizes are not added.

From what I can see, on today's Firefox and Chrome 108+MathML, simply removing the `mathsize` attribute entirely provides a healthy and good-looking rendering. Is that a possible easy refactor here?

### LINKED ISSUE #2051: [MathML] remove reliance on mathvariant
With MathML Core [issue#182](https://github.com/w3c/mathml-core/issues/182), we no longer have mathvariant available, and newly prefer a combination of Unicode and CSS for this kind of styling.

An example of the previous mixed state of affairs is available in [ar5iv#184](https://github.com/dginev/ar5iv/issues/184).

I suspect we should add a selection of `math_variant` (or similarly named) CSS classes for the cases where Unicode does not have available codepoints.

brucemiller:
First step, clean up erroneous, missing and/or redundant font declarations which lead to erroneous or redundant mathvariant declarations in the MathML. Typically happens when the appropriate font is not propagated while synthesizing symbols.

Then add classes, like `ltx_mathvariant_italic`, for cases where the mathvariant cannot be handled by remapping the Unicode.

Fixes #2021 
Fixes #2051

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -731,6 +731,9 @@ sub stylizeContent {
     $class = ($class ? $class . ' ' : '') . 'ltx_font_oldstyle'; }
   elsif ($font =~ /smallcaps/) {
     $class = ($class ? $class . ' ' : '') . 'ltx_font_smallcaps'; }
+  elsif ($variant) {    # Any left-over mathvariant? Punt to CSS
+    $class = ($class ? $class . ' ' : '') . 'ltx_mathvariant_' . $variant; }
+
   if ($opacity) {
     $cssstyle = ($cssstyle ? $cssstyle . ';' : '') . "opacity:$opacity"; }
 
@@ -738,8 +741,8 @@ sub stylizeContent {
   my %props = ($tag eq 'm:mo' ? opdict_lookup($text, $role) : ());
 
   # Resolve stretch & size
-  $stretchy = undef if ($tag ne 'm:mo');                          # Only allowed on m:mo!
-  $size     = undef if $stretchy;                                 # Ignore size, if we're stretching.
+  $stretchy = undef if ($tag ne 'm:mo');             # Only allowed on m:mo!
+  $size     = undef if $stretchy;                    # Ignore size, if we're stretching.
   my $stretchyhack = undef;
   if ($text =~ /^[\x{2061}\x{2062}\x{2063}]*$/) {    # invisible get no size or stretchiness
     $stretchy = $size = undef; }
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -721,8 +721,8 @@ sub stylizeContent {
     $variant = ($plane1hack && ($variant ne $u_variant) && ($variant =~ /^bold/)
       ? 'bold' : undef); }                       # Possibly keep variant bold
                                                  # Use class (css) to patchup some weak translations
-  if ($text =~ /^\p{Format}*$/) {                # Formatting (eg. InvisibleTImes)
-    $font = $variant = $color = $bgcolor = $opacity = undef; }
+  if ($text =~ /^\p{Format}*$/) {                # Only Formatting (eg. InvisibleTImes) (or empty)
+    $font = $variant = $color = $bgcolor = $opacity = undef; }    # Needs no viz. styling attributes
   elsif (!$font) { }
   elsif ($font =~ /caligraphic/) {
     # Note that this is unlikely to have effect when plane1 chars are used!
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -715,13 +715,15 @@ sub stylizeContent {
   my $u_variant  = $variant
     && ($plane1hack ? $plane1hackable{$variant}
     : ($plane1 ? $variant : undef));
-  my $u_text = $u_variant && unicode_convert($text, $u_variant);
+  my $u_text = ($tag ne 'm:mtext') && $u_variant && unicode_convert($text, $u_variant);
   if ((defined $u_text) && ($u_text ne '')) {    # didn't remap the text ? Keep text & variant
     $text    = $u_text;
     $variant = ($plane1hack && ($variant ne $u_variant) && ($variant =~ /^bold/)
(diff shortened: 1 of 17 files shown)


lib/LaTeXML/Post/MathML.pm
```diff
@@ -729,15 +731,18 @@ sub stylizeContent {
     $class = ($class ? $class . ' ' : '') . 'ltx_font_oldstyle'; }
   elsif ($font =~ /smallcaps/) {
     $class = ($class ? $class . ' ' : '') . 'ltx_font_smallcaps'; }
+  elsif ($variant) {    # Any left-over mathvariant? Punt to CSS
+    $class = ($class ? $class . ' ' : '') . 'ltx_mathvariant_' . $variant; }
```
dginev, line 735: Should this check also unset `$variant = undef;` after it is moved to a class value?

As it is, I see that one of the tests has both a class and mathvariant attribute, as in:
```xml
<mtext class="ltx_mathvariant_italic" mathvariant="italic">
```

But in MathML Core the only prescribed use of mathvariant is for the `normal` value. So the `class` addition should suffice. That was my suggestion in https://github.com/brucemiller/LaTeXML/issues/2051#issuecomment-1676419924

brucemiller: Right, well as I understood the original issue, and the implications of the (eventually) linked mozilla issue, it was that Core will ignore (most) mathvariants, and that Full *may* ignore it, and that CSS could be used as a fallback. Moreover, that browsers should be able to cope with there being both CSS and mathvariant.

Keeping the mathvariant at least records for MathML-Full agents which don't support CSS that the token was (apparently) intended to belong to a certain semantic class.  Removing it seems somehow wrong, even though in the current scheme of things it is likely always to be ignored (but harmless?)

dginev: Since it is ignored in practice, keeping it until there is some actual issue caused by it is likely harmless, as you say. And indeed it is a MathML Full use.

I've heard rumors of hypothetical MathML Full agents without CSS support, but I wonder if/when we'll have some to test with...

Linking the current MathML Core text on [mathvariant](https://w3c.github.io/mathml-core/#dfn-mathvariant) just in case:
> The mathvariant attribute, if present, must be an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match of normal.

---

No further comments on the PR, looks good to merge.

dginev (approved):

brucemiller:
Following offline discussions, simplified the `m:mtext` (gets CSS, but not mathvariant), and avoid visual styling of formatting codepoints such as invisible-times.  @dginev please take another peak if all looks OK.

lib/LaTeXML/Post/MathML.pm
```diff
@@ -715,13 +715,15 @@ sub stylizeContent {
...
     : ($plane1 ? $variant : undef));
-  my $u_text = $u_variant && unicode_convert($text, $u_variant);
+  my $u_text = ($tag ne 'm:mtext') && $u_variant && unicode_convert($text, $u_variant);
   if ((defined $u_text) && ($u_text ne '')) {    # didn't remap the text ? Keep text & variant
     $text    = $u_text;
     $variant = ($plane1hack && ($variant ne $u_variant) && ($variant =~ /^bold/)
       ? 'bold' : undef); }                       # Possibly keep variant bold
                                                  # Use class (css) to patchup some weak translations
-  if    (!$font) { }
+  if ($text =~ /^\p{Format}*$/) {                # Formatting (eg. InvisibleTImes)
```
dginev, line 724: this regex is a little odd - does it mean starting with `\p{Format}` or empty? If so, you can replace the `*$` with `?`

dginev (approved):
Cool! Good to merge. 

Tiny regex quibble, but it's just a minor performance-related note.

lib/LaTeXML/Post/MathML.pm
brucemiller: I clarified the comment; should mean that the text contains only formatting characters, or is empty.

Issue #1864, opened 2022-06-01
LaTeXML cannot find TeX


teepeemm:
Copied from my email to the listserv:
I'm having trouble getting LaTeXML to find my TeX installation.

```
: echo '\\documentclass{article}\\usepackage{algorithm}\\begin{document}\\end{document}' | latexml -
...
Warning:missing_file:algorithm Can't find package algorithm at algorithm.sty.ltxml; line 20
...
<?latexml searchpaths="(current working directory)"?>
```

I can specify the path via:
```
: echo '\\documentclass{article}\\usepackage{algorithm}\\begin{document}\\end{document}' | latexml - --path=/usr/local/texlive/2020/texmf-dist/tex/latex/algorithms
...
Warning:missing_file:ifthen Can't find package ifthen at ifthen.sty.ltxml; line 27
...
<?latexml searchpaths="/usr/local/texlive/2020/texmf-dist/tex/latex/algorithms,(current working directory)"?>
```
So I could stack several --path arguments, but that gets unwieldy.

A bit under a year ago, I had a similar question, and the answer that worked at the time was to set the environment variable LATEXML_KPSEWHICH.  I've tried that:
```
: export LATEXML_KPSEWHICH=$(which kpsewhich)
: echo $LATEXML_KPSEWHICH                    
/Library/TeX/texbin/kpsewhich
: echo '\\documentclass{article}\\usepackage{algorithm}\\begin{document}\\end{document}' | latexml -
...
Warning:missing_file:algorithm Can't find package algorithm at algorithm.sty.ltxml; line 20
...
<?latexml searchpaths="(current working directory)"?>
```

The part that leaves me really confused is that in the LaTeXML directory, all the tests are fine:
```
: make test
...
t/000_test_context.t ...... 1/1 #
#   LaTeXML version 0.8.6
#   XML::LibXML 2.0207; libxml2 20912
#   XML::LibXSLT 1.99; libxslt 10134
#   TeX 3.14159265 (TeX Live 2020)
...
All tests successful.
Files=34, Tests=441, 479 wallclock secs ( 0.15 usr  0.10 sys + 448.36 cusr 17.33 csys = 465.94 CPU)
Result: PASS
```
(Having skipped t/97_manifest.t and t/170_grammar_coverage.t because I don't have continuous integration.)

[...]

PR #2294, opened 2024-01-02
remove unnecessary pathname_is_nasty check on search paths

### LINKED ISSUE #2293: kpsewhich fails if parent directory contains parentheses
Just tested with the 29 Dec master, it seems like latexml struggles if the current path contains parentheses. I believe this has to do with some failed escaping when calling kpsewhich.

Minimal example:
```latex
\documentclass{article}
\usepackage[T1]{fontenc}
\begin{document}
  Parentheses in parent directory.
\end{document}
```
If you run this in a path containing e.g. `/tmp/test(parentheses)`, LaTeXML will fail with
```
Warning:missing_file:t1enc Can't find definitions t1enc
  at t1enc.def.ltxml; line 24
  search paths are /private/tmp/test(parentheses)
```

xworld21:
Fix #2293 (and its older duplicate #1864).

The `pathname_is_nasty` check on the `$paths` array is not needed: the paths are inserted into an environment variable, not the command line.


brucemiller:
Superseded by #2297

Issue #2049, opened 2023-02-06
`DontPrintSemicolon` breaks newline parsing for algorithm2e


MilesCranmer:
Hey @dginev @brucemiller,

I wanted to report an edge case bug I am seeing with `algorithm2e.sty.ltxml` environments, which I believe is due to `\DontPrintSemicolon` (which turns off the printing of `;` at the end of each line) interfering with how `algorithm2e.sty.ltxml` parses the contents of an algorithm. Here is an example:
```latex
\usepackage[linesnumbered,lined,ruled,commentsnumbered]{algorithm2e}

...

\begin{algorithm}[h!]
\algsize
\DontPrintSemicolon
\Data{$P$, a population of expressions}
\Data{$\data$ as in \cref{alg:pysr}}
\Result{a single expression (the winner of the tournament)}
\Param{$n_s$, the tournament size (=12)}
\Param{$p_\text{tournament}$, the probability of selecting the fittest individual (=0.9)}
\BlankLine
\Fn{\Tournament{$P, \data$} \algcomment{\href{https://github.com/MilesCranmer/SymbolicRegression.jl/blob/0bb8b2bb58a4338729c84e3f52575ac475c86a09/src/Population.jl\#L100-L121}{[code]}}}{
    $Q \gets $ a random subset of size $n_s$ of $P$\;
    \While{$\Length{Q} > 1$}{
        {$E \gets \GetFittest{Q}$}\;
        \If{$\Rand{} < p_\text{tournament}$}{
            \Break
        }
        remove $E$ from $Q$\;
    }
    \KwRet{$E$}
}
\BlankLine\;
\Fn{\GetFittest{$P$} \algcomment{\href{https://github.com/MilesCranmer/SymbolicRegression.jl/blob/0bb8b2bb58a4338729c84e3f52575ac475c86a09/src/Population.jl\#L86-L95}{[code]}}}{
    $\ell_\text{best} \gets \infty$ \;
    \For{$E$ \KwTo $P$}{
        $C \gets$ complexity of $E$\;
        $\ell \gets \text{accuracy of }E$ \;
        \;
        \algcomment{include adaptive parsimony:}\;
        ${\ell \gets \ell}\times\exp(\text{recent number of expressions seen at size }C)$\;
        \;
        \If{$\ell < \ell_\text{best}$}{
            $\ell_\text{best} \gets \ell$ \;
            $E^\ast \gets E$\;
        }
    }
    \KwRet{$E^\ast$}
}
[...]

PR #2296, opened 2024-01-04
emulate \DontPrintSemicolon in algorithm2e.sty

dginev:
Fixes https://github.com/dginev/ar5iv/issues/423

Since algorithm2e.sty is emulated in Perl, with some very special considerations for the ends of lines, I think this is the best way to also support the `\DontPrintSemicolon` without losing the line endings.

There is probably a possible refactor that relies on the native interpretation more, but that could be deferred...

Running example:
```tex
\documentclass{article}
\usepackage{algorithm2e}
\begin{document}
\begin{algorithm}
\DontPrintSemicolon
line 1

line 2\;
line 3\;
\For{$t = 1, 2, \ldots, T$} {
    line 1\;
    line 2\;
    line 3\;
}
\end{algorithm}
\end{document}
```

### DIFF
--- a/lib/LaTeXML/Package/algorithm2e.sty.ltxml
+++ b/lib/LaTeXML/Package/algorithm2e.sty.ltxml
@@ -77,9 +77,9 @@ for my $env (qw(algorithm2e algorithm algorithm*)) {
       Let('\parbox', '\lx@algo@parbox');
       Let("\\\\",    '\lx@algo@par');
       Let('\strut',  '\lx@algo@strut');
-      AssignCatcode("\r" => CC_SPACE);    # NOT CC_EOL, so newlines don't turn to \par!
-                                          #    AssignCatcode("\r" => 13);
-          #    Let(T_ACTIVE("\r"), '\relax');    # More appropriate than \par, I think?
+      # AssignCatcode("\r" => CC_SPACE);    # NOT CC_EOL, so newlines don't turn to \par!
+      #    AssignCatcode("\r" => 13);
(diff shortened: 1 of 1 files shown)


dginev:
I also made the printing switches global assignments and spotted a curious discrepancy - algorithm2e will also create a listing line if there is a regular `\par` in the flow, but our binding had deactivated `\r` (to a space), so the par breaks weren't getting seen.

I probably need to fetch some more tests before this is really solid, since the comments do a great job of warning there may be too many newlines created if one isn't careful.

dginev:
As Bruce pointed out on the call today, this fixes the exact report from #2049

brucemiller:
This looks plausible, I think. Is it ready for review?

dginev:
Doing a quick run through the [ctan examples](https://ctan.org/tex-archive/macros/latex/contrib/algorithm2e/doc)... It looks like the PR introduces some extra trailing newlines at the ends of indented blocks. Except no, the same extra line-breaks exist in the master branch. So no visible regressions.

I think the minimal fix here is ready for review+merge, and there are certainly more details one can fine-tune to get the examples rendering perfectly. But maybe not for this release?

dginev:
A comment for future reference (since I don't have an issue to attach it to):

Besides the empty trailing line, another detail that could be polished is the vertical algorithm bars, which are currently emulated via what seems to be `|` ascii art. I remembered that Twitter of all places has a nice solution for those using flexbox. I extracted it into a codepen (try horizontally shrinking/growing the window to demo the effect):
https://codepen.io/dginev/pen/QWoNgON

Maybe something to try a little later on.

PR #2300, opened 2024-01-11
SkipSpaces between two OptionalMatch arguments in \pdfliteral

dginev:
Minimal fix for https://github.com/arXiv/html_feedback/issues/258

The main issue was a need for adding the `SkipSpaces` parameter between two `OptionalMatch:` parameters in `\pdfliteral`, so that the arguments get read in appropriately, and then ignored (for now).

The minimal test snippet is:
```tex
\documentclass{article}
\begin{document}
\pdfliteral direct {2 Tr 0.5 w}%
test%
\pdfliteral direct {0 Tr 0 w}%
\end{document}
```

This kind of pattern would benefit from a test, but I think that can wait until we actually try to emulate some of the directives (e.g. turning `test` bold here).

### DIFF
--- a/lib/LaTeXML/Core/Parameter.pm
+++ b/lib/LaTeXML/Core/Parameter.pm
@@ -95,6 +95,9 @@ sub read {
       "Missing argument " . Stringify($self) . " for " . Stringify($fordefn),
       "Ended at " . ToString($gullet->getLocator));
     $value = T_OTHER('missing'); }
+  elsif ($value && $$self{optional} && $$self{type} =~ /^OptionalMatch/) {
+    # experiment: skip spaces after a successful OptionalMatch read
+    $gullet->skipSpaces; }
   return $value; }
 
 # This is needed by structured parameter types like KeyVals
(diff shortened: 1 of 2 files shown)


dginev:
Btw this makes me wonder if `OptionalMatch:literal` should also skip spaces in the case where it successfully matches the literal.

brucemiller:
On 1/11/24 16:34, Deyan Ginev wrote:
> Btw this makes me wonder if |OptionalMatch:literal| should also skip 
> spaces in the case where it successfully matches the literal.

Seems worth a try to see what tests explode, but I suspect that this is 
not the standard behavior of (eg) "*"

dginev:
@brucemiller OK, being a bit braver for the sake of exploration. I added a commit that appeared to pass all tests, as well as the motivating `\pdfliteral` example.

I could only think of `Parameter::read` as the place to add the check however, since `OptionalMatch` doesn't have a dedicated reader function - which makes it look a little out of place.

brucemiller (approved):

PR #2301, opened 2024-01-17
fix URLs for badges in the README

dginev:
The badges in the README had URLs that no longer function, minor fixes:
 - our CI workflow is no longer named CI, so we need 2 badges - one for Linux, one for Windows
 - the CC0 license badge can simply point to the official URL, the old one has a "bad signature" today

### DIFF
--- a/README.pod
+++ b/README.pod
@@ -2,8 +2,9 @@
 
 =head1 L<LaTeXML|http://dlmf.nist.gov/LaTeXML/>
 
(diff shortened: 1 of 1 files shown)


dginev:
The PR changes can be previewed at:
https://github.com/brucemiller/LaTeXML/blob/1057c393380d04882df426ca0c763d6c8f9b88ac/README.pod

brucemiller:
Thanks!

PR #2302, opened 2024-01-23
Subfigure model enhancements

### LINKED ISSUE #2036: maths disappears within figures with multiple elements
Very annoying regression:
```latex
\documentclass{article}
\begin{document}
\begin{figure}
  \begin{center}
  First
  \end{center}
  \[ Second \]
  \begin{center}
  Third
  \end{center}
\end{figure}
\end{document}
```
The "Second" line appears in the XML, but is absent in the HTML output.

### LINKED ISSUE #2107: newlines and indentation in figures
Hi!
I need to have breaks followed by indentation (Works in latex), and it needs work the same in text and in figures. It doesn't now. (See attached file).

<pre>
\begin{figure}
\begin{flushleft}
  In fig: This should be left\\
  \mbox{} ~~~~~This should be indented on the next line
\end{flushleft}
  \caption{A figure}
\end{figure}
</pre>

Does not obey \\ (when the window is wide enough). This is with (LaTeXML version 0.8.7) tested on multiple browsers (Safari, Firefox, Chrome).

Also I'm not sure why figure tries to center. This is not the default in latex. (But I know how to fix it by editing .ltx_flex_figure in LeTexML.css)

Here is a small test example you can run:

[latexmltest.tex.zip](https://github.com/brucemiller/LaTeXML/files/11520592/latexmltest.tex.zip)

### LINKED ISSUE #2154: figure loses items
An itemize inside a figure loses all the items. Try the attached file. There are lots of information in the \items but they just don't appear in the html, but are in the xml

Eg "Landmark" is  in aside.xml but not in aside.html

Tested on mac using a git pull from today. To reproduce
Uzip attached aiside.tex.zip
latexml  --documentid=aside --destination=aside.xml aside.tex
latexmlpost --format=html5 --pmml --destination=aside.html aside

David
[aside.tex.zip](https://github.com/brucemiller/LaTeXML/files/12099789/aside.tex.zip)

dginev:
Fixes #2036 . Fixes #2107 . Fixes #2154 .
  - Also in ar5iv:
     - Fixes https://github.com/dginev/ar5iv/issues/118 
     - Fixes https://github.com/dginev/ar5iv/issues/197
     - Fixes https://github.com/dginev/ar5iv/issues/252
     - Fixes (upto sizing) https://github.com/dginev/ar5iv/issues/243
     - Fixes (upto styling) https://github.com/dginev/ar5iv/issues/263
     - Fixes (upto styling) https://github.com/dginev/ar5iv/issues/275
     - Fixes https://github.com/dginev/ar5iv/issues/298
     - Fixes (tables+figures) https://github.com/dginev/ar5iv/issues/367

This PR redesigns the subfigure constructs in LaTeXML, via:
 - Recognize (and heuristically assemble if needed) "panels", marked with a class value `ltx_subfigure_panel`, during Core processing. 
   - We continue to check sizes and infer breaks, when adjacent panels exceed `\textwidth`. 
   - But we now also recognize cases where very narrow adjacent pieces belong in the same panel (such as an image followed by `\footnote` or a single `.` dot char).
   - Move the generation of "sized" classes for flexbox to XSLT. With the PR all considerations specific to flexbox are now contained to `LaTeXML-para-xhtml.xsl`.
   - Better handling for multi-caption figures in HTML, using a "partitioning strategy" based on captions. Not yet propagated to JATS or other output formats.
   - Better handling for `{minipage}` subfigures, thanks to the new `_CaptureBlock_` mechanism.
   - Ideally the XSLT has been made more robust, covering a range of edge cases, and no longer silently dropping unexpected content (rather silently copying it over).
   - More tests, and likely a couple of more to come before merging, as there is a lot of diversity of markup that can end up surprising if unexpected.

Transitioning into public WIP, and welcoming code review by @brucemiller .

I have added 3 new tests so far, but sadly they test only half of the new changes (the XML half, without the HTML pieces). I'd recommend building HTML locally and exploring the grids via the browser inspector. The `t/complex/figure_mixed_content` test especially contains some really tough constructions from arXiv that haven't been perfected fully just yet.


lib/LaTeXML/Post/Graphics.pm
```diff
@@ -134,18 +134,22 @@ sub getGraphicsSourceTypes {
...
-    #
-    # If this is *NOT* a known graphics type, it would be best to treat it as
-    # a name suffix (e.g. name.1 from name.1.png)
-    if ((length($reqtype) > 0) && !(grep { $_ eq lc($reqtype) } $$self{graphics_types})) {
-      $name .= ".$reqtype"; }
-    my $file  = pathname_concat($dir, $name);
-    my @paths = pathname_findall($file, paths => $LaTeXML::Post::Graphics::SEARCHPATHS,
-      # accept empty type, incase bad type name, but actual file's content is known type.
-      types => ['', $self->getGraphicsSourceTypes]);
+    # if we already have a usable candidate, save ourselves the work
```
dginev, line 137: I also added two graphics-related upgrades:
 - If a graphic was available in the `candidates` attribute, we should use that path information in post-processing, which speeds things up along.
 - I used an example image distributed with texlive for the tests ( "example-image-a", it seemed appropriate), and realized we weren't finding those images. So added a fallback call to `pathname_kpsewhich` to Image.pm

lib/LaTeXML/Core/Document.pm
```diff
@@ -1419,8 +1419,20 @@ sub addSSValues {
   return; }
 
 sub addClass {
-  my ($self, $node, $class) = @_;
-  return addSSValues($self, $node, class => $class); }
+  my ($self, $node, $value) = @_;
+  return addSSValues($self, $node, class => $value); }
+
+sub removeClass {
```
brucemiller, line 1425: May be useful in the future to generalize this to `removeSSValues` and add a layer like `addClass`

lib/LaTeXML/Post/Graphics.pm
brucemiller: Hmm... using these images in the test cases is introducing a system-dependence that causes failure.

dginev: Ah, I think this is due to the floating texlive directory name (based on texlive version). I suppose I should switch to the PNG I see in the manifest, `t/graphics/none.png` ?

But I think the `candidates` upgrade is actually healthy, especially in docs with lots of images (avoids a lot of duplicated searching)... Or?

dginev:
My new-and-improved caption partitioning works in the tests I have added, but regressed on a no-caption test with explicit breaks (e.g. the minimal equation test from #2036 ).

While the `ltx:caption` partitions the `<div class="ltx_flex_figure">...</div>` flex subfigures, the `ltx:break` elements need to act as partitions for counting the `ltx_flex_size_*` class values (1,2,3,4,many) which anchor the different form factors for CSS. And I dropped that code while refactoring. Will add a commit that reintroduces that in a bit, as well as the test which revealed the need.

dginev:
Alright, the `ltx_flex_size_[value]` once again respects `ltx:break` boundaries. The `t/structure/figure_grids.tex` test is particularly informative for these class values (in HTML).

I will check more ar5iv examples tomorrow, I think I fully vetted the related latexml issues today.

lib/LaTeXML/Core/Document.pm
dginev: Done, thanks!

dginev:
I have just concluded reviewing the ar5iv issues, and recorded the one the PR resolves fully.

There are still some cases where the current flex design comes out simplistic, as it is geared to each (subfigure) panel in a row to be equally sized with all others. E.g. in a row of 4 panels, each one gets the same class `ltx_flex_size_4`. That is not a great fit for less usual constructs, such as the figures from https://github.com/dginev/ar5iv/issues/266 and https://github.com/dginev/ar5iv/issues/275

Supporting them goes beyond the scope of the current PR, but it is likely that another pass will be needed in the future, maybe also connected to providing more information about sizing through CSS classes. If we have enough information to construct advanced CSS selectors, the current HTML scaffolding may be correct/sufficient as-is.

brucemiller (approved):
Scary, but impressive. And seems to improve many cases, w/o breaking anything that I can see. Thanks!!!

PR #2303, opened 2024-01-24
--noinvisibletimes: choose zero-width space over invisible times

### LINKED ISSUE #1941: do not emit invisible times before and after text
LaTeXML seems just a bit too eager to introduce invisible times operators:
```bash
latexmlmath "a\\textrm{ is less than}b"
```
produces
```xml
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="a\textrm{ is less than }b" display="block">
  <mrow>
    <mi>a</mi>
    <mo>⁢</mo>
    <mtext> is less than </mtext>
    <mo>⁢</mo>
    <mi>b</mi>
  </mrow>
</math>
```
My superficial impression – not backed up by any corpus or careful consideration – is that in the majority of situation, the text should not be surrounded by operators.

dginev:
Fixes #1941 . At least it allows it to be fixed, via a new option.

Here is a draft following some offline discussions on allowing a customization switch for preferring zero-width space (U+200B) over invisible times (U+2062) in implied `<mo>` elements.

The PR minimally touches up the character in post-processing, via the `MathML::stylizeContent` sub.

An easy test is:
```
$ latexmlc 'literal:2x' --whatsin=math --dest=test.html \
  --noinvisibletimes
```

I have exposed the option in latexmlpost and latexmlc. But not yet in latexmlmath (maybe later).

### DIFF
--- a/bin/latexmlpost
+++ b/bin/latexmlpost
@@ -36,7 +36,7 @@ my ($format, $extension, $is_html, $urlstyle) = (undef, undef, undef, 'server');
 my ($numbersections) = (1);
 my ($stylesheet, $defaultresources, $icon, @css, @javascript, @xsltparameters);
 my ($mathimagemag) = (1.75);
-my ($linelength, $plane1, $hackplane1, $mathstylize) = (undef, undef, undef, undef);
+my ($linelength, $plane1, $hackplane1, $invisibletimes) = (undef, undef, undef, undef);
 my ($dographics, $svg, $picimages) = (undef, undef, undef);
 my @graphicsmaps = ();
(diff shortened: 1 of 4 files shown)


brucemiller:
I think I tend to prefer one of the variants we discussed, like `--noinvisibletimes`, for example. There isn't really a good solution, but there is clearly at least a temporary need. 

So why don't you take a fresh look, change the option name if you want to, and then go ahead & merge it whichever way you feel good about :>

dginev:
`--noinvisibletimes` works, thanks!

PR #2304, opened 2024-01-24
pgfmath parse for width,depth,height

dginev:
Added in the hope of fixing https://github.com/dginev/ar5iv/issues/431 , but it only resolved the first error in that issue.

Minimal example enabled by the PR:
```tex
\documentclass{article}
\usepackage{pgf}
\begin{document}

\pgfmathparse{height("Some Lovely Text")} \pgfmathresult

\pgfmathsetmacro\MathAxis{height("$\vcenter{}$")}

\MathAxis

\end{document}
```


brucemiller (approved):
sure, ok. looks good.

PR #2309, opened 2024-02-01
Sandbox run 01.2024, fixes for Fatal cases (10)

dginev:
This PR lifts a first small selection of the unfortunate Fatal regressions in the arXiv sandboxes with the latest master:

1. Add a guard for `\xyoption` to avoid loading the same option twice (as inspired by the original sty, which also does so)
2. allow more recursion in deeply nested argument reads, even if a true infinite loop lurks behind that - the final Fatal printed gets a more informative trace in the genuine loop cases.
3. add a guard in `revertScript` to only call `->defined_as` on Token objects (and not Tokens).
4. recognize and guard the advanced use of `\multirow` where the height param can be a negative or fractional number. Not implemented yet, but won't Fatal anymore.
5. Fix a surprisingly common regression with xparse and lipsum conflicting (really my hacky expl3 "pool" conflicting with the expl3.sty binding). I believe this was the trickiest issue to detect here. And yes, 1% of sandbox articles included lipsum in a conflicting setting.
    - The minimal example is simply:
    ```tex
    \documentclass{article}
    \usepackage{xparse}
    \usepackage{lipsum}
    \begin{document}
    \end{document}
    ```
    I think eventually the expl3 "pool" should disappear entirely, but I will defer that until we start loading `latex.ltx` raw. I think that would be a better home for that removal.

6. A [connected change](https://github.com/dginev/ar5iv-bindings/commit/a1698f0bd41f86ea695f12f6d461c0023e3f69b1) outside of this PR is updating the `biblatex.sty` experimental binding to catch up to the new internals of url and hyperref - by simply not using them and loading hyperref to get a working `\href` and `\url`. 
7. guard literal keyword regexes in listings.sty with `quotemeta()`
8. guard failed parses from `realizeXMNode` coming back with a non-node object
9. ensure `import.sty` paths work with ZIP inputs, by making requested paths absolute w.r.t the "SOURCEDIRECTORY" parameter (source is unpacked to a temp dir for ZIP).
10. Recognize math mode in ulem.sty macros.

This concludes round 1 of my debugging of Fatals. I will rerun the sandboxes with the updated commits and circle back with a comment.

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -4444,19 +4444,19 @@ DefPrimitiveI(T_SUB,   undef, sub { scriptHandler($_[0], 'SUBSCRIPT'); });
 # at the least they're ugly; in some applications they affect "round trip" processing.
 # OTOH, direct use of \@@POSTSUPERSCRIPT, etal, MAY need to have extra braces around them.
 # So, when reverting, we're going to a bit of extra trouble to make sure we have ONE set
-# of braces, but no extras!!  [Worry about lists of lists...]
+# of braces, but no extras!!
 sub revertScript {
   my ($script) = @_;
-  my @tokens   = $script->revert;
-  my @t        = @tokens;
+  # We need to handle lists of lists, see arXiv:2210.11051
+  my @tokens = Tokens($script->revert)->unlist;
+  my @t      = @tokens;
   my $l;
   if ($t[0]->defined_as(T_BEGIN)) {
     $l++; shift(@t); }
   while (@t && $l) {
     my $t = shift(@t);
-    if (ref $t eq 'LaTeXML::Core::Token') {    # jump over pre-wrapped Tokens, see arXiv:2210.11051
-      if    ($t->defined_as(T_BEGIN)) { $l++; }
-      elsif ($t->defined_as(T_END))   { $l--; } } }
+    if    ($t->defined_as(T_BEGIN)) { $l++; }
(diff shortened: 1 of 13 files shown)


dginev:
The PR's impact on our arXiv sandboxes is now documented here:
https://gist.github.com/dginev/abb72c989f4474a3a33300ed23ac6c86

lib/LaTeXML/Package/TeX.pool.ltxml
```diff
@@ -4453,8 +4454,9 @@ sub revertScript {
     $l++; shift(@t); }
   while (@t && $l) {
     my $t = shift(@t);
-    if    ($t->defined_as(T_BEGIN)) { $l++; }
-    elsif ($t->defined_as(T_END))   { $l--; } }
+    if (ref $t eq 'LaTeXML::Core::Token') {    # jump over pre-wrapped Tokens, see arXiv:2210.11051
```
brucemiller, line 4457: Interesting that the leading comment ends with "Worry about list of lists". That `->revert` would yield a list of both `Token` and `Tokens` almost seems like the real bug. Either way, your patch doesn't protect against *leading* `Tokens`. It's also unclear that simply skipping over `Tokens` is the right thing; what if there are unbalanced or leading or trailing braces in there?

I'd think a simpler, though slightly inefficient, approach would be
```
my @tokens = Tokens($script->revert)->unlist;
```

dginev: Excellent suggestion - and indeed more correct. 

Done: Tested with two articles, seems to work well.

brucemiller (approved):
Looks good! Thanks!!

PR #2311, opened 2024-02-02
Set lspace and rspace for zero-wdith space mo 

dginev:
A bit of an unfortunate follow-up to #2303 , based on observing that we get the default `0.27...em` spacing for any zero-width Unicode chars not yet in the MathML operator dictionary.

See the [discussion in mathml-core](https://github.com/w3c/mathml-core/issues/209#issuecomment-1924012036) for more details, as well as [an example codepen](https://codepen.io/dginev/pen/zYbjGNj).

Setting the lspace/rspace attributes explicitly feels like we are reverse-patching the defaults, but without the rendering looking normal, it seems that it causes more harm than good to switch away from invisible times.

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -755,6 +755,8 @@ sub stylizeContent {
     $stretchy = $size = undef;
     # If requested, switch implied mo to space
     if ($text eq "\x{2062}" and not($invisibletimes)) {
+      # and set lspace/rspace to 0 until we have browser interop.
+      $props{lspace} = $props{rspace} = $props{force_lspace} = $props{force_rspace} = '0em';
       $text = "\x{200B}"; } }
   if ($size) {
     if ($size eq ($LaTeXML::MathML::SIZE || 'text')) {    # If default size, no need to mention.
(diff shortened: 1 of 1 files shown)


xworld21:
Now that I see this... why did you go for U+200B as opposed to U+2063? (Conversely, why did I suggest U+2063 in the first place https://github.com/brucemiller/LaTeXML/issues/1941#issuecomment-1783782218? I don't remember.)

xworld21:
> (Conversely, why did I suggest U+2063 in the first place [#1941 (comment)](https://github.com/brucemiller/LaTeXML/issues/1941#issuecomment-1783782218)? I don't remember.)

To answer myself: because U+2063 is mentioned in the operator dictionary. U+200B is not.

dginev:
For U+2063 the assumption is that we _know_ the operator truly was used as a separator, as it would for say delimiting the elements of [a permutation cycle](https://en.wikipedia.org/wiki/Cyclic_permutation). The math-oriented invisible characters in Unicode are meant to imply some certainty of having captured the intention - which may then be used (spoken) by accessibility tools.

On the other hand, a discussion in the MathML Core meetings had broad consensus that the zero-width space U+200B doesn't carry any implied meaning, and could be used as a placeholder. The preference was that we also avoid using an empty `<mo/>` element, since those are traditionally attributed to errors in XML pipelines. In an effort to please as many perspectives as possible (including ours: we have cases where we don't know what the operators stands for and want it silent in AT), we went for U+200B.

I find all of this to be quite useful, since it helps MathML Core end up with better defaults for more corners of Unicode than it has already covered.

brucemiller (approved):
Yikes, we just can't win, can we?

PR #2314, opened 2024-02-07
detach tikz figure test, update extra tests

### LINKED ISSUE #2313: TikZ no longer causes newpage with figures
In TeXLive 2023, `make test` is failing with
```
t/80_complex.t ............ 7/15 
#   Failed test 't/complex/figure_mixed_content'
#   at /Users/timothy.prescott/Documents/programming/github/LaTeXML/blib/lib/LaTeXML/Util/Test.pm line 120.
# Difference at line 921 for t/complex/figure_mixed_content
#       got : '  </paragraph>'
#  expected : '    <pagination role="newpage"></pagination>'
t/80_complex.t ............ 11/15 # Skip: t/complex/si is only checked in continuous integration. (use make test CI=true)
t/80_complex.t ............ 15/15 # Looks like you failed 1 test of 15.
t/80_complex.t ............ Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/15 subtests 
	(less 1 skipped subtest: 13 okay)
```

`./tools/maketests --compare t/complex/figure_mixed_content` returns
```
========================================
Compare Expected t/complex/figure_mixed_content to current output XML:
Missing: 920,920d919
<     <pagination role="newpage"></pagination>
Failed!
```
I was able to trim it down to
```tex
\documentclass{article}
\usepackage{tikz}
\begin{document}
\begin{figure}
content
\end{figure}
\end{document}
```
which has xml
```xml
<?latexml class="article"?>
<?latexml package="tikz"?>
  <figure xml:id="fig1">
    <p>content</p>
  </figure>
  <pagination role="newpage"/>
</document>
```
The newpage is not present in TeXLive 2023 or if TikZ is not used.

dginev:
Fixes #2313 

This addresses the short-term symptom of the main test suite failing. texlive 2022 and texlive 2023 have differences in pgf and tikz, which may lead to (sometimes reasonable) markup discrepancies in the latexml-generated XML under each texlive.

For now, we avoid testing entirely in such fragile situations. Instead we have an extra target, which can be optionally ran via:
```
CI=true make extratest
```

The extra tests are currently expecting a `texlive 2022` to succeed.

Some of these extras also needed an update following the recent schema changes in latexml, so I've added that update to this PR. I have also checked that the regular tests pass under texlive 2022 and texlive 2023.

Note that the CI tests **did not** pass under texlive 2023 before this PR, and still do not with it. Those details are tracked by issue #2044, which we deferred to the next milestone (0.8.9)


brucemiller (approved):
OK, I guess this is the best interim solution. Thanks!

PR #2315, opened 2024-02-07
Sandbox Fatal fixes 01.2024, round 2

dginev:
This PR is a (final) follow-up to the just merged #2309 .

I address another handful of Fatal issues, and one minor binding issue:
 - allow deeper recursion in digesting reads of parameters
 - allow deeper recursion in stringify
 - Guard against infinite loops in complex listings
     - The examples were in cases that deal with `QUOTED_RE`, but I think there are more general possible pitfalls related to `colnum` not getting incremented and sanity-checked in every possible arm of the loop conditional.
     - Finding all of the tricky corner cases is a little hard without a deep dive, so I added a distinctive Error message. We can collect all papers exhibiting the issue after a rerun and build a tougher test suite.
 - IEEEtran.cls.ltxml needs to support an optional arguments for its list envs, and also rebind the usual env names. This fixes https://github.com/arXiv/html_feedback/issues/417
 - `\@finalstrut` and `\@textsuperscript` were being used, added to LaTeX.pool.
 - A very tricky case in pgfmath was that a single dot `.` is recognized as zero. So much so that `\pgfmathparse{.-.2ex}` is a valid expression.
     ```tex
      \documentclass{article}
      \usepackage{pgf}
      \begin{document}
      \pgfmathparse{.-.2ex}\pgfmathresult
      \end{document}
     ```
 - IEEEtran.cls also does not depend on `\sc` internally, and a handful of arXiv papers in that class redefined `\sc` to be something completely different, wreaking some Fatal havoc. I made two IEEEtran-specific primitives to avoid such clashes.
 - There was a very unpleasant infinite loop with OmniBus loading natbib late, during `\bibitem` expansion. For some reason not entirely clear to me the `\bibitem` DefMacro from natbib didn't stick, and the OmniBus definition expanded into itself in a loop. 
    - I created a natbib-specific `\lx@nat@bibitem`, and used a `let` to bind `\bibitem` to it during `natbib.sty.ltxml` load, which appears to avoid the problem.
- Last *and* least, I double-checked `$text` is always defaulted to an empty string before used in regexes for MathML.pm, which avoids a flurry of pointless perl warnings.

Happy to rerun the sandboxes once we merge here, and triple-check we are in good shape for a release tag.

### DIFF
--- a/lib/LaTeXML/Package/xcolor.sty.ltxml
+++ b/lib/LaTeXML/Package/xcolor.sty.ltxml
@@ -290,7 +290,7 @@ sub DecodeColor {
   my $color_re = qr/($color_expr_re)(($func_expr_re)*)/;
 
   my $color;
-  if ($expression =~ /^$color_re$/) {
+  if ($expression =~ /^$color_re\s*$/) {
     #DG: Dear reader, I present to you: maintenance hell:
     my $prefix     = $2 || $10;
     my $name       = $3 || $11;
--- a/lib/LaTeXML/Common/Font.pm
+++ b/lib/LaTeXML/Common/Font.pm
@@ -22,7 +22,7 @@ use LaTeXML::Common::Font::Metric;
 use LaTeXML::Common::Font::StandardMetrics;
 use LaTeXML::Common::Color;
 use List::Util qw(min max sum);
-use base       qw(LaTeXML::Common::Object);
+use base qw(LaTeXML::Common::Object);
 
 # Note that this has evolved way beynond just "font",
 # but covers text properties (or even display properties) in general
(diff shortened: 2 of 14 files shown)


dginev:
The changes of this PR appear to improve conversion stability in our sandboxes, bringing roughly 1% improvement to the Fatal severity. Details at:
https://gist.github.com/dginev/abb72c989f4474a3a33300ed23ac6c86

Ready for review.

lib/LaTeXML/Package/xcolor.sty.ltxml
```diff
@@ -290,7 +290,7 @@ sub DecodeColor {
   my $color_re = qr/($color_expr_re)(($func_expr_re)*)/;
 
   my $color;
-  if ($expression =~ /^$color_re$/) {
+  if ($expression =~ /^$color_re\s*$/) {
```
dginev, line 293: Minor, but the trailing space wasn't recognized and was creating 50+ needless errors. Example from arXiv:1705.04144:
```tex
\documentclass{article}
\usepackage{xcolor}
\begin{document}
\color{rgb, 255:red, 190; green, 190; blue, 190 } Test here.
\end{document}
```

PR #2319, opened 2024-02-12
use U+2217 for multiplication asterisk

### LINKED ISSUE #2318: asterisk in math should be rendered as U+2217
Something I noticed recently and surprised me greatly: LaTeXML outputs a simple asterisk * for the character *, whereas it should be using U+2217 (ASTERISK OPERATOR).

- code: `$a*b$`
- LaTeXML: image
- LaTeX: image
- MathJax (uses U+2217): image

xworld21:
Fix #2318 the 'right way' by patching the existing `DefMathI` call.

### DIFF
--- a/lib/LaTeXML/Package/TeX.pool.ltxml
+++ b/lib/LaTeXML/Package/TeX.pool.ltxml
@@ -4806,12 +4806,12 @@ DefMathI('+', undef, '+', role => 'ADDOP', meaning => 'plus');
 DefMathI('-', undef, '-', role => 'ADDOP', meaning => 'minus');
 ## Redefine, if we want Unicode minus
 ##DefMathI('-', undef, "\x{2212}", role => 'ADDOP',   meaning  => 'minus');
-DefMathI('*', undef, '*', role => 'MULOP',   meaning => 'times');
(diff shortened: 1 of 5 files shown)


dginev:
I am still wondering how to make the decision which piece of LaTeXML should carry the information, since we have so many possible points of intervention. Just above your change there is a commented out line:

```perl 
# Redefine, if we want Unicode minus
#DefMathI('-', undef, "\x{2212}", role => 'ADDOP',   meaning  => 'minus');
```

In that case the Unicode minus is instead "stylized" during the MathML post-processing [here](https://github.com/brucemiller/LaTeXML/blob/893f0ce9e2fe47ee228d33d9e1c8e3248c2e8400/lib/LaTeXML/Post/MathML.pm#L698-L699). 

Were we/Are we trying to keep some of these assumptions out of XMath?

xworld21:
> Were we/Are we trying to keep some of these assumptions out of XMath?

I was puzzled by that U+2212, now I get it. In that case, it seems more consistent to make the change in MathML.pm, I guess. The PR is trivial in either case.

brucemiller (approved):
Looks good; Thanks!

PR #2325, opened 2024-02-16
Tikz cd

brucemiller:
This PR provides a binding for `tikz-cd`, after fixing some infrastructure.

The critical bug was that `ALIGN_STATE` which tracks `{}` during scanning to enable or disable certain alignment tokens (`&,\cr,...`) was being done too late (in Stomach rather than Gullet).   It's easy to miss some corner cases, so some sandbox testing is warranted.

This is a WiP since the resulting `svg` are displayed too low; there would seem to be one too many adjustments involving `pgf@pixminy` (?) , although other tikz pictures seem mostly correct.

PR #2327, opened 2024-02-21
tikz-related update to jheppub.sty.ltxml 

dginev:
This PR originates from testing #2325 . It can be merged as a follow-up.

The jheppub.sty.ltxml binding was a little outdated. I noticed a case where `tikz-cd.sty` wasn't loading despite having fresh latexml support (arXiv:2112.10658).

As it turns out, `jheppub.sty` now has that as a dependency, as well as a handful of others. Added here.

### DIFF
--- a/lib/LaTeXML/Package/jheppub.sty.ltxml
+++ b/lib/LaTeXML/Package/jheppub.sty.ltxml
@@ -20,14 +20,20 @@ use warnings;
 use LaTeXML::Package;
 
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-RequirePackage('hyperref');
-RequirePackage('color');
-RequirePackage('natbib');
 RequirePackage('amsmath');
 RequirePackage('amssymb');
+RequirePackage('amsthm');
+# RequirePackage('aliascnt');
 RequirePackage('epsfig');
 RequirePackage('graphicx');
+PassOptions('natbib', 'sty', 'numbers', 'compress');
+RequirePackage('natbib');
+RequirePackage('xcolor');
+RequirePackage('hyperref');
 RequirePackage('inst_support');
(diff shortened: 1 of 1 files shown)


dginev:
Ah, wrong - this is more sinister. The arXiv package attaches its own `jheppub.sty`, with a completely unmodified header from the official 2013 original, but *seemingly* modified contents?

I think I have unearthed that today's official jheppub.sty is located at:
https://jhep.sissa.it/jhep/help/JHEP/TeXclass/DOCS/jheppub.sty

And yet that file doesn't have any mention of tikz... The arXiv:2112.10658 one has the header lines:

```
%% jheppub.sty
%% Copyright 2013 SISSA Medialab
%
% This work may be distributed and/or modified under the
% conditions of the LaTeX Project Public License, either version 1.3
% of this license or (at your option) any later version.
% The latest version of this license is in
%   http://www.latex-project.org/lppl.txt
% and version 1.3 or later is part of all distributions of LaTeX
% version 2005/12/01 or later.
%
% This work has the LPPL maintenance status `author-maintained'.
%
% The Current Maintainer of this work is
% SISSA Medialab <info@medialab.sissa.it>
%
% This work consists of the file jheppub.sty.
\NeedsTeXFormat{LaTeX2e}
\ProvidesPackage{jheppub}[2013/03/21 r534]
```

But I am suspicious whether that `r534` jheppub actually has the tikz mention. Closing here until I am sure this can be solved "officially".

dginev:
Right, here is another online copy of the r534 versioned jheppub.sty, which seems to be the only current hit in google:

https://github.com/HSF/documents/blob/master/CWP/papers/HSF-CWP-2017-02_training/latex/jheppub.sty

There is no mention of tikz at all. So I am lead to believe the sty in arXiv:2112.10658 is a modified version distributed without changing anything in the sty metadata... There isn't much we can do here that wouldn't be really dirty, so I'll shelve this (and sustain some `undefined:{tikzcd}` errors that stem from it).

PR #2332, opened 2024-02-28
Fix spelling errors in some perl modules.

hpreusse:


dginev (approved):
Thank you for the cleanup!

PR #2340, opened 2024-03-20
adjust reversions of \braket to use | separators

### LINKED ISSUE #2334: `\braket{}` tex changed by `latexml`
Given the following `main.tex`:

```latex
\documentclass{article}
\usepackage{braket}

\begin{document}
$\braket{a|b}$
\end{document}
```

Running `latexml test.tex --dest=test.xml` yields an XML file including the line

```xml
...
    <p><Math mode="inline" tex="\braket{a}{b}" text="inner-product@(a, b)" xml:id="p1.m1">
...
```

The thing to note here is that the math in the `tex`  attribute does not match the original math in the TeX source. (This is breaking some parsing I am doing).

Thanks in advance!

dginev:
Fixes #2334 . Just a small set of reversion tweaks.

### DIFF
--- a/lib/LaTeXML/Package/braket.sty.ltxml
+++ b/lib/LaTeXML/Package/braket.sty.ltxml
@@ -55,13 +55,13 @@ sub splitBraketArg {
 
 #======================================================================
 DefMacro('\braket{}', sub {
-    my ($gullet, $arg) = @_;
-    my ($codes, @args) = splitBraketArg($arg, 2);
+    my ($gullet, $arg)  = @_;
+    my ($codes,  @args) = splitBraketArg($arg, 2);
     my $n = scalar(@args);
     return Invocation(T_CS('\lx@braket@' . $codes), @args)->unlist; });
 DefMacro('\Braket{}', sub {
-    my ($gullet, $arg) = @_;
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/braket.sty.ltxml
```diff
@@ -75,53 +75,53 @@ DefMath('\lx@Braket@{}', '\left\langle#1\right\rangle',
...
 DefMath('\lx@braket@V{}{}', '\langle#1\,|\,#2\rangle',
-  meaning => 'inner-product',
-  alias   => '\braket');
+  meaning   => 'inner-product',
+  reversion => '\braket{#1|#2}');
 DefMath('\lx@braket@D{}{}', '\langle#1\,\|\,#2\rangle',
-  meaning => 'inner-product',
-  alias   => '\braket');
+  meaning   => 'inner-product',
+  reversion => '\braket{#1|#2}');
```
brucemiller, line 82: The "D" (double) versions should revert to `\|`, rather than simply `|` (for the "V" versions).
Same comment applies below.

dginev: Thank you for catching that. It should be enacted now.

I also rebased to the latest master, for a simpler merge.

PR #2341, opened 2024-03-28
Use em rather than percent for MathML mo minsize, maxsize

brucemiller:
Use em rather than percent for MathML mo minsize, maxsize to avoid interpretation issues.

Also, tweak operator dictionary assumptions about what browsers will assume is stretchy.

### DIFF
--- a/lib/LaTeXML/Post/MathML.pm
+++ b/lib/LaTeXML/Post/MathML.pm
@@ -645,6 +645,9 @@ sub pmml_infix {
 my %default_token_content = (
   MULOP => "\x{2062}", ADDOP => "\x{2064}", PUNCT => "\x{2063}");
 
+# chars that all browsers will know are stretchy (notably, apparently, NOT "|")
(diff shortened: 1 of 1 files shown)


dginev (approved):
Code looks good.

The only part to wonder out loud is whether we'll encounter less issues with the `em` formulation compared to the `%` formulation.

Now that there is traction to correct the `%` behavior to match MathML 3, I remember you mentioned that theoretically that behavior is more accurate. But that assumes perfect implementations...

brucemiller:
Actually, I'm not sure if % is theoretically more accurate than em.  In practice, em is.

As far as "encounter(ing) more issues"; well, other issues may be more apparent. Currently Safari ignores %, so essentially treats it as stretchy, which comes out looking reasonable even if it wasn't the author's intention. However, with ems, Safari seems to quantize the size in an odd way, so sometimes it doesn't look as intended either. Chrome's % tends to produce extremely large fences; with ems they come out in the correct size, but sometimes that seems to accentuate the symmetric bug.  So one could make a strategy based on which bugs they expect to be fixed when?

PR #2342, opened 2024-04-01
recognize \bgroup as an opener for readBalanced $require_open=true

### LINKED ISSUE #2335: pdftex_expanded test regression
While comparing the test runs between perl and rust, noticed that the same unexpected error showed up in both. Namely when converting the test

```
$ latexml t/expansion/pdftex_expanded.tex
```

We get:

```
Error:expected:{ Expected opening '{' at pdftex_expanded.tex; line 57 col 16
Error:unexpected:<endgroup> Attempt to pop last locked stack frame at pdftex_expanded.tex; line 57 col 0
```

Looks like a regression, possibly informative one? 
pdftex converts the file without any unusual messages.

dginev:
Fixes #2335 

Based on a small pdftex test, the current latexml behavior is correct. 
We are just emitting an extra error for the case where `\bgroup` was used as the opener for a balanced read with `$require_open` set to true.

<details>

```tex
\def\foo#1{bar #1 baz}

% does NOT treat \bgroup as {
\foo\bgroup hello}

% treats \bgroup as {
T5.8:
\expanded\bgroup\meaning\egroup}

% also treats \bgroup as {
\lowercase\bgroup SMALL}

\bye
```

</details>

Recognizing tokens that were `\let` to `{` seems to be appropriate, even if I am not sure it matches 100% the original implementation.

I also updated the POD with slightly more recent information.

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -415,7 +415,7 @@ sub readBalanced {
   # Does we need to expand to get the { ???
   if ($require_open) {
     my $token = ($expanded ? readXToken($self, 0) : readToken($self));
-    if ((!$token) || ($$token[1] != CC_BEGIN)) {
+    if ((!$token) || ($$token[1] != CC_BEGIN && !Equals($STATE->lookupMeaning($token), T_BEGIN))) {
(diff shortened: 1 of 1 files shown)


dginev:
Background: I freshly reproduced this error in the Rust port, so decided to come back and understand the root cause better.

PR #2343, opened 2024-04-03
Redefinitions

brucemiller:
This PR fixes a problem with definition locking, particularly when preloading.  The idea (I think!) should be that redefinitions of locked definitions from raw tex (or style files) should generally be ignored, but those from a binding should be respected, even if it is indirectly loaded from another style file.  This wasn't always happening when bindings were preloaded.

Getting this sorted out revealed some un-intended redefinitions. One was the unfortunately named `bibliography` counter (used for ID's within bibliographies)  which is now named `@lx@bibliography`. And also the function `AddToMacro` presumably should allow what is effectively a redefinition.


brucemiller:
The actual fix is relatively simple, but awkward to explain & test. I created a rather convoluted test case `t/daemon/complex/testlocks`.  I created the expected output using `tools/maketests --xml t/daemon/complex/testlocks`. However, the actual testing fails; it looks like it's getting the processing-instructions in random order.  @dginev could you have a look?

dginev:
Took a look. You've ran into a limitation of the daemon testing setup: when the directories get checked, the testing harness will look for *both* `.spec` and `.tex` input files, which means that you've actually added two separate tests - `testlocks.spec` and `testlocks.tex`, both of which will be tested against `testlocks.xml`. Naturally, one of the two runs has to fail, if the spec does anything non-trivial.

The way I've worked around this with the other daemon tests (which mostly share a small number of input `.tex` snippets) is to put all sources a directory level up, where the testing harness isn't looking, in `t/daemon`. So moving that up and changing the source line to:
```
source = ../testlocks.tex
```

Should align with the current testing setup. Definitely room to document that somewhere...

brucemiller:
Ah, ok, that makes sense; I was just trying to hide the clutter of all the extra style files & bindings. Thanks!

dginev (approved):

PR #2357, opened 2024-04-22
cleaning up a few outdated KeyVals corners

dginev:
There are more subtle refactors and updates possible in KeyVals, but here is a PR that cleans up a couple of the more obvious pieces which are now outdated.

[Found while porting]


brucemiller (approved):
ha, that one's almost funny :>

PR #2358, opened 2024-04-24
[WiP] Tex pool reorg

brucemiller:
Split the (huge) `TeX.pool` into more manageable, modular components:
* TeX primitives, grouped according to David Bausum's "families" of primitives
* LaTeXML-specific utilities, helpers, etc
* LaTeXML binding for `plain.tex`.

Future (or continuing) work will 
* regularize *all* LaTeXML-specific control-sequences to start with `\lx@`
* support reading `plain.tex` and `latex.ltx` directly for the bulk of definitions, after establishing what needs LaTeXML-specific binding, eg for semantics.


dginev (approved):
This looks surprisingly well-cataloged! Kudos to David Bausum for the thematic structure.

I generally quite like it. One note is that this kind of PR is the perfect place for subtle typos to sneak in, since git won't micromanage the large chunks of text moving from one file to another, and we can't really inspect it all manually. So I hope copy-paste was bug free :> The tests passing is a big encouragement.

One curious question: Is there an obvious performance footprint from reorganizing? Or are `make test` times comparable to master?

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -433,6 +433,7 @@ sub build_kpse_cache {
   foreach my $path (split(/$KPATHSEP/, $texpaths)) {
     $path =~ s/^!!//; $path =~ s|//+$|/|;
     push(@filters, $path) if -d $path; }
+  my $filterre = '(?:' . join('|', map { "\Q$_\E"; } @filters) . ')';
```
dginev, line 436: Edge case note: I think there is a subtle behavior change here if `@filters` is empty. The new regex will then evaluate to `(?:)`, which apparently matches against every string. So in that case `$skip` would be always false, rather than always true. I think?

brucemiller:
@dginev here's another take, per our discussions; give it another look.

PR #2362, opened 2024-05-02
Use smaller symbols for \triangleleft and \triangleright

DominikPeters:
The current unicode symbols used by LaTeXML for `\triangleleft` and `\triangleright` are larger than the symbols used when LaTeX is used. I propose changing the unicode symbols [U+25C1 ◁ White Left-Pointing Triangle](https://www.compart.com/en/unicode/U+25C1) to [U+25C3 ◃ White Left-Pointing Small Triangle](https://www.compart.com/en/unicode/U+25C3), and similarly for right-pointing.

Example document:
```latex
\documentclass{article}
\begin{document}
    Test string: $x \bigtriangleup y \bigtriangledown z \triangleleft w \triangleright v$
\end{document}
```

PDF:
image
LaTeXML:
image
LaTeXML with this pull request:
image


dginev:
@DominikPeters this PR would also need a few of the tests updated to match the new characters - see a local `make test` run for details.

Btw the presented visual reasoning for transitioning to using the small Unicode character appears reasonable to me on a surface read. But @brucemiller may have more thoughts.

xworld21:
As further datapoint, unicode-math uses U+25C1, and in fact loading unicode-math makes the triangles bigger in the PDF output. Not sure if this is unicode-math redefining the macro, or instead altering the font maps.

brucemiller:
Hmm, well, maybe.

Of course, Unicode symbols don't come with a size (nor TeX commands, for that matter); the fonts that render them do. And so different fonts will have different sizes.  There may be a reasonable *expectation* of size, particularly expecting a set of shapes to have the same size, or relative sizes for normal compared to "big" or "small".

Given the somewhat random collection of symbols used in TeX, along with a different, though related, random collection in Unicode, it's kinda hard to come up with a completely consistent mapping. Apparently the `unicode-math` developer encountered the same mystery?

So, while the example you showed seems compelling --- and you may be quite correct --- I'd be interested in looking at all the various triangles, boxes, big, small, etc  (as defined in a variety of style files) and comparing the observed relative sizes in both pdflatex and with common (eg STIX) unicode fonts.before deciding on changes,

xworld21:
I was curious so I checked: https://www.unicode.org/reports/tr25/ seems quite explicit about how big the triangles should be (see Table 2.5). Although I agree that one should check whether the fonts are following that idea or not. http://mirrors.ctan.org/macros/unicodetex/latex/unicode-math/unimath-symbols.pdf renders each character in several fonts (Latin Modern, STIX, Fira Math, etc). For U+25C1, it seems like they generally agree with the spec, that is, the left/right triangles are roughly the same size as their up/down counterparts. That should clarify the HTML side, at least.

brucemiller:
Hmm. Thanks for the links @xworld21 that should have clarified things, but alas!!!! So, if you look at the basic cm fonts, the `\lefttriangle, \righttriangle` come from `cmmi` while the `\bigtriangleup, \bigtriangledown` come from `cmsy` and are decidedly larger than the former two.  Curiously, the tables in the `unimath-symbols.pdf` show them all similar size for each font shown (but different sizes across different fonts), and have "(large)" in the description of the left & right triangles.  So, clearly, you'll get different relative sizes for things when you select special fonts.
But it also appears that for cm fonts, Knuth intended to have the left & right smaller than the up down, even adding "big" to the names.  So, I guess I have to concede that @DominikPeters is right here...

xworld21:
> So, clearly, you'll get different relative sizes for things when you select special fonts.

Will this be addressed in your new approach to fonts? It sounds like you need a secondary font map 'glyph -> unicode' which uses the right codepoint depending on the actual font (and keep track of the font itself on top of the font family). This would probably matter also for the chancery vs roundhand script characters.

brucemiller:
On 9/22/24 08:30, Vincenzo Mantova wrote:
>     So, clearly, you'll get different relative sizes for things when you
>     select special fonts.
> 
> Will this be addressed in your new approach to fonts? It sounds like you 
> need a secondary font map 'glyph -> unicode' which uses the right 
> codepoint depending on the actual font (and keep track of the font 
> itself on top of the font family). This would probably matter also for 
> the chancery vs roundhand script characters.

The current code allows you do do that, I think, but the upcoming font 
code mainly refines it and makes it more consistent. But setting up 
these mappings is so tedious & error prone,  that I'd wonder if it was 
worthwhile, even so. Finding an appropriate Unicode-point that 
convincingly corresponds to the glyph in a particular font, is hard 
enough; finding 2 or more where one looks like this font, and the other 
looks like that one, is generally hard or rare.

Well, except for the 2 current cases: geometric things where it's hard 
to infer the designers intentions; and the chancery/roundhand 
distinction. The latter is a special case, but I think it it will fit 
the scheme, once I'm in a position to test.

xworld21:
> the chancery/roundhand distinction. The latter is a special case, but I think it it will fit the scheme, once I'm in a position to test.

I have spotted a few other things in glyphtounicode.tex, like some fonts swapping `\phi` and `\varphi`. I haven't actually tried the fonts, but it's something to keep in mind, at least.

brucemiller:
I've made an alternative PR #2422, that uses an intermediate size U+22B2, U+22B3 for the triangles, and also updates the affected tests.  @DominikPeters Let me know if this looks reasonable to you.

PR #2363, opened 2024-05-02
Fix authblk line breaks when font has been customized

DominikPeters:
Author lines typeset using the `authblk` package can have an unexpected behavior when the `\Authfont` has been set to a custom font size and a line break `\authorcr` is used to separate authors.

Example document:
```latex
\documentclass{article}
\usepackage{authblk}
\renewcommand\Authfont{\Large}
\title{Use of authblk}
\author[1]{Alice}
\author[1]{Bob}
\author[1]{\authorcr Charlie} % \authorcr = line break
\author[1]{Dominik}
\affil[1]{Example University}
\begin{document}
	\maketitle
\end{document}
```

| PDF | LaTeXML | LaTeXML with this pull request |
| --- | --- | --- |
| image | image | image
|  | `<br class="ltx_break"><span class="ltx_text" style="font-size:144%;">Charlie</span>` | `<br class="ltx_break">Charlie`

The reason for the difference is that the `authblk` binding does not actually apply the `\Authfont` to the author name text in its current version, except after `\authorcr`. 
https://github.com/brucemiller/LaTeXML/blob/f75b0c6b06ff9191178a008c62a95f2b9c1dd087/lib/LaTeXML/Package/authblk.sty.ltxml#L27
This pull request removes this:
https://github.com/brucemiller/LaTeXML/blob/ca7e2146e73b9e566b797992487cc8102f0fe14a/lib/LaTeXML/Package/authblk.sty.ltxml#L27

Example to make it clear that normally `\Authfont` is ignored: let's set
```latex
\renewcommand\Authfont{\bfseries}
```
Then LaTeXML outputs the following:
image

LaTeXML with this pull request gives the same output as in the above table, right-most cell.

### DIFF
--- a/lib/LaTeXML/Package/authblk.sty.ltxml
+++ b/lib/LaTeXML/Package/authblk.sty.ltxml
@@ -24,7 +24,7 @@ DefMacro('\Authfont',  '\normalfont');
 DefMacro('\Authsep',   ',');
 DefMacro('\Authand',   ' and ');
 DefMacro('\Authands',  ', and ');
-DefMacro('\authorcr',  '\\\\\Authfont');
+DefMacro('\authorcr',  '\\\\');
 
 # package bookkeeping
 DefConditional('\ifnewaffil');


brucemiller:
Would it be "better" to use `\Authfont` consistently, versus using it not-at-all?

dginev:
> Would it be "better" to use \Authfont consistently, versus using it not-at-all?

From my usual corpus perspective, if the author is captured in the structured frontmatter, it can be styled with CSS. The TeX-based customization using `\Authfont` may make it harder to interplay with a generic style, so - if used consistently - it should ideally deposit a `class` attribute rather than a `style` directvie.

DominikPeters:
I guess one could merge this PR and open an issue saying that \Authfont is being ignored. Then someone could work on implementing that separately at some point. The current behavior is just clearly not right.

brucemiller:
OK, fine. Ganging up on me!!
Thanks for the contribution!

PR #2364, opened 2024-05-02
Limit length of captions in wrapfigures using CSS

DominikPeters:
The [current bindings](https://github.com/brucemiller/LaTeXML/blob/f75b0c6b06ff9191178a008c62a95f2b9c1dd087/lib/LaTeXML/Package/wrapfig.sty.ltxml#L21) for the `wrapfig` package ignore the author-specified width. Together with the standard css, this leads to a bad presentation of wrapfigures with long captions. This pull request adds CSS (setting [`width: min-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/min-content) on the `<figure>` element) to limit the length of the caption of a wrapfigure to the length of the image. (Probably the more elegant solution is to take into account the width parameter of the `wrapfigure` environment, but this is a simple fix for the moment.)

Example:
```latex
\documentclass{article}
\usepackage{wrapfig,tikz}
\begin{document}
    \begin{wrapfigure}{r}{0.4\textwidth}
        \centering
        \tikz{\fill[purple] (0,0) rectangle (4,2);}
        \caption{A red rectangle deserving a long caption.}
    \end{wrapfigure}
    This is a\foreach \wordnum in {0,...,300} { test}.
\end{document}
```

PDF:
image

LaTeXML (note that if the caption gets longer, the figure can take up up to 100% of the width of the page):
image

LaTeXML with this pull request:
image

### DIFF
--- a/lib/LaTeXML/resources/CSS/ltx-article.css
+++ b/lib/LaTeXML/resources/CSS/ltx-article.css
@@ -52,6 +52,10 @@
 /* experimental: flex model, to center short captions, justify long ones */
 .ltx_table,
 .ltx_figure { display:flex; flex-direction:column; align-items:center; }
+.ltx_table.ltx_align_floatright,
+.ltx_table.ltx_align_floatleft,
+.ltx_figure.ltx_align_floatright, 
+.ltx_figure.ltx_align_floatleft { width:min-content; }
 .ltx_table .ltx_caption,
 .ltx_figure .ltx_caption { text-align:justify; }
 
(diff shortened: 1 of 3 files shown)


brucemiller:
Seems reasonable on the surface. @dginev how does this interact with your work?

dginev:
Well, it is hard to say for certain without adding a few more tests. The one test I had checked in (`t/complex/figure_mixed_content.tex`) has a `{wrapfigure}` that contains an `{algorithm}` in a minipage, which is already a nice counter-example:

Using the mainline latexml, and ar5iv.css, that figure renders as:


with the PR, because the figure contains textual content, the `min-content` directive compresses down to what is essentially the length of the longest word:


So we may want to make such CSS enhancement a little more precise, applying only to the image and table cases. And check in tests for those respectively.

brucemiller:
So, it seems that the wrapfigures request of width is probably a relevant constraint.

dginev:
Let's close here since `width: min-content;` really isn't the right approach. 

I have opened a issue where we can discuss further at #2578

Issue #2354, opened 2024-04-12
JATS: Use Commonmeta to generate JATS metadata


castedo:
This is a rough feature idea.

SUMMARY:
Use a JSON-compatible file format (JSON, YAML, TOML, even [JSOML](https://pypi.org/project/jsoml/) :sweat_smile:) that conforms to https://commonmeta.org or similar in parallel to TeX files to generate JATS XML.

USER SCENARIO:
Technically competent authors that keep article metadata separate from .tex files, for example:
https://github.com/sabinomaggi/ten-years-challenge-pulsed-drive/blob/submission/article/metadata.yaml
Simple scripts can convert such files into .tex files to be `\input`.

This approach might be a good solve for #1648, #2351, #2352, #2353.

dginev:
This suggestion sounds a little experimental to the usual LaTeX ecosystem, which as you know is where latexml places its main focus. Having structured inputs is lovely, and life would be easier if all of a document's frontmatter/backmatter was already in a standard structured format before the conversion starts. But in LaTeX it generally isn't, and conventions can vary wildly between different class files for `\documentclass`.

I can certainly imagine a new `commonmeta.sty` package with its own `commonmeta.sty.ltxml` support for latexml, which deals with its own kind of metadata. 

The current organizational thinking for latexml is to start such experimental extensions outside of the core code, to avoid eclipsing the "usual" LaTeX needs.

castedo:
> start such experimental extensions outside of the core code

Yeah, that make sense to me to keep the core code (in Perl) separate from any code implementing the initial rough feature idea here.

What I'm thinking now is just hashing out some terminology and extra documentation. For instance, for the medium-term, LaTeXML could output JATS XML which is lacking important metadata *by design*. Then documentation can describe various ways that this limited JATS can be enhanced to have all kinds of fancy metadata.

In particular, now I am thinking a quick cheap recommendation is for an author to literally hand-code the XML that goes into the article meta and then use a trivial XSLT to merge the LaTeXML output JATS with the hand-coded XML with article metadata. Then additional documentation can suggest other approaches with 3rd party software that are fancier and more DRY (like using JSON/YAML etc...).

What you think @dginev?

castedo:
BTW, what does "backmatter" mean?

dginev:
[...]

PR #2365, opened 2024-05-18
hyperref and hyperxmp metadata improvements

xworld21:
Add remaining `\hypersetup` keywords and add `xml:lang` if `pdfmetalang` or `pdflang` were specified (and LaTeXML should probably use `pdflang` as document language if not already specified elsewhere). Note that this also fixes the wrong mapping of `pdfsubject` – it should be `dcterms:description` not `dcterms:subject`.

The metadata could then be used elsewhere (e.g. JATS #2354, EPUB), but this PR is only about preserving the data in the XML output.

Note that the implementation is incomplete: certain properties (such as authors) are implemented as lists by hyperxmp (`<rdf:Seq><rdf:li>...`, sometimes `<rdf:Bag><rdf:li>...`, or `<rdf:Alt><rdf:li>` for alternative languages). I haven't touched any of that as LaTeXML implements RDF essentially as strings.


brucemiller:
This all looks plausible; I don't use XMP myself, but this should probably help those who do.  @dginev do you have any thoughts on this? (other than not being an XMP fan, as I recall :> )

dginev:
I have a classic comment: it would be nice to have a test, which will have the additional effect of self-documenting how these XMP enhancements can be used by authors. I haven't used hyperxmp.sty myself, so I trust @xworld21 knows more than me here.

dginev:
I am also new to PRISM, but so far I find it palatable - there is a [W3C Member Submission](https://www.w3.org/submissions/2020/SUBM-prism-20200910) with a reasonable [W3C Team Comment](https://www.w3.org/submissions/2020/01/Comment/).

One observation while quickly skimming that, the [PRISM Namespaces](https://www.w3.org/submissions/2020/SUBM-prism-20200910/prism-basic.html#_Toc46322878) section lists the URI added in this PR as a `basic:` prefix, as there appear to be another ~10 other PRISM-related namespace URIs. Maybe we want to qualify that in latexml as `prismbasic:` or such?

xworld21:
> it would be nice to have a test

Excellent idea, and it's making me discover details I'd missed (e.g. only some properties respect `pdfmetalang`, others support a language by prefixing `[en]`).

Is it ok if I include the sample document included in the hyperxmp documentation? Would there be license issues with that?

dginev (approved):
Looks good on my end.

PR #2368, opened 2024-05-27
remove redundant \expandafter definition

dginev:
Spotted while porting. There appear to be two \expandafter definitions following each other. 

As the first one wasn't getting used (immediately overwritten by the second), this PR removes it for clarity.


dginev:
I also removed a redundant `\relpenalty` (TeXbook value seems to be 500)

brucemiller:
I'll have to remember to take a look at `\newread` & `\newrite`, as there are some subtle differences between `plain.tex` and `plain.pool` approaches which need sorting.

brucemiller:
Looks good; thanks!

PR #2369, opened 2024-05-29
Base primitive improvements

brucemiller:
Fixes various little fidelity problems that turned up while playing with `plain.tex`.
* recognize more font names
* initialize State more like INITEX does (UC,LC codes, etc)
* a bit better with `\nonscript`, '\fam`,
* make `\write` expand in the special `\the` style ('XGeneralText` parameter type)
* Various primitives, '\mathopen`, etc, should digest args
* Fixup `\left,\right` processing to digest args while still grouping between them.


lib/LaTeXML/Core/Document.pm
```diff
@@ -250,6 +250,7 @@ sub canAutoClose {
     || (($t == XML_ELEMENT_NODE)                              # otherwise must be element
     && !$node->getAttribute('_noautoclose')                   # without _noautoclose
     && ($node->getAttribute('_autoclose')                     # and either with _autoclose
+      || $node->getAttribute('_autoopened')                   # or was autoopened
```
dginev, line 253: Does this mean that the implied "default" now is for any Tag() marked as `autoOpen => 1`, it will also lead to `autoClose => 1` ?

I see we have a special attribute `_noautoclose` that prevents autoclose, but it makes me wonder if we should also respect an explicit `autoClose => 0` from Tag(). Or document that we won't, or...

lib/LaTeXML/Engine/TeX_Math.pool.ltxml
```diff
@@ -670,53 +692,67 @@ DefConstructor('\mathinner{}', "?#1(<ltx:XMWrap role='ATOM'>#1</ltx:XMWrap>)()",
...
   );
 
-# With new treatment of Simple Symbols as just Box's with assigned attributes,
-# we're not getting whatsits, and so we're not looking them up the same way!!!
-# TEMPORARILY (?) hack the Delimiter map
-foreach my $entry (values %DELIMITER_MAP) {
-  $DELIMITER_MAP{ $$entry{char} } = $entry; }
+# The \lx@left, \lx@right versions are like \left,\right but without any extra grouping
+DefConstructor('\lx@left TeXDelimiter',
+  "?#hint(<ltx:XMHint/>)(#1)",
```
dginev, line 727: What is the idea behind the empty XMHint elements? I see they are used in a few definitions here. I wonder if they should carry more information to be useful - or are they already useful when empty?

I see the code adds them when the `\left`/`\right` is applied to a dot, which indeed has no visible trace in rendering. Maybe an attribute on XMHint could indicate some of that?

t/parse/parens.xml
```diff
@@ -537,21 +537,21 @@
         </Math>
       </equation>
       <equation xml:id="S5.Ex21">
-        <Math mode="display" tex="\left|{a\over b}\right|" text="absolute-value@(a / b)" xml:id="S5.Ex21.m1">
+        <Math mode="display" tex="\left{\delimiter 69632778}{a\over b}\right{\delimiter 86414091}" text="delimited-⟨⟩@(a / b)" xml:id="S5.Ex21.m1">
```
dginev, line 540: Hm, we may want to avoid the `{` and `}` entirely when reverting delimiters. The tex attribute on the Math above encounters an error in pdftex. What succeeded was the plainer form from parens.tex:
```tex
$\left\delimiter 69632778 {a\over b}\right \delimiter 86414091$
\bye
```

Since we have recently fixed a report about `tex` attribute reuse with TeX, I thought I would bring this up.

dginev (commented):
I am a little out of my depth on the TeX details, but tried to provide a first surface review. 

`plain.tex` itself will certainly be the harsher arbiter here.

lib/LaTeXML/Core/Document.pm
brucemiller: When you put it like that, it does seems too much, although it's probably correct most/all of the time?  The problem element in this case was the `ltx:text` opened for font switches, so better to explicitly mark them as auto-closable, as well.  patched...

lib/LaTeXML/Engine/TeX_Math.pool.ltxml
brucemiller: I inherited the hint from the previous code, but I doubt the hint is ever actually used. Probably put there either so there's at least 1 element, or "just in case".

t/parse/parens.xml
brucemiller: Yeah, good catch. patched..

dginev (approved):

PR #2380, opened 2024-07-17
update xsl rule for moderncv personal position/title

dginev:
A bit of a minor tweak that makes the moderncv.cls output capable of handling a simple plain-text position (originating in the `\title` macro). The previous version was expecting a multiline construct in a `\vbox`, which is a bit too specific.

Funny story how I stumbled on it...


dginev:
Also, this *could* be refactored to a dedicated rule specific to `ltx:contact[@role='position']`, let me know if that's preferable to the `xsl:if` test.

brucemiller:
seems reasonable, although there may be a more general approach (later).

PR #2384, opened 2024-08-06
Extra safety when unreading alignment-related characters for \aligned

### LINKED ISSUE #2383: Robustness to stacked aligned environments
There appears to have been a change of behavior where a particular stacked alignment configuration triggers an edge case error. If I am diagnosing this right, the necessary conditions are:
- two stacked aligned environments, such as `{eqnarray}` and `{aligned}`,
- a body with braced wrappers, such as `{ARG}`.

A minimal example is:
```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}

Safe:

\begin{eqnarray}
\begin{aligned}
  A  % this is OK
\end{aligned}
\end{eqnarray}

In contrast:

\begin{eqnarray}
\begin{aligned}
{A}
\end{aligned}
\end{eqnarray} % an ERROR is emitted here

Also broken:

\begin{eqnarray}
\begin{aligned}
 A \\
\end{aligned}
\end{eqnarray}  % an ERROR is emitted here

\end{document}
```

The original document contained this exact stacking, for a 2-line display equation. It may have been an author mistake, but the error cascade is quite problematic - the example I was examining had the entire conversion Fatal as a result. This works well in pdflatex.

dginev:
Fixes #2383 . Quite subtle, but I think I got it - I can make the minimal example from the issue into a test if preferred.

Details: The key problem was that when the `->unread` of a `{` happened, the *actual* aligned state got modified, rather than the "safely neutralized" local state, which appears to be the primary purpose of this parameter type.

### DIFF
--- a/lib/LaTeXML/Package/amsmath.sty.ltxml
+++ b/lib/LaTeXML/Package/amsmath.sty.ltxml
@@ -590,8 +590,8 @@ DefPrimitive('\@ams@aligned@bindings', sub {
 DefParameterType('alignsafeOptional', sub {
     my ($gullet, $default, $inner) = @_;
     my ($tok, $value);
-    { local $LaTeXML::ALIGN_STATE = 1000000;
-      $tok = $gullet->readNonSpace; }
+    local $LaTeXML::ALIGN_STATE = 1000000;
+    $tok = $gullet->readNonSpace;
     if    (!defined $tok) { }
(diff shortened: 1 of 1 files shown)


brucemiller:
Your observations seem right, and this patch doesn't cause any surprises that I can detect. Thanks!

PR #2396, opened 2024-08-17
restore viewBox in pgf pictures

### LINKED ISSUE #2376: add viewBox to inline SVGs

xworld21:
Fix #2376 for PGF/TikZ.

The [commit that removed `@viewBox`](https://github.com/brucemiller/LaTeXML/commit/b1d4af1958c1e1e0f222b3e3deb933e04a6e3b92 ) said 'AVOID scaling, in particular viewBox so that text size matches containing document'. I can't quite tell what issue was being addressed: if `@viewBox` and `@width`, `@height` agree, no scaling happens. Rather, having the correct `@viewBox` *enables* CSS to change width and height, and that can only happen intentionally. So I am guessing this PR is fine.

### DIFF
--- a/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
+++ b/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
@@ -85,9 +85,9 @@ DefConstructor('\lxSVG@insertpicture{}', sub {
     else {
       $document->openElement('ltx:picture');
       $document->openElement('svg:svg',
-        version => "1.1",
-        width   => $props{pxwidth}, height => $props{pxheight},
-####        viewBox  => "$props{minx} $props{miny} $props{pxwidth} $props{pxheight}",
+        version  => "1.1",
+        width    => $props{pxwidth}, height => $props{pxheight},
+        viewBox  => "$props{minx} $props{miny} $props{pxwidth} $props{pxheight}",
         overflow => "visible");
       my $x0 = -(0 + $props{minx});
       my $y0 = $props{pxheight} + $props{miny};


brucemiller:
No, sadly just adding the `viewbox` line alone doesn't work, even though you'd reasonably think that it should.  When there are graphics and text combined into the figure, somehow, somewhere, inconsistencies between the dimensions & ems used in the TeX processing clashes with the dimensions & ems used by the browser, and stuff ends up looking sometimes quite horrible.  I'll have to go dig through my sample set to find an example.  
Forcing px and omitting the viewbox was the only way I got things to look decent, although I'll freely admit that I doubt this is the Correct solution. Maybe `rem` is part of it?

xworld21:
> sadly just adding the `viewbox` line alone doesn't work, even though you'd reasonably think that it should. When there are graphics and text combined into the figure, somehow, somewhere, inconsistencies between the dimensions & ems used in the TeX processing clashes with the dimensions & ems used by the browser, and stuff ends up looking sometimes quite horrible. I'll have to go dig through my sample set to find an example

Thanks, the example would be really helpful! I really struggle to see how adding the viewbox can affect the output (or rather, how omitting it helps in any way). As far as I understand, the viewbox `0 0 width height` is essentially the 'default' value of the SVG anyway, so it is a no-op. Provided CSS is not manipulating the images further.

(I have enabled the viewbox in BookML a few weeks ago with no problems, but it's not too significant... most of my few users are still using dvisvgm for TikZ so they won't notice the difference.)

xworld21:
> Forcing px and omitting the viewbox was the only way I got things to look decent, although I'll freely admit that I doubt this is the Correct solution. Maybe `rem` is part of it?

PS: I have done some tweaks in BookML about this by saving the 8/9/10/11/12pt class option, the dpi (default 100 in LaTeXML), and readjusting the SVGs in CSS (requires viewbox). The results are reasonable as long as the fonts are similar enough. I specifically tested with Latin Modern in both PDF and HTML, and got a perfect match, of course.

brucemiller:
Looks good! Thanks!!

PR #2400, opened 2024-08-18
update \XC@mcolor in pgf.sty rather than pgfsys-latexml.def

### LINKED ISSUE #2398: TikZ pictures in title, authors, etc have no color
The following
```latex
\documentclass{article}
\usepackage{tikz}
\title{Title}
\author{Author\begin{tikzpicture}\filldraw[red] (0,0) circle (4pt);\end{tikzpicture}}
\date{1 January 1980}
\begin{document}
  \maketitle

  \begin{tikzpicture}\filldraw[red] (0,0) circle (4pt);\end{tikzpicture}
\end{document}
```
should have a red disk next to the author name. However, it comes out black (or more precisely, it takes the color of the surrounding text). This happens only when `\author{}` is specified in the preamble; behaviour seems normal if `\author{}` is after `\begin{document}`

In preamble:
```xml
<svg:g fill="#000000" stroke="#000000" ...>
```
In document body (note the additional `@color`):
```xml
<svg:g color="#FF0000" fill="#FF0000" stroke="#FF0000" ...>
```

(First spotted in https://github.com/brucemiller/LaTeXML/pull/2395#issuecomment-2294924608.)

xworld21:
The change guarantees that `\XC@mcolor` is updated after color/xcolor are loaded, but also before finishing loading PGF, as opposed to at the end of the preamble. This also improves the usage of `\XC@mcolor`: in principle, other packages may be adding stuff to it, so it should be extended rather than redefined.

This fixes #2398, and hopefully does not break anything.

### DIFF
--- a/lib/LaTeXML/Package/pgf.sty.ltxml
+++ b/lib/LaTeXML/Package/pgf.sty.ltxml
@@ -37,6 +37,11 @@ DefMacro('\pgfsetcolor{}', '\pgfsetcolor@orig{#1}\lxSVG@set@color');
 DefPrimitive('\lxSVG@set@color', sub {
     MergeFont(color => LookupValue('color_pgfstrokecolor')); });
 
+# This seems to be needed to actually set colors (code similar to xxcolor.sty)
+if (XEquals(T_CS('\XC@mcolor'), T_CS('\relax'))) {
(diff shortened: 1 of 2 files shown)


brucemiller:
Yeah, that looks about right! Thanks!!

PR #2401, opened 2024-08-18
respect PGF/TikZ bounding boxes

### LINKED ISSUE #2399: TikZ bounding box is not respected
Using `\useasboundingbox` does not achieve the desired result. In TikZ, that command sets a bounding box (`\pgf@picminx` etc) and stops updating it to accommodate new paths that go past the boundary. Strangely enough, LaTeXML keeps enlarging the bounding box anyway. For now, the only (?) way to get the correct bounding box is to run `\pgfresetboundingbox` at the end, followed by the desired `\useasboundingbox`.

(This has actual consequences. For example, if a figure is clipped to a small bounding box, LaTeXML will clip correctly but leave a big bounding box, resulting in lots of empty space.)

Ugly example where the two figures should be identical:
```latex
\documentclass{article}
\usepackage{tikz}
\begin{document}
  text before:
  \begin{tikzpicture}
    \pgfresetboundingbox
    \useasboundingbox (0,0) rectangle (3,2);
    \draw[blue, very thick] (0,0) rectangle (3,2);
    \draw[green, very thick] (-2,-2) rectangle (4,3);
  \end{tikzpicture}
  :text after

  text before:
  \begin{tikzpicture}
    \draw[blue, very thick] (0,0) rectangle (3,2);
    \draw[green, very thick] (-2,-2) rectangle (4,3);
    \pgfresetboundingbox
    \useasboundingbox (0,0) rectangle (3,2);
  \end{tikzpicture}
  :text after
\end{document}
```

HTML:
image

PDF:
image

xworld21:
Allow PGF and TikZ to set \pgf@relevantforpicturesizefalse, which is how pictures can override their own bounding box (see 'use as bounding box').

Fixes #2399. As usual: does it break something else?

The undesirable side effect is that figures that purposefully draw outside of the bounding box will overflow into the surrounding text. That might have been the original motivation for locking `\ifpgf@relevantforpicturesize` to true.


dginev:
> As usual: does it break something else?

I carried over some tikz tests that Bruce had prepared some time back, which you can use to double-check for any obvious regressions. They were added in #2314 and have a separate make target, via:

```
CI=true make extratest
```

They won't run in CI by default either, since they take a long while to complete...

lib/LaTeXML/Package/pgf.sty.ltxml
```diff
@@ -38,9 +38,9 @@ DefPrimitive('\lxSVG@set@color', sub {
     MergeFont(color => LookupValue('color_pgfstrokecolor')); });
 
 # This makes this conditional always true;  Why???
-DefConditionalI('\ifpgf@relevantforpicturesize', undef, sub { 1; });
-DefMacroI('\pgf@relevantforpicturesizefalse', undef, '');
-DefMacroI('\pgf@relevantforpicturesizetrue',  undef, '');
+# DefConditionalI('\ifpgf@relevantforpicturesize', undef, sub { 1; });
+# DefMacroI('\pgf@relevantforpicturesizefalse', undef, '');
+# DefMacroI('\pgf@relevantforpicturesizetrue',  undef, '');
```
brucemiller, line 43: Cool! You might as well just remove those lines, as well as the obsolete comment preceding them.

PR #2404, opened 2024-08-21
Accents

brucemiller:
This PR is another round in improving the fidelity of lower-level TeX and plain.tex. It deals with accents. Accents in TeX fonts basically show an accented space which gets overlaid on top of a character to be accented (potentially with some positioning tweaks).  To deal with Unicode and MathML, we need  to distinguish 3 kinds of thing:
* a combining char to follow a letter
* a standalone char (or string) which looks like the TeX glyph. This will either be a "spacing" analog of the accent, or at worst a space followed by the combiner
* an "unwrapped" char which is just the accent part which would be the content of a Math token used as an over/under script.
So, we've set up a table within `LaTeXML::Util::Unicode` to store this info, and use it wherever accenting operations are needed.

### DIFF
--- a/lib/LaTeXML/Engine/TeX_Character.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Character.pool.ltxml
@@ -14,7 +14,7 @@ package LaTeXML::Package::Pool;
 use strict;
 use warnings;
 use LaTeXML::Package;
-use Unicode::Normalize;
+use LaTeXML::Util::Unicode;
 
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 # Character Family of primitive control sequences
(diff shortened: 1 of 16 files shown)


lib/LaTeXML/Util/Unicode.pm
```diff
@@ -24,6 +23,73 @@ sub UTF {
   my ($code) = @_;
   return pack('U', $code); }
 
+my $NBSP = UTF(0xA0);
+sub NBSP { return $NBSP; }
```
dginev, line 27: Is this attempting to guard the lexical `$NBSP` from some sort of external redefinition?

Surprised it wasn't introduced as a package-level `our $NSBP;` declaration. That can be added to `@EXPORT` too, if desired.

lib/LaTeXML/Util/Unicode.pm
```diff
@@ -24,6 +23,73 @@ sub UTF {
...
+    unwrapped => ',', pos => 'UNDERACCENT' },
+  { name => 'ogonek', combiner => "\x{0328}", standalone => "\x{02DB}",
+    unwrapped => "\x{02DB}", pos => 'UNDERACCENT' },                            # not even math???
+);
+# Set up a hash keyed on both standalone & combiner chars
+our %accent_data = ();
+foreach my $entry (@accent_data) {
+  $accent_data{ $$entry{standalone} } = $entry;
+  $accent_data{ $$entry{combiner} }   = $entry;
+}
```
dginev, line 86: Oh, this is rather dangerous. Could you change the name of either `@accent_data` or `%accent_data`?

I was thrown for a bit not knowing what happened, seeing the hash indexing syntax in `$accent_data{$char}`. 

Such a setup is bound to generate a subtle bug one day, when someone uses the wrong delimiters in haste:
```perl
my @a = ('x','y','z');
my %a = ('1'=>'a', '2'=>'b', '3'=>'c');
print $a[2]; # z
print $a{2}; # b
```

dginev: I think the hash variant can be kept, and you can even fully expand the loop that adds `standalone` and `combiner` in code, to avoid looping the setters every time latexml is ran...

Simplest is to run it once, print with Dumper, then copy the result back into the file. And on second read - aren't `standalone` and `combiner` already set in the explicit hashes? Maybe the loop is a leftover? Except that the loop sets them as equal, while the hash often has the values as different. Hm, I wish I understood a little more...

dginev (approved):
I left some perl-oriented comments.

Would need to study accents a bit more before I can say anything substantive on the TeX emulation side of affairs.

lib/LaTeXML/Engine/TeX_Character.pool.ltxml
```diff
@@ -70,43 +70,31 @@ sub applyAccent {
...
-      applyAccent($stomach, $letter, $combiner, $acc, $reversion); }
-    else {
-      Warn('unexpected', "accent$n", $stomach, "Accent '$n' not recognized");
-      Box(ToString($letter), undef, undef, $reversion); } });
+    my $encoding = LookupValue('font')->getEncoding || 'OT1';
+    my $char     = ($encoding ? FontDecode($n, $encoding) : chr($n));
+    if (my $entry = unicode_accent($char)) {
+      applyAccent($stomach, $letter, $$entry{combiner}, $$entry{standalone},
+        Invocation(T_CS('\accent'), $num, $letter)); }
+    else {    # Unknown accent ?  Really should OVERLAY it on top of $letter???
```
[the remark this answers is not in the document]
brucemiller: OK, put in a simplistic handling for this case, along w/CSS and an extra line in test.

lib/LaTeXML/Util/Unicode.pm
brucemiller: Thanks, switched over to cleaner `\N{NBSP}` approach (along with `use charnames ':full';` for older perls)

lib/LaTeXML/Util/Unicode.pm
brucemiller: Renamed a bit to make it less scary.

dginev (approved):
Thank you for the updates, the PR looks ready to merge. ✅ 

I tried another two alternatives for the "unknown accent" markup, visible in [this codepen](https://codepen.io/dginev/pen/GRbdVba). Not a focus at the moment, but something to mull over in the background. `<ruby>` and `display: inline-grid;` are curious tools.

PR #2409, opened 2024-08-26
prune an empty \bibitem when auto-opened without need

### LINKED ISSUE #2403: Spurious bibitem in {thebibliography} with custom macros
This is a subtle issue that arises in `.bbl` files which also contain some custom bookkeeping macros - things like conditionals and assignments to registers - just before some or all concnrete `\bibitem` cases.

A distilled minimal example is:
```tex
\documentclass{article}
\begin{document}
\begin{thebibliography}{1}

\relax\bibitem[Author et~al.(2019)Author, Other]{Author2019}
  Some content.

\end{thebibliography}
\end{document}
```

which produces the bibliography:
```xml
  <bibliography xml:id="bib">
    <title>References</title>
    <biblist>
      <bibitem xml:id="bib.bib1">
        <tags>
          <tag>[1]</tag>
          <tag role="refnum">1</tag>
        </tags>
        <bibblock/>
      </bibitem>
      <bibitem key="Author2019" xml:id="bib.bibx1">
        <tags>
          <tag>[Author et al.(2019)Author, Other]</tag>
          <tag role="refnum">Author et al.(2019)Author, Other</tag>
        </tags>
        <bibblock>
Some content.
</bibblock>
      </bibitem>
    </biblist>
  </bibliography>
```

It should be possible to avoid creating the leading empty `[1]` bibitem.

dginev:
Fixes #2403 

With Bruce's help, I zeroed in on the cause of the issue, namely `\par@in@bibliography`. That code, which defensively auto-opens a `\bibitem`, recognizes too few of the legitimate cases with macro use (as described in the issue).

Since it is rather high difficulty to both handle "sloppy" uses of `{thebibliography}` as well as the advanced macro uses before a real `\bibitem`, this PR resigns to undoing the error after it is caused.

Namely, this first stab suggests to check in the `afterDigest` hook of `\lx@bibitem` whether the previous recorded box was an _empty_ use of `\bibitem`. In the case it was, the assembled previous box is pruned from the digested list, and its `RefStepID` is reused.

Seems to work well on the few cases I tried, open to adjustments.

### DIFF
--- a/lib/LaTeXML/Engine/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Engine/LaTeX.pool.ltxml
@@ -4263,11 +4263,22 @@ DefConstructor('\lx@bibitem[] Semiverbatim',
     . "#tags"
     . "<ltx:bibblock>",
   afterDigest => sub {
+    # check if the previous bibitem had an empty body,
+    # in which case prune it and reuse its ID and counters.
+    # this happens due to auto-open being a bit fragile (see issue #2403)
+    my $pruned_prev = 0;
+    if (my $prev_item = $LaTeXML::LIST[-1]) {
+      if (ref $prev_item eq 'LaTeXML::Core::Whatsit' and
+        $prev_item->getDefinition->getCS->getCSName eq '\lx@bibitem' and
+        not($prev_item->getArg(1)) and IsEmpty($prev_item->getArg(2))) {
+        $pruned_prev = 1;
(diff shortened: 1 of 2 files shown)


brucemiller (approved):
This seems pretty tricky, but I can't think of a better way to deal with it, other than understanding all the cases where an empty bibitem is created and somehow avoiding it.  So, let's go with it.. Thanks!

PR #2411, opened 2024-09-03
Plain fonts

brucemiller:
This PR gets most of the low-level font related commands better cooperating, in particular control sequences defined by `\font` are now represented by a `FontDef` which then will change to the font, and is also recognized by `\the` and `\meaning`, as well as the accessors `\fontdimen`, etc.  These commands are also stored in `\textfont,\scriptfont,\scriptscriptfont` for use with `\fam` and the `\mathcode` mechanism to choose fonts for math tokens in a particularly interesting way, which now mostly works.  Consequently most of plain.tex mostly works.

This mostly cooperates with the higher-level LaTeX style of font selection currently used by LaTeXML (by specifying family, shape, etc, w/o explicitly naming fonts or font files); later work will need to integrate these better.


brucemiller:
Interesting pragmatic, if counter-intuitive, development:  When alphanumerics appear in math, we've been keeping their content ASCII, but recording the (presumably semantic) styling, which gets dealt with appropriately by the post-processor. That's not so convenient for certain more exotic styles (Caligraphic, Blackboard bold) in text mode, where the style ultimately must be handled with CSS.  This last set of commits makes this distinct handling more explicit and consistent,  even when lower-level TeX commands such as `\char` are used.

lib/LaTeXML/Engine/LaTeX.pool.ltxml
```diff
@@ -2804,8 +2809,9 @@ DefMacro('\UseTextAccent{}{}', '{\fontencoding{#1}#2{#3}}');
 DefPrimitive('\DeclareMathAccent DefToken {}{} {Number}', sub {
     my ($stomach, $cs, $kind, $class, $code) = @_;
     $class = ToString($class);
-    my $info  = LookupValue('fontdeclaration@' . $class);
-    my $glyph = FontDecode($code->valueOf, ($info ? $$info{encoding} : $class));
+    my $info = LookupValue('fontdeclaration@' . $class);
+    #    my $glyph = FontDecode($code->valueOf, ($info ? $$info{encoding} : $class));
+    my ($glyph) = FontDecode($code->valueOf, ($info ? $$info{encoding} : $class));
```
dginev, line 2814: This array unwrap looks a little scary/fragile...
Should we check for `wantsarray` in FontDecode? 

We could even avoid computing the font details if we are only requesting the scalar `$glyph` back.

lib/LaTeXML/Engine/TeX_Debugging.pool.ltxml
```diff
@@ -102,14 +102,15 @@ DefMacro('\meaning Token', sub {
       $type =~ s/^LaTeXML:://;
       # Pre-step: We can't extract the bodies of definitions which are defined via Perl subroutines.
       # So do the next best thing -- represent them as their tokens.
-      if ($type =~ /(primitive|conditional|constructor)$/i) {
-        $definition = $definition->getCSorAlias;
-        $type       = ref $definition;
-        $type =~ s/^LaTeXML:://;
-        if (my $fontinfo = LookupValue('fontinfo_' . ToString($definition))) {
-          $meaning = 'select font ' . ($$fontinfo{fontname} || 'fontname');
+      if ($type =~ /(fontdef)$/i) {
```
dginev, line 105: `(fontdef)` doesn't need the parens, I assume they're leftover

lib/LaTeXML/Engine/TeX_Debugging.pool.ltxml
```diff
@@ -102,14 +102,15 @@ DefMacro('\meaning Token', sub {
...
+        if (my $fontinfo = $definition->isFontDef) {
+          $meaning = 'select font ' . ($$fontinfo{name} || 'fontname');
```
dginev, line 107: Nice, this is a very healthy improvement for `\meaning`.

lib/LaTeXML/Engine/TeX_Fonts.pool.ltxml
```diff
@@ -27,69 +27,79 @@ use LaTeXML::Package;
...
-DefParameterType('FontToken', sub {
-    my ($gullet) = @_;
-    my $token = $gullet->readToken;
+# A Defined Font, or Font Identifier,
+# defined by \font, OR \textfont, \scriptfont,\scriptscriptfont, or even \font itself
+# Actually returns the fontinfo (hash) associated with the font.
+DefParameterType('FontDef', sub {
+    my ($gullet)  = @_;
+    my $readtoken = $gullet->readToken;
+    my $token     = $readtoken;
```
dginev, line 36: I think `$readtoken` is never used. Leftover?

dginev (approved):
This is very impressive (and partially beyond my understanding).

I added some minor comments related, largely just perl tidbits.
I will study some more the new test - which looks very helpful.

Looks good to merge!

lib/LaTeXML/Engine/LaTeX.pool.ltxml
brucemiller: I'd been tempted to minimize the change to the API by doing that, but Perl tricks you. A previous call had been `Box(FontDecode(...),...)` where you'd think you were getting *only* the `$glyph`, but actually wantarray is true!

lib/LaTeXML/Engine/LaTeX.pool.ltxml
```diff
@@ -2772,11 +2772,16 @@ DefPrimitive('\DeclareTextSymbol DefToken {}{Number}', sub {
...
+    DefPrimitiveI($ecs, undef, sub {
+        my ($glyph, $adjfont) = FontDecode($code, $encoding);
+        Box($glyph, $adjfont, undef, $cs); });
     return; });
 
-# hmmm... what needs doing here; basically it means use this encoding as the default for the symbol
-DefMacro('\DeclareTextSymbolDefault DefToken {}', '');    # '\DeclareTextSymbol{#1}{?}');
+DefPrimitive('\DeclareTextSymbolDefault DefToken {}', sub {
+    my ($stomach, $cs, $encoding) = @_;
+    $encoding = ToString(Expand($encoding));
```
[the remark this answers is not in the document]
brucemiller: Interestingly, there's already an `Expanded` parameter type, but it's slightly special: requires braces and treats `\the` specially.  Perhaps it wants a clearer name, since "expanded" is so generic and would be tempting to use for what you suggest here.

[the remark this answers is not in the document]
dginev: Right, what is the existing `Expanded` parameter type meant to be I wonder? I half-remember it being related to `\edef`, but that is the parameter `DefExpanded` nowadays.

In the spirit of the `XToken` parameter, maybe a name such as `XPlain` would be a good mnemonic for what I'm after? But maybe such a name is hard to "explain".

lib/LaTeXML/Engine/LaTeX.pool.ltxml
dginev: Well the `Box()` auto-unlisting pattern is also a dangerous thing to do in Perl :>

`Box(scalar(FontDecode(...)),...` doesn't have quite the same ring to it, I admit.

lib/LaTeXML/Core/Definition/FontDef.pm
```diff
@@ -0,0 +1,74 @@
...
+  my ($self) = @_;
+  return $STATE->lookupValue($$self{fontID}); }
+
+sub invoke {
+  my ($self, $stomach) = @_;
+  my $current = $STATE->lookupValue('font');
+  if (my $fontinfo = $STATE->lookupValue($$self{fontID})) {
+    # Temporary hack for \the\font; remember the last font def executed
+    $STATE->assignValue(current_FontDef => $$self{cs},                                     'local');
+    $STATE->assignValue(font            => $STATE->lookupValue('font')->merge(%$fontinfo), 'local');
```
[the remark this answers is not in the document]
brucemiller: Done! Thanks!

lib/LaTeXML/Engine/LaTeX.pool.ltxml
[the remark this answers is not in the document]
brucemiller: Used in `\pdfstrcmp`, but I think `XGeneralText` is actually what should be used.  Need to do a bit more checking, and followup with another PR. Maybe take up your initial suggestion?

lib/LaTeXML/Engine/TeX_Debugging.pool.ltxml
brucemiller: thanks!

lib/LaTeXML/Engine/TeX_Fonts.pool.ltxml
brucemiller: yup; thanks!

PR #2418, opened 2024-09-20
Fix incorrect conversion of mathtools symbols like \coloneq and \Coloneq

### LINKED ISSUE #2410: Incorrect conversion of mathtools symbols like \coloneq and \Coloneq
Hello,

I have noticed that LaTeXML currently converts certain symbols from the `mathtools` package incorrectly. Specifically, the following conversions are observed:

- `\coloneq` is converted to `:-`
- `\eqcolon` is converted to `-:`
- `\Coloneq` is converted to `::-`
- `\Eqcolon` is converted to `-::`

However, in the actual `mathtools` package, these symbols are converted as follows:

- `\coloneq` should convert to `:=`
- `\colondash` should convert to `:-`
- `\eqcolon` should convert to `=:`
- `\dashcolon` should convert to `-:`
- `\Coloneq` should convert to `::=`
- `\Colondash` should convert to `::-`
- `\Eqcolon` should convert to `=::`
- `\Dashcolon` should convert to `-::`

It would be great if the LaTeXML implementation could be updated to match the `mathtools` package’s definitions.

Thank you for your attention to this issue.

fiveseven-lambda:
Closes #2410.

### DIFF
--- a/lib/LaTeXML/Package/mathtools.sty.ltxml
+++ b/lib/LaTeXML/Package/mathtools.sty.ltxml
@@ -722,19 +722,27 @@ DefMathI('\vcentcolon',    undef, ':', role => 'RELOP');
 DefMathI('\ordinarycolon', undef, ':', role => 'RELOP');
 
 DefMath('\dblcolon', "::", role => 'RELOP');
-# taken from txfonts.sty.ltxml
+
 DefMath('\coloneqq',    "\x{2254}",   role => 'RELOP');
 DefMath('\Coloneqq',    "\x{2A74}",   role => 'RELOP');
-DefMath('\coloneq',     ":-",         role => 'RELOP');
-DefMath('\Coloneq',     "::-",        role => 'RELOP');
+DefMath('\coloneq',     "\x{2254}",   role => 'RELOP');
+DefMath('\Coloneq',     "\x{2A74}",   role => 'RELOP');
 DefMath('\eqqcolon',    "\x{2255}",   role => 'RELOP');
 DefMath('\Eqqcolon',    "=::",        role => 'RELOP');
-DefMath('\eqcolon',     "-:",         role => 'RELOP');
-DefMath('\Eqcolon',     "-::",        role => 'RELOP');
+DefMath('\eqcolon',     "\x{2255}",   role => 'RELOP');
+DefMath('\Eqcolon',     "=::",        role => 'RELOP');
 DefMath('\colonapprox', ":\x{2248}",  role => 'RELOP');
 DefMath('\Colonapprox', "::\x{2248}", role => 'RELOP');
+DefMath('\approxcolon', "\x{2248}:",  role => 'RELOP');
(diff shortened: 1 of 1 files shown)


dginev:
Thank you for contributing this @fiveseven-lambda but it isn't yet finished. Could you (or should we) grep for the macros you are fixing over the entirety of latexml's Package/ directory?

For example, here is what Github finds for `\coloneq`:
https://github.com/search?q=repo%3Abrucemiller%2FLaTeXML%20coloneq&type=code

It would be really nice if we take this chance to sync up the Unicode output across all bindings.

fiveseven-lambda:
Regarding `\coloneq`, I have checked that it is also provided by `txfonts` and `mathabx`, and those implementations seem to be functioning correctly without issues. All of these use U+2254 for the symbol "≔", so it seems that the output is already consistent across these packages.
Could you provide specific instances where the same symbol is mapped to different Unicode characters? I would be happy to investigate and help ensure consistency if such cases exist.

dginev:
Indeed, subtle. I am mostly concerned the kind of oversight you are fixing may have happened elsewhere. It didn't help that in the original `.sty` packages _some_ of the macros have matching definitions while others (intentionally) don't.

I have quickly checked the macros you've introduced, and most don't appear in existing bindings, so this PR may be good to go.

dginev (approved):

fiveseven-lambda:
I see, thank you.
If I come across any other issues with the bindings in the future, I will make sure to report them.

brucemiller:
This is all rather curious. As far as I can tell, the bindings were correct *when* *implemented*. For whatever reason (mistake?), the packages `txfonts, pxfonts, mathtools` originally all defined `\coloneq` to give ":-", whereas `\coloneqq` gave ":=".  OTOH, `abxfonts, unicode-math` have `\coloneq` giving the more intuitive ":=".

Now it seems that `newtxmath` which (with `newtxtext`) replaces `txfonts` has `\coloneq` give ":=". We don't yet have a binding for these. (`txfonts` remains unchanged, giving ":-").

`mathtools` itself has been modified (*in place*) to redefine `\coloneq` as ":=". The online doc for the latter mentions it as a "breaking change Summer 2022", although that update hasn't yet propogated to my system, curiously.  That change also adds the new symbols you defined `\approxcolon`, etc. 

And to add yet another curiosity, while `mathtools` *also* updated `\Coloneq` to give "::=", `newtxmath` apparently did *not*, so that it still gives "::-" (like `txfonts` did).

It's awkward for packages to introduce breaking changes, but I suppose if we interpret that ":=" was the original intention, we probably should merge your changes, right?

brucemiller:
OK, cool, let's go with it. Thanks for the report and patch!!

fiveseven-lambda:
@brucemiller 

> although that update hasn't yet propogated to my system, curiously.

Is it sure that you are using TeX Live of this year? TeX Live requires a full new install every year.
See https://tex.stackexchange.com/questions/107017/why-does-tex-live-require-yearly-updates

PR #2420, opened 2024-09-22
Stray comments

brucemiller:
When characterizing certain blocks of XML, we need to be careful of unexpected comment nodes, since the exact order of appearance can be affected by other code changes. For example, this can affect marking nodes due to `\centering` or similar.  This PR fixes a couple of such cases, and updates some test cases which seem to have been missing `ltx_centering` (or similar).

### DIFF
--- a/lib/LaTeXML/Engine/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Engine/LaTeX.pool.ltxml
@@ -1383,8 +1383,11 @@ sub applyAligningContext {
   if (my $container = LookupValue('ALIGNING_NODE')) {
     my ($node, $previous) = @$container;
     my @children = $node->childNodes;
-    while (my $skip = shift(@children)) {
-      last if !$previous || $previous->isSameNode($skip); }
+    while (@children) {
+      my $skip = $children[0];
+      last if !$previous;
+      shift(@children);
+      last if $previous->isSameNode($skip); }
     while (my $child = shift(@children)) {
       setAlignOrClass($document, $child, $align, $class) if $child->nodeType == XML_ELEMENT_NODE; } }
   return; }
--- a/lib/LaTeXML/Engine/TeX_Box.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Box.pool.ltxml
@@ -368,8 +368,8 @@ sub insertBlock {
   my $inline    = $is_svg || $document->canContain($context_tag, '#PCDATA');
   my $container = $document->openElement('ltx:_CaptureBlock_', '_vertical_mode_' => 1, %blockattr);
   $document->absorb($contents);
-  my @nodes     = $container->childNodes;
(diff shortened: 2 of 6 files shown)


lib/LaTeXML/Engine/LaTeX.pool.ltxml
```diff
@@ -1383,8 +1383,11 @@ sub applyAligningContext {
   if (my $container = LookupValue('ALIGNING_NODE')) {
     my ($node, $previous) = @$container;
     my @children = $node->childNodes;
-    while (my $skip = shift(@children)) {
-      last if !$previous || $previous->isSameNode($skip); }
+    while (@children) {
+      my $skip = $children[0];
+      last if !$previous;
```
dginev, line 1388: since `$previous` is not changed during this `while` loop, maybe the check that it's defined can be lifted up? As in:

```perl
if ($previous) {
  while (my $skip = shift(@children)) {
    last if $previous->isSameNode($skip);
  }
}
```

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -368,8 +368,8 @@ sub insertBlock {
   my $inline    = $is_svg || $document->canContain($context_tag, '#PCDATA');
   my $container = $document->openElement('ltx:_CaptureBlock_', '_vertical_mode_' => 1, %blockattr);
   $document->absorb($contents);
-  my @nodes     = $container->childNodes;
-  my @node_tags = map { $document->getNodeQName($_); } @nodes;
+  my @nodes     = grep { $_->nodeType != XML_COMMENT_NODE; } $container->childNodes;
```
dginev, line 371: should we make a new method akin to [`element_nodes`](https://github.com/brucemiller/LaTeXML/blob/d3d3fcc8d84cbcd789dbdd626effb1c786bbf771/lib/LaTeXML/Common/XML.pm#L92) that gets "material" child nodes?
Actually that may be easier with an accept-list rather than a reject-list, as in

```perl
sub content_nodes {
  my ($node) = @_;
  return ($node ? grep { my $t = $_->nodeType; $t == XML_ELEMENT_NODE || $t == XML_TEXT_NODE } 
    $node->childNodes : ()); }
```

dginev (approved):
Interesting subtlety. I left a couple of comments, mostly wondering if we will need this kind of enhanced treatment of "content" children somewhere else.

But it also seems good to merge as-is. Thanks!

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
brucemiller: Good suggestion!

brucemiller:
Thanks for the good suggestions; adopted.

Yes, it's possible there are other cases; a superficial scan suggests it's done correctly in most cases, but it's an easy thing to overlook when developing new code and you're obsessed with finding the nodes you want.

PR #2421, opened 2024-09-23
Expansions

brucemiller:
This PR improves clarity of when expansion is full or partial, and improves naming of utility functions and Parameter types.   Reminder: partial expansion is what `\edef` does which defers expansion of `\protected` definitions, and the result of expanding `\the`.
Now the `Expand()` function fully expands, and there is a new `ExpandPartially()` function. Similarly, the `Expanded` and `ExpandedPartially` Parameter types expand fully or partially.   Finally, a few commands have been corrected to the right versions, such as some pdfTeX commands have been updated to use `XGeneralText` (a partially expanding version of `GeneralText`).


lib/LaTeXML/Core/Gullet.pm
```diff
@@ -333,18 +333,20 @@ sub unread {
...
 # For arguments to \if,\ifx, etc use $for_conditional true,
 # which handles \noexpand and CS which have been \let to tokens specially.
 sub readXToken {
-  my ($self, $toplevel, $for_conditional) = @_;
+  my ($self, $toplevel, $for_conditional, $fully_expand) = @_;
   $toplevel = 1 unless defined $toplevel;
-  my $autoclose      = $toplevel;    # Potentially, these should have distinct controls?
-  my $for_evaluation = $toplevel;
+  my $autoclose = $toplevel;    # Potentially, these should have distinct controls?
+                                #  my $fully_expand = $toplevel;
```
dginev, line 348: some leftover comments for removal

lib/LaTeXML/Engine/Base_ParameterTypes.pool.ltxml
```diff
@@ -148,11 +148,26 @@ DefParameterType('XUntil', sub {
...
-# This reads a braced tokens list, expanding as it goes,
-# but expanding \the-like commands only once.
+# Simulate reading a plain argument, and then fully expanding it.
+# Similar to when \csname is used
 DefParameterType('Expanded', sub {
     my ($gullet) = @_;
-    $gullet->readBalanced(1, 0, 1); },
+    my $token = $gullet->readToken;
+    if (!$token->equals(T_BEGIN)) {    # if non-brace, wrap the token in {}, to isolate it
+      $gullet->unread($token, T_END); }
```
dginev, line 157: Scary coding pattern. unread `T_END` then invoke readBalanced with `$require_open` set to "false".

Not to say it is wrong. Just wondering if it is scary enough to get its own named subroutine `readExpanded` in Gullet. Which can then take the `$expanded` arg as an input arg.

I think this PR uses the technique 4 times, so maybe there's some value in naming it...

lib/LaTeXML/Package.pm
```diff
@@ -899,17 +899,31 @@ sub generateID_nextid {
 #
 #======================================================================
 
-# Return $tokens with all tokens expanded
+# Return $tokens with all tokens fully expanded
+our $T_EXPAND_END_CHECK = T_OTHER('END EXPANSION');
```
dginev, line 903: Shouldn't that be a `T_MARKER` ?

dginev (approved):
Nice consistency upgrade and bug fixes!

I left some minor comments, but the PR is good to merge as-is (the new test is a big help).

lib/LaTeXML/Package.pm
brucemiller: Well, yeah, but `T_MARKER` never make it out of `Gullet`. OTOH, perhaps if there's a helper function from within `Gullet`, something safer could be arranged? Either way, it's all about error recovery.

lib/LaTeXML/Engine/Base_ParameterTypes.pool.ltxml
brucemiller: I was initially inclined to think that it *should* be (slightly) scary, rather than hidden away, since it's doing something rather peculiar. The use case is you're reading a "plain" argument, like `#1`, which would also accept a single token, and then asking it to be expanded. What if it's a macro expecting arguments?
Alternatively, it may want a helper function if there's a sensible (but likely convoluted) error recovery; At this point I don't know what that would be. (but see below)

lib/LaTeXML/Package.pm
dginev: Right, but if the `T_OTHER` makes it out of Gullet and encounters some processing in the Error realm, it may make it out into the final XML. Which we _may_ appreciate, but I suspect the average user won't...

That said, the last time I used a magic token I invented my own catcode (`CC_SMUGGLE_THE`), so I don't really have a clear head on the topic. Your call, it works as-is.

brucemiller:
Thanks for your comments; they caused me to rethink a couple of points, so please take another look and re-review.

dginev (approved):
The new approach is quite elegant, very nice idea! 

Looks good to merge in my eyes.

PR #2422, opened 2024-09-24
Triangles

brucemiller:
This PR finds a nice in-between size for left & right triangles (`\triangleleft,\triangleright`), U+22B2, U+22B3, that are smaller than U+25B3, U+25BD (`\bigtriangleup,\bigtriangledown`), as they should be, but still larger than `mathabx`'s `\smalltriangleup, \smalltriangledown, \smalltriangleleft, \smalltriangleright`) U+25B5,U+25Bf, U+25C3,U+25B9, as the seemingly also should be.

If acceptable, this is an alternative to #2362 

closes #2362


dginev (approved):
Sounds well-triangulated to me.

brucemiller:
Let's go with this for the time being; we can always adjust again later.

PR #2423, opened 2024-09-28
Kerns

brucemiller:
A small PR providing more consistent treatment of kerns and skips, and plausibly implements `\unkern,\unskip,\lastkern,\lastskip`.  Several test cases are updated due to which & how many spacing codepoints find their way into text.


t/alignment/algx.xml
```diff
@@ -31,7 +31,7 @@
       <listingline xml:id="algx1.l2"><tags>
           <tag><text fontsize="80%">2:</text></tag>
           <tag role="refnum">2</tag>
-        </tags>     <text font="bold">return</text> 7
+        </tags>  <text font="bold">return</text> 7
```
dginev, line 34: The leading space reduction in the algx test, and the algorithmic snippet in the [figure_mixed_content](https://github.com/brucemiller/LaTeXML/blob/2bef09e696a4b8d5b77f1ce38badce95c9165dca/t/complex/figure_mixed_content.tex#L264-L288) test rang a bell, where we've had two arXiv reports mentioning the leading whitespace in listings is getting wrongly removed. That's [arXiv/html_feedback 1712](https://github.com/arXiv/html_feedback/issues/1712) and [arXiv/html_feedback 2201](https://github.com/arXiv/html_feedback/issues/2201) for reference.

Maybe the underlying trickiness comes from even trying to model whitespace as actual characters inside an XML document (rather than some `<mspace>`-like element with a spacing value). But since we're doing characters, I wonder if it's worth taking a closer look at what exactly changed in those `{algorithmic}`-related tests, and if it is the same cause as the `{lslistings}` reports.

dginev: I now understand that the line I am commenting on had 6 ascii spaces (U+0020) converted to two alternative space chars: (U+2003)(U+2002)

I should probably take some time next to study how successful various fonts are in getting the "expected" widths displayed. GitHub's review interface seems to be unsuccessful?

brucemiller: The trigger for this PR was noticing that some kerns & skips were being dropped (they show up a *lot* in plain tex stuff); followed by improving the accuracy of converting them to spaces and filling in a few unimplemented commands.
That said, I too was surprised at being reminded that these spaces are (still) actually relied on for layout in listings. Even so, maybe that's enough to fix the cases you mention; I checked the 2nd one and the algorithms now come out indented nicely.
It would seem better in a *few* cases (not all kerns & skips!) to use an element w/width and CSS, rather than just clever spaces. The constructors for `\hskip,\kern` could do that, if it could detect the situation.  But I'm not sure how to define it; in this case possibly following an `ltx:tags` would be sufficient, but I haven't thought it through.

lib/LaTeXML/Engine/TeX_Glue.pool.ltxml
```diff
@@ -26,19 +26,22 @@ use LaTeXML::Package;
...
-    my $n = int(($ems + 0.3) / 0.333);           # 10pts per space...?
-    return (UTF(0xA0) x $n); } }
+  my $ems     = $dimen->ptValue / $fs;
+  my $s       = '';
+  for (my $i = $#spaces ; ($i >= 0) && ($ems > 0) ; $i--) {
+    my $w = $spaces[$i][0];
+    if ($ems + 0.01 > $w) {
+      my $n = int(($ems + 0.01) / $w);
+      $ems -= $n * $w; $s .= $spaces[$i][1] x $n; } }
+  return $s; }
```
dginev, line 44: Currently this loop would avoid adding any character if the requested dimension is narrower than `0.167em`, is that correct? Just wondering if there are legitimate cases where someone wants an even smaller amount of space. The hair space (U+200A) could fill in smaller gaps. I've seen one mention it is traditionally 1/16em or 0.0625.

dginev (approved):
The PR code looks healthy on review ✅ 

I indulged in some discussion, you're welcome to merge whenever.

lib/LaTeXML/Engine/TeX_Glue.pool.ltxml
brucemiller: Indeed, my revision makes it easy to add more spacing chars to the mix, but I didn't actively hunt down any extras. You have a list you'd like to include?

dginev: I do not have a list really. More of a learning-as-I-review process.

brucemiller:
I tweaked the set of space chars to be a little cleaner. There's lots of space (cough) for over-thinking here; we're really just approximating the space requested.

brucemiller:
Would be nice if you could check whether the 1st arXiv issue is also fixed; and record resolutions (however that's being done)

dginev:
> Would be nice if you could check whether the 1st arXiv issue is also fixed; and record resolutions (however that's being done)

They're not latexml issues, so nothing to record here. The arXiv issue will get closed when an updated LaTeXML reconverts the article *and* it goes live, which is easily a month away from today, likely longer. The CSS issue Vincenzo identified is likely to get resolved a little quicker, but likely still weeks away.

lib/LaTeXML/Engine/TeX_Glue.pool.ltxml
```diff
@@ -121,8 +136,13 @@ DefPrimitiveI('\vfilneg', undef, undef);
...
 #----------------------------------------------------------------------
 # \lastskip         iq is 0.0 pt or the last glue or muglue on the current list.
-
-DefRegister('\lastskip' => Glue(0), readonly => 1);
+DefRegister('\lastskip' => Dimension(0), readonly => 1, getter => sub {
+    my $box;
+    for (my $i = $#LaTeXML::LIST ; $i > 0 ; $i--) {
+      my $box = $LaTeXML::LIST[$i];
+      last if !$box || !$box->getProperty('isSkip');
+      return $box->getProperty('width'); }
```
dginev, line 144: Rust's clippy gave me a fun message while porting this loop:
```
error: this loop never actually loops
```

I didn't notice that while reviewing -- may be worth a second look? @brucemiller

dginev: Maybe the `last` should have been a `next` ?

PR #2424, opened 2024-10-08
Simplify \lastskip,\lastkern (suggested by D.Ginev)

brucemiller:
Simplify `\lastskip` (and also simlar `\lastkern`); as mentioned in last comment in #2423. The code was correct but hadn't been cleaned up after development.


brucemiller:
Actually, the loops were needed, just unfinished: They needed `next` clauses to skip over Comments!

dginev (approved):
Looks better, and good to merge ✅ 

I'm a tiny bit tempted to ask for a tricky test exercising `\lastskip` and `\lastkern`, but just voicing the thought for now.


One comment:

> Actually, the loops were needed, just unfinished

Let me note again that this was caught by the Rust compiler during porting. It noticed the previous code for the loops only executed one step and exited -- something I (an actual human) missed, while reviewing the other PR. I'm quite excited this happened.

brucemiller:
The commands are tested, just not the comments special cases (since comments are left out of tests due to variability).  Yeah, compilers are good; Thanks Rust! :>

Issue #2429, opened 2024-10-14
greek babel broken on texlive 2023


dginev:
Using the greek package option of babel now produces a Fatal error in texlive 2023 on my device. Minimal example:

```tex
\documentclass{article}
\usepackage[greek,english]{babel}

\begin{document}
\textgreek{dry run}
\end{document}
```

latexml log:
```
latexmlc (LaTeXML version 0.8.8; revision 2233d877)
processing started Mon Oct 14 17:38:09 2024
Error:undefined:\IfFormatAtLeastTF The token T_CS[\IfFormatAtLeastTF] is not defined. at greek.ldf; line 601 col 20
Info:latex:(babel)  Package babel Info: Importing at babel-english.tex; line 11 col 0
Info:latex:(babel)  Package babel Info: Importing at babel-greek.tex; line 11 col 0
Fatal:misdefined:\bbl@line Expansion of '\bbl@line' has unbalanced {}
	at babel-greek.tex; line 11 col 0 - line 11 col 1
	Expansion is sigma.final.3.0 = { ()"()[σΣ] } }sigma.final.3.1 = { remove }[transforms.posthyphenation]diaeresis.hyphen.1.0 = { |([ϊΐϋΰ]) }diaeresis.hyphen.1.1 = { remove }diaeresis.hyphen.1.2 = { no = {1}, pre = -, post = {1|ϊΐϋΰ|ιίυύ} }
	In Core::Definition::Primitive[\read Num... /home/deyan/perl5/lib/perl5/LaTeXML/Engine/TeX_FileIO.pool.ltxml; line 76
	 <= Core::Stomach[@0x644d3b9e7748] <= Core::Definition::Primitive[Begin] <= Core::Stomach[@0x644d3b9e7748] <= Core::Definition::Primitive[Begin] <= Core::Stomach[@0x644d3b9e7748] <= Core::Gullet[@0x644d3b9e0138] <= Core::Stomach[@0x644d3b9e7748] <= Core::Definition::Constructor[\usepac... <= Core::Stomach[@0x644d3b9e7748]
Conversion complete: 1 error; 1 fatal error; 1 undefined macro[\IfFormatAtLeastTF] (See /home/deyan/git/my-LaTeXML/test.latexml.log)

```

```bash
$ tex --version
TeX 3.141592653 (TeX Live 2023/Debian)
kpathsea version 6.3.5
```

CC @brucemiller some fresh food for thought for our recent conversation touching on texlive regressions.

[...]

PR #2430, opened 2024-10-14
trim whitespace in loading LaTeX classes and packages

dginev:
Minimal example:
```tex
\documentclass[
	aps, pra]{revtex4-2}
\begin{document}
Hello world.
\end{document}
```

This leading `\par` in the options of `\documentclass` was the first encountered problem in a [recent arXiv report](https://github.com/arXiv/html_feedback/issues/2294), which turned out to be also related to a greek babel regression, reported in #2429 .

I took the opportunity to sanitize (by trimming whitespace) the options argument to `\documentclass, \documentstyle, \usepackage, \RequirePackage, \LoadClass, \PassOptionsToPackage, \PassOptionsToClass, \ExecuteOptions`.

This may be removed soon after in a world where we load all of latex.ltx raw, but I realized that only after completing the code changes. So filing the PR in any case, for your consideration.

### DIFF
--- a/lib/LaTeXML/Engine/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Engine/LaTeX.pool.ltxml
@@ -55,11 +55,16 @@ DefConstructor('\documentclass OptionalSemiverbatim SkipSpaces Semiverbatim []',
   beforeDigest => sub { onlyPreamble('\documentclass'); },
   afterDigest  => sub {
     my ($stomach, $whatsit) = @_;
-    my $options = $whatsit->getArg(1);
+    my $options = ToString($whatsit->getArg(1));
     my $class   = ToString($whatsit->getArg(2));
+    my @options;
+    if ($options) {
+      $options =~ s/^\s+//;
+      $options =~ s/\s+$//;
+      @options = split(/\s*,\s*/, $options); }
     $class =~ s/\s+//g;
     LoadClass($class,
-      options => [($options ? split(/\s*,\s*/, ToString($options)) : ())],
+      options => [@options],
       after   => Tokens(T_CS('\AtBeginDocument'), T_CS('\warn@unusedclassoptions')));
     return; });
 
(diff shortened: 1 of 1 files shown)


brucemiller:
This kinda screams for either a Trimmed type of argument, or at least that the option processing routines would trim their arguments.

dginev:
Done, thanks. I tried a couple of approaches, then realized the argument handling happens at different stages - some options are Tokens, others are already Boxes (typically a List). So doing the sanitization after the string phase is easier in the short run.

Parameter types could improve too, but `OptionalTrimmedSemiverbatim` seemed to be a mouthful.

dginev:
I also found 5 places in 5 bindings which can benefit from sharing this new helper function. I haven't committed them here yet, let me know if you'd like them added (in authblk, cleveref, elsarticle, enumitem and ntheorem).

brucemiller:
Hmm (again). So actually all(?) of these are cases of keyval or option lists, but at a somewhat special top-level, which may not be quite the same as the usual keyval processing.  And a 3rd hmm is that these are commands that call for special treatment anyway, so are likely to always need  LaTeXML bindings even if we read latex.ltx.  I wonder if some use or adaptation of `OptionalKeyVals` would be the best long-term approach?

dginev:
The parameter type name that resonates with me is `PackageOptions`, since the comma list convention is a top-level latex convention for classes and packages.

With that name we could even avoid the leading `Optional`, since they are always optional.

I am also wondering if we should be comparing to `DirectoryList` here, rather than `KeyVals`.

brucemiller:
Options aren't always optional, conceivable. Think of something like the various `\setkeys` commands that are out there. And I believe some packages accept keyval options, and probably more will in the future; possibly even `\documentclass` ? So, I'd think keyval is still the right basis, to the extent we can keep things compatible across all the use cases.

Of course, there are subtle differences between the implementation packages and appropriateness for `\usepackage`. See
https://tex.stackexchange.com/questions/26771/a-big-list-of-every-keyval-package

dginev:
Ok, but aren't these key-val related considerations out of scope for this PR? Currently we treat "package options" as comma-separated lists, which is an expectation that goes all the way down into [InputDefinitions](https://github.com/brucemiller/LaTeXML/blob/76e50b767039ce5c9f71051f28bb335e08c021c0/lib/LaTeXML/Package.pm#L2542).

I fear switching to a KeyVals object will have to propagate some large refactor all the way into Package.pm, while this PR just wanted to land a bug fix for whitespace handling. 

Creating a new parameter type is fine by me, which can be refactored later as needed. But I don't think it is realistic to do much more than that before v0.8.9 is out.

dginev:
How about we merge here without extending scope? Or close?

brucemiller (approved):
This certainly looks reasonable; it may be we'll want to do more or differently later on, but it seems fine for now.

PR #2433, opened 2024-10-24
scoped \href arguments to avoid leaking color

dginev:
I stumbled onto a simple color leakage [arxiv report](https://github.com/arXiv/html_feedback/issues/2409) yesterday and found a quick fix.

But I am also a little unsatisfied by how specific the fix is - it only contains one macro, rather than all possible cases where this containment may be needed. But I am not sure which cases those are.

So for now I'm offering the quick fix for consideration. And adding a test.

Minimal example reproducing the issue:
```tex
\documentclass{article}
\usepackage{hyperref}
\usepackage{xcolor}
\begin{document}
A link: \href{https://example.com}{\color{blue}here}.

\section{Color should not leak into this heading}
This text should be black.

\end{document}
```


brucemiller:
Doesn't the `DefConstructor(... bounded=>1)` do what you need?  And wouldn't it be convenient to just add the test to `hypertest`?

dginev:
Good point, done.

I am still wondering if refactoring the `sub { $_[0]->bgroup; }` to the sub reference `\&LaTeXML::Core::Stomach::bgroup` wouldn't add a nice speedup, but at least in `make test` it isn't that visible, so I'll leave it out for now.

PR #2436, opened 2024-11-14
Rearrange2

brucemiller:
This PR rearranges a bit of code to more appropriate modules; more consistent utilities and naming.  Improves the tracing and debugging commands.

### DIFF
--- a/lib/LaTeXML/Package/article.cls.ltxml
+++ b/lib/LaTeXML/Package/article.cls.ltxml
@@ -101,5 +101,17 @@ DefMacro('\appendix', '\@appendix');
 # Actually we should be using section counter
 DefPrimitive('\@appendix', sub { startAppendices('section'); });
 
+#======================================================================
+DefPrimitiveI('\tiny',         undef, undef, font => { size => 5 });
+DefPrimitiveI('\scriptsize',   undef, undef, font => { size => 7 });
+DefPrimitiveI('\footnotesize', undef, undef, font => { size => 8 });
+DefPrimitiveI('\small',        undef, undef, font => { size => 9 });
+DefPrimitiveI('\normalsize',   undef, undef, font => { size => 10 });
+DefPrimitiveI('\large',        undef, undef, font => { size => 12 });
+DefPrimitiveI('\Large',        undef, undef, font => { size => 14.4 });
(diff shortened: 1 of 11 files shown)


dginev (changes requested):
It looks great overall, but now that you have raised awareness to `\tiny` ... `\Huge` having moved out of TeX.pool.ltxml in #2358, I worry they will quietly disappear in cases they used to work in before. Ideally, we would like to make them widely available to all `.cls` cases in latexml that used to work in the past.

So at the least we should add them to `OmniBus.cls.ltxml`? 

But I also wonder if they wouldn't be a reasonable fit in the font-related block of `LaTeX.pool.ltxml` itself for now, until the next evolution of `.cls` support.

lib/LaTeXML/Package/article.cls.ltxml
```diff
@@ -101,5 +101,17 @@ DefMacro('\appendix', '\@appendix');
...
+DefPrimitiveI('\scriptsize',   undef, undef, font => { size => 7 });
+DefPrimitiveI('\footnotesize', undef, undef, font => { size => 8 });
+DefPrimitiveI('\small',        undef, undef, font => { size => 9 });
+DefPrimitiveI('\normalsize',   undef, undef, font => { size => 10 });
+DefPrimitiveI('\large',        undef, undef, font => { size => 12 });
+DefPrimitiveI('\Large',        undef, undef, font => { size => 14.4 });
+DefPrimitiveI('\LARGE',        undef, undef, font => { size => 17.28 });
+DefPrimitiveI('\huge',         undef, undef, font => { size => 20.74 });
+DefPrimitiveI('\Huge',         undef, undef, font => { size => 29.8 });
+
```
xworld21, line 115: I should flag that I plan to open a related issue: `\tiny`, ..., `\Huge` are responsible for changing `\baselineskip`, and the fact that LaTeXML does not do that has very visible side effects with TikZ.

brucemiller:
Good catch! `slides` is the only one not based on `LoadClass`.  I kinda hated the duplication, but theoretically classes define these differently; and having them defined when the "shouldn't" be causes confusion to the (eventual) `latex.ltx` loader. 
@xworld21 I'm looking forward to further tikz refinements; there's going to be a big push on all that stuff in the future.

PR #2440, opened 2024-11-24
Move rule

brucemiller:
Update the schema to move `ltx:rule` from being only allowed in Inline contexts, but also in Block and Para contexts. This avoids unnecessary `ltx:p,ltx:para`.  Make full-width `\hrule` close any current `ltx:p`.  And update some tests.


dginev (approved):
Looks good.

It is a potential breaking change, so would also need a mention in the release Changelog.

Issue #2392, opened 2024-08-13
figure distortion


CaseyMiller:
Figures become distorted. The example shows spheres in pane a becoming distorted to ellipsoids. Figure files attached and the source is

\documentclass[11pt]{article}

% Graphics support
\usepackage{graphicx}
\usepackage{subcaption}

\begin{document}

\begin{figure*}[t]
   \centering
    \begin{subfigure}[b]{0.33\linewidth}
        \includegraphics[width=\linewidth]{BCC.png}
        \caption{}
    \end{subfigure}
    \begin{subfigure}[b]{0.35\linewidth}
        \includegraphics[width=\linewidth]{BCE.png}
        \caption{}
    \end{subfigure}
    \begin{subfigure}[b]{0.3\linewidth}
        \includegraphics[width=\linewidth]{PSF_25S.png}
        \caption{}
    \end{subfigure}
    \\
    \begin{subfigure}[b]{0.4\linewidth}
        \includegraphics[width=\linewidth]{PEF_25E.png}
        \caption{}
    \end{subfigure}
    \begin{subfigure}[b]{0.4\linewidth}
        \includegraphics[width=\linewidth]{PSF_1000S.png}
        \caption{}
    \end{subfigure}
    \caption{Figure example.} 
    \label{fig:ex2}
\end{figure*}

\end{document}
![BCC](https://github.com/user-attachments/assets/f8914b59-b8de-4439-9771-b113e5d1fd56)
![BCE](https://github.com/user-attachments/assets/ec0c02bd-d01b-47bb-8bd4-ab23b588bf07)
![PEF_25E](https://github.com/user-attachments/assets/4814b86b-cc70-4834-a95c-caf2b5238b5f)
![PSF_25S](https://github.com/user-attachments/assets/d9022869-5e16-4e4b-a746-5a83af63966b)
![PSF_1000S](https://github.com/user-attachments/assets/189220c6-9ef5-40ed-8565-afc4bb0964d5)

CaseyMiller:
[figure.pdf](https://github.com/user-attachments/files/18149696/figure.pdf)
[figure.zip](https://github.com/user-attachments/files/18149707/figure.zip)

Included example, pdf produced, and html.

dginev:
The attached HTML keeps the row of 3, row of 2 arrangement from the PDF, and marks each of the 5 panels with `ltx_img_square`, which I think is close to the original image aspect ratios.

[...]

PR #2447, opened 2024-12-17
typo that disabled magnify option of latexml.sty

dginev:
Encountered during the mega discussion in #2392 .

Sad, unfortunate, etc.


brucemiller:
Oh, my! Well, good catch...

PR #2448, opened 2024-12-17
Leaders

brucemiller:
A slightly hacky implementation of `\leaders`, at least with the correct parameters, and some refinement to stretchy accents (that TeX deals with internally with leaders).


brucemiller:
You'd think, but from the TeX perspective, stretchy is more of the exception. Sometimes requiring quite some trickery to achieve.

PR #2449, opened 2024-12-18
Amscd

brucemiller:
This PR fixes the `amscd` package to make the horizontal connectors (typically arrows) stretch.   Since MathML (currently?) only stretches horizontally the participants in `mover/munder` (see https://github.com/w3c/mathml-core/issues/270) some extra support (hack) is needed in the MathML postprocessing.  But the end result now looks like it should.


lib/LaTeXML/Package/amscd.sty.ltxml
```diff
@@ -101,22 +97,34 @@ DefConstructor('\cd@stack Undigested {} ScriptStyle ScriptStyle', sub {
...
+      role => 'ARROW', stretchy => 'true', meaning => 'rightarrow', width => Dimension('30pt')); });
+DefPrimitive('\lx@amscd@rightarrow', sub {
+    Box("\x{2192}", undef, undef, '\rightarrowf',
+      role => 'ARROW', stretchy => 'true', meaning => 'rightarrow', width => Dimension('30pt')); });
+DefPrimitive('\lx@amscd@leftrightarrow', sub {
+    Box("\x{2194}", undef, undef, '\leftrightarrow',
+      role => 'ARROW', stretchy => 'true', meaning => 'leftrightarrow', width => Dimension('30pt')); });
+DefPrimitive('\lx@amscd@equals', sub {
+    Box("=", undef, undef, T_OTHER('='),
+      role => 'ARROW', stretchy => 'true', meaning => 'equals', width => Dimension('30pt')); });
```
dginev, line 127: As a clarifying question:

What is the intended meaning of an `XMTok` that has a set positive width (30pt here), with `stretchy="true"` at the same time? 

Does the XMath `width` act as `min-width` in this circumstance?

lib/LaTeXML/Post/MathML.pm
```diff
@@ -535,7 +535,14 @@ sub pmml_maybe_resize {
...
   my $xoff   = $node->getAttribute('xoffset') || ($parent && $parent->getAttribute('xoffset'));
   my $yoff   = $node->getAttribute('yoffset') || ($parent && $parent->getAttribute('yoffset'));
-  if ($width || $height || $depth || $xoff || $yoff) {
+  my $role   = $node->getAttribute('role')    || ($parent && $parent->getAttribute('role'));
+  my $unstretchy = (($node->getAttribute('stretchy') || ($parent && $parent->getAttribute('stretchy'))) || 'yes') eq 'no';
+
+  # First, special case hack for stretchy arrows, with specified width
+  # Stretchiness (currently) only has effect within munder/mover!!!!!
+  if ($width && (($role || '') eq 'ARROW') && !$unstretchy) {    # SPECIAL CASE HACK
+    $result = ['m:mover', {}, $result, ['m:mspace', { width => $width }]]; }
```
dginev, line 544: I wonder if a well-named `class` attribute for this special case won't be useful in the future. Would make it a little clearer for people inspecting the markup, and easy to check when it gets refactored as MathML Core improves.

Maybe `<mover class="ltx_math_stretched">` ?

dginev: The `$unstretchy` check may not be booleany enough. If we go with strecthy="true" by default, and when explicit, then everything else is false (including invalid values), then:

```perl
  my $stretchy_attr = $node->getAttribute('stretchy') || ($parent && $parent->getAttribute('stretchy')) || "true";
  my $unstretchy = $strechy_attr neq "true";
```

brucemiller: Somehow I knew you'd suggest that, so why didn't I just head you off by implementing it?  But actually, something like `\ltx_horizontally_stretched` might be even better; I was a little concerned about the code doing something weird with vertically stretching cases. (Alas, no visible distinction in MathML)

t/ams/cd.xml
```diff
@@ -205,15 +221,15 @@
                 </XMCell>
                 <XMCell/>
                 <XMCell align="center">
-                  <XMTok fontsize="160%" name="uparrow" role="ARROW">↑</XMTok>
+                  <XMTok fontsize="160%" name="||" role="VERTBAR">∥</XMTok>
                 </XMCell>
                 <XMCell/>
                 <XMCell align="center">
-                  <XMTok fontsize="160%" name="uparrow" role="ARROW">↑</XMTok>
+                  <XMTok fontsize="160%" name="||" role="VERTBAR">∥</XMTok>
```
dginev, line 228: The `name="||"` got me curious, so I checked the DefMath, which indeed uses that name for `\|`. 

```perl
DefMathI('\|',        undef, "\x{2225}", role => 'VERTBAR',  name    => '||');
```

In the TeXbook, I see its definition marked with "vertical line" in Appendix E, p. 420:
```
\def\|{\leavevmode\hbox{\tt\char‘\|}} % vertical line
```

Would `vertical-lines` be a better name in this case?

lib/LaTeXML/Post/MathML.pm
brucemiller: Actually, if a class is going to be used to detect the "special case", then we can drop the test of `stretchy` completely.

t/ams/cd.xml
brucemiller: It *might* be, but `name` is a bit of an orphan: it's somewhere between the content, reversion and meaning; it probably shouldn't even exist at all, but it does get used for various checks.
At some point in the not too distant future, a scan of all uses needs to be analyzed and figure out where to go with it. Perhaps it should end up being part of intent infrastructure?  So: Hold onto that thought!

dginev (approved):

brucemiller:
OK, I added `class=ltx_horizontally_stretchy` to the arrows, and use that to detect the special case in postprocessing. So take another look.

dginev (approved):
Even better, LGTM.

PR #2450, opened 2024-12-18
Overline fix

brucemiller:
Fix an oversight in #2448, namely that `\overbrace,\underbrace` include `\limits` (handled by setting `scriptpos=mid`), which puts any following scripts above or below. However, lines and arrows should *not* do this!


dginev (approved):
Sounds good.

Our testing framework can't really track these changes all the way to the final rendering which is a pity. One day...

PR #2454, opened 2024-12-22
Fixes #2453

### LINKED ISSUE #2453: font encoding should be expanded in \DeclareSymbolFont
At this line
https://github.com/brucemiller/LaTeXML/blob/13018504e00ba934bc067357c372b72951227494/lib/LaTeXML/Engine/LaTeX.pool.ltxml#L2830-L2836
the `$enc` parameter should be expanded (it could be e.g. `\encodingdefault`). The LaTeX sources have a suggestive `\edef\reserved@b{#2}` which I think is doing the expansion.

Minimal example where it causes issues:
```latex
\documentclass{article}
\usepackage{fontenc}
\DeclareSymbolFont{numbers}{\encodingdefault}{\sfdefault}{\mddefault}{n}
\DeclareMathSymbol{0}\mathalpha{numbers}{"30}
\begin{document}
  Test.
\end{document}
```
The call to `\DeclareMathSymbol` will not expand `\encodingdefault` and generate an error ('Info', for some reason). PDFLaTeX will work fine.

dginev:
I trusted the diagnostic in #2453 and changed the respective parameter type to one emulating `\edef` expansion.

Quick and simple PR.

### DIFF
--- a/lib/LaTeXML/Engine/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Engine/LaTeX.pool.ltxml
@@ -2827,7 +2827,7 @@ DefPrimitive('\DeclarePreloadSizes{}{}{}{}{}', undef);
 # The next font declaration commands are based on
 # http://tex.loria.fr/general/new/fntguide.html
 # we ignore font encoding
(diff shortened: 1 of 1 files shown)


brucemiller:
Seems quite plausible. Thanks!

Issue #1541, opened 2021-06-01
SVG output: use current color?


flyn-org:
I use LaTeXML to generate HTML5 that is eventually integrated into a site that supports "dark" and "light" modes. This process includes a number of tikz figures that LaTeXML translates into SVG. The SVG includes color definitions such as #000000, but this causes a problem when the site is placed into dark mode, as the black lines are hidden by the black background.

Would it be possible to use "[currentColor](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/color)" as the default color for SVG figures? This allows the figure to adopt the "foreground" color, which the dark and light modes change. I found that a simple replacement of "#000000" with "currentColor" in the generated HTML5 worked nicely.

PR #2456, opened 2024-12-23
only adjust black color during box operations

### LINKED ISSUE #2455: color from surrounding scope overrides colors inside boxes
This was quite difficult to pin down. Emitting a box apparently inherits the surrounding colour, overriding whatever happens in the box.
```latex
\documentclass{article}
\newbox\mybox
\usepackage{xcolor}
\begin{document}
  \color{black}

  \global\setbox\mybox=\vbox{The final word should be \textcolor{green}{green}}
  \box\mybox
\end{document}
```
You get
```xml
  <para vattach="bottom" xml:id="p1">
    <p><text color="#000000">The final word should be green</text></p>
  </para>
</document>
```
If I remove `\color{black}`, I get the expected
```xml
  <para vattach="bottom" xml:id="p1">
    <p>The final word should be <text color="#00FF00">green</text></p>
  </para>
</document>
```
(Context: I am experimenting with reading beamer raw. Against all odds, it actually works, except that all colours are missing due to this bug!)

dginev:
Fixes #2455 .

The issue is really excellent in scope and triage - I added the test case here as well.

That said, the resolution is just a little tricky, because I am not sure if there is a clear design for testing "a color having been set". I tried a couple of generic solutions, but kept losing test coverage. Then I saw we are using a `Black->equals` test to assert the default color scheme. As it turns out, using that test keeps all tests passing and also covers the reported example.

But I can't at all claim that I am certain this is the best approach.


brucemiller:
Hmm, you're right: being black doesn't mean that it wasn't explicitly set to black. I want to look a bit more...

xworld21:
> Hmm, you're right: being black doesn't mean that it wasn't explicitly set to black. I want to look a bit more...

There are a few packages that use `Black` (as opposed to `['rgb', 0, 0, 0]` which *is* different). Grepping for `Black` brings up:
- diagbox: `Black` is used as default linecolor
- color/xcolor: use `Black` as default or if parsing/lookup fails
- framed/makecell/ntheorem: use `Black` as framecolor
- LaTeX.pool: uses `Black` as default color and as fill/stroke for `\circle`;
- xy: uses `Black` as default color
- wasysym: uses `Black` as background
- Package.pm: uses `Black` if color lookup fail

So it seems safe to interpret `Black` as 'currentColor' (in SVG terminology); if TeX code explicitly sets the color to black, it will actually use `['rgb', 0, 0, 0]`. With that in mind, it should really be called `CurrentColor` instead of `Black` (which would lead to a fix for #1541, by the way).

If anything, `adjustBoxColor` may have to adjust all other occurrences of `Black` too (linecolor, framecolor, background, fill, stroke) depending on how the normal TeX engines behave.

PR #2457, opened 2024-12-24
handle empty case in Tokens::stripBraces

### LINKED ISSUE #2452: empty group in keyval should return empty value
Example:
```latex
\documentclass{article}

\usepackage{keyval}
\makeatletter
\define@key{createvalue}{empty}{\def\myempty{#1}}
\setkeys{createvalue}{empty={}}

\begin{document}
  \ifx\myempty\@empty Empty.\else Not empty.\fi
\end{document}
```
LaTeX returns 'Empty', LaTeXML gives 'Not empty'.

dginev:
Fixes #2452

The classic loops in Tokens::stripBraces were missing an edge case where the braces held nothing - and an empty Tokens would be the appropriate response.

Interestingly, the bug reported in #2452 was absent from the Rust port, as I had used Rust's Iterator abstraction (with [peekable](https://doc.rust-lang.org/std/iter/struct.Peekable.html)), which safely avoids the need of index arithmetic.


dginev:
I also added the report from the issue as a new test here, to spot any regressions (e.g. during porting)

t/keyval/keyvalemptyvalue.tex
```diff
@@ -0,0 +1,10 @@
+\documentclass{article}
+
+\usepackage{keyval}
+\makeatletter
+\define@key{createvalue}{empty}{\def\myempty{#1}}
+\setkeys{createvalue}{empty={}}
```
xworld21, line 6: I suggest adding `{{}}` (and maybe `{{{}}}` too), to be sure that the *correct* number of braces is stripped out. When I reported the issue, I fixed it locally by stripping all braces, but that was wrong too.

dginev: Great point, the check only comes back empty for `{}` and `{{}}` in keyval.

The stripBraces fix is correct for the intended semantics of "stripBraces", but it looks like keyval should only strip them twice at most. Maybe I can add a "max_layers" parameter here...

dginev: Alright, a little scary but I updated the test to request exactly 2 braced strips for this case, extending `stripBraces` to allow for a number of layers.

brucemiller:
This is scary stuff since, in real TeX/LaTeX, the brace stripping is almost always an artifact of exactly what sequence of helper macros with and without passed arguments being RE-braced (as `{#1}`). So, you generally just feel Lucky when these corner cases work out.
That said, what you've done seems straightforward and correct; and if some error shows up, it's probably more about errors cancelling, than about introducing any. So...

PR #2459, opened 2024-12-27
\unvbox and friends request a block unwrap

### LINKED ISSUE #2458: \unvbox does not unbox the box
Couldn't let it sit until after the bank holidays for obvious reasons.

The implementation of the `\un*box` commands is wrong at https://github.com/brucemiller/LaTeXML/blob/13018504e00ba934bc067357c372b72951227494/lib/LaTeXML/Engine/TeX_Box.pool.ltxml#L586-L591
They should replace the boxes with their contexts, but `$stuff->unlist` is typically a one element list consisting of the `<ltx:block>` wrapper. The difference is visible here:
```latex
\documentclass{article}
\newbox\mybox
\begin{document}
  Vertical box below.

  \setbox\mybox=\vbox spread 3cm{This box is 3cm high}\box\mybox

  Another vertical box below, or is it?

  \setbox\mybox=\vbox spread 3cm{This box should be 3cm high.}\unvbox\mybox

  Was it 3cm?
\end{document}
```
The XML output may look the same (LaTeXML won't save the height), but the `@vattach` attribute is the visible symptom that the box has not been unwrapped.

A hasty workaround is to call `$whatsit->getArg(2)` instead of `->unlist`, and to stick some kind of newline (`\nointerlineskip`?). A proper solution should at least involve some error checking (TeX throws an error when the box types are mismatched, e.g. `\mybox=\hbox{}\unvbox\mybox`) and some thinking about the type of newline required. I haven't checked the TeX book for the details.

(Raw beamer uses a lot of `\unvbox` and it creates nested blocks like there's no tomorrow. `->getArg(2)` flattens the output nicely which suggests I am generally right.)

dginev:
Fixes #2458 .


xworld21 (commented):
Three quick comments related to the originating issue:
- could there be any data in the wrapper box that gets lost in this way? in TeX, a box is merely a box, but is that true for LaTeXML?
- `\unvbox` and `\unvcopy` must also switch to vertical mode, or it will cause other unintended side effects (per TeX book 'The vertical list inside that box is appended to the current vertical list, without changing it in any way')
- the `\un.(copy|box)` commands should error out if the box is of the wrong horizontal/vertical type

dginev:
@xworld21 I am not sure your questions apply to the level of emulation that is in latexml at present. Vertical mode has not been implemented in a "strict" sense, as matching the exact layout algorithm from TeX was never a main latexml goal. This could be contrasted to converters such as tex4ht and rustex, where the emulation follows TeX much more closely.

All of this could evolve of course, but the PR here tries to minimally match the `\vbox` and `\hbox` DefConstructors, where the outer whatsit will indeed be lost/unwrapped, while the "content" argument will remain in its place.

xworld21:
> I am not sure your questions apply to the level of emulation that is in latexml at present

I meant to say it needs to introduce a newline or start a new paragraph. I applied your patch (or rather, a worse version I cooked up before your PR) and you get e.g. consecutive bits of text with no space between them. The `\unvbox`, `\unvcopy` constructors must also introduce a newline or close the current paragraph. I figure injecting a `\par` is close enough.

I'll produce some examples to show the difference.

dginev:
I came up with this curious test, and indeed, the PR should add a vertical annotation of some kind to emulate it correctly:

```tex
\documentclass{article}
\newbox\boxA
\newbox\boxB
\newbox\boxC
\newbox\boxD
\begin{document}
  
\setbox\boxA=\vbox spread 1cm{1cm Box}
\setbox\boxB=\vbox spread 2cm{2cm Box}
\setbox\boxC=\vbox spread 3cm{3cm Box}
\setbox\boxD=\vbox{\unvbox\boxA\unvbox\boxB\unvbox\boxC}

combined vbox: \fbox{\box\boxD}
\end{document}
```

And actually, the master branch of latexml appears to create rather decent XML for it, as in:
```xml
    <p>combined vbox: <inline-block framed="rectangle" vattach="bottom">
        <p vattach="bottom">1cm Box</p>
        <p vattach="bottom">2cm Box</p>
        <p vattach="bottom">3cm Box</p>
      </inline-block></p>
```

So I am also wondering if the lack of an unwrap isn't already handled reasonably enough...

dginev:
New approach ready for review.

Studying a couple of examples more closely, I think the main issue is with having too much verbosity in the XML wrappers, especially as the `\un*` primitives explicitly are unwrapping requests.

Luckily, the `insertBlock` utility already has code that smartly unwraps unneeded containers, but it requires that no new properties are _required_ on the container.

So, I added a special property `_unwrap_block` for the `\un*` primitives, which `\vbox` and `\vtop` consult to call `insertBlock` without any props. Which in turn leads to no additional container nodes, and no verbosity from `vattach="bottom"` attributes.

dginev:
Closing in favour of doing TeX right :>

PR #2460, opened 2024-12-28
Allocate

brucemiller:
This PR makes register allocation (eg `\newcount` and friends) a bit safer against load order or accidental redefinitions. A new utility `\lx@alloc@` useful for plain's `\alloc@` and LaTeX's `\e@alloc` avoids allocating a `\count` (etc) slot if it has already been assigned to a register. `NewCounter(...)` avoids redefining counters.

### DIFF
--- a/lib/LaTeXML/Engine/TeX_Registers.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Registers.pool.ltxml
@@ -70,6 +70,29 @@ DefPrimitive('\muskipdef SkipSpaces Token SkipSpaces SkipMatch:=', sub {
 DefPrimitive('\toksdef SkipSpaces Token SkipSpaces SkipMatch:=', sub {
     shorthandDef($_[0], $_[1], '\toks', Tokens()); });
 
+# Candidate for use defining plain's \alloc@ and latex's \e@alloc
+our %stored_registers = (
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Engine/TeX_Registers.pool.ltxml
```diff
@@ -70,6 +70,29 @@ DefPrimitive('\muskipdef SkipSpaces Token SkipSpaces SkipMatch:=', sub {
...
+    $type    = ToString($type);
+    $tracker = ToString($tracker);
+    my $next;
+    if ($stored_registers{ ToString($allocator) }) {
+      $next = LaTeXML::Package::allocateRegister($type);
+      $next =~ s/^\Q$type\E//; }
+    else {
+      my $xnext = $STATE->lookupValue($tracker) || Number(0);
+      $next = $xnext->valueOf + 1; }
+    my $value = Number($next);
```
dginev, line 87: `$value` is never used?

dginev (approved):
Looks healthy, as far as I can tell :> Thanks!

lib/LaTeXML/Engine/TeX_Registers.pool.ltxml
brucemiller: Indeed! removed; thanks!

PR #2461, opened 2024-12-28
Everymath

brucemiller:
Use `\everymath, \everydisplay` correctly. The latter is used (alone) for displayed math, the former for inline math. Add some tests.


dginev:
Well, it is certainly _more_ correct than before. But there are likely further hidden inaccuracies for more custom math environments.

As an example test of a subtlety, the left-arrow here surprised me in the PDF (and neither arrow shows up via this PR):

```tex
{
 \everymath={\leftarrow}
 \everydisplay={\rightarrow}
 \begin{equation}
  a+b
 \end{equation}
}
```

dginev (approved):
Good to merge, just flagging that there are more icebergs out there...

brucemiller:
Ah, yes. Painfully obvious, now that you point it out.  Thanks!
So, here's a slightly scarier, but more complete version.

lib/LaTeXML/Engine/LaTeX.pool.ltxml
```diff
@@ -2045,15 +2045,9 @@ DefConstructorI('\lx@begin@display@math', undef,
...
-  beforeDigest => sub {
-    $_[0]->beginMode('display_math');
-    if (my @everymath_toks = $STATE->lookupDefinition(T_CS('\everymath'))->valueOf->unlist()) {
-      $_[0]->getGullet->unread(@everymath_toks); }
-    if (my @everydisplay_toks = $STATE->lookupDefinition(T_CS('\everydisplay'))->valueOf->unlist()) {
-      $_[0]->getGullet->unread(@everydisplay_toks); }
-    return; },
-  properties  => sub { RefStepID('equation') },
-  captureBody => 1);
+  beforeDigest => sub { $_[0]->beginMode('display_math'); },
```
dginev, line 2048: Very nice simplification. The cherry on the cake would be to switch the few cases of 
```
beforeDigest => sub { $_[0]->beginMode('display_math'); }
``` 

to the abbreviated form of 

```
mode => 'display_math'
```

dginev (approved):
Good to merge, nice upgrade.

lib/LaTeXML/Engine/LaTeX.pool.ltxml
brucemiller: Ha! I fell into that trap! (& wasted time). It's subtly different. That short form simply switches the mode (typically within an environment which establishes a grouping), whereas the commands in question like `\lx@begin@displaymath` come in pairs which will do `beginMode` and `endMode`.

dginev: Ah! `mode` also ends the mode at the end, indeed that slipped my mind. Apologies.

PR #2463, opened 2024-12-30
More precise expansion for \typeout

### LINKED ISSUE #2221: token registers
I think I've dug as far into this rabbit hole as I'm able...

Working on #2123, I determined that `InputDefinitions('changes', type => 'sty', noltxml => 1)` doesn't work because `changes.sty` uses presetting keys, and xkeyval.sty.ltxml doesn't support that.  I went one step further and tried to `InputDefinitions('xkeyval', type => 'sty', noltxml => 1);`.  This results in changes loading without error, but the example posted in that issue still doesn't work.  I think I chased it down to xkeyval.sty working with token registers.  The following tex
```tex
\def\twoexpansions{3}
\def\oneexpansions{\twoexpansions}
\def\zeroexpansions{\oneexpansions}
\toks78\expandafter{\zeroexpansions}
\typeout{toks78 is: \the\toks78}

\documentclass{article}
\begin{document}
toks78 is: \the\toks78
\end{document}
```
will create a pdf displaying "toks78 is: 3", but the log file will say "toks 78 is: \oneexpansions".  On the other hand, LaTeXML will have "toks78 is: 3" in both locations.

dginev:
Fixes #2221 .

The unfortunate aspect is that the example reported in the issue identifies only a surface symptom problem, where `\typeout` was being too eager with expanding its argument before printing. But isn't otherwise related to the deeper issues related to changes.sty loads.

Although I suspect the main work of improving core interpretation will improve that front.

### DIFF
--- a/lib/LaTeXML/Engine/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Engine/LaTeX.pool.ltxml
@@ -4714,9 +4714,9 @@ DefPrimitiveI('\makeglossary', undef, undef);
 # C.11.6 Terminal Input and Output
 #======================================================================
 
-DefPrimitive('\typeout{}', sub {
+DefPrimitive('\typeout ExpandedPartially', sub {
     my ($stomach, $stuff) = @_;
-    Note(ToString(Expand($stuff)));
+    Note(ToString($stuff));
     return; });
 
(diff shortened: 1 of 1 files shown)


brucemiller:
Cool! Thanks!

PR #2464, opened 2024-12-30
Deactivate \keywords after \maketitle

### LINKED ISSUE #2263: llncs: `\and`  ignored after `\maketitle`
MWE:

```latex
\documentclass{llncs}

\begin{document}
\keywords{Lorem \and ipsum}

\maketitle

\keywords{Lorem \and ipsum}
\end{document}
```

LaTeXML output:

```xml
  <keywords name="Keywords">Lorem  and ipsum</keywords>
  <keywords name="Keywords">Lorem ipsum</keywords>
```

dginev:
Fixes #2263 .

That issue could also be closed as wontfix, as it is not really a "healthy" use of LaTeX syntax, since using \keywords after \maketitle is asking for trouble.

But since trouble is often a part of life, and we have some idiosyncratic code that _caused_ the behavior, I took a pass ironing it out a little further.

Namely, after `\maketitle` executes in a document, we try to deactivate a number of frontmatter macros. This PR adds `\keywords` to that list of macros, but also redoes the deactivation as a subroutine, so that deactivated macros can _also_ consume their arguments. The guard in question seems to [have been around since 2007](https://github.com/dginev/LaTeXML/commit/e1a454b00e8ef28cf2f673559df0d782da361301#diff-75ceaa5c4d7a9cd00a7663c800339bf9eec6355a16d8931c6c03d155852c6ff6R414), so I am conservatively sticking with it here.

### DIFF
--- a/lib/LaTeXML/Engine/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Engine/LaTeX.pool.ltxml
@@ -1243,6 +1243,7 @@ DefMacroI('\maketitle', undef,
     . '\global\let\title\relax'
     . '\global\let\author\relax'
     . '\global\let\date\relax'
+    . '\global\let\keywords\relax'
     . '\global\let\and\relax', locked => 1);
 
 DefMacro('\@thanks', '\@empty');
--- a/lib/LaTeXML/Engine/LaTeX.pool.ltxml
+++ b/lib/LaTeXML/Engine/LaTeX.pool.ltxml
@@ -1231,20 +1231,22 @@ DefMacro('\@add@conversion@date', '\@add@frontmatter{ltx:date}[role=creation]{\t
 # Doesn't produce anything (we're already inserting frontmatter),
 # But, it does make the various frontmatter macros into no-ops.
 DefMacroI('\maketitle', undef,
-  '\lx@frontmatterhere'
-    . '\@startsection@hook'
-    . '\global\let\thanks\relax'
-    . '\global\let\maketitle\relax'
-    . '\global\let\@maketitle\relax'
-    . '\global\let\@thanks\@empty'
-    . '\global\let\@author\@empty'
-    . '\global\let\@date\@empty'
-    . '\global\let\@title\@empty'
-    . '\global\let\title\relax'
-    . '\global\let\author\relax'
-    . '\global\let\date\relax'
-    . '\global\let\keywords\relax'
-    . '\global\let\and\relax', locked => 1);
+  '\lx@frontmatterhere\@startsection@hook\lx@deactivate@frontmatter');
+DefPrimitive('\lx@deactivate@frontmatter', sub {
+    Let('\maketitle',  '\relax',  'global');
+    Let('\@maketitle', '\relax',  'global');
+    Let('\@thanks',    '\@empty', 'global');
+    Let('\@author',    '\@empty', 'global');
+    Let('\@date',      '\@empty', 'global');
+    Let('\@title',     '\@empty', 'global');
+    DefMacro('\thanks[]{}', Tokens(), scope => 'global');
+    DefMacro('\title[]{}',  Tokens(), scope => 'global');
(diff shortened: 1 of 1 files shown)


brucemiller:
hmm, well, it's the wrong fix, isn't it? In the cited issue, the pdf actually has keywords appearing twice (whether or not that's healthy). But the underlying issue seems to be that `article.cls` resets `\and` (to `\relax`), but `llncs.cls` does not.

dginev:
> In the cited issue, the pdf actually has keywords appearing twice

How did you build that PDF? In my simple `pdflatex` run of the issue snippet I saw a single entry:



Now, the _source_ certainly had `\keywords` used twice, and my understanding is that the _second_ use did nothing at all, it just set an internal macro and moved on.

dginev:
But my PR appears to have accidentally unlocked `\maketitle`, which would have brought some pain in arXiv. I have just returned the lock, looking forward to more discussion.

dginev:
> the pdf actually has keywords appearing twice

On a third(!) look, I was fooled by missing a new page getting shipped. Indeed, the keywords appear twice in the PDF for this example.

I still think my PR attempt is a good _patch_ for the current approach of latexml, or do we really want to enable `\keywords` to deposit metadata before and after `\maketitle` in latexml?

brucemiller:
>  or do we really want to enable |\keywords| to deposit 
> metadata before and after |\maketitle| in latexml?

I'd invert the question: Do we really want to disable that, when LaTeX 
itself does not?  It's maybe even ironic, since the original issue is 
that latexml (in `LaTeX.pool`) disabled `\and` while `llncs.cls` does not.

In any case, I think the whole duplicate `\keywords` is a red herring 
introduced simply to demonstrate that `\maketitle` had unexpected 
side-effects.

dginev:
> In any case, I think the whole duplicate `\keywords` is a red herring

Yes, but we've patched this school of herrings before, so might as well be consistent... We can keep the llncs-sepcific issue open until some better solution comes to mind, but I think the PR itself may avoid an edge case in the mid-term.

brucemiller:
what edge case? What is the problem you're trying to solve?

dginev:
I am trying to disable all frontmatter-depositing macros after `\maketitle` has been encountered, as is the current intent of the `\maketitle` binding in LaTeX.pool.

The edge case this avoids is depositing duplicate metadata unintentionally.

The difference to latex this (continues to) cause is that in some .cls files the frontmatter macros after `\maketitle` indeed do deposit ink to the page, where latexml will silently ignore them. I think we can stay consistent with that behavior (namely, include `\keywords` to the neutralization code) until a larger redesign takes place.

brucemiller:
On 1/23/25 9:17 AM, Deyan Ginev wrote:
> I am trying to disable all frontmatter-depositing macros after | 
> \maketitle| has been encountered, as is the current intent of the | 
> \maketitle| binding in LaTeX.pool.

I think you're inferring more intent than is there; the intent is to 
emulate LaTeX's `\maketitle`.

> The edge case this avoids is depositing duplicate metadata unintentionally.

But the metadata was *intentionally* duplicated.  And your patch doesn't 
avoid duplication, it avoids it completely where you apparently don't 
want it.

> The difference to latex this (continues to) cause is that in some .cls 
> files the frontmatter macros after |\maketitle| indeed do deposit ink to 
> the page, where latexml will silently ignore them. I think we can stay 
> consistent with that behavior (namely, include |\keywords| to the 
> neutralization code) until a larger redesign takes place.

LaTeXML silently ignores metadata? I don't recall having seen this. It 
seems like a completely different bug (that would be more confusing to 
fix later, if we patch things to perpetuate the bug). Shouldn't we just 
fix that bug?

dginev:
> emulate LaTeX's `\maketitle`.
 
I think you mean article.cls here, which is where the `\maketitle` binding code appears to be copied from.

Ok, if there is a moratorium on improving that, I will let that be as it is.

PR #2465, opened 2024-12-31
To attribute

brucemiller:
This PR provides an option `toAttribute` to `DefConstructor,DefEnvironment` (and `Whatsit`s generally) to customize the conversion of a whatsit to an attribute value; its sytax is analogous to that of the `reversion` option, being a sub or string, possibly containing `#1..`. 
This is useful for objects which need the flexibility of a `Whatsit` (as opposed to simple `Box`) but which are sometimes placed in an attribute value that ends up in the text (or html) of the conversion; In those cases, the value should contain neither XML, nor the TeX reversion (as `toString` typically does). The motivating example is `\hskip`.
The `setAttribute` method of `Document`, when given an object reference as value, will use `toAttribute` to turn it into a string, but we need to be careful *not* to use `ToString` on such values before they get there.

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -498,10 +498,10 @@ sub serialize_aux {
       (map { $_ . '="' . $atnodes{$_} . '"' } sort keys %atnodes)
     );
     my $noindent_children = ($heuristic
-        # This emulates libxml2's heuristic
-        #     ? $noindent || grep { $_->nodeType != XML_ELEMENT_NODE } @children
+      # This emulates libxml2's heuristic
+      #     ? $noindent || grep { $_->nodeType != XML_ELEMENT_NODE } @children
       ? $noindent || grep { $_->nodeType == XML_TEXT_NODE } @children
-        # This is the "Correct" way to determine whether to add indentation
+      # This is the "Correct" way to determine whether to add indentation
       : $model->canContain(getNodeQName($self, $node), '#PCDATA'));
     return join('',
       ($noindent ? '' : $indent), $start,
@@ -1376,12 +1376,15 @@ sub makeError {
 # [xml:id and namespaced attributes are always allowed]
 sub setAttribute {
   my ($self, $node, $key, $value) = @_;
-  return if (ref $value) && ((!blessed($value)) || !$value->can('toAttribute'));
   if (ref $value) {
     if ($key eq '_box') {
       return $self->setNodeBox($node, $value); }
     elsif ($key eq '_font') {
       return $self->setNodeFont($node, $value); }
+    elsif ((!blessed($value)) || !$value->can('toAttribute')) {
+      Warn('unexpected', (ref $value), $self,
+        "Don't know how to encode $value as an attribute value");
+      return; }
     else {
       $value = $value->toAttribute; } }
   if ((defined $value) && ($value ne '')) {    # Skip if `empty'; but 0 is OK!
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -1376,8 +1376,14 @@ sub makeError {
 # [xml:id and namespaced attributes are always allowed]
 sub setAttribute {
   my ($self, $node, $key, $value) = @_;
-  return                       if (ref $value) && ((!blessed($value)) || !$value->can('toAttribute'));
-  $value = $value->toAttribute if ref $value;
+  return if (ref $value) && ((!blessed($value)) || !$value->can('toAttribute'));
+  if (ref $value) {
+    if ($key eq '_box') {
+      return $self->setNodeBox($node, $value); }
+    elsif ($key eq '_font') {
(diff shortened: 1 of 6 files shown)


lib/LaTeXML/Core/Document.pm
```diff
@@ -1376,8 +1376,14 @@ sub makeError {
 # [xml:id and namespaced attributes are always allowed]
 sub setAttribute {
   my ($self, $node, $key, $value) = @_;
-  return                       if (ref $value) && ((!blessed($value)) || !$value->can('toAttribute'));
-  $value = $value->toAttribute if ref $value;
+  return if (ref $value) && ((!blessed($value)) || !$value->can('toAttribute'));
```
dginev, line 1379: Would it be useful to emit a `Warning` if this unusual case is encountered, where `$value` was e.g. a perl array ref or hash ref? Could be some rare mistake that gets caught quickly with a warning message.

lib/LaTeXML/Core/Whatsit.pm
```diff
@@ -218,6 +218,31 @@ sub beAbsorbed {
...
+  my ($self) = @_;
+  my $props  = $$self{properties};
+  my $defn   = $self->getDefinition;
+  my $spec   = $$props{toAttribute} || $$defn{toAttribute};
+  if (!defined $spec) {
+    return $self->toString; }    # Default
+  elsif (ref $spec eq 'CODE') {    # If handled by CODE, call it
+    $spec = &$spec($self, $self->getArgs); }
+  # Now, similar to substituteParameters, but creating a string.
+  $spec =~ s/#(#|[1-9]|\w+)/ toAttribute_aux($self,$1)/eg;
```
dginev, line 232: Hm, I think I understand the code here, but I am wondering if we should have a an example DefConstructor that uses this argument-based `toAttribute` string form, and a test that exercises it.

lib/LaTeXML/Core/Whatsit.pm
```diff
@@ -218,6 +218,31 @@ sub beAbsorbed {
...
+
+sub toAttribute_aux {
+  my ($self, $code) = @_;
+  my $value;
+  if    ($code eq '#') { return $code; }
+  elsif ((ord($code) > ord('0')) && (ord($code) <= ord('9'))) {
+    $value = $self->getArg(ord($code) - ord('0')); }
+  else {
+    $value = $self->getProperty($code); }
+  $value = $value->toAttribute if (defined $value) && (ref $value);
```
dginev, line 243: minor, but `ref undef` works fine, so `defined $value` is redundnat.

lib/LaTeXML/Package.pm
```diff
@@ -3682,6 +3686,14 @@ the one defined in the C<prototype>.  This is a convenient alternative for
 reversion when a 'public' command conditionally expands into
 an internal one, but the reversion should be for the public command.
 
+=item C<toAttribute=E<gt>I<texstring> | I<code>($whatsit,#1,#2,...)>
+
+specifies the conversion of the invocation back into plain text for an attribute value
+(the default is C<toString>).
+The I<textstring> string can include C<#1>, C<#2>...
+The I<code> is called with the C<$whatsit> and digested arguments
+and must return a string.
```
dginev, line 3695: Possibly a doc string like this is also needed for the DefEnvironment addition.

dginev (commented):
Interesting upgrade, looks healthy.

I have a naming comment (no surprise there). The `toNoun` names are typically used for methods in latexml. The alternative form specifiers tend to have some noun name as with `reversion`, `alias`, `sizer`...

Indeed, `toAttribute` names methods implemented by our Box-like objects. I think it would be helpful to name the config parameter in the Definition objects slightly differently.

(Aside: one of my related questions was wheter we want a generic `->toAttribute` method in Definition.pm, which defaults to `->toString`)

A simple distinction would be to swap `to` for `as`, namely `asAttribute`. Another is `intoAttribute` (Rust has standardized [.into()](https://doc.rust-lang.org/std/convert/trait.Into.html) as one reason I thought of that).

The main point of my comment is that the distinction between `reversion` and `revert()` is a nice model to follow for `?` and `toAttribute()`. Maybe `attributeSpec` or `attributeForm` gets even closer to that...

lib/LaTeXML/Core/Document.pm
brucemiller: good idea; done!

lib/LaTeXML/Core/Whatsit.pm
brucemiller: The conceit is that it should work just like `reversion`.

lib/LaTeXML/Core/Whatsit.pm
brucemiller: indeed; fixed.

lib/LaTeXML/Package.pm
brucemiller: yep, done.

brucemiller:
Yeah, the naming is generally off for all these toNoun things, but I like your `attributeForm` suggestion.

PR #2475, opened 2025-01-03
use RTL to shift \llap content and fix \llap, \rlap extreme wrapping

### LINKED ISSUE #2474: \llap does not shift to the left, creates newlines
`\llap` is supposed to put its content to the left of the current position, but the HTML generated by LaTeXML does not do that. The resulting box will also pile all text in a column because of `width:0.0pt`. Just try the following:

```latex
\parindent=0pt

Sufficiently long text \llap{with overlap}
```

The overlap direction can be fixed with `direction:rtl;`; to be safe, one must also create a further wrapper which restores the intended direction. The characters piling up can be fixed with `white-space: nowrap`.

Note also that the `\parident=0pt` should not be necessary... the LaTeXML CSS adds `text-indent` even inside the boxes created by `\hbox` (I am too lazy right now to create another issue!).

xworld21:
Fix #2474 (?), unless you object to `\@ADDCLASS` here.


dginev:
@xworld21 I am not sure the RTL trick doesn't stray a bit too far astray from its original CSS purpose. I have recently been playing around with the `float:` property, which may be a little closer to the TeXbook description:

> ‘\llap{something}’ creates a box of width zero,
with ‘something’ extending just to the left of that box.

This indeed works on all kinds of content, including tables and equations.

I tried a simple TeX snippet:
```tex
testing some \llap{llap} and \rlap{rlap} text here.
\bye
```

and hand-crafted some HTML using `float:` and your class-based approach. This result appears to match the PDF perfectly:

```html
<html>
  <style>
    .llap-container,
    .rlap-container {
      width: 0;
      display: inline-block;
      vertical-align: bottom;
    }
    .llap {
      float: right;
    }
    .rlap {
      float: left;
    }
  </style>
  <body>
    <p>testing some <span class="llap-container"><span class="llap">llap</span></span> 
and <span class="rlap-container"><span class="rlap">rlap</span></span> text here.
    </p>
    </body>
</html>
```

That said, I suspect the _usual_ way of applying these macros will be really bad for accessible outcomes (especially screen readers), as the CSS trickery never carries to the accessibility tree. So I would also discourage basing any important new work on these macros...

dginev:
A related aside to my last comment: being able to deduce the context of `\hss` use and map that to a class attribute that floats the outcome left/right, or centers it (for `\hss #1 \hss`), also seems to me to be as useful as having functional `\llap` and `\rlap`.

xworld21:
> ```
>     .llap {
>       float: right;
>     }
> ```

That is infinitely better, thanks!

I am wondering if we should just promote `\llap` to an inline-block constructor with `sizer => 0` and let XSLT create the wrappers when it sees `ltx_[lr]lap`. It should solve multiple problems at once – no need to hack around with the zero-width space, the inline width can be moved to the class, and it is more semantic.

> That said, I suspect the _usual_ way of applying these macros will be really bad for accessible outcomes (especially screen readers), as the CSS trickery never carries to the accessibility tree. So I would also discourage basing any important new work on these macros...

The tests include a few legit cases about preserving alignment (and, well, I bumped into this with the raw beamer stuff) which should work fine with screen readers I think. I assume screen readers read floats in a sensible way. I'll check. As to using \llap for e.g. creating new symbols, yeah, the output won't make sense accessibility wise.

xworld21:
> That is infinitely better, thanks!

Oh, I take that back. Vertical alignment changes with float, and if you add `text-decoration: underline` the difference is quite visible. The rtl trick has the advantage of keeping the text on the same line.

dginev:
I am not sure I follow. Adding `text-decoration: underline;` to the example I shared seems to display well? As in, it matches a PDF where `\underline` was used in the source.

Maybe it wasn't obvious - I had added `vertical-align: bottom;` to restore the alignment. There could be trickier cases, would be nice to translate those into tricky tests.

xworld21:
> Maybe it wasn't obvious - I had added `vertical-align: bottom;` to restore the alignment. There could be trickier cases, would be nice to translate those into tricky tests.

I... I did get two misaligned underlines before, but I can't reproduce it again! I guess I must have changed the inline block size accidentally (most likely with some debug border), and that would have the same effect on `direction:rtl`, most likely. Back to floats it is until we find an actual reproducible problem.

lib/LaTeXML/resources/XSLT/LaTeXML-inline-xhtml.xsl
```diff
@@ -33,22 +33,41 @@
...
-        <xsl:with-param name="context" select="$innercontext"/>
-      </xsl:apply-templates>
-      <xsl:apply-templates>
-        <xsl:with-param name="context" select="$innercontext"/>
-      </xsl:apply-templates>
-      <xsl:apply-templates select="." mode="end">
-        <xsl:with-param name="context" select="$innercontext"/>
-      </xsl:apply-templates>
-    </xsl:element>
+    <xsl:variable name="content">
```
xworld21, line 36: Just proof of concept, this would have to be implemented for ltx:inline-block and ltx:p too (the other possible constructors from `\hbox`).

PR #2477, opened 2025-01-06
implement TikZ/PGF baseline support

### LINKED ISSUE #2462: TikZ: baseline attribute in tikzpicture
In LaTeX, it is possible to shift an inline tikpicture with the keyword `baseline` either by an absolute value or by specifying a coordinate point (as in the example below). It seems that this feature is not supported in LaTeXML.

Here is a minimal example:

```tex
\documentclass{article}
\usepackage{tikz}

\begin{document}
Text before
\begin{tikzpicture}[baseline=(baseline point)]
	\coordinate (baseline point) at (0, -1em);
	\draw (0, 0) circle[radius=2em];
\end{tikzpicture}
text after
\end{document}
```

Crop from the PDF file, `pdflatex` output:
image

Crop from the HTML file, `latexmlc` output:
image

xworld21:
Fix #2462. Note that it does not fix the baseline issues in the 'unit tests by Silviu'. This is a good sign, because those tests do not modify the baseline, ever; they are broken for a different minx/miny related reason.

### DIFF
--- a/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
+++ b/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
@@ -77,13 +77,15 @@ DefConstructor('\lxSVG@insertpicture{}', sub {
     if ($document->getNodeQName($current) =~ /^svg:/) {    # Already in svg:svg ?
           # Seemingly this positioning is correct; otherwise inherit from context
       my ($x, $y) = (-$props{minx}, -$props{miny});
+      $y -= $props{depth}->pxValue if $props{depth}->pxValue;
       my $gnode = $document->openElement('svg:g',
         transform   => "matrix(1 0 0 1 $x $y)",
         _scopebegin => 1, class => 'ltx_nestedsvg');
       $document->absorb($content);
       $document->closeElement('svg:g'); }
     else {
-      $document->openElement('ltx:picture');
+      my $picture = $document->openElement('ltx:picture');
+      $document->setAttribute($picture, 'imagedepth', $props{depth}->pxValue) if $props{depth}->pxValue;
       $document->openElement('svg:svg',
         version => "1.1",
         width   => $props{pxwidth}, height => $props{pxheight},
@@ -114,12 +116,13 @@ DefConstructor('\lxSVG@insertpicture{}', sub {
     my $miny    = LookupRegister('\pgf@picminy');
     my $width   = LookupRegister('\pgf@picmaxx');
     my $height  = LookupRegister('\pgf@picmaxy');
+    my $depth   = Dimension(Expand(T_CS('\pgf@shift@baseline')));
     my $w       = max($width->pxValue,  1);
     my $h       = max($height->pxValue, 1);
     my $content = $whatsit->getArg(1);
     my ($cwidth, $cheight, $cdepth) = $content->getSize;
-    Debug("TIKZ size: " . ToString($width) . " x " . ToString($height)) if $LaTeXML::DEBUG{pgf};
-    Debug("TIKZ pos: " . ToString($minx) . " x " . ToString($miny))     if $LaTeXML::DEBUG{pgf};
+    Debug("TIKZ size: " . ToString($width) . " x " . ToString($height) . " + " . ToString($depth)) if $LaTeXML::DEBUG{pgf};
+    Debug("TIKZ pos: " . ToString($minx) . " x " . ToString($miny)) if $LaTeXML::DEBUG{pgf};
     Debug("TIKZ BODY: " . ToString($cwidth) . " x " . ToString($cheight)
         . " + " . ToString($cdepth)) if $LaTeXML::DEBUG{pgf};
     Debug("BODY is " . ToString($content)) if $LaTeXML::DEBUG{pgf};
@@ -130,7 +133,7 @@ DefConstructor('\lxSVG@insertpicture{}', sub {
     $whatsit->setProperty(miny     => 0);
     $whatsit->setProperty(width    => $width);
     $whatsit->setProperty(height   => $height);
-    $whatsit->setProperty(depth    => Dimension(0));
+    $whatsit->setProperty(depth    => $depth->add($miny->negate));
     $whatsit->setProperty(pxwidth  => $w);
     $whatsit->setProperty(pxheight => $h);
     # or tikz macro (see corescopes)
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
```diff
@@ -77,13 +77,15 @@ DefConstructor('\lxSVG@insertpicture{}', sub {
     if ($document->getNodeQName($current) =~ /^svg:/) {    # Already in svg:svg ?
           # Seemingly this positioning is correct; otherwise inherit from context
       my ($x, $y) = (-$props{minx}, -$props{miny});
+      $y -= $props{depth}->pxValue if $props{depth}->pxValue;
```
xworld21, line 80: Actually I haven't tested this, how do I do that?

lib/LaTeXML/resources/XSLT/LaTeXML-picture-xhtml.xsl
```diff
@@ -95,6 +95,15 @@
...
+        <xsl:attribute name="style">
+          <xsl:if test="@imagedepth">
+            <xsl:value-of select="concat('vertical-align:',-@imagedepth,'px')"/>
+            <xsl:if test="svg:svg/@style">;</xsl:if>
+          </xsl:if>
+          <xsl:value-of select="svg:svg/@style"/>
+        </xsl:attribute>
+      </xsl:if>
       <!-- but copy other svg:svg attributes -->
       <xsl:for-each select="svg:svg/@*">
```
xworld21, line 108: Actually:
```suggestion
      <xsl:for-each select="svg:svg/@*[name() != 'style']">
```

lib/LaTeXML/resources/RelaxNG/LaTeXML-common.rng
```diff
@@ -408,7 +408,7 @@ Note that, unlike \TeX, this is the total height, including the depth (if any).<
 baseline of the content shown in the image.
 When displayed inilne, an image with a positive \attr{depth}
 should be shifted down relative to the baseline of neighboring material.</a:documentation>
-        <data type="integer"/>
+        <data type="float"/>
```
brucemiller, line 411: Why?

xworld21: The depth is almost never an integer, and it goes into `vertical-align: NNpt;` so there is no harm in having decimals (and there is actual harm in  rounding it, as the SVG will be visibly off).

This applies to width/height as well. For scalable images, fractional sizes make sense and rounding causes artifacts. This is off topic for this PR, though.

brucemiller: I'm generally a bit confused about possibly overloading and/or conflating things, here. `@imagedepth` goes with `@imagewidth,@imageheight` for images, whereas `@height,@depth,@width` are more generic, and probably are what should be used.  Is pgf responsible for setting `pgf@shift@baseline`?

lib/LaTeXML/resources/XSLT/LaTeXML-picture-xhtml.xsl
```diff
@@ -95,8 +95,17 @@
...
+      <xsl:if test="@imagedepth | svg:svg/@style">
+        <xsl:attribute name="style">
+          <xsl:if test="@imagedepth">
+            <xsl:value-of select="concat('vertical-align:',-@imagedepth,'px')"/>
+            <xsl:if test="svg:svg/@style">;</xsl:if>
+          </xsl:if>
+          <xsl:value-of select="svg:svg/@style"/>
+        </xsl:attribute>
+      </xsl:if>
       <!-- but copy other svg:svg attributes -->
```
brucemiller, line 107: It's probably my fault for being slow with the PR reviews, but now we've got different variants of adjusting depth or baseline in the XSLT (#2397)
If we had a `@depth`, and used `<xsl:call-templates name="add_attributes"/>`, instead of `add_classes`, we'd get the adjustment for free, I think. Do we get too many other CSS things added?

lib/LaTeXML/resources/RelaxNG/LaTeXML-common.rng
xworld21: Yes, PGF sets `\pgf@shift@baseline` internally. I believe I used `@imagedepth` because it's the only relevant attribute that I had available for the task. I think. I admit I don't remember now.

lib/LaTeXML/resources/XSLT/LaTeXML-picture-xhtml.xsl
xworld21: Yes, unfortunately they overlap, but they are trying to do the same thing. I wager that this version here is the better one, for instance because `-` is outside of the string!

xworld21: > we'd get the adjustment for free

I didn't reply to this one, but of course that sounds like a good idea (adding `@depth` to the schema and implementing the special behaviour in `add_attributes`).

lib/LaTeXML/resources/RelaxNG/LaTeXML-common.rng
xworld21: PS: I checked why I was using `@imagedepth`. This is because `<svg:svg>` has no `@depth` attribute, so the information has to go in `<ltx:picture>`, which already has the standard `@imagewidth` etc. I think that having *both* attributes on `<ltx:picture>` would be confusing.

brucemiller: `imagedepth` is for image (files, usually raster), and is in pixels; I'm less worried at this point about fractional pixels. Either way, *something* is missing!

brucemiller: I can get the right appearance with a simpler subset of what you've got here: Basically, in the afterDigest for `\lxSVG@insertpicture`, I set `style` to the vertical align based on what's in ``pgf@shift@baseline`.  So, then you don't need any extensions to schema or xslt. And that kinda makes sense, since `svg:svg` doesn't have a `depth` or `baseline` or `y-shift` or whatever attribute, preferring we use CSS. And on the other hand `ltx:picture` is a bit of a mess, having one foot in ancient pstricks code, and the other in modern svg.

PR #2480, opened 2025-01-08
more correct emulation of \box and \copy primitives

### LINKED ISSUE #2412: pgf layers do not get cleared
A baffling scope bug, seemingly related to some issues reported on ar5iv. I spotted this in the atoms-and-orbitals test: the content of pgf layers (created via `pgfonlayer`) does not clear when closing a group.
```latex
\documentclass{article}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
    \pgfdeclarelayer{background}
    \pgfsetlayers{background,main}
    \draw[red] (-1,0) -- (1,0);
    \begin{pgfonlayer}{background}
    \filldraw (0,0) circle (4pt);
    \end{pgfonlayer}
\end{tikzpicture}
\begin{tikzpicture}
    \pgfdeclarelayer{background}
    \pgfsetlayers{background,main}
    \draw[red] (-1,0) -- (1,0);
\end{tikzpicture}
\end{document}
```
renders as
image
You can see that the black circle is still in the background layer even if it's a new image.

dginev:
Fixes #2412 .

Two key upgrades:
 - deeply clone the box returned during `\copy`
 - for `\box`, add a new `removeValue` method in State, which (idempotently) unbinds the nearest binding of a value in the undo table.

This gets the tests discussed in the issue to pass, and matches the pgf rendering in the PDF build (really great diagnostic by @xworld21 !)

I would need a second opinion by @brucemiller on whether this kind of "removeValue" method makes sense for latexml's State. Its code is inspired by the global assignment mode of `assign_internal`, which will unbind all values in all frames, and then assign a value on the cleaned up 1-element stack.

I am wondering if many of the current uses of `AssignValue($key, undef)` are _meant_ as this `RemoveValue` behavior, which could motivate changing local uses of assign_internal with undef to also unbind from the nearest assigned frame -- without creating a new method.

Edit: Another direction for refining the code is to add a `'local'` vs `'global'` argument for `removeValue`, to indicate whether one or all bindings should be deleted.

But I am feeling confident I am following the `\box` unbind semantics accurately now, as explained in the TeXbook. What was missing was unbinding the actual assignment - we were adding a local undef assignment, while also keeping the original assignment in place, when it happened in a higher frame.

Thoughts?


lib/LaTeXML/Core/State.pm
```diff
@@ -227,6 +227,23 @@ sub shiftValue {
   assign_internal($self, 'value', $key, [], 'global') unless $$vtable{$key}[0];
   return shift(@{ $$vtable{$key}[0] }); }
 
+sub removeValue {
```
xworld21, line 230: You could use this code to implement `scope => 'samelevel'` assignments (with a better name, if you can find it), which would be equivalent to `removeValue` when assigning `\undef`. Should you one day discover another assignment 'in place'.

lib/LaTeXML/Core/State.pm
```diff
@@ -227,6 +227,23 @@ sub shiftValue {
...
+  my ($self, $key) = @_;
+  if (my $vvec = $$self{value}{$key}) {
+    my $value;
+    # undo ONLY the most recent binding for this value.
+    my $frame;
+    my @frames = @{ $$self{undo} };
+    while (@frames) {
+      $frame = shift(@frames);
+      if (my $n = $$frame{value}{$key}) {    # Undo the bindings, if $key was bound in this frame
```
xworld21, line 239: Is this idempotent? I.e. if applied twice, will it remove the *two* previous bindings? That would be too much for `\box`.

dginev: Curious idea - we don't yet have a test for that. Something like:

```
\newbox\myboxD
\setbox\myboxD\vbox{D won't appear outer.}
\begingroup
  \setbox\myboxD\vbox{D will appear inner.}
  \box\myboxD
  \box\myboxD
\endgroup
\bye
```

dginev: Added a test. It appears to work as expected?

lib/LaTeXML/Core/State.pm
dginev: That's also a possibility. My main worry is that `assign_internal` is already somewhat entangled, so adding more to it just makes life harder... Will wait for feedback from Bruce.

I enjoy the idea of having a dedicated removal method, for what that's worth. But since the removal isn't global, the name is a bit misleading. It's really `UnbindLastValueAssignment`, but this is not a Java project...

lib/LaTeXML/Core/State.pm
dginev: Scratch that, the test wasn't precise enough -- there is indeed a bug where unbinding twice will pop the outer frame, nice catch @xworld21

dginev: Ok, using `exists` and a `0` value for the undo counter becomes idempotent. The new test now has correct timing and matches the PDF.

But the code got extra scary. Maybe good, maybe not?

lib/LaTeXML/Core/State.pm
xworld21: > That's also a possibility. My main worry is that `assign_internal` is already somewhat entangled, so adding more to it just makes life harder... Will wait for feedback from Bruce.

I was thinking the opposite: this code is very similar to `assign_internal` and would be better placed there for ease of maintainance... ?

I am also wondering if removing values has other side effects, e.g. will this make `\ifcsname` false? That's not really possible in TeX, is it? (This has no real consequence of course since `removeValue` is only used for the box registers.)

dginev: > this code is very similar to assign_internal and would be better placed there for ease of maintainance... ?

Only if it can be DRY ("do not repeat yourself"). Otherwise just growing the body of the function will make it more impenetrable.

xworld21: Going back to the DRY idea, couldn't this code be folded into the global assignment? If I parse correctly, your new code and global assignment differ by only two lines (`last;` and `$$frame{value}{$key} = 0`). So a `samelevel` scope could be implemented by adding the line `last if $samelevel` in `assign_internal` (but it wouldn't *remove* the value as you do here, of course).

dginev: I am open to the idea.

Would likely seek a slightly better name ... To fit better with `global` and `local`, maybe `initial` is a good keyword? Fishing for inspiration from [CSS initial](https://developer.mozilla.org/en-US/docs/Web/CSS/initial)

xworld21: I agree that `samelevel` is a terrible name... let's think about how one could *document* the new scope. The assignment in scope `samelevel` means something like:

- in the last active scope where `$name` was assigned a value (so `last`?)
- in the scope where `$name` received its current value/the current definition of `$name` was made (so `current_def` or `current_value`?)
- in the same scope where `$name` was assigned the current value (so `same`?)
- in the scope where `$name` was initially assigned the current value (so `initial`?)

I am not totally convinced by `initial`. It sounds too similar to global.

dginev: Well, the primary purpose we'll use this for is still "removing" or "unsetting" the value (CSS has an "unset" property related to initial which also comes to mind). So having an associated `scope=>"unset"` could also be intuitive when looked at (+ documented) from the right perspective, if the "initial" is too confusing. The thing that bothers me a little is that such calls really don't take a "value" argument, so unifying the interfaces will be inevitably clunky. So I can also see as reasonable to keep the code paths separate.

brucemiller:
I really need to think about what TeX is really doing, conceptually, to choose a good name & API. I expect that whatever this is should be kept fairly internal, and *not* exported to `Package`, since except for this special case of a box, removing a value is almost certainly  *not* what you'll want to do.
The cloning is curious for a different reason. It's only needed since `adjustBoxColor_internal` (unwisely?) modifies the boxes in place; and that's different than TeX, because we've diverged from TeX by embedding color into the Font, and embedded the Font into the box. (TeX handles color via a much different hack).

brucemiller:
Ah, not quite there; As mentioned in some other issue, there's the problem of when `Black` actually should have meant "current color".  Consider this variant of your tests:
```
\setbox\myboxC\vbox{{\color{black}Black, stays black, then black.}}
\copy\myboxC
{\color{red}\copy\myboxC}
\copy\myboxC
```

brucemiller:
I'm suspecting that perhaps Font's `$DEFCOLOR` should be undef, rather than `Black` (and maybe many other `Black`, as well), but that'll involve tracking a number of surprises.

lib/LaTeXML/Core/State.pm
brucemiller: I may be missing something, but I think you're overcomplicating it. It's not really "removing", nor "unbinding", but setting to "<void>" (in Knuth's terms) in the same frame, or "inplace" as @xworld21 suggested. So, adding the following clause to `assign_internal` (eg. after "local") gets the right functionality, and is arguably clearer all-round, since it keeps things together where things can be more easily compared:
```
  elsif ($scope eq 'inplace') {                          # Special case for \box & friends
    if (exists $$self{$table}{$key}) {        # If the value was previously assigned AT ALL
      $$self{$table}{$key}[0] = $value; }     # Simply replace the value in its frame
    else {                                    # Otherwise, push new value & set 1 to be undone
      $$self{undo}[0]{$table}{$key} = 1;
      unshift(@{ $$self{$table}{$key} }, $value); } }    # And push new binding.
```
Then, in `\box` update the assignment:
```
    AssignValue($box, undef, 'inplace');
```
BTW: `\unhbox` and `\unvbox` need the same change.

dginev: > setting to "void" (in Knuth's terms) 

Are we sure we know what that means in latexml terms? That's _very_ slippery to me.

brucemiller: Why wouldn't we know what void means? (again, you're overcomplicating)

dginev: Because I read the original code:
https://github.com/brucemiller/LaTeXML/issues/2412#issuecomment-2578251006

If we can't agree on the basic API here, I'll just close - feel free to add the version of the code you like.

brucemiller: Fine, I'll submit a patch; the code I suggested fixes #2412, and the atomic orbitals.
The color propagation is a separate problem, with 2 parts. It is a result of the color being embedded in the font in box, and thus to change it we're modifying the *stored* box (which cloning is one possible fix); But also, we should only be changing the color if it had been the "default black", rather than unconditionally.  Alas, we don't yet have a clear handle on that default.

PR #2481, opened 2025-01-08
use UnTeX in \patchcmd replacements

### LINKED ISSUE #2468: etoolbox: \patchcmd strips spaces because of ToString
`\patchcmd` mangles expansions that contain whitespace, because it is using `ToString` at
https://github.com/brucemiller/LaTeXML/blob/f82b733eade2c81f24567c349929f2a696cb66cb/lib/LaTeXML/Package/etoolbox.sty.ltxml#L1299
I understand that it should be `UnTeX` instead. Minimal broken example:
```latex
\documentclass{article}
\usepackage{etoolbox}
\def\example{{Normal to \bfseries italic}}
\begin{document}

\example

\patchcmd\example{\bfseries}{\textit}{}{}

\example

\end{document}
```

dginev:
Fixes #2468 

This was both diagnosed and solved by @xworld21 , I am just filing the PR.

I can confirm the desired improvement is indeed there when using `UnTeX`.


lib/LaTeXML/Package/etoolbox.sty.ltxml
```diff
@@ -1296,15 +1296,15 @@ DefMacro('\patchcmd [] DefToken {}{}{}{}', sub {
         Info('unexpected', 'patchcmd', $gullet, "Patchcmd is not supported on LaTeXML-native definitions, will not patch " . ToString($cs));
         return $failure;
       }
-      my $string        = ToString($expansion);
-      my $search_string = ToString($search);
+      my $string        = UnTeX($expansion);
+      my $search_string = UnTeX($search);
```
brucemiller, line 1300: You'll probably want to add the optional `$suppress_linebreak` argument here, (ie. `,1`) just in case.

dginev: Oh, sure!

dginev: Thanks for beating me to it.

brucemiller (approved):
Now I like it!

Issue #2439, opened 2024-11-22
vertical space in multiline environments is ignored


physikerwelt:
In comparison to MathJax, LaTeXML, and TEMML seem to ignore vertical spacing in align environments.
```
$$\begin{aligned}{A\\[600pt]
B\end{aligned}$$
vs
$$\begin{aligned}{A \\
B\end{aligned}$$
```

MathJax adds rowspacing=... in those cases. Is there a reason why LateXML does not do that, or is this a bug?

dginev:
It looks like an unintentional limitation to me.

brucemiller:
The extra space on `\\[600pt]` gets recorded (as `padding`) in the current row in the `Alignment` object; there's code which wants to add CSS (to the cell?) but it seems to be getting lost.
FWIW: MathML Core doesn't support `rowspacing`.

dginev:
> FWIW: MathML Core doesn't support rowspacing.

Luckily both Firefox and Chrome support `padding-bottom` on `<mtd>`, here is an example codepen:

https://codepen.io/dginev/pen/EaYdYbV

dginev:
@physikerwelt I think Bruce already touched on this above, but let me expand:

We haven't had any sort of "official decision" on that exact attribute, but a decision we made back for the v0.8.8 release was for LaTeXML to now target "MathML Core" as the presentation MathML dialect.

As it happens, `rowspacing` is non-core, confirmed by the [MathML 4 attributes table](https://www.w3.org/TR/mathml4/#attributes-23), so if we can model it with CSS, we probably will do that and only that.

PR #2483, opened 2025-01-10
more detailed warning for toAttribute in Document::setAttribute

dginev:
Follow-up to #2465 , adding a little more detail to the warning message.

I encountered the warning while diagnozing something independent, on the example:
```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}
$$\begin{aligned}A \end{aligned}$$
\end{document}
```

Which with this PR will now print to the `.latexml.log` file the following.

@brucemiller if you have any suggestions how to handle the array-reference attribute, I can add that here if you prefer.

```
Warning:unexpected:ARRAY While setting "columnwidths", don't know how to encode ARRAY(0x5728582ae2f8) as an attribute value.
	at test.tex; line 5 col 2 - line 5 col 34
	$VAR1 = [
	          bless( [
	                   491521
	                 ], 'LaTeXML::Common::Dimension' )
	        ];
	In Core::Document[@0x572857366e40] /tmp/test.tex; from line 5 col 2 to line 5 col 34
	 <= Alignment[] <= Core::Document[@0x572857366e40] <= Core::Definition::Constructor[\lx@beg... <= ...
Warning:unexpected:ARRAY While setting "rowheights", don't know how to encode ARRAY(0x5728582ae4f0) as an attribute value.
	at test.tex; line 5 col 2 - line 5 col 34
	$VAR1 = [
	          bless( [
	                   447828
	                 ], 'LaTeXML::Common::Dimension' )
	        ];
	In Core::Document[@0x572857366e40] /tmp/test.tex; from line 5 col 2 to line 5 col 34
	 <= Alignment[] <= Core::Document[@0x572857366e40] <= Core::Definition::Constructor[\lx@beg... <= ...
 0.02 sec)
```

### DIFF
--- a/lib/LaTeXML/Core/Document.pm
+++ b/lib/LaTeXML/Core/Document.pm
@@ -20,6 +20,7 @@ use LaTeXML::Common::Error;
 use LaTeXML::Common::XML;
 use LaTeXML::Util::Radix;
 use Unicode::Normalize;
+use Data::Dumper;
 use Scalar::Util qw(blessed);
 use base         qw(LaTeXML::Common::Object);
 
@@ -1383,7 +1384,7 @@ sub setAttribute {
       return $self->setNodeFont($node, $value); }
     elsif ((!blessed($value)) || !$value->can('toAttribute')) {
       Warn('unexpected', (ref $value), $self,
-        "Don't know how to encode $value as an attribute value");
+        "While setting \"$key\", don't know how to encode $value as an attribute value.", Dumper($value));
       return; }
     else {
       $value = $value->toAttribute; } }


dginev:
P.S. The reason I was studying that example is to check on the cause behind #2439

brucemiller:
Good suggestion! I've incorporated it into #2495, since I was already messing with this message.
As far as a general handling of arrays, the obvious thing would be to join the `->toAttribute`s of the elements, but the appropriate separator probably depends on the attribute.

dginev:
>  the appropriate separator probably depends on the attribute.

For `columnwidths` and `rowheights`, I assume are commas OK for delimiters? I can add a follow-up PR with that.

Closing here, since the change has landed via #2495 (thanks!)

brucemiller:
> For |columnwidths| and |rowheights|, are commas OK for delimiters? I can 
> add a follow-up PR with that.

Are those ever *intentionally* set as attributes? (as opposed to 
accidentally, due to being in a hash of properties).  They don't seem to 
be in the schema, nor directly addressed by the XSLT.

dginev:
Well, judging by Alignment.pm's documentation, yes? There is a documented function `openContainer`, as:

> `openContainer  = sub($doc,%attrib);` creates the container element with given attributes

And then:
```perl
&{ $$self{openContainer} }($document, ($attr ? %$attr : ()),
    cwidth       => $$self{cwidth}, cheight => $$self{cheight}, cdepth => $$self{cdepth},
    rowheights   => $$self{rowheights},
    columnwidths => $$self{columnwidths},
  );
```

So this _appears to be_ intentionally reaching the XML attributes?

brucemiller:
Alas, those "attributes" aren't necessarily all actual XML attributes, 
just data for the element openner that does ultimately open the XML element.
In any case, if you're after rowspacing (the space between rows) the 
array of rowheights isn't directly what you need, anyway.

brucemiller:
Unintended consequences of sloppy documentation? (or design?)  I'm confused: I don't get any warnings on your example, and you shouldn't, since the schema doesn't allow those as attributes so they never should be turned to strings.

dginev:
@brucemiller Oh nice, master is silent indeed - you must have fixed/changed something! Cool!

You can reproduce exactly what this PR reports using commit fa8191e757fcac70716db9f532346496a4c8a449 (which is the commit of #2465 , this PR was made using that latexml checkpoint)

brucemiller:
Oh! yeah (duh); that was exactly the problem: the value was being turned to string (and complained about) *before* verifying that the value was even going to be used.

dginev:
Let's consider this resolved then - thank you!

PR #2484, opened 2025-01-10
[minor] update to iftex 2022/02/03

dginev:
As prompted by the recent iftex.sty github discussion, updating with the latest no-op macros from that package.


brucemiller:
Seems safe; Thanks!
I almost wonder whether the various `\RequireXXX` ought to at least issue a warning?

PR #2485, opened 2025-01-11
Pgfmathparse

### LINKED ISSUE #2482: \pgfmathparse calls every function *twice*
Tweak any PGF function to have a side effect, for example edit pgfmath.code.tex.ltxml as follows:
```latex
    e          => sub { NoteSTDERR('evaluating `e`'); $E; },
```
Then run the innocuous
```latex
\pgfmathparse{e}\pgfmathresult
```
LaTeXML will print 'evaluating `e`' twice!

Aside from the performance impact, which is probably negligible, this is making it impossible to replicate PGF's random number generator because `rand` gets called twice every time.

xworld21:
Fix #2482 and lots of other small issues with the parser.

My intention was to simply 'linearise' the grammar, i.e., make sure it never backtracks when parsing the rules. This should improve performance quite a bit (the parser is now guaranteed to run in linear time). I am wondering how much of that can be done for the maths grammar, which has dramatic performance issues.

However, while writing some tests, I discovered that the parser is very broken. So I fixed the easy stuff (wrong functions, string handling) and added a new test to exercise the basic functionality (right now, just the functions).

There are still major issues, e.g. you can't even evaluate `bin(185)` without LaTeXML claiming an overflow. Strings should become tokens in case they contain unexpanded primitives. The blacklist is very fragile. Maybe the LaTeXML parser should fall back to the PGF one when it fails.


dginev:
Thank you for looking into this! Could you also add the new test files to MANIFEST, so that we get CI working?

The one comment I will leave for archival is that - yes, indeed - the perl handling for pgf math has been all of fragile, buggy and incomplete. But importantly - so has the raw interpretataion of the TeX sources of that package. So incremental progress here is good (and I am especially grateful you took the time to add so many new tests).

xworld21:
The math parser is now slightly less broken, but still very broken. The operator precedences are all slightly wrong – they should have been set reading the `\pgfmathdeclareoperator` commands in the sources.

Still wrong: the factorial operator binds higher than multiplication, but lower than power; the radians 'unit' is between addition and multiplication. Hopefully I got the rest right.

@dginev are there notable examples from arXiv that might need all of this?

dginev:
> @dginev are there notable examples from arXiv that might need all of this?

Oh yes, certainly, pgf is a very commonly used package. Looking at the [loaded files](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/info/loaded%5Ffile?all=false) report for ar5iv, pgfmathparser.code.tex gets loaded in 436,976 articles.

If you want to fish for examples, I remember from #2237 that most/all pgfplots use the math parser, and there are ~59,000 articles using that library. That previous PR patch was based on arXiv:2104.00602, and you can a few other reported pgf-related issues at the ar5iv tracker. But naturally arXiv readers do not distill the problems down to the unit test level. It is possible that converting the pgf/tikz [showcase sites](https://tikz.dev/) are a better way to fish out relevant examples.

But yes, this is very much a sizeable need in arXiv.

lib/LaTeXML/Package/pgfmath.code.tex.ltxml
```diff
@@ -646,87 +651,133 @@ BEGIN {
...
-    | FUNCTION simplefactor
-           { LaTeXML::Package::Pool::pgfmath_apply($item[1], $item[2]); }
-    | FUNCTION0                   { LaTeXML::Package::Pool::pgfmath_apply($item[1]); }
-    | NUMBER UNIT          { LaTeXML::Package::Pool::pgfmath_convert($item[1],$item[2]); }
-    | NUMBER REGISTER      { LaTeXML::Package::Pool::pgfmath_apply('*', $item[1], $item[2]); }
+
+<skip:'[\s\{\}]*'>            # braces ignored during parse...
+
+formula : predicate conditional[$item{predicate}](?)
+  { $item[2][0] // $item{predicate}; }
```
xworld21, line 676: Perl 5.28-5.34 interprets this line *and the following ones until comparison* as part of the action, hence the failures.

xworld21: The problem seems to be caused by `//`, I replaced it with `defined ? :`.

xworld21:
Uhm, the `<pagination role="newpage"></pagination>` error changes with TeX Live version. Is there a standard solution for that particular problem?

dginev:
@xworld21 OK, tried on a server with an old texlive, it looks like there is an extra `\clearpage` somehow added to the end of even the smallest pgf-loading document:

```tex
\documentclass{article}
\usepackage{pgf}
\begin{document}\end{document}
```

What you can do for the CI test is to add `\let\clearpage\relax` after `\usepackage{pgf}`, as I assume you won't be needing that feature in this particular math-oriented setup. That works in my setup.

xworld21:
> What you can do for the CI test is to add `\let\clearpage\relax` after `\usepackage{pgf}`

Done!

In the meanwhile, I changed my mind. I can definitely fix the radians operator at least, and implement gcd. I'll do more tweaks and throw in some long expressions without parentheses in the tests, to catch other precedence errors.

brucemiller:
Interestingly, our raw TeX is gradually good enough that using *only* the raw `pgfmath.code.tex` actually works (except for some shading errors), but it is sooooo slow.

xworld21:
By the way, this should be draft as well – I discovered that it breaks pgfplots somehow! Maybe because of `ExpandedPartially`? I don't know right now, I need to test it more carefully and squash the remaining bugs before calling it done.

xworld21:
I am afraid I didn't take note of *which* PGF plot was broken by this PR. I tried a few examples after rebasing to master and it looks fine. This needs to be tested in the wild.

xworld21:
> However, the [dominoes test](https://texample.net/dominoes/) which is mentioned in the issue regresses - the arrangement seems to flip around. I half-remember this being an old problem before #2237 got merged.

Ok, the problem comes from the line
```latex
\pgfpoint{sin(\yy)*(\xx)}{-((\xx)/75)^2+(\zz)/100*(\xx)}
```
Switching to 
```latex
\pgfpoint{sin(\yy)*(\xx)}{-1*((\xx)/75)^2+(\zz)/100*(\xx)}
```
is fine. I guess the unary `-` is not associating properly and it becomes part of the left operand of `^`. Probably.

xworld21:
The dominoes look good now. Ideally, one should really rewrite the grammar to follow pgfmathparser.code.tex more closely rather than patching the existing grammar, but let's shelve that for later.

Further note: I am fairly convinced that back in March, I pasted a pgfplots example *without* adding a necessary usepackage at the beginning, which made me think that the new parser was breaking things. False alarm, most likely!

dginev:
I have started testing this PR (with some of the others tucked in) on our arXiv sandboxes. 

There seems to be a widespread regression (Fatals with 100 errors) due to no longer parsing some pgfmath epxressions.

As a first example, arXiv:1511.02101 has errors of the kind:
```
Error:pgfparse:pgfparse Parse of '70.40544pt-\pgfdecoratedcompleteddistance-1sp' failed
	at inclusionvcd.tex; line 399 col 0 - line 399 col 178
	LTX: '70.40544'
	Left: -\pgfdecoratedcompleteddistance-1sp
	Next token is T_CS[\advance] ( == Core::Definition::Primitive[\advance Variable SkipKeyword:by])
	In Core::Definition::Primitive[\pgfmaths... 
```

I will try to add more details here later today.

xworld21:
> So, to my limited expectations, this PR appears to be an improvement and is good to merge.

Let me add: there are more inconsistencies. For instance, PGF parses `2^3^2` as `(2^3)^2` but `2^-3^2` as `2^(-3^2)`, and that is not accounted for yet here. However, I am in favour of merging as is because it should be a substantial net improvement, hopefully with negligible regressions. We can add tests and improvements later, and eventually replace the grammar with something that tracks pgfmathparser almost line by line.

By the way, should we merge the new test with the stress_pgfmath, that I somehow did not notice?

dginev:
> By the way, should we merge the new test with the stress_pgfmath, that I somehow did not notice?

Oh, sure. I think reusing that simplifies admin a little, they are exercising the same functionality.

dginev:
With the latest fixes we are closer to (but still a bit weaker than) the tikz-cd arXiv sandbox. That has just under 20,000 articles and at time of writing goes from 8.98% Fatal to 9.99% Fatal articles, as well as from 33.32% Error to 39.97% Error articles.

Also somewhat worryingly the conversions appear to be slower than before (anecdotally for now). That said, the changeset I am testing includes more than only this PR, so likely not all causes are rooted here. I will investigate some more today. 

Edit: Some of the issues were indeed elsewhere, but there are still a number of parsing errors that should be valid parses. I will collect them during Q1 of 2026 (hopefully a promise I can keep).

PR #2489, opened 2025-01-18
minor: \bibsep is a glue register

dginev:
Very minimal, fixes  https://github.com/arXiv/html_feedback/issues/2651

The scarier question is "how do we catch all of these DefRegister subtleties in general, in all bindings?"

Minimal snippet:
```tex
\documentclass{article}
\usepackage{natbib}
\begin{document}
\bibsep.9ex plus.1ex minus.05ex
\end{document}
```


brucemiller:
Thanks! (I can't answer your scary question)

Issue #342, opened 2013-01-05
Support LaTeX-internal utility macros


dginev:
[Originally Ticket 1666]

I am incrementally testing the native interpretation of the XY package in LaTeXML. Currently, the simplest trivial xy drawings convert well (modulo box arithmetic), but as things get more even slightly elaborate, I am running into errors concerning undefined LaTeX utility macros.

For the moment, I have stumbled on errors of not having the \z@ and \m@ne macros. I dug around online and found a really nice [project](https://bitbucket.org/martin_scharrer/macros2e) that documents the 2e macros that would be useful to authors. [PDF](https://bitbucket.org/martin_scharrer/macros2e/downloads/macros2e.pdf)

So, I will try to gradually add support for what seems needed by XY on the branch and we can use the ticket for discussing a general strategy. As a footnote, we should really take care of the Box magic needed by Tikz and XY.

PR #2495, opened 2025-01-20
Cases

brucemiller:
This PR makes `\MakeUppercase,\MakeLowercase` more consistent with current LaTeX implementation, particularly as to using Unicode mappings rather than `\uccode,\lccode`. The mechanisms used for defining mappings and exclusion rely on control sequences being robust (being `protected` and redirecting through a cs with space appended), so a number of commands have been updated.
In the process a couple of other patches: Definition's alias is now consistently a `Token`, rather than string; reduced unnecessary debugging noise, warnings, etc.


lib/LaTeXML/Engine/plain.pool.ltxml
```diff
@@ -159,8 +159,6 @@ RawTeX(<<'EoTeX');
   \mathchardef\@m=1000
   \mathchardef\@M=10000
   \mathchardef\@MM=20000
-  \countdef\m@ne=21\relax
-  \m@ne=-1
```
dginev, line 163: Why do we lose `\m@ne` ? I have a very vague old memory of stumbling on an error with it a long time ago, as it turns out in #342 .

lib/LaTeXML/Package.pm
```diff
@@ -164,6 +164,8 @@ sub coerceCS {
   if    (ref $cs) { }
   elsif ($cs =~ s/^\\csname\s+(.*)\\endcsname//) {
     $cs = T_CS('\\' . $1); }
+  elsif ($cs =~ s/^(.)$//) {    # Match an active char
+    ($cs) = TokenizeInternal($1)->unlist; }
```
dginev, line 168: [minor, optional] I would have probably checked `length($cs) == 1` to avoid the regex hoop, but it works either way.

lib/LaTeXML/Package.pm
```diff
@@ -639,9 +642,13 @@ sub NewCounter {
     NewCounter($within); }
   my $cs       = T_CS("\\c\@$ctr");
   my $prevdefn = $STATE->lookupMeaning($cs);
-  if ($prevdefn && ((ref $prevdefn) eq 'LaTeXML::Core::Definition::Register')
-    && (($$prevdefn{address} || '') =~ /^\\count/)) {
-    Info('unexpected', $cs, undef, "Counter $ctr was already allocated, skipping"); }
+  if ($prevdefn && ((ref $prevdefn) eq 'LaTeXML::Core::Definition::Register')) {
+    my $a = ($$prevdefn{address} || '') =~ /^\\count/;
```
dginev, line 646: `$a` can get commented out too here.

dginev (approved):
Impressive changeset. I read through the full PR, left some minor comments.

lib/LaTeXML/Core/Definition.pm
```diff
@@ -39,13 +39,12 @@ sub getCS {
 
 sub getCSName {
   my ($self) = @_;
-  return (defined $$self{alias} ? $$self{alias} : $$self{cs}->getCSName); }
+  return ($$self{alias} || $$self{cs})->getCSName; }
 
-# NOTE: Need to clean up alias; really should already be Token (or Tokens?)
-# and is not always a CS!
+# NOTE:  alias should be Token (or Tokens?)
 sub getCSorAlias {
```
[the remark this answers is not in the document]
brucemiller: Meaning it should have been `getAliasorCS`? Perhaps you're right. It's only used in a couple of places, so we could change it, without too much danger. OTOH, starting with "getCS" seems a good thing. `getCSName` is documented, btw, but who reads PODs. :>
I suspect that `getCS` has to return the name that the definition was made under, and not the alias.

lib/LaTeXML/Core/Document.pm
```diff
@@ -1376,41 +1376,45 @@ sub makeError {
...
-        my $prefix = $node->lookupNamespacePrefix($ns);    # namespace already declared?
-        if (!$prefix) {                                    # if namespace not already declared
-          $prefix = $$self{model}->getDocumentNamespacePrefix($ns, 1);             # get the prefix to use
-          getDocument($self)->documentElement->setNamespace($ns, $prefix, 0); }    # and declare it
-        if ($prefix eq '#default') {    # Probably shouldn't happen...?
+    if ($key eq 'xml:id') {                       # If it's an ID attribute
+      $value = recordID($self, $value, $node);    # Do id book keeping
+      ## and bypass all ns stuff
+      $node->setAttributeNS($LaTeXML::Common::XML::XML_NS, 'id', $value); }
+    elsif ($key =~ /:/) {                         # ANY namespaced attribute
```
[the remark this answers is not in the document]
brucemiller: I thought about that, but the code clarity suffered so much...

lib/LaTeXML/Core/Document.pm
```diff
@@ -1376,41 +1376,45 @@ sub makeError {
...
+  elsif ($key eq '_font') {
+    return $self->setNodeFont($node, $value); }
+  ## Next, verify the attribute allowed by Model, else internal or namespaced
+  elsif (($key =~ /:/) || ($key =~ /^_/)
+    || $model->canHaveAttribute($model->getNodeQName($node), $key)) {
+    ## OK, we're going to use the value, so make sure it's a string.
+    if (ref $value) {
+      if ((!blessed($value)) || !$value->can('toAttribute')) {
+        Warn('unexpected', (ref $value), $self,
+          "Don't know how to encode $value as an attribute value for $key");
```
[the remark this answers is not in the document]
brucemiller: Good suggestion! I'll incorporate that into the next commit.

lib/LaTeXML/Engine/LaTeX.pool.ltxml
```diff
@@ -2916,47 +2915,26 @@ DefPrimitiveI('\tenlnw',   undef, undef, font => { family => 'linew10' });
...
-DefMacroI('\ss', undef, '\lx@utf@ss');
-DefMacroI('\dh', undef, '\lx@utf@dh');    # in latex?
-DefMacroI('\DH', undef, '\lx@utf@DH');
-DefMacroI('\dj', undef, '\lx@utf@dj');
-DefMacroI('\DJ', undef, '\lx@utf@DJ');
-DefMacroI('\ng', undef, '\lx@utf@ng');
-DefMacroI('\NG', undef, '\lx@utf@NG');
-DefMacroI('\th', undef, '\lx@utf@th');
-DefMacroI('\TH', undef, '\lx@utf@TH');
+# At least all things on uclclist need to be macros; robust achieves that
```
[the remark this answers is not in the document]
brucemiller: See the magic in Package's `defRobustCS`.

lib/LaTeXML/Engine/plain.pool.ltxml
brucemiller: Nope, we just lost the redefinition, which allocated it in the wrong slot, besides!

lib/LaTeXML/Package.pm
brucemiller: OK, thanks.

lib/LaTeXML/Package.pm
brucemiller: Thanks!

lib/LaTeXML/Core/Definition.pm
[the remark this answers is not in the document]
dginev: `getAliasorCS` yes in terms of order, but naturally that reads weird due to the CamelCase convention. A small editor tweak would make it `getAliasedCS`, which reads well and implies the right priority.

Whether that fits better than `getCS` - up to you - but I think either would be an improvement.

PR #2498, opened 2025-01-25
omit \lxSVG@closescope in reversion

### LINKED ISSUE #2476: \lxSVG@closescope leaks into reversion, breaks --pictureimages
As per subject: very simple TikZ figures will break --pictureimages because `\lxSVG@closescope` does not disappear from the reversion. I am not planning to use --pictureimages, I am just trying to understand how pgfsys-latexml.def is supposed to be written. I gather that the intention is for the `@tex` tag to rebuild an equivalent output.

xworld21:
Fix #2476. It doesn't fix `--pictureimages` in general: it only makes it compile *something* successfully. Unfortunately, the output is cropped the wrong way, but that's a different issue altogether.


dginev:
Sure. I guess we can motivate the empty reversion with `\lxSVG@begingroup@` having an empty reversion.

dginev (approved):

PR #2502, opened 2025-01-26
ensure that \kern inside an SVG only affects the following content

xworld21:
When `\kern` happens inside an SVG group, the current code shifts the entire group, which goes wrong as soon as `\kern` appears *after* the content (hence `HACK HACK HACK` in the comments, I guess). I changed it so that it creates an extra `<svg:g>` around the previous content and translates it the opposite way before applying the new transform.

In practice, this fixes the horizontal alignment of arrow labels in xymatrix, because they use `\kern` before *and* after the label: we go from
image
to
image

Vertical alignment is obviously off; it becomes pretty good with #2488, but still visibly wrong due to unrelated depth shenanigans.

### DIFF
--- a/lib/LaTeXML/Engine/TeX_Kern.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Kern.pool.ltxml
@@ -27,16 +27,40 @@ use LaTeXML::Package;
 # \lastkern         iq is 0.0 pt or the last kern on the current list.
 
 # \kern is heavily used by xy.
-# Completely HACK version for the moment
 # Note that \kern should add vertical spacing in vertical modes!
 DefConstructor('\kern Dimension', sub {
     my ($document, $length, %props) = @_;
     my $parent = $document->getNode;
     if ($document->getNodeQName($parent) eq 'svg:g') {
       if (my $x = $length->pxValue) {
-        # HACK HACK HACK
         my $transform = $parent->getAttribute('transform');
         $parent->setAttribute(transform => ($transform ? $transform . ' ' : '') . "translate($x,0)");
+        if (my @nodes = $parent->childNodes) {
+          # nodes have already been deposited, wrap those in svg:g to avoid shifting them
+          # the code below is essentially Document::wrapNodes, but we add the transform attribute
+          # *before* running afterClose, or collapseSVGGroup will immediately remove the wrapper
+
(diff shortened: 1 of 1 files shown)


dginev:
Workflow question: Do you have a branch with all/most of your tikz-related PRs merged? I am curious to test them in tandem. Such a branch will inevitably be a diverged one from master (as we'd like to reintegrated PR-by-PR), but could make other upgrades easier to test while the process of merging converges...

xworld21:
> Do you have a branch with all/most of your tikz-related PRs merged?

No, they are essentially orthogonal to each other, but I might as well make one for testing purposes. Even if the merged branch diverges from the smaller PRs, it should still be a close approximation. I just need a moment to rebase...

lib/LaTeXML/Engine/TeX_Kern.pool.ltxml
```diff
@@ -27,16 +27,40 @@ use LaTeXML::Package;
...
+          $document->afterOpen($new);
+          $parent->replaceChild($new, $nodes[0]);
+
+          if (my $font = $document->getNodeFont($parent)) {
+            $document->setNodeFont($new, $font); }
+          if (my $box = $document->getNodeBox($parent)) {
+            $document->setNodeBox($new, $box); }
+          foreach my $node (@nodes) {
+            $new->appendChild($node); }
+          $document->afterClose($new) unless $leave_open; }
```
brucemiller, line 63: So, if I understand the problem, if there are already children in the current `svg:g`, you don't want to modify the transform of *that* `g`, but what you really want to do is simply open a *new* one; the following children will go there and be translated, right?  You just need `_autoclose=1`, so that that new group will be closed appropriately.

However, if you *had* tried that, it would have failed, since there's some mixup with `svg:g` being opened and closed wrongly. I think I've sorted that out and can make an alternative PR, or try to explain where some other changes are needed (`TeX_Box` and `pgfsys-latexml`).

xworld21: > if you _had_ tried that, it would have failed

Yep, that would have been my preferred approach, but nodes would just disappear. Might have gain an extra gray hair in the process.

> I think I've sorted that out and can make an alternative PR

By all means, if you know how to tweak `svg:g` so that it closes at the right time, please do!

PR #2514, opened 2025-02-10
Fold 2 calls to kpsewhich; avoid regexp in tightest loop

brucemiller:
Small optimization to `build_kpse_cache`; seemingly worthwhile since it's invoked on every run.


brucemiller:
Maybe by a half for simple cases that easily map to `substr`; more for 1st char, less for last.

PR #2367, opened 2024-05-25
convert ltx:graphics candidates to file URIs

xworld21:
Fix #2355 (allegedly) by making `candidates` a comma-separated list of URIs instead of plain file names.

I do not have conclusive proof about Windows, because the relevant tests seem to require a minimum texlive version, and I am using MikTeX. This would explain why CI did not catch the issue.

I suggest taking this PR further: make `candidates` a *space* separated list (since now spaces are %-encoded), and add this restriction to the schema (with something like `list(anyURI)`).

PR #2519, opened 2025-02-23
use URI::file for (de)serialising paths in XML

### LINKED ISSUE #2355: Install Fails on Windows 10 via "cpan LaTeXML"
# Issue
The installation of LaTeXML fails when attempting to install it via the `cpan LaTeXML` instruction. Specifically it fails during the make test phase of installation. This is performed with a fresh intsall of Strawberry Perl.

# Details

The log files indicates that the issue is with the `structure.t` and `complex.t` tests. Examining those specific entries, we find:

```
t/50_structure.t .......... 26/44 #   Failed test 't/structure/figure_grids'
#   at C:\STRAWB~1\cpan\build\LaTeXML-0.8.8-1\blib\lib/LaTeXML/Util/Test.pm line 120.
# Difference at line 10 for t/structure/figure_grids
#       got : '    <graphics candidates="..\graphics\none.png" class="ltx_figure_panel" graphic="../graphics/none.png" options="width=213.96474pt,keepaspectratio=true" xml:id="g1"></graphics>'
#  expected : '    <graphics candidates="../graphics/none.png" class="ltx_figure_panel" graphic="../graphics/none.png" options="width=213.96474pt,keepaspectratio=true" xml:id="g1"></graphics>'
```

and

```
t/80_complex.t ............ 6/15 #   Failed test 't/complex/figure_dual_caption'
#   at C:\STRAWB~1\cpan\build\LaTeXML-0.8.8-1\blib\lib/LaTeXML/Util/Test.pm line 120.
# Difference at line 14 for t/complex/figure_dual_caption
#       got : '      <graphics candidates="..\graphics\none.png" graphic="../graphics/none.png" options="width=411.93767pt,keepaspectratio=true" xml:id="S0.F2.g1"></graphics>'
#  expected : '      <graphics candidates="../graphics/none.png" graphic="../graphics/none.png" options="width=411.93767pt,keepaspectratio=true" xml:id="S0.F2.g1"></graphics>'

t/80_complex.t ............ 7/15 #   Failed test 't/complex/figure_mixed_content'
#   at C:\STRAWB~1\cpan\build\LaTeXML-0.8.8-1\blib\lib/LaTeXML/Util/Test.pm line 120.
# Difference at line 81 for t/complex/figure_mixed_content
#       got : '      <graphics candidates="..\graphics\none.png" graphic="../graphics/none.png" options="width=85.35826pt,keepaspectratio=true" xml:id="S0.F1.sf1.g1"></graphics>'
#  expected : '      <graphics candidates="../graphics/none.png" graphic="../graphics/none.png" options="width=85.35826pt,keepaspectratio=true" xml:id="S0.F1.sf1.g1"></graphics>'
```

Comparing the test results to the expected values, it is clear that the build is either producing slashes in the wrong direction or the expected direction of the slashes is incorrect. I suspect that this is likely an issue with Windows and Linux using different slash directions to denote paths.

# Log
```
Running make test for BRMILLER/LaTeXML-0.8.8.tar.gz
"C:\Strawberry\perl\bin\perl.exe" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib\lib', 'blib\arch')" t/*.t
t/000_test_context.t ...... #
t/000_test_context.t ...... 1/1 #   LaTeXML version 0.8.8
#   XML::LibXML 2.0209; libxml2 21004
#   XML::LibXSLT 2.002001; libxslt 10137
#   TeX 3.141592653 (TeX Live 2024)
t/000_test_context.t ...... ok
t/002_unit_findfile.t ..... ok
t/003_unit_imagemagick.t .. 1/? # Skip: This unit test only examines basic Image::Magick conversion, when installed.
t/003_unit_imagemagick.t .. ok
t/00_unittest.t ........... ok
t/05_tokenize.t ........... ok
t/10_expansion.t .......... ok
t/12_grouping.t ........... ok
t/170_grammar_coverage.t .. skipped: Only checked in continuous integration. (use make test CI=true)
t/20_digestion.t .......... ok
t/22_fonts.t .............. ok
t/30_encoding.t ........... ok
t/32_keyval.t ............. ok
t/33_keyval_options.t ..... ok
t/40_math.t ............... ok
t/50_structure.t .......... 25/44
t/50_structure.t .......... 26/44 #   Failed test 't/structure/figure_grids'
#   at C:\STRAWB~1\cpan\build\LaTeXML-0.8.8-1\blib\lib/LaTeXML/Util/Test.pm line 120.
# Difference at line 10 for t/structure/figure_grids
#       got : '    <graphics candidates="..\graphics\none.png" class="ltx_figure_panel" graphic="../graphics/none.png" options="width=213.96474pt,keepaspectratio=true" xml:id="g1"></graphics>'
#  expected : '    <graphics candidates="../graphics/none.png" class="ltx_figure_panel" graphic="../graphics/none.png" options="width=213.96474pt,keepaspectratio=true" xml:id="g1"></graphics>'
t/50_structure.t .......... 30/44 # Skip: t/structure/glossary is only checked in continuous integration. (use make test CI=true)
t/50_structure.t .......... 43/44 # Looks like you failed 1 test of 44.
t/50_structure.t .......... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/44 subtests
        (less 2 skipped subtests: 41 okay)
t/52_namespace.t .......... ok
t/53_alignment.t .......... ok
t/55_theorem.t ............ ok
t/56_ams.t ................ ok
t/65_graphics.t ........... ok
t/70_parse.t .............. ok
t/80_complex.t ............ 5/15
t/80_complex.t ............ 6/15 #   Failed test 't/complex/figure_dual_caption'
#   at C:\STRAWB~1\cpan\build\LaTeXML-0.8.8-1\blib\lib/LaTeXML/Util/Test.pm line 120.
# Difference at line 14 for t/complex/figure_dual_caption
#       got : '      <graphics candidates="..\graphics\none.png" graphic="../graphics/none.png" options="width=411.93767pt,keepaspectratio=true" xml:id="S0.F2.g1"></graphics>'
#  expected : '      <graphics candidates="../graphics/none.png" graphic="../graphics/none.png" options="width=411.93767pt,keepaspectratio=true" xml:id="S0.F2.g1"></graphics>'

t/80_complex.t ............ 7/15 #   Failed test 't/complex/figure_mixed_content'
#   at C:\STRAWB~1\cpan\build\LaTeXML-0.8.8-1\blib\lib/LaTeXML/Util/Test.pm line 120.
# Difference at line 81 for t/complex/figure_mixed_content
#       got : '      <graphics candidates="..\graphics\none.png" graphic="../graphics/none.png" options="width=85.35826pt,keepaspectratio=true" xml:id="S0.F1.sf1.g1"></graphics>'
#  expected : '      <graphics candidates="../graphics/none.png" graphic="../graphics/none.png" options="width=85.35826pt,keepaspectratio=true" xml:id="S0.F1.sf1.g1"></graphics>'
t/80_complex.t ............ 11/15 # Skip: t/complex/si is only checked in continuous integration. (use make test CI=true)
t/80_complex.t ............ 15/15 # Looks like you failed 2 tests of 15.
t/80_complex.t ............ Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/15 subtests
        (less 1 skipped subtest: 12 okay)
t/81_babel.t .............. ok
t/82_moderncv.t ........... ok
t/83_expl3.t .............. 1/3 # Skip: t/expl3/tilde_tricks is only checked in continuous integration. (use make test CI=true)
t/83_expl3.t .............. 2/3 # Skip: t/expl3/xparse is only checked in continuous integration. (use make test CI=true)
t/83_expl3.t .............. ok
t/84_slides.t ............. ok
t/90_latexmlpost.t ........ ok
t/91_latexmlc_api.t ....... ok
t/92_profiles.t ........... ok
t/931_epub.t .............. ok
t/93_formats.t ............ ok
t/94_runtimes.t ........... ok
t/95_complex_config.t ..... ok
t/96_fatal.t .............. ok
t/97_manifest.t ........... skipped: Only checked in continuous integration. (use make test CI=true)

Test Summary Report
-------------------
t/50_structure.t        (Wstat: 256 (exited 1) Tests: 44 Failed: 1)
  Failed test:  26
  Non-zero exit status: 1
t/80_complex.t          (Wstat: 512 (exited 2) Tests: 15 Failed: 2)
  Failed tests:  6-7
  Non-zero exit status: 2
Files=35, Tests=452, 991 wallclock secs ( 0.09 usr +  0.06 sys =  0.16 CPU)
Result: FAIL
Failed 2/35 test programs. 3/452 subtests failed.
gmake: *** [makefile:1669: test_dynamic] Error 255
  BRMILLER/LaTeXML-0.8.8.tar.gz
  C:\STRAWB~1\c\bin\gmake.exe test -- NOT OK
//hint// to see the cpan-testers results for installing this module, try:
  reports BRMILLER/LaTeXML-0.8.8.tar.gz
Stopping: 'install' failed for 'LaTeXML'.
```

xworld21:
Use URI::file for serialising and deserialising paths written to XML for `@graphic`, `@candidates`, and the processing instructions 'searchpaths' and 'graphicspath'. It also replaces previous manual uses of URI::file.

This prevents potential issues regarding:
- OS-specific path separators: backslashes on Win32 will never leak into XML
- absolute paths: URI::file will add `file://` when the paths is absolute, including on Windows for paths starting with e.g. `C:` (or longpaths, but LaTeXML does not support those _yet_)
- spaces, commas, Unicode, etc: URLs are automatically encoded and decoded
- lists of URLs: multiple URLs are now space separated, which is safe, since spaces within URLs will be percent-encoded

Note that I have replaced the previous implementation of `pathname_to_url` as it didn't make sense to have two. I have adjusted CrossRef accordingly.

This supersedes #2367 and fixes #2355 (well, I haven't tried it on a Windows machine yet, but I am sure it does).


t/complex/candidates.tex
```diff
@@ -0,0 +1,8 @@
+\documentclass{article}
+\usepackage{graphicx}
+\title{Multiple candidate test}
+\begin{document}
+
+\includegraphics{sunset spaced}
```
xworld21, line 6: By the way: the commit hooks break on file with spaces.

lib/LaTeXML/Core.pm
```diff
@@ -212,7 +212,7 @@ sub convertDocument {
           my @dedup = ();
           while (my $check = shift(@copy)) {
             unshift(@dedup, $check) if !(grep { $_ eq $check } @dedup); }
-          $document->insertPI('latexml', searchpaths => join(',', @dedup)); } }
```
dginev, line 215: > lists of URLs: multiple URLs are now space separated, which is safe, since spaces within URLs will be percent-encoded

Could you share the motivation for dispencing with the comma? That was safe too I think.

There may be code in various places external to latexml expecting the current separator char (since they get serialized out), so this will be a backwards-incompatible change.

xworld21: > Could you share the motivation for dispencing with the comma? That was safe too I think.

Not if the file name contains a comma!

dginev: Er, a URI-encoded comma is %2C right?

xworld21: > this will be a backwards-incompatible change.

True, this would have to be documented. On the bright side, using file URLs means that serialisation is a known quantity (e.g. libexslt implements `str:decode-uri` and `str:encode-uri`).

xworld21: > Er, a URI-encoded comma is %2C right?

Good point, but, I just checked and URI::file does not encode the comma (it's not a special character after all).

t/complex/candidates.tex
xworld21: And CI too!

lib/LaTeXML/Core.pm
dginev: Right, but I feel even less comfortable knowing that now. 

I'd expect either no attempt at encoding, or a complete URI-encoding for methods that claim `...to_url` (or similar). I've used [URI::Encode](https://metacpan.org/pod/URI::Encode) for that in the past.

I think the master version of the method you modified was mainly interested in using `/` separator chars for the paths and did no encoding. 

We certainly didn't support all OS-legal filenames, that's definitely a good critique. But I am actually aware of two projects I need to go and refactor from splitting on commas to splitting on space, and then also add a URI-decode step, which is painful. I wonder who else is relying on the current convention, and how quietly this change will lead to breakage for people...

xworld21: (Absolute paths will keep starting with file:// though.)

xworld21: I changed `pathname_to_url` so that the comma is also encoded, and the file:// part is omitted. That preserves the previous behaviour on Unix-like systems except for some special characters getting encoded. I believe the behaviour is correct also w.r.t. Win32 UNCs, but that's harder to test properly.

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -211,13 +213,34 @@ sub pathname_absolute {
...
+
+sub pathname_from_url {
+  return unless defined $_[0];
+  return URI->new($_[0], 'file')->file; }
+
+sub pathname_from_urls {
+  return unless defined $_[0];
+  my @urls     = split(/,/, $_[0]);
+  my @nonempty = grep { $_ } @urls;
+  return map { pathname_from_url($_) } @nonempty; }
```
dginev, line 236: Apologies in advance for the naming nit-picking.

`pathname_from_urls` probably wants to indicate the many-to-many map. So an extra `s` - 
 `pathnames_from_urls`

Similarly for the `_to_` variant.

Also, we probably want to decide if we want to drop the `_file` from `_file_url`, or the opposite - to use it everywhere. As indicated, a URL is ambiguous. Indeed a relative pathname on linux such as `./subdir/a/b/c/test.png` is also a match for a relative URL.

So maybe all of your new methods should be `fileurl`/`fileurls` as suffixes... But maybe don't rush to take my feedback before we hear from @brucemiller .

xworld21: > probably wants to indicate the many-to-many map. So an extra `s` -
> `pathnames_from_urls`

Makes sense. I was just uncertain as to whether `pathname_` was meant as a namespace as opposed to an actual word.

xworld21: Looking back at (and rebasing) this PR, a summary of why the PR looks like this:
- the URL without `file:` must remain for backward compatibility with external tools that do not expect it
- the `file:` scheme can be omitted because those 'urls' are used in contexts where `file:` is implied (e.g. graphics candidates)
- `pathname_to_url(s?)` is the version without `file:` scheme, for backward compatibility with the previous `pathname_to_url` (which was never about producing complete well formed URLs!)
- `pathname_to_file_url` exists only for those rare situations in which the `file:` scheme is actually required (DTD and document URI)

So the naming problem really comes from the original `pathname_to_url` not producing URLs in the first place. We could rename `pathname_to_url` to `pathname_urlencode`, which is more accurate.

PR #2526, opened 2025-03-06
Have math rewrite patterns NOT distinguish between text and display style

brucemiller:
That is, `\lxDeclare[...]{$\sum_x$}` shouldn't match the pattern whether it is in textstyle or displaystyle. Seems safe and useful.


dginev (approved):
So a rewrite will match e.g. a summation whether its limits are positioned as scripts, or under/over it? Indeed, that level of precision is likely too detailed to be useful.

Good to merge.

PR #2527, opened 2025-03-11
Unset box

### LINKED ISSUE #2412: pgf layers do not get cleared

brucemiller:
Unlike how TeX treats other registers, when `\box` is used, TeX discards the box data, for efficiency. Previously, LaTeXML simply assigned the box to `undef` but that assignment took place in the current group; but what TeX is doing is assigning it "in place", that is, in the group or frame where it was last stored. This PR corrects the behavior, and improves the test case.

This PR is an alternative to #2480.
Fixes #2412 

Incidentally, a side issue was raised within #2412 regarding color changes propagating backwards into stored boxes. I think that issue is essentially a duplicate of #2455; that issue is not addressed here.


lib/LaTeXML/Core/State.pm
```diff
@@ -169,6 +169,12 @@ sub assign_internal {
     else {                                    # Otherwise, push new value & set 1 to be undone
       $$self{undo}[0]{$table}{$key} = 1;
       unshift(@{ $$self{$table}{$key} }, $value); } }    # And push new binding.
+  elsif ($scope eq 'inplace') {                          # Special case for \box & friends
+    if (exists $$self{$table}{$key}) {        # If the value was previously assigned AT ALL
+      $$self{$table}{$key}[0] = $value; }     # Simply replace the value in its frame
+    else {                                    # Otherwise, push new value, like local
+      $$self{undo}[0]{$table}{$key} = 1;
+      unshift(@{ $$self{$table}{$key} }, $value); } }    # And push new binding.
```
dginev, line 177: Why have the `else` at all? If this is called with `undef` on a value that wasn't defined before, it will create a new local definition of `undef`?

How about raising a warning and returning as a no-op instead?

brucemiller: Although the current use case has value undef, the API is straightforward, and may be useful for other values someday.
What problem are you expecting?

dginev: I expect subtle bugs, coming from extending an internal removal detail (voiding a box in place) to a new assignment API.

dginev (commented):
Also something that caught my eye:

As implemented, setting `\globaldefs` will override the `inplace` scope to be `global` or `local`.

That's a rare concern, also since it would have broken all "named" scopes, but we haven't had a bug report encountering that yet.

lib/LaTeXML/Core/State.pm
brucemiller: Sorry, but that's vague enough that it applies to anything, including your suggestions above.

dginev (approved):

brucemiller:
Hmm, `\globaldefs` does get used in texmf; Named scopes was essentially a OmDoc thing, right?
So, perhaps to be safe(r), the `\globaldefs` clause should only kick in when scope is undef, global or local?

dginev:
> So, perhaps to be safe(r), the \globaldefs clause should only kick in when scope is undef, global or local?

That sounds right to me, indeed.

lib/LaTeXML/Core/State.pm
dginev: My suggestion above was to limit this new feature to the use case that motivated it - removing a value in the "special" case for voiding a box in place. I should have made an additional suggestion - to also require that `undef` is passed in and nothing else.

So rather than extend for future use, I would restrict to the motivating use and send a Warning if used otherwise.

dginev (approved):

lib/LaTeXML/Core/State.pm
xworld21: For what it's worth: I agree with @dginev that new APIs should be careful even if there are no plans of using them beyond clearing boxes. Here, how you treat undefined names seems arbitrary compared to the documentation of `inplace`: my intuitive understanding of 'same frame' is that you bubble up until the frame containing the definition, and so `inplace` should become `global` if no such definition has ever existed. I don't really see why `inplace` should become `local`. That should also be closer in spirit to Knuth's implementation (which simply assigns null to the register, and thus would act globally if the register were empty).

In all of this, I am also interpreting 'name is defined' as 'key exists' (thus assigning undef makes the *value* undefined but leaves the name defined), which matches your code here.

brucemiller:
Thanks; good catch!

lib/LaTeXML/Core/State.pm
dginev: Thanks @xworld21 , we'll convince him next time 🤝

brucemiller: Hmm, ok. global makes slightly more sense as a starting point if/when we end up needing the capability. PR coming...

PR #2530, opened 2025-03-12
If assigning 'inplace', and there's a value, assign globally

brucemiller:
Patch following discussion in #2527.  If we assign "inplace", with an actual value (as opposed to clearing), and there is no previous assignment, we'll assign globally. A reasonable default, if we end up ever needing the feature.

### DIFF
--- a/lib/LaTeXML/Core/State.pm
+++ b/lib/LaTeXML/Core/State.pm
@@ -173,8 +173,13 @@ sub assign_internal {
   elsif ($scope eq 'inplace') {                          # Special case for \box & friends
     if (exists $$self{$table}{$key}) {        # If the value was previously assigned AT ALL
       $$self{$table}{$key}[0] = $value; }     # Simply replace the value in its frame
-    else {                                    # Otherwise, push new value, like local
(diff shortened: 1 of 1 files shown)


brucemiller:
For some reason, I can't add @xworld21 to reviewr

dginev:
@brucemiller same, I think the repository settings are too restrictive. In fact I know they are, because I don't have any access to them - you are the only one who can investigate there. 

It's the "Settings" tab at the rightmost top of the repo page.

brucemiller:
Ah, the docs are somewhat hard to parse, but I *think* the problem is that for free accounts, you can only request *one* reviewer. And I haven't discovered a way to UN-request the review from @dginev

dginev:
That's not quite it, since you can't request external reviews on any of the other PRs that don't have a reviewer set either, for example https://github.com/brucemiller/LaTeXML/pull/2364

dginev (approved):
PR code matches its description.

xworld21:
Good that you didn't wait for my review: I don't understand frames at all...

PR #2531, opened 2025-03-12
Setbox adjustments

### LINKED ISSUE #2455: color from surrounding scope overrides colors inside boxes

brucemiller:
In TeX, `\setbox` doesn't record the current color or background in the box it saves, but when the box  is used, it inherits the then current colors. In LaTeXML, we *do* record those colors (perversely, within the font). So, we have to be careful to preset them to defaults before `\setbox` creates the box content. When using the box we should update only those colors which were the inherited defaults. Moreover, we must create a copy of the boxes, rather than modifying them in place, since some commands (eg. `\copy`) still keeps the box in the register.  This PR achieves those aims.

Fixes #2455
And is a better alternative to the approaches in:
Closes #2480
Closes #2456


lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -515,85 +519,75 @@ DefRegister('\dp Number', Dimension(0),
...
+  my $adj = $box;
+  if($type eq 'ARRAY'){
+    $adj = [map { adjustBoxColor_rec($fg, $bg, $cache, $_); } @$box]; }
+  elsif($type eq 'HASH'){
+    $adj = {map { $_=>adjustBoxColor_rec($fg, $bg, $cache, $$box{$_}); } keys %$box}; }
+  # NASTY access to internal structure; but worth a whole API for this one hack???
+  elsif ($type =~ /^LaTeXML::Core::(?:Box|List|Whatsit|Alignment)$/) { # blessed hashes!
+    $$cache{$box} = $adj = bless {%$box}, $type; # Cache BEFORE recursion!
+    map { $$adj{$_} = adjustBoxColor_rec($fg, $bg, $cache, $$box{$_}); } keys %$box; }
+  elsif($type eq 'LaTeXML::Common::Font'){
```
dginev, line 548: A Font shouldn't make it as an argument to a box-oriented subroutine, surely? This makes porting this to typed languages rather painful, but it is also rather confusing to think about. Why is a font object passed in as `$box`?

brucemiller: You're right, bad naming. It's oriented to (digested) objects, not just boxes. Fixed.

dginev: Is a Font object a Digested object?

I do not understand the design here.

brucemiller: A Font object is *inside* of a Digested object, and it is exactly the Font objects inside of the digested objects that we're adjusting.

dginev (approved):
I have read this little function five times now and I still find it confusing. I'll approve the PR, but I am really not looking forward to porting it.

Did you check if the tests from my older PR all continued to pass? I noticed that you didn't carry them over.

brucemiller:
Sure, I did, and more! (there were other tests in your PR, but those were covered in #2527)

brucemiller:
Maybe this improved comment helps a little?

t/graphics/xcolors.tex
```diff
@@ -246,12 +246,19 @@ \section{Color Compararisons}
...
+\setbox\mybox=\hbox{{\color{black}\colorbox{white}{Black on White!}}}
 \copy\mybox
-\setbox\mybox=\hbox{Should be Red!}
-{\color{red}\copy\mybox}
+
+\setbox\mybox=\hbox{Red on Green!}
+{\color{red}\colorbox{green}{\copy\mybox}}
+
+\setbox\mybox=\hbox{Red on Green, but \color{green}\colorbox{blue}{Green on Blue!}}
+{\color{red}\colorbox{green}{\copy\mybox}}
```
xworld21, line 261: I suggest adding a test where the color is set outside of the box, before `\setbox`, and likewise before using `\copy`.

brucemiller: done!

dginev (approved):

PR #2533, opened 2025-03-18
Better \kern in svg

brucemiller:
Currently a `\kern` in svg adds the shift to the containing `svg:g`, but if there already were children, they get shifted as well. We really want to create a *new* `svg:g` in that case. Alas, there was clumsiness in other places dealing with `svg:g`, namely `\hbox` would wrap it's content in `svg:g`, but didn't actually close the correct container when done, so that got simplified as well. And finally, a few `svg:g` that got created by the driver, behind the scenes, needed to be marked as `autoclose`, so that they could be properly closed.

Thanks, @xworld21 for the excellent debugging, but I think this ends up being simpler, in the long run, than #2502.
check if this does what you expect.

Closes #2502


dginev:
Since reviewing `_autoclose` effects is nearly impossible without context, should we add a test exercising this? Maybe in `t/tikz` which runs only with the "extra" test target.

brucemiller:
It should show up in the `xy` test, but that doesn't seem to be there. There should be a simple version of it, though

dginev:
The reason we have them in the "extra" tests is that we can't reproduce the exact same XML markup across texlives (as the internal `tex` attribute values change with the changes to pgf). So `t/graphics` isn't quite doable.

dginev (approved):
Thank you for the test - it makes the difference nicely visible. Good to merge!

New PR version on the left, `master` on the right, using ar5iv.css

Screenshot from 2025-03-18 19-44-57

PR #2534, opened 2025-03-20
add viewBox in xy, with baseline adjustment

### LINKED ISSUE #2376: add viewBox to inline SVGs

brucemiller:
This PR adds viewbox to XY's svg along with baseline adjustment (via `vertical-align`). It also wraps the svg in an inline-block so that the baseline works even when within `mtext` (even in Firefox). We also add a couple of interesting test cases to the xy tests.

This is an alternative to #2397 and steals most of the code, but deals with the baseline adjustment differently. Thanks @xworld21 for excellent debugging and useful testcase!!!

Closes #2397 
Fixes #2376


lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -191,7 +191,7 @@ a.ltx_LaTeXML_logo { text-decoration: none; }
 
 .ltx_tabular .ltx_tabular { width:100%; }
 .ltx_inline-block,
-.ltx_inline-logical-block { display:inline-block; }
+.ltx_inline-logical-block { display:inline-block; text-indent:0px; }
```
xworld21, line 194: Right, I have seen text-indent getting inherited and cause issues! This will have some side effects, but all good I hope. It's just a bit weird to see it in a xypic PR.

brucemiller: Indeed, but not as weird as the effect it caused on a test case! :>

dginev (approved):

PR #2536, opened 2025-03-23
add @rowdepths to Alignment

xworld21:
Following up on https://github.com/brucemiller/LaTeXML/pull/2488#discussion_r2001739762, simply record `@rowdepths` among the data collected by Alignment.


brucemiller (approved):
Looks good! We may want to followup later distinguishing total-height from height, and possibly this gives us a better handle on (& alternative to) rowspacing. But for now, this is good! Thanks!

PR #2537, opened 2025-03-27
Fix use of \pgf@shift@baseline

brucemiller:
A slight correction now, so we're all on the same page. When the `baseline` pgf keyword is used,  it provides an *alternative* baseline to `\pgf@picminy', rather than adding to it; `\pgf@shift@baseline` defaults to `\pgf@pixminy`. Thus, when we set the `vertical-align`, we use the difference `miny-shift`, rather than the `shift`. (The shifted circle was deceptive, due to symmetry).

This is a correction to #2535
Oh! And I semi-accidentally snuck in @xworld21's baseline adjustment to the pgf alignments. Thanks!
Attn: @xworld21


dginev (approved):

brucemiller:
My, what a suspiciously fast review! :>

PR #2540, opened 2025-03-30
prepare for 00README.json new metadata directives from arXiv

dginev:
This small PR prepares latexml for the new `00README.json` metadata convention that succeeds the `00README.XXX` in arXiv's Submission v1.5 system.

This is not a core TeX enhancement, but follows the healthy practice of testing LaTeXML against the arXiv corpus for its annual releases. Pack.pm has existing support for `00README.XXX` when processing ZIP inputs, which this PR extends to `00README.json`.

I am marking this PR as a draft, until the new file format is officially documented at the [arXiv 00README documentation page](https://info.arxiv.org/help/00README.html). Tested on a couple of valid+invalid JSON files, the key target being entries of the form:

```json
{
   "sources" : [
      {
         "usage" : "toplevel",
         "filename" : "main.tex"
      }
   ]
}
```

### DIFF
--- a/lib/LaTeXML/Util/Pack.pm
+++ b/lib/LaTeXML/Util/Pack.pm
@@ -15,8 +15,9 @@ use warnings;
 
 use LaTeXML::Util::Pathname;
 use File::Spec::Functions qw(catfile);
-use File::Path qw(rmtree);
+use File::Path            qw(rmtree);
 use IO::String;
+use JSON::XS     qw(decode_json);
 use Archive::Zip qw(:CONSTANTS :ERROR_CODES);
 
 use base qw(Exporter);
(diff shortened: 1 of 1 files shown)


dginev:
The new metadata file is now in production testing, and its documentation is publicly visible:
https://info.arxiv.org/help/00README.html#00readme-json-file-format

Marking the PR as ready to review, thanks in advance!

dginev:
Here is an example of such an article, as the transition is proceeding:

https://arxiv.org/abs/2504.07866

brucemiller:
I was confused by the example, as I was expecting to see the metadata exposed in some fashion, but couldn't detect it. And also, surprised that there was no HTML version.

But none of that affects the apparent reasonableness of the PR, so... Thanks!

dginev:
The arXiv HTML deployment has not had updates for nearly a year, which may dispel some of the confusion.

We lead from ar5iv's side, then eventually the code updates will reach the arXiv installation at a yet-to-be-derermined milestone.

PR #2541, opened 2025-04-03
Foreign Objects and overall size and placement of SVG

brucemiller:
This PR improves SVG generation. A new debugging option (`--debug=svg`, as well as a more verbose `svg_verbose`) prints some internal information but also draws boxes to show the overall size and placement of where the svg *should* go, and even more interesting, where `svg:foreignObject` go. This approach reveals some interesting things, in particular that `xy` is making positioning adjustments via font parameters that we hadn't yet defined; that both `xy` and `pgf` sometimes put pictures in surprising places with respect to the current baseline (eg in `xy` whether you use `$\xymatrix{...}$` vs plain `\xymatrix{...}`, and so on).  Moreover, in conjunction with browser's Inspector, it reveals that where the browser actually places the content of an `svg:foreignObject` may not be what you expect. For example, they apparently process the contents in block mode, but also we're getting left & right padding for isolated things like "$=$" where TeX wouldn't.

This PR is an alternative to #2488, although some of what was learned there may still be useful to improve the positioning of the contents of foreignObjects w.r.t their container.

@xworld21 Please review/comment! Thanks for a great many clues & examples!

Closes #2488.


dginev (approved):
What a small PR - must have been easy! ;-)

Leaving my very grateful approval stamp, and I will proceed to explore some of the interesting tests available.

Thank you for all of the breakthroughs @xworld21 !

brucemiller:
Here's a handy test file that collects up @xworld21 samples and some others: [xw.zip](https://github.com/user-attachments/files/19596149/xw.zip)

brucemiller:
Not sure what the test issue is: fails locally too; and even master does, even though it passed the tests when the previous PR was merged...

dginev:
The test checked in the repository includes `«  Citation extérieure  »` which seems to have 2 `No break space (U+00A0)` chars on each side. Meanwhile the PR emits no such space chars. Curious - which piece changed? I am reproducing on texlive 2022.

dginev:
Oh, right, the `$nominal_fontinfo` likely changed the spacing logic in the tests. Leaving you to decide whether it is the test or the code that needs tweaking @brucemiller

lib/LaTeXML/Engine/TeX_Fonts.pool.ltxml
```diff
@@ -45,30 +45,32 @@ DefParameterType('FontDef', sub {
...
-    1.01,               # delimiter 2
-    0.25,               # math axis height
-    ];
+  0.25,        # Slant
+  0.5,         # Space
+  0.0,         # Stretch
+  0.0,         # Shrink
+  0.430555,    # x-height
+  1.000003,    # quad
+  0.0,         # extra space
```
dginev, line 57: Now I know too much: seeing a `0.0` next to a field titled "extra space", begs the question... Why not some positive value?

dginev (approved):

xworld21 (changes requested):
Ah, this is not working on my machine
image
My CSS is using Latin Modern Math at 10pt. The result should in principle as close as possible to actual TeX, but it obviously isn't.

I think we are still missing a proper foundation for what LaTeXML should be doing. I suggested choosing a specific anchor – middle of baseline – and offloading to CSS the hard work of twisting the HTML to sit there. This code is still anchoring everything to the top left corner, which reacts badly to sizing issues.

My comments below offer some potential explanations as to why I am getting bad results (funnily enough, I wrote them before even testing the PR...). In short: the Perl code looks fine to me except I am not convinced by that `\lineskip`, while I do not believe that the CSS approach is sound.

brucemiller:
> I think we are still missing a proper foundation for what LaTeXML should be doing.
Actually, I think we're finally on the road to that proper foundation, but we've not completed it yet. Your misplaced "=" is a critical example, but you can't really understand what's going on until you run with `--debug=svg`, and use the browser's inspector. What you'll see is an aqua box showing where LaTeXML (pretending to be TeX) thinks the "=" should go. And to my eye (in Firefox & Chrome) it is exactly where it should be. That seems to me to be a good basis.

However, using the inspector, we see that the way the browser has drawn the content is a bit different; it's positioned to the right, for sure. In this case, we could do `<mo lspace="0">=</mo>`, (since TeX doesn't pad here, but MathML does). Or we could:
> I suggested choosing a specific anchor – middle of baseline – and offloading to CSS the hard work of twisting the HTML to sit there. This code is still anchoring everything to the top left corner, which reacts badly to sizing issues.

Yes, that might be better, by translating again, relative to the (now, hopefully) correct position and then centering.

My strategy in this PR was to *first* get the `foreignObject` positioned correctly, and *only* then worry about how the browser renders its content.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -333,11 +364,17 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
...
+      my $linesep     = $STATE->lookupDefinition(T_CS('\lineskip'))->valueOf;
+      my $ht = $h->larger($lineheight); # Max of lineheight, since FO is block-like
+      my $y  = $ht->subtract($linesep);   # Draw DOWN from the top
+      my $H  = $ht->add($d);
+      Debug("SVG:foreignObject ".$w->pxValue.' x '.$h->pxValue.' x '.$d->pxValue."; "
+        . " => H=".$H->pxValue." @ y=".$y->pxValue
+        . " for Whatsit=".ToString($whatsit)) if $LaTeXML::DEBUG{svg_verbose};
+      $node->setAttribute(width     => $w->pxValue) unless $node->hasAttribute('width');
+      $node->setAttribute(height    => $H->pxValue) unless $node->hasAttribute('height');
+      $node->setAttribute(transform => "matrix(1 0 0 -1 0 ".$y->pxValue.")");
```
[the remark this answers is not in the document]
brucemiller: We definitely need better comments, but not until I understand it! Especially the mythical "CSS box" :>

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -333,11 +364,17 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
...
       my ($w, $h, $d) = $whatsit->getSize;
-      my $y  = $STATE->lookupDefinition(T_CS('\baselineskip'))->valueOf->pxValue;
-      my $ht = $h->add($d);
-      $node->setAttribute(width     => $w->pxValue)  unless $node->hasAttribute('width');
-      $node->setAttribute(height    => $ht->pxValue) unless $node->hasAttribute('height');
-      $node->setAttribute(transform => "matrix(1 0 0 -1 0 $y)");
+      my $lineheight  = $STATE->lookupDefinition(T_CS('\baselineskip'))->valueOf;
+      my $linesep     = $STATE->lookupDefinition(T_CS('\lineskip'))->valueOf;
+      my $ht = $h->larger($lineheight); # Max of lineheight, since FO is block-like
+      my $y  = $ht->subtract($linesep);   # Draw DOWN from the top
```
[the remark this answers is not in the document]
brucemiller: Yes and no. TeX doesn't use `\baselineskip` here since it's positioning relative to the baseline already (it does use it for spacing separate lines). But the browser (apparently) draws foreign objects as blocks, relative to their top. 

The actual computation is a bit heuristic, and there may be a more accurate version, but something like this is required.  And if `\baselineskip` isn't set correctly, that should probably be fixed, rather than avoided :>

lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
```diff
@@ -176,18 +174,19 @@ DefMacroI('\pgf@shift@baseline@smuggle', undef, '0pt');
...
 # 1. Beginning and ending a stream ===================================#
 #=====================================================================#
+# Adapted from pgf/systemlayer/pgfsys.code.tex
 DefMacro('\pgfsys@typesetpicturebox{}', <<'EoTeX');
-% \pgf@picmaxx,\pgf@pixmaxy are now the SIZE of the picture!!!!!!
-\advance\pgf@picmaxy by-\pgf@picminy\relax%
+   \pgf@ya=\pgf@shift@baseline\relax%
+   \advance\pgf@ya by-\pgf@picminy\relax%
+   % \pgf@picmaxx,\pgf@pixmaxy are now the SIZE of the picture!!!!!!
+   \advance\pgf@picmaxy by-\pgf@picminy\relax%
```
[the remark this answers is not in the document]
brucemiller: Yes, just reordered to look more like the current pgf code.

lib/LaTeXML/Package/xy.tex.ltxml
```diff
@@ -89,31 +92,51 @@ DefConstructor('\lx@xy@svgnested Digested',
...
   properties => sub {
     my ($x0, $y0, $x1, $y1) = @{ LookupValue('saved_xy_range')
         || [Dimension(0), Dimension(0), Dimension(0), Dimension(0)] };
-
     Debug("XY: " . ToString($x0) . ' ' . ToString($y0) . ' ' . ToString($x1) . ' ' . ToString($y1))
-      if $LaTeXML::DEBUG{xy};
+      if $LaTeXML::DEBUG{svg_verbose};
     my $w         = $x1->subtract($x0);
     my $h         = $y1->subtract($y0);
+    if($w->valueOf < 0){        # Rarely, the range hasn't actually been set?!?
```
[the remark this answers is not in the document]
brucemiller: It showed up in `\xymatrix{A}`, and I think LaTeXML is doing what xy does; it's just that xy doesn't particularly care about width & height for this case (but svg does)

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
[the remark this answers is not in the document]
xworld21: > The actual computation is a bit heuristic, and there may be a more accurate version, but something like this is required. And if `\baselineskip` isn't set correctly, that should probably be fixed, rather than avoided :>

To summarize my take on this, I am advocating for a strict separation between emulating TeX and taking care of the browser. E.g. `\baselineskip` is a register that measures line spacing in TeX, not in HTML. Unless you take a wider approach like setting `line-height: \baselinestretch` as well, I don't see why this should be the correct number to use.

There is also a problem of loss of information. You add the line skip, but I need to undo it because I want to use `text-box` (something I would do in BookML), I can't, because the shift has been merged with the depth information. That's why I was playing with saving the additional values in CSS variables.

xworld21:
After more experimentation, I'd like to ask that the code is written with potential use of `text-box` in mind. `text-box` gives the best results with arbitrary fonts. However, it requires *undoing* the depth shift used in this PR (not just the lineskip!). If that depth information can be added to XML somehow, so I can extract it and use it in media queries, that would be great (say a data attribute, or a CSS variable... whichever method looks reasonable).

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -436,8 +436,21 @@ span.ltx_framed       { display:inline-block; text-indent:0; } /* avoid padding/
...
+ SVG (pgf/tikz & xy) basics
+ Layout depends heavily on TeX's fonts which are more compact than typical browser's fonts
+ */
 
+svg text,
+svg foreignObject {
+    letter-spacing: -1px;
+}
+svg foreignObject span{
+    line-height: 0.9;
```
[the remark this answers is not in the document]
xworld21: This is the effect of those values on my example:
image

brucemiller:
Actually, I find the description of `text-box` a bit confusing. They refer to "space to trim from a text element's block container", and I'm wondering if we even have "text elements" in the intended sense, here. We have Math, sometimes blocks, inline-blocks, etc. 

I'm also not clear which "depth shift" you're wanting to access.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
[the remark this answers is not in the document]
xworld21: Following up on `line-height`: to make this PR work with my choice of font (Latin Modern 10pt), it is probably enough to set `line-height` on each foreignObject to match its height. That's more or less what I was doing in my PR as well, *but*, you are now making sure that fO is always tall enough to fit the content. That should solve the problems I was having.

We could push this further: set height & line height to *twice* `\baselineskip`, so that there is plenty of room for different fonts. Then you should reliably end up with content at the correct baseline without resorting to `text-box`. Of course you'll need to reset `line-height` in children inline-block's.

Side note: the height of the foreign object does not have to be the total height. It could be just the height. That would allow a smoother upgrade to `text-box` down the line.

brucemiller:
So `text-box` doesn't affect math? That's what I feared, and (to my eye) the placement of math is one of the two most egregious current bugs.

And maybe it's a matter of perspective, but what you're calling the "depth shift", to me, moves the box *up* by the *height* of the box (since svg is starting from the top).

xworld21:
> So `text-box` doesn't affect math? That's what I feared, and (to my eye) the placement of math is one of the two most egregious current bugs.

It does affect maths: it gets deposited and aligned to whichever baseline is determined by the choice of trimming. In my experiment it works exactly as it should.

xworld21:
I should clarify text-box with a picture: this is what happens if I add MathML to the MDN playground example of text-box
image
The second line is `text-box-trim: both` and it only contains a math tag.

(Playground in question: https://developer.mozilla.org/en-US/docs/Web/CSS/text-box-trim#result)

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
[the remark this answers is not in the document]
xworld21: I did a little experiment on your PR. I can get pretty good `text-box` results by adding a flex wrapper, coupled with
```css
.ltx_foreignObject_container {
  width: 100%; height: 100%; }

@layer lx-foreignObject {
  .ltx_foreignObject_content {
    text-wrap: nowrap; width: max-content; }
  .ltx_foreignObject_content .ltx_inline-block {
    text-wrap: revert-layer; }
  @supports (text-box: trim-end cap alphabetic) {
    foreignObject {
      translate: 0 calc(-1 * var(--lx-depth)); }
    .ltx_foreignObject_container {
      display: flex; justify-content: center; align-items: end; }
    .ltx_foreignObject_content {
      text-box: trim-end cap alphabetic; }
    .ltx_foreignObject_content .ltx_inline-block {
      text-box: revert-layer; }
  }
}
```
It looks like a lot, but what it is doing is:
- LaTeXML saves the depth in the variable `--lx-depth`; then CSS uses `--lx-depth` to move the bottom edge of the foreignObject up to the intended baseline. (I have also removed the `\lineskip` shift completely.)
- `flex` moves the content bottom edge to the bottom edge of the foreignObject, and it centers the content horizontally. This requires two extra div's in XSLT, as in my PR.
- `text-box` trims the bottom edge of the content to its baseline, so that anything below the baseline will overflow under the bottom edge. Hence HTML text will sit on the same baseline as the equivalent TeX text, regardless of font metrics.
- `@layer` is an orthogonal idea. This is the correct way of reverting values that could be inherited from the foreignObject: it makes it as if the foreignObject never existed. This is in case e.g. `text-box` is customised somewhere else and the foreignObject would break inheritance. It is probably overkill, but it would make sense for e.g. resetting `line-height`.

This is robust against changing font sizes: the images look acceptable even when changing font size between 8pt and 12pt. They become very crowded obviously – my point is that slightly different font metrics will look good. For changing font sizes, ideally images should be sized in rem units, which is out of scope for this PR.

I still have no idea how to achieve the same without `text-box` and without the exact browser font metrics. It really feels like CSS has no access to that kind of information (which is why `text-box` was created, I suppose).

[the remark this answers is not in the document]
xworld21: PS: and xy looks excellent! I didn't say this, but it looks like you really fixed the underlying issues.

brucemiller:
I did some playing around with `text-box`, but couldn't get anything quite as convincing as your image above (something fishy with depth? and no good in Firefox). OTOH, I also did some playing around with `flex`, and am starting to appreciate it.  Following some hints @xworld21 gave in comments above, I came up with the following strategy.

Give the `foreignObject` a large size, maybe 3 times the content width & height (min `\baselineskip`  for safety) and also translate it so that it's centered over the Aqua Box. Add a single `span` wrapping the content, with the following CSS:
```
svg foreignObject > .ltx_svg_wrapper {
    display:flex;
    flex-wrap:wrap;
    align-content:center;
    justify-content:center;
    width:100%;
    height:100%;
}
```
And this should center the content horizontally and vertically within the FO, and lines up well with the AquaBox. Well, pretty good in Chrome, almost right in Firefox.  Is this something like what you were thinking? Or alternatively, does it sound workable?  Should I add a commit for this?

Of course, this is still not completely immune to font differences, but is close so long as the ratio of heights to depths is similar.

And, of course, there are still some font metric & size computation issues to sort out on the pure TeX side, but this version is a bit less affected.

brucemiller:
Incidentally, the `span` wrapper only seems to be necessary because I can't seem to get `display:flex` to take when applied directly to the `foreignObject`.

xworld21:
> this should center the content horizontally and vertically within the FO

We are converging! This was the first version of my PR, before I went off chasing baselines. Time for some Unicode art.
```
                 CSS Box

   TeX Box    NW'───N'───┐
               │half-lead│
NW────N────┐   ├─────────┤
 │  height │   │  ascent │
 │         │   │         │
 │    C    │   │    C'   │
 ├────B────┤   ├────B'───┤ ⇐ baseline
 │  depth  │   │ descent │
 └────S────┘   ├─────────┤
               │half-lead│
               └────S'───┘
```
I am pretty confident that the above is an accurate representation of the TeX vs CSS box (except that height/depth are not quite the same as ascent/descent: height/depth are based on the actual characters in the box, ascent/descent are determined by the font). So far we have more or less converged on three strategies:
- set NW = NW': this is what LaTeXML does now. Even with identical font metrics, it's not quite right (one should also add `line-height:0` which is unreliable and causes inheritance issues). We are trying to move past that.
- set C = C': my first attempt, and now also by Bruce. Very easy to do with flex, and very predictable (and does not even need messing with heights and line heights). It produces visible misalignment of e.g. multiple labels above arrows.
- set B = B': my subjective opinion is that this is the best option. When font metrics are not to far, pictures can pass of as generated by TeX (for the purposes of alignment; rectangles drawn around text like in tcolorbox will obviously never be quite right). However, B=B' really seems _impossible_ without `text-box`; even guessing from the TeX side is extremely unreliable. We both tried but we haven't come up with a robust non-text-box approach.

On the practical side, @brucemiller would it make sense to merge the XY updates first? I would like to push my own take on the CSS, but it's getting tricky to keep my PR in sync. Same for the SVG debug flags.

Given that B=B' seems impossible without text-box, my suggestion is to do C=C' but enhance it to B=B' with `@supports (text-box: ...)` (or at least, leave enough metadata in the XML so that it can be customised in CSS).


Re details:

> (something fishy with depth? and no good in Firefox)

Did you apply `translate: 0 $depth`? In the above picture, you need to raise the bottom of the FO (line 'S') up to the baseline (line 'B'). Depending on how you set up the FO height & translation, you may have to shift up by $depth or $depth+$linesep.

> maybe 3 times the content width & height

This shouldn't be required. If you size the foreignObject to width/height+depth and the outer flex wrapper to 100% width/height, it should all work out naturally (text wrap could be an issue, maybe the inner wrapper should have fit-content or max-content?). That should also simplify how you adjust the depth for text-box vs non-text-box.

> Incidentally, the `span` wrapper only seems to be necessary because I can't seem to get `display:flex` to take when applied directly to the `foreignObject`.

Yes, flex does not work on foreignObject. Not sure why. Also, foreignObject CSS works slightly differently in WebKit for whichever reason. It's safer to use flex on an additional wrapper.

brucemiller:
Thanks for the clarifications (& validations)!

My reasoning with the excessively large "half-lead" was that, as I understand it, flex here only can center the content by distributing the "extra" space in the container. In some cases, the TeX box is quite small and leaves no wiggle room without some generous padding. OTOH, what I hate about what I sketched is that the XML, which *should* be an abstract, but "faithful" representation of the document, no longer is (although the ultimate HTML may (or may not :> ) looks reasonable).

So, maybe it's best at this point to merge the PR as it is, knowing that there's more to do (or at most with a couple of adjustments to make further exploration easier). Then we get onto a common footing.

xworld21:
> flex here only can center the content by distributing the "extra" space in the container

When I tried it, flex was very happy to overflow in all directions! My PR never needed extra padding.

brucemiller:
Interesting; you're right it doesn't seem to require the excess padding. So, this actually allows a lot of code simplification; the size & position of the `svg:foreignObject` seems to "make sense", and the heavy lifting is (currently) carried by CSS flex.  I also added `data-(width|height|depth)` attributes to the `<span class="ltx_foreignObject_wrapper">`  (temporarily?) to support further experimentation, as @xworld21 requested.

I'm thinking this is a good ratchet click, with room for further refinement later. Y'all agree?

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -331,14 +362,32 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
...
+      # Size of FO is width and total height (+ depth)
+      $node->setAttribute(width     => $W->pxValue) unless $node->hasAttribute('width');
+      $node->setAttribute(height    => $H->pxValue) unless $node->hasAttribute('height');
+      # Set top of FO height above current, as if baseline will line up w/outside
+      $node->setAttribute(transform => 'matrix(1 0 0 -1 '.$x->pxValue.' '.$y->pxValue.')');
+      $node->setAttribute(overflow  => 'visible');
+      $node->setAttribute('_wrapped'=>1);
+      # Add a wrapper, so that CSS has a chance of tweaking/centering
+      my $wrapper = $document->wrapNodes('ltx:text',$node->childNodes);
+      $document->addClass($wrapper, 'ltx_foreignObject_wrapper');
```
xworld21, line 385: Interesting that you are adding this to the XML!

xworld21 (approved):
I haven't tried it yet, but it looks good to me! For me it's fine to go ahead, and I'll keep hacking at it until I find something reasonable.

The last idea I have is to determine the 'ascent/descent' length of Computer Modern, then translate the FO up by ((ascent-height)-(descent-depth))/2 so that the center of the TeX *line* matches the center of the CSS line (regardless of baselineskip!). This might work even better than text-box.

I don't see anything corresponding to ascent/descent in the `$nominal_font_info` though.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
dginev: tiny naming nit: @brucemiller could we avoid that capital `O`? 

It matches the SVG element name, but makes our CSS class suite noisier. either just `ltx_foreignobject_wrapper` or `ltx_foreign_object_wrapper` would be fine.

dginev (approved):
I agree this is in a nice and mergeable state, so feel free to go ahead. Ideally we just tweak the new CSS class name to not have CamelCase in it.

When viewing the attached `xw` example set, to get the final "Flow" example working with the ar5iv.css today I had to add an additional rule that shrinks down my font size, as in:

```css
svg foreignobject > .ltx_foreignObject_wrapper {
    & .ltx_p {
        line-height: 0.9rem;
        font-size: 0.8rem;
        text-align: left;
    }
}
```

ar5iv currently sets `"Noto Sans"` for the requested `.ltx_font_sansserif`

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
brucemiller: Yeah, I was surprised that I left it that way, too! Thanks for calling me out. I moved the wrapping to XSLT, which is more appropriate, and you originally had it. And I went ahead & made 2 wrappers, with your class names, as you had it.

brucemiller:
@dginev the span around that example has `style="font-size:80%;"`; is that getting lost, or you needed it twice? (fwiw: I haven't tracked it down, but I suspect we're not yet getting specifically sans metrics being used here(?))

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -436,8 +436,21 @@ span.ltx_framed       { display:inline-block; text-indent:0; } /* avoid padding/
...
+.ltx_foreignobject_container {
+    display:flex;
+    flex-wrap:wrap;
+    align-content:center;
+    justify-content:center;
+    width:100%;
+    height:100%;
+}
+.ltx_foreignobject_content {
+    display:contents;
```
xworld21, line 452: Does this mean that every child of `_content` becomes a flex item? Isn't that risky w.r.t. alignment and wrapping? Does text + math + text render correctly? (I should really just test it...)

brucemiller: To be honest, I really don't understand it completely. Its purpose seems to be to make that inner span sorta "transparent" as far as CSS layout is concerned(?) Without that span it behaved "correctly"; with it, the vertical position was off until I set display to `contents`. Perhaps there's a better way.

xworld21: Ok I'll test this tonight. I have a number of situations in mind that would break here (e.g. whitespace between successive math elements would disappear). I don't remember having this issue, but I was implicitly using `display:block` (my PR had div instead of span).

brucemiller: I did try `block`, but that didn't work. Perhaps the `flex` approach really doesn't want the extra layer?  If so, making the outer span block (& full sized) and moving all the flex CSS to the *inner* span would be the right thing. Or, just have a single wrapper.

dginev:
@brucemiller the habitual use of `style` attributes for hanging various overrides is something I should again flag as: "Oh it would be so much more convenient if we had consistent use of classes to attach these to".

I updated to the latest state of the PR and retried - indeed the font was already small enough. What was still activating from ar5iv was the large `line-height` however. So I reduced that to match the font-size. But, to ensure it continues matching it, I set both `line-height` and `font-size`.

I almost wish if you absolutely had to use style attributes like this, you would use them to set variables, as with `style="--fo-size-scale: 0.8"` and then I can set the CSS as:

```css
.ltx_foreignobject_container {
  & .ltx_p {
    line-height: calc(var(--fo-size-scale)*1rem);
    font-size: calc(var(--fo-size-scale)*1rem);
    text-align: left;
  }
}
```

Tested that, works nicely.

brucemiller:
Weird. I'm not sure what "still activating" means, here; LaTeXML (currently) doesn't set it, so you're getting some other value that you've set, or the default which is typically something like 1.2 (which is normal, not large).  If you want to cramp it down, use unit-less `line-height:1`,  relative to the font-size like my earlier *hack*. Avoid "em", which MDN points out has poor inheritance. And certainly not "rem", which forces you to unnecessarily reset the font-size, and even if LaTeXML used variables, they'd be relative, so you're almost guaranteed to set the wrong size.

So, while an eventual move to a good set of classes, and even variables, will be a good thing, this one's a disincentive.

dginev:
> I'm not sure what "still activating" means, here; LaTeXML (currently) doesn't set it,

ar5iv does, that wasn't clear from my previous comment? It's set to `1.5rem` on regular paragraphs.

dginev:
> So, while an eventual move to a good set of classes, and even variables, will be a good thing, this one's a disincentive.

I am not quite sure what you are seeing, but over here it mostly sounds `!important`

lib/LaTeXML/resources/CSS/LaTeXML.css
brucemiller: the math drops down with display block

brucemiller:
rem is great, for some things, but I'd have thought that `line-height:1.5rem` would be a bad idea, generally. Why not unitless `line-height:1.5`, relative to the current font-size?  I don't really get where you're going with the snark. If a document has `style:font-size:80%` scattered up the document, *or* has `class="ltx_smaller"`, *or* has `style:--font-scale:80%`, and you insist on using `rem` deep within a svg, you're going to end up with lots of `!important` *and* things displayed in the wrong size.  But, as you say, not my call.

dginev:
1. The difference between `style="80%"` and `style="--varname: 0.8"` is that the former requires an override, the latter provides a well-advertised handle. (both for cases where the outer context CSS had reasons to change the style)
2. `rem` is nice because a designer can predictably scaffold the full scope of their page, exactly without contextual queues (and if context is really needed, there is `calc()`). The examples of stacked `em` and `%` are hard to interact with unless they are embraced everywhere - the more they stack the harder it gets. What happens quite often in arXiv is "an author used this kind of construct in this new unexpected place, let's re-arrange the sizes". A paragraph in a foreign object in SVG is a perfect example of that.
4. It's not about `1.5rem` vs `1.5` unitless. It is about having a CSS rule written about a completely different context (regular paragraphs) that I need to have an alternative for in a new context. The reliance on `style` attributes nearly always complicates that.

brucemiller:
1. But if style or class are nested, you get the "Cascade", while with the variable, you only get the last.
2. With rem, I'd think the designer gets a "predictable scaffold" if they've done the entire document in absolute units. I don't quite see why you think you have to "re-arrange sizes" .  If anything, a paragraph embedded in an svg is a perfect example of something you should *not* be mucking around with, if you don't understand its specific history, context and nesting.
3. Context is exactly the good thing about "cascading" and inheritance.

lib/LaTeXML/resources/CSS/LaTeXML.css
xworld21: Two interesting bugs. One, `display:contents` is logically wrong for the reasons I mentioned. This is what happens when the node label is `$5$ or $5$`
image

The other bug is about box sizing. In this picture, the aqua box is correct. However, the foreignObject is the box around containing "or" only! Which is really strange. How did the foreignObject and the aqua box get out of sync?

Getting back to contents vs block:
- `display: content` works when the node is a simple math tag. Then the CSS box around the tag is essentially equivalent to the TeX box and it works out. However, it won't lay out text like in a block, so it fails as in the screenshot.
- `display: block` lays out the content normally, which would be preferred. However, it is affected by the mismatch between C being the center of the TeX _box_ (minimum rectangle that fits all the glyphs, with depth), and C' being the center of the CSS _inline box_ (minimum rectangle that fits the line vertically, and the glyphs horizontally). So with lowercase characters, the C is much lower than C' and so it get shifted down.

I don't have a good answer yet (except for `text-box`, obviously). Really, the only idea could be to pick ascent and descent from the CM metrics; say the max height and depth of A-Za-z. Then you can shift the foreignObject back up so that C is more or less where C' should be.

xworld21: The mismatch between aqua box and FO comes from these lines. The aqua box get its size (correctly) from HBoxContents
https://github.com/brucemiller/LaTeXML/blob/048f31a314513b7dc446b63b077c79946bea832a/lib/LaTeXML/Engine/TeX_Box.pool.ltxml#L312
while the foreignObject autoclose handler uses the `$whatsit` which I gather is *supposed* to be the entire content between opening and closing, but apparently isn't:
https://github.com/brucemiller/LaTeXML/blob/048f31a314513b7dc446b63b077c79946bea832a/lib/LaTeXML/Engine/TeX_Box.pool.ltxml#L367-L368

Which raises the question: can the `\hbox` constructor create the foreignObject by itself? That would seem less error prone.

xworld21: > the only idea could be to pick ascent and descent from the CM metrics;

I think I am on the right track. I have hardcoded the Latin Modern Math metrics into the following CSS:
```css
foreignObject { translate: 0 calc((8.06pt - attr(data-height type(<length>)) + attr(data-depth type(<length>)) - 1.94pt) / 2); }
```
8.06pt and 1.94pt are the 'ascent' and 'descent' of the OTF version of Latin Modern.

This looks pretty good on Chrome, and it survives changing font family and size in CSS. It also looks good in Firefox *after you replace attr() by hand* (Firefox does not support attr()).

I want to say it works in Safari as well, but WebKit has some baffling bug in which when you zoom the page, the pt units inside an SVG *don't rescale*. Once we have a robust solution in place, we'll have to come up with WebKit-specific workarounds.

xworld21: @brucemiller I have updated #2488 with better flex styling (see align-*items*), the above trick with hardcoded ascent/descent (but in em units), text-box progressive improvements (multiline content may be broken on WebKit until their bug is resolved). I have also changed the test about foreignObject having non-SVG children.

It looks reasonable in all browsers except for that '5 or 5' bug, which requires fixing whatever goes wrong in `$whatsit->getSize`, and the particular situations that trigger the WebKit bug around `vertical-align:bottom` in a trimmed text box.

brucemiller: I'll have to look more closely at #2488 to see the details, but I can see the "5 or 5" problem. The `\hbox` only opens an `svg:g`, w/o assuming a `foreignObject` will be created; but it *does* have the correct whatsit. OTOH, the code upon closing an `svg:foreignObject` is only seeing the last whatsit, as-if it's the one responsible for closing it. Getting the "complete" whatsit works out much better, but I have to figure out the proper way to get it.

xworld21: > I'll have to look more closely at #2488 to see the details

Let me also convert it to draft: I went all in with modern CSS features in the `text-box` branch (well, `@layer` and `@supports` are not particularly new by now, but it's not the typical CSS shipped with LaTeXML). I expect you to cherry pick the bits you like, rename the CSS variables, etc!

brucemiller: The first commit [ecd63ee](https://github.com/brucemiller/LaTeXML/pull/2541/commits/ecd63eed8862eaeefa6dcd358c3f03d50e71378c) makes sure to catch all the boxes that contribute to auto-opened elements, so that fixes the "S or S" problem. 

The next commit https://github.com/brucemiller/LaTeXML/pull/2541/commits/d0e7ac6a20310b591bd7b59aed80cb2947140d3a benefits from studying #2488 (though I may have missed things). And there's a couple of speculative bits where I could use some feedback.

I think I've understood @xworld21  suggestions regarding flex, but alas the browser doesn't vertically center the Ink, it centers the "line box"; and with Math, I'm getting rather tall boxes where the actual math is distinctly *not* in the center!  However, setting CSS `line-height:0` produces a compact box that sits exactly in the aqua border. Cool, but how to set it back? `line-height:revert` on the children didn't work. I used `line-height:normal`, but that's likely surprising for designers.

I still have some size measuring patches under development, but even that won't fix where text within a given width breaks differently in the browser. I added some `width:110%` tweaks which gets a nice result (Even an `!important` to override `minipage`'s explicit width (Don't look @dginev )).  But this isn't a good long-term solution. Probably some measurements (foreignObject, minipage, others (or all)) should be expressed in the em of the font used in the content?

xworld21: > However, setting CSS `line-height:0` produces a compact box that sits exactly in the aqua border.

That's only true in Chrome, although I don't quite understand why. Firefox is more true to the spec, and will still align the content to a hypotetical center line computed based on ascent and descent. Firefox has other problems too: `line-height: 0` does not take, one needs `line-height: 0.01` (and you need to explicitly inherit it in math content).

Note that my solution, based on ascent/descent metric from computer modern, is transparent to `line-height`: you can set it to whichever value (or not set it at all), and the output will look right.

lib/LaTeXML/resources/XSLT/LaTeXML-picture-xhtml.xsl
```diff
@@ -146,7 +146,15 @@
         <!-- If foreignObject in a DIFFERENT namespace, copy as foreign markup -->
         <xsl:when test="local-name()='foreignObject'
                         and not(namespace-uri(child::*) = $SVG_NAMESPACE)">
```
xworld21, line 148: I should flag that this test is still wrong.

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -436,7 +436,34 @@ span.ltx_framed       { display:inline-block; text-indent:0; } /* avoid padding/
...
+    position:relative; left:-5%;
+}
+/* line height 0 just for this layer, to keep the line-box compact */
+.ltx_foreignobject_content {
+    display:block;
+    line-height:0;
+}
+/*    line-height:revert;  Why doesn't this work? */
+.ltx_foreignobject_content * {
+    line-height:normal;
```
xworld21, line 461: If all of the content is inside a node, I believe this will negate `line-height:0` altogether. Revert is too heavy handed: it will go back to the UA stylesheet which is probably `inherit`. revert-layer has better semantics for this.

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -436,7 +436,34 @@ span.ltx_framed       { display:inline-block; text-indent:0; } /* avoid padding/
...
+}
+/* will have been sized by FO, so override the explicit size; AND shift back to right */
+.ltx_foreignobject_content .ltx_minipage {
+    width:110% !important;
+    position:relative; left:5%}
```
xworld21, line 466: Re font sizes: I believe within SVG we should prioritise matching font sizes, as in, passing an explicit `font-size:NNpt` as an inline style. It wouldn't be enough, so +10% is a good idea, but it doesn't make sense to support using completely different nominal sizes.

lib/LaTeXML/resources/XSLT/LaTeXML-picture-xhtml.xsl
dginev: Removing it is a nice experiment (since we have quite a number of tests).

xworld21: As far as I know, the foreignObject is only created via autoopen/autoclose, and the handler is not *supposed* to generate a foreignObject if the children are all SVG tags. On the other hand, I don't think the schema can enforce this, so there won't be any warnings if LaTeXML creates a redundant foreignObject by mistake.

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -436,7 +436,34 @@ span.ltx_framed       { display:inline-block; text-indent:0; } /* avoid padding/
...
+
+/* Use width 110% in a couple of places to inhibit unexpected line breaks
+  due to TeX/browser font differences */
+.ltx_foreignobject_container {
+    display:flex;
+    align-items:center;
+    justify-content:center;
+    height:100%;
+    width:110%;
+    position:relative; left:-5%;
```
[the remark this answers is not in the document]
xworld21: Yes, that's the bug, and a very old one at that: https://bugs.webkit.org/show_bug.cgi?id=23113

Any use of anything that tweaks coordinates *within* a foreignObject (translate, position:relative, etc) will reset the content to the top left corner of the SVG, with disastrous effects. I found the issue some time ago, because MathJax uses position:relative on its hidden assistive MathML, so all my SVGs were broken!

You can translate the foreignObject itself. The complication is that `-5%` will mean a completely different thing! You'll need to do some arithmetic in XSLT to precompute the lengths, e.g. inject an extra `translate()` in the transform and modify `@width` and `@height`.

lib/LaTeXML/Core/Document.pm
```diff
@@ -1842,9 +1843,11 @@ sub appendElementBox {
     setNodeBox($self, $p, $newbox);
     # AND, propogate to autoOpen'd ancestors due to same initial box (See appendTextBox)
     while (($p = $p->parentNode) && ($p->nodeType == XML_ELEMENT_NODE)
-      && $p->getAttribute('_autoopened')
-      && ((getNodeBox($self, $p) || '') eq $origbox)) {
-      setNodeBox($self, $p, $newbox); } }
+      && $p->getAttribute('_autoopened')) {
+        my $obox = getNodeBox($self,$p);
+        last if $obox && (($box eq $obox) || ($box eq ($obox->unlist)[-1]));
+        setNodeBox($self,$p, ($p ? List($obox,$box) : $box)); } }
```
xworld21, line 1849: This fixes the 5 or 5 problem, but I found another box-related bug: my tikzcd example contains
```latex
A \arrow[r, "\phi\eta" near start, "\phi" near start, "\psi"',"\eta" near end] & {\hbox{\large B}}
```
The foreignObject around the B is much wider than the aqua box. Exactly twice as large (coincidence?).

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -331,14 +362,29 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
...
+      my $y = $h;
+      my $f = $whatsit->getFont;
+      Debug("SVG:foreignObject ".$w->pxValue.' x '.$h->pxValue.' x '.$d->pxValue."; "
+        . " => H=".$H->pxValue." @ y=".$y->pxValue
+        . " for Whatsit=".ToString($whatsit)) if $LaTeXML::DEBUG{svg_verbose};
+      # Size of FO is width and total height (+ depth)
+      $node->setAttribute(width     => $W->emValue(undef,$f).'em') unless $node->hasAttribute('width');
+      $node->setAttribute(height    => $H->emValue(undef,$f).'em') unless $node->hasAttribute('height');
+      # Set top of FO height above current, as if baseline will line up w/outside
+      $node->setAttribute(transform => 'matrix(1 0 0 -1 '.$x->pxValue.' '.$y->pxValue.')');
```
xworld21, line 381: I don't think this is a good idea at this stage. The transform attribute has to be in pixel, and it has to match the physical size, because of the upside down y shenanigans. Now when em changes, the FO will shift by an unpredictable amount.

lib/LaTeXML/resources/CSS/LaTeXML.css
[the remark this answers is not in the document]
dginev: > Yes, that's the bug, and a very old one at that: https://bugs.webkit.org/show_bug.cgi?id=23113

Thanks, I have left a comment there to indicate our potential interest in the patch.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
xworld21: Thanks! Although, I think those values should be px as well. Best approach I have come up so far is:
```css
foreignObject { translate: 0 calc(0.306em + (var(--lx-depth) - var(--lx-height)) / 2); }
```
which take the point "C", push it down to "B", then push it up by where the center would have been if TeX computed lines like CSS. Well, I'll need to draw another picture to clarify...

My PR is still wrong because `0.306em` needs to be px as well. I haven't thought about the correct approach. Probably two additional variables --lx-ascent, --lx-descent.

Pixel values are also needed to switch to text-box, because one needs to translate up by `var(--lx-depth)`, to move the point "S" up to "B".

xworld21: Updated box picture:
```
   TeX Box       CSS Box

              NW'═══N'═══╗
               ║half-lead║
 ┌─────────┐   ║─────────║
 │  ascent │   ║  ascent ║
 │         │   ║         ║
NW════N════╗   ║         ║
 ║  height ║   ║         ║
 ║   CLN   ║   ║   CLN'  ║ ← center of line (when line-height≥1)
 ║   CBX   ║   ║         ║
 ║────B────║   ║────B'───║ ← accessible via text-box
 ║  depth  ║   ║         ║
 ╚════S════╝   ║ descent ║
 └─────────┘   ║─────────║
               ║half-lead║
               ╚════S'═══╝
```

The double lines are the borders of each box as seen from CSS.

The single lines of the TeX box can be computed by adding the ascent/descent metrics somewhere during construction.

Finding the single lines of the CSS box is... tricky and cannot be done reliably. The half-lead is (1lh-1em)/2, but only if line-height is set to a number ≥1. The baseline position is _typically_ 0.5cap below the center, but only for fonts that approximately satisfy (ascent - cap-height) = descent. I gather that it is not a rule.

The baseline is special: if you have text-box-trim, then you can actually use it from CSS for alignment.

My CSS line with `0.306em - ...` is 'simply' shifting CBX up to CLN, so that flex ends up putting CLN' on top of CLN.

brucemiller: Interesting. But the width value is very intentionally in em, since it is used for laying out inline text. With em, you have a fighting chance of getting reasonable line breaks (but of course, no guarantee).  The y translation, otoh, is pure heuristic; perhaps it should be px rather than em? But since the "True" value depends on things we have no access to (short of text-box), it hardly seems worth being overly precise.

xworld21: > The y translation, otoh, is pure heuristic; perhaps it should be px rather than em?

The y translation is not necessarily heuristic. It depends on the strategy;
- CLN = CLN': exact, needs px values of ascent/descent/depth
- B = B' (text-box): exact, needs px value of depth
- B ~ B' (w/o text-box): approximate, needs px value of ascent/descent/depth + 0.5cap

The heuristic is all in that 0.5cap though, not in the depth value.

The width in em units seems good for the reasons you mention. Although I consider that a temporary measure until SVGs can be sized in rem units (which is now possible since viewBox works).

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -436,12 +436,36 @@ span.ltx_framed       { display:inline-block; text-indent:0; } /* avoid padding/
...
+    height:100%;
+    width:var(--fo_width);
+}
+/* Also horizontally for single math, since Chrome adds a lot of left/right spacing. */
+.ltx_foreignobject_container:has( > .ltx_foreignobject_content > math:only-child) {
+    justify-content:center;
+}
+/* If multiple children, give more space to avoid unexpected line breaks
+   due to font differences. */
+.ltx_foreignobject_container:has( > .ltx_foreignobject_content > :not(:only-child)) {
```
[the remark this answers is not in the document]
brucemiller: No, foreignObjects are not only from `\hbox`, so you don't want to prevent wrapping unconditionally.

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -436,12 +436,36 @@ span.ltx_framed       { display:inline-block; text-indent:0; } /* avoid padding/
 .ltx_nounicode    { color:red; }
 .ltx_nodisplay    { display:none; }
 /*======================================================================
- SVG (pgf/tikz ?) basics */
-
+ SVG (pgf/tikz & xy) basics
+ */
+/* Heuristically adjust for content uncentered within line-box */
+foreignObject {
+    translate: 0 0.1em;
```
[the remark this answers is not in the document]
brucemiller: exactly. As I said, it's heuristic for now.

lib/LaTeXML/resources/CSS/LaTeXML.css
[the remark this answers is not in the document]
xworld21: What are the other examples of foreignObject? I thought that xy and pgf were both going through hbox.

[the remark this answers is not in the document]
brucemiller: `foreignObject` is explicitly opened by `\hbox`, but also is `autoOpen`, so just about anything non-svg might open one if we're already in svg.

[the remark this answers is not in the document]
dginev: You could use an alignment as well right? Pretty much anything is allowed with the right escapes.

```tex
\documentclass{article}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
  \matrix[ampersand replacement=\&] {
      \node (test) [shape=rectangle,draw] {\begin{tabular}{cc}
              a & b \\
              c & d
          \end{tabular}};
      \\
  };
\end{tikzpicture}
\end{document}
```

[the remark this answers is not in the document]
xworld21: > You could use an alignment as well right?

Interesting! PGF is still going through `\hbox`, so it's not quite the example I was asking for, but it reveals a different bug. The box height is pretty much spot on. The depth is wrong. LaTeXML wants to treat tabular as inline-table with vertical-align:middle, hence the depth of should be about half the height of the table. Instead, it comes out as 0.

This is not the first time I encounter depth 0 with vertical boxes that should have some depth. I'll try to submit an issue with something more definite.

PR #2542, opened 2025-04-06
do not test directories for writeability on Windows

### LINKED ISSUE #2504: failure to write in subdirectory on Win32
A simple
```
latexml --dest=subdir/dest.xml source.tex
```
can fail on Windows with error
```
Couldn't create writable destination directory subdir/:
```
which seems to come from the line
https://github.com/brucemiller/LaTeXML/blob/70de862ff1997e228e8a0ee33ba61d1800adf189/bin/latexml#L193-L194

However, `-w $dir` can (and does!) generate false negatives on Windows, because Perl literally tests for the wrong thing (see https://perldoc.perl.org/perlport#-X).

The trivial fix is to skip the test on Windows altogether. The proper fix is to actually open the file for writing, which of course should not be done in `validate_args`.

xworld21:
FIx #2504 the quick way.

### DIFF
--- a/bin/latexml
+++ b/bin/latexml
@@ -190,7 +190,8 @@ sub validate_args {
   if ($destination) {
     $destination = pathname_canonical($destination);
     if (my $dir = pathname_directory($destination)) {
-      if (!(pathname_mkdir($dir) && -w $dir)) {
+      # -w is not about writeability on Windows, see https://perldoc.perl.org/perlport#-X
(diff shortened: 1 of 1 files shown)


dginev (approved):

brucemiller (approved):
Yikes, Yuck!  I second the comment "Would be nice if..."
But I guess if @dginev  is ok with it, and our (non)resident windows user @xworld21 obviously is, I guess I'm ok with it too. Thanks!

PR #2543, opened 2025-04-06
treat Cygwin as Unix, do not use ShellQuote on NetWare

xworld21:
While I was looking at Windows things, I remember that LaTeXML is currently broken on Cygwin. This PR unbreaks it, although it comes with a big caveat: it assumes that kpsewhich is also a Cygwin binary.

Also, I was wrong to use Win32::ShellQuote on NetWare. Not that anybody was complaining.


dginev (approved):
Nice

PR #2549, opened 2025-04-22
Box management

brucemiller:
This PR improves the management of the boxes (Box, List, Whatsit) responsible for XML elements, in particular, the accumulation of boxes when an XML element is automatically opened.

It fixes the duplicated boxes responsible for too-wide foreign objects mentioned in #2541.

PR #2555, opened 2025-05-06
Sizing

brucemiller:
This PR establishes a better connection between TeX fonts (and `\font`, fontcs's, font metric and parameters) and LaTeXML's more logically (LaTeX) oriented `LaTeXML::Common::Font` objects. This allows for better access to font parameters and metrics improving the sizing of digested boxes (important for tikz, xy). It also provides a richer set of preinstalled font metrics. So far, we're just scratching the surface by dealing better with horizontal, vertical and "paragraph" mode (although we still haven't implemented proper TeX modes), and better sizing of math sub & superscripts, and alignment structures.  Basically this deals with the most egregious issues seen in the test case used in #2541, #2549.


dginev (approved):
I have walked through all code changes - impressive.

My understanding in terms of TeX algorithm accuracy is very limited, but I think I examined the perl aspects carefully and found everything neat and reasonable.

Will do some testing shortly, but feel free to merge whenever.

Edit: tested with some of the t/graphics files and sizes seemed healthy. The diffs in the test files are also encouraging (heights getting slightly larger, and/or having part of their value migrating into depth)

PR #2559, opened 2025-05-10
some MathML operator dictionary fixes and tests

xworld21:
I noticed some typos in the MathML operator dictionary code and fixed them.

To be sure, I added a couple of artificial tests in t/post/simplemath.xml to verify that the code is working as intended (e.g. now `&&` is correctly recognised as binary operator, so it does not get lspace/rspace attributes).

### DIFF
--- a/lib/LaTeXML/Post/MathML/OperatorDictionary.pm
+++ b/lib/LaTeXML/Post/MathML/OperatorDictionary.pm
@@ -102,7 +102,7 @@ sub lookup_category {
   # Otherwise, if it has two characters:
   if ($len == 2) {
     my $code2 = ord(substr($content, 1, 1));
-    # [NOT YET HANDLED] If Content is the surrogate pairs corresponding to
+    # [ADDED TO $Content_form] If Content is the surrogate pairs corresponding to
     #   U+1EEF0 ARABIC MATHEMATICAL OPERATOR MEEM WITH HAH WITH TATWEEL
     #   or U+1EEF1 ARABIC MATHEMATICAL OPERATOR HAH WITH DAL and Form is postfix,
     #   exit with category I.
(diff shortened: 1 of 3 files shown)


lib/LaTeXML/Post/MathML/OperatorDictionary.pm
```diff
@@ -116,8 +116,8 @@ sub lookup_category {
     #   then replace Content with the Unicode character "U+0320
     #   plus the index of Content in Operators_2_ascii_chars"
     #   and move to step 3.
-    elsif (my ($p) = grep { $$Operators_2_ascii_chars[$_] ? ($_) : (); } 0 .. $#$Operators_2_ascii_chars) {
-      $code1 += $p;
+    elsif (my $p = $$Operators_2_ascii_chars{$content}) {
+      $code1   = 0x320 + $p;
```
dginev, line 120: Subtle edge case detail: 

```perl
elsif (my ($p) = (0)) {...}
``` 
from the previous code would have evaluated as true. 

But the hash lookup approach would be false for the 0 index. Probably gfood to add a test for `!!`.

dginev: Can we just hardcode the exact final character codes in `$Operators_2_ascii_chars` and avoid the runtime arithmetic? Should be safer (and won't have 0 values).

dginev (changes requested):

lib/LaTeXML/Post/MathML/OperatorDictionary.pm
xworld21: Good catch! Now harcoded directly in the hash.

dginev (approved):
I think this PR indeed follows MathML Core.

It is not easy to penetrate which of the details are actually relevant and how to test them - I am not sure the `!!` test I suggested actually reveals anything of interest, as it generates a simple `<mo>` without attributes. But the code looks safe now to my reading ✅

xworld21:
> I am not sure the `!!` test I suggested actually reveals anything of interest, as it generates a simple `<mo>` without attributes

Actually, the lack of attributes is the best signal that the dictionary is working as intended. It's recognising that MathML thinks of `!!` as postfix, so there is no need for `lspace=0`. If you test with LaTeXML 0.8.8, it will behave correctly with `!!` by accident (it looks up `!`, which is also postfix), but it will do the wrong thing with `&&` (it tries `(` which has 'Default' category when infix, so it adds redundant lspace/rspace to make it a binary operator).

This is still fragile in so far as LaTeXML does *not* compute the infix/postfix/prefix form in the way MathML and could omit lspace/rspace in the wrong situation. If I find a way to trigger a bug there, I'll file an issue with the example.

brucemiller (approved):
seems to be reasonable and following spec, as best as I can understand. Thanks!

PR #2562, opened 2025-05-13
allow setting a maximum number of warnings

### LINKED ISSUE #2308: Maximum number of warnings?
The current record-holder for largest asset in ar5iv is `arXiv:1501.02683`, which has a ZIP of 647 MB in size.

To add insult to injury, the bundle contains a single `cortex.log` file, containing the latexml conversion logging information - and totaling a whopping 7.7 million lines. 

As it turns out, those lines are almost entirely the same warning message, emitted in a busy loop until the timeout is reached (45 minutes). It reads:
```
Warning:expected:<number> Missing number, treated as zero
        at 1501.02683.tex; line 2441 col 0 - line 2441 col 12
        while processing \count@
        Next token is T_CS[\relax] ( == Core::Definition::Primitive[\relax]) more: \ifnum\count@<\value{proofcount}
\advance\count@\@ne\relax\expandafter\iterate\fi\let\iterate\relax
        In Core::Gullet[@0x559bfe9cba78] /dev/shm/hN2Sba2fTg/1501.02683.tex; 
 from line 2441 col 0 to line 2441 col 12
```

It is very clearly a pathological case of expansion gone wrong. But it suggests an idea for a new guard, similar to the "100 Error" guard that terminates execution. Maybe 10,000 Warnings?

Curious if @brucemiller finds this a good idea, and how many warnings could be considered a "healthy maximum".

dginev:
Fixes #2308 - following the approach for maximum number of errors.

The cost is calling a `lookupValue` every time we emit a Warning, which is possibly OK?

If this seems acceptable, I plan on combining it with a run-specific binding, for example `ar5iv.sty.ltxml` can include the line:

```perl
AssignValue('MAX_WARNINGS', 10_000, 'global');
```

So it is meant as an internal capability for the moment.

### DIFF
--- a/lib/LaTeXML/Common/Error.pm
+++ b/lib/LaTeXML/Common/Error.pm
@@ -383,6 +383,10 @@ sub Warn {
   my $formatted = generateMessage("Warning:" . $category . ":" . ToString($object),
     $where, $message, 0, @details);
   _printline($formatted);
+  # Optional: practical guards for max warnings, (e.g. to avoid infinite loops overflowings disks)
+  if (my $maxwarnings = ($state && $state->lookupValue('MAX_WARNINGS'))) {
+    if (($state->getStatus('warning') || 0) > $maxwarnings) {
+      Fatal('too_many_warnings', $maxwarnings, $where, "Too many warnings (> $maxwarnings)!"); } }
   return; }
 
 # Informational message; results likely unaffected


brucemiller (approved):
Seems useful in extreme cases. Thanks!

dginev:
For archival, here is a first example of this internal feature used for the ar5iv builds:
https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/fatal/too%5Fmany%5Fwarnings

We've already hit 8 articles stopped at 10,000 warnings from the June 2025 sources of arXiv.

PR #2566, opened 2025-05-19
Refine zero-width attachments in subfigure algorithm

dginev:
This is motivated by a peculiar report for [arXiv:2408.03371, table 2](https://ar5iv.labs.arxiv.org/html/2408.03371v1#S2.T2).

As it turns out, it is a `{deluxetable*}` with an explicit `\tablenotetext` note following the tabular data.

The current code treats `ltx:note` a little too respectfully and upgrades them (wrongly) to subfigure panels in certain branches of the `arrange_panels_and_breaks` algorithm. There was also a subtlety where elements best treated standalone (on a subfigure row) were getting broken out too early, so a following note couldn't reattach to them.

I added a new test for the example which should **NOT** use the subfigure algorithm (this PR's contribution). I also caught a mistake in the existing complex subfigure test file, where a note was getting treated as a panel. With this PR, there is a new grouping block element that attaches it to the previous `ltx:itemize`.


brucemiller:
Hmm, conflicts. And I'm wondering if the modes work otherwise affects the best approach (or need) here?

dginev:
Ah, I forgot I had done this by now - it was once again needed in #2591, so maybe I should revive this PR first and decide if we can merge. It would still be needed, at least from my current understanding of the recent changes.

dginev:
@brucemiller ready for review. Let's also do this before #2591 , I will rebase that branch after we've decided how to handle things here.

PR #2567, opened 2025-05-23
[Feature] Extend input kinds with  --whatsin=directory

dginev:
This PR introduces a new feature, allowing LaTeXML to take a directory on the filesystem as input. The two expected call patterns are:

```
latexmlc --dest=example.zip directory/
latexmlc --dest=example.zip --whatsin=directory directory
```

The feature reuses the same logic for the `--whatsin=archive` ZIP inputs, but also refactors the code in a way that makes it easy to add other archive processors (for example gz and tar) at a later date. The interface for a processor requires method implementations for:
 - `find_file($name)`
 - `full_filename($name)`
 - `find_tex_files()`
 - `unpack_source()`
 - `cleanup()`

I still need to figure out what to do with some of the more half-baked utility functions, such as `get_archive` and `pack_collection`, but ~~maybe~~ I suggest those can get postponed for a next iteration.

Renaming the `archive` keyword to `zip` would also become more pressing when there are at least two different archival processors. The directory one simply reuses the detection logic from the ZIP input, but without deleting anything (as ZIP also erases the sandbox directory at the end of processing).


dginev:
The last commit https://github.com/brucemiller/LaTeXML/pull/2567/commits/8c1dffd2eae4e7ca2998c2b214a01de6d06dc6bb ensured a very specific edge case is logged correctly, namely:
```
latexmlc /tmp/empty/ --dest=test.zip --log=test.log
```

The expectation is to write only a single `test.log` file which contains the Fatal+invalid error that there was no TeX input file, then that is in turn added to `test.zip` and written to disk. No other assets are created.

Previously, latexmlc would also write a `test.log` on disk, alongside the one in `test.zip` - with slightly different contents. I don't think that is useful. I also got a chance to refactor some special-cased code in LaTeXML.pm, as the detection can be done and logged once in `Core::digestFile` while the writing of the log file can always be done by the Writer post-processor.

This is encountered surprisingly often in arXiv conversions, where some of the source bundles have ill-formed contents (e.g. an encrypted and inaccessible `main.synctex` file)

lib/LaTeXML/Util/Pathname.pm
```diff
@@ -373,9 +373,9 @@ sub candidate_pathnames {
           qr/^\Q$name\E\Q$ext\E$/]); } }
   # Now, combine; precedence to leading directories.
   foreach my $dir (@dirs) {
-    opendir(DIR, $dir) or next;
-    my @dir_files = readdir(DIR);
-    closedir(DIR);
+    opendir(my $dir_handle, $dir) or next;
+    my @dir_files = readdir($dir_handle);
+    closedir($dir_handle);
```
dginev, line 378: this change was a recommendation (severity 5) by perl critic - using a lexical var name.

dginev:
Ready for review.

brucemiller (approved):
I didn't follow every detail, but overall it looks sensible and useful. Thanks!

PR #2569, opened 2025-05-29
robust path join in import.sty.ltxml

dginev:
Fixes [arXiv:2103.14899](https://ar5iv.labs.arxiv.org/html/2103.14899), reported in [ar5iv:535](https://github.com/dginev/ar5iv/issues/535).

The concrete use example was:
```tex
\subimport{./}{Main/Tex/01_introduction}
\subimport{./}{Main/Tex/02_related_works}
\subimport{./}{Main/Tex/03_proposed}
```

The issue was particularly visible when using the `.zip` input to latexml (appending the raw `.` char was invalidating the unpacked sandbox path, hence all relative imports were failing).

A bit of care to always use a path separator improves the arXiv article to Warning status.

### DIFF
--- a/lib/LaTeXML/Package/import.sty.ltxml
+++ b/lib/LaTeXML/Package/import.sty.ltxml
@@ -13,7 +13,6 @@
 package LaTeXML::Package::Pool;
 use strict;
 use warnings;
-use Config;
 use LaTeXML::Package;
 use LaTeXML::Util::Pathname;
 #======================================================================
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/import.sty.ltxml
```diff
@@ -31,8 +32,11 @@ DefPrimitive('\lx@set@path OptionalMatch:* {}', sub {
 DefPrimitive('\lx@append@path OptionalMatch:* {}', sub {
     my ($stomach, $star, $path) = @_;
     $path = ToString(Expand($path));
+    return unless $path;
     if (my @paths = @{ LookupValue('SEARCHPATHS') }) {
-      my $newpath = shift(@paths) . $path;
+      my $lead_path = shift(@paths);
+      $lead_path =~ s/\/+$//;
+      my $newpath = join($Config::Config{'path_sep'}, ($lead_path, $path));
```
brucemiller, line 39: Shouldn't something in `Util::Pathname`, like `pathname_concat`, be covering this? There's already some pathname separator stuff in there (although possibly it should be based on `Config`).

PR #2571, opened 2025-06-01
Info and empty box fallback when \pgfsys@hbox fails to find a ready box

### LINKED ISSUE #2326: pgfplots legend regression
This was a well-hidden regression as we don't have pgfplots tests in place.

Minimally, it is reproduced by the error from converting the snippet:
```tex
\documentclass{article}
\usepackage{pgfplots}
\usepgfplotslibrary{groupplots}

\begin{document}
\begin{tikzpicture}
\begin{groupplot}
  \nextgroupplot[legend to name = {complegend}]
\end{groupplot}
\end{tikzpicture}%
\end{document}
```

Which has regressed to emit an error:
```
Error:undefined:\pgfplots@legend@to@name The token T_CS[\pgfplots@legend@to@name] is not defined. at min.tex; line 17 col 0
```

which can then lead to a Fatal-grade error cascade, as with converting [arXiv:1709.07394](https://ar5iv.labs.arxiv.org/html/1709.07394).

I had a careful bisecting session and traced back the original cause to the Scan Toks feature commit https://github.com/brucemiller/LaTeXML/commit/4757b39ddd0977099ef7c640ce546206adf735d6 . In my testing this variation has been broken in all commits since July 2023, so it isn't related to any recent PR work.

Tested with texlive 2022, but I suspect this is generic enough to be reproducible on other texlives.

dginev:
Closes #2326 

I compared the renderings, and the minimal example, as well as Figure 11 from arXiv:1709.07394 render with parity between HTML and PDF, if a simple fallback is added to `\pgfsys@hbox`.

I am not sure if there is a good reason for the box to be missing on lookup so I added an `Info` message when that happens. But a graceful failure mode is to provide a default empty box instead of `undef`, which gets zero sizes and a healthy output.

Since perltidy reformatted the file a little, it is easier to look at the `?w=1` changeset here:

https://github.com/brucemiller/LaTeXML/pull/2571/files?w=1

### DIFF
--- a/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
+++ b/lib/LaTeXML/Package/pgfsys-latexml.def.ltxml
@@ -96,7 +96,7 @@ DefConstructor('\lxSVG@insertpicture{}', sub {
         transform   => "translate($x0,$y0) matrix(1 0 0 -1 0 0)",
         _scopebegin => 1);
       addSVGDebuggingBox($document,
-        $props{minx},$props{miny},$props{width},$props{height},'#FF00FF')
+        $props{minx}, $props{miny}, $props{width}, $props{height}, '#FF00FF')
         if $LaTeXML::DEBUG{svg};
       $document->absorb($content);
       while ($document->maybeCloseElement('svg:g')) { }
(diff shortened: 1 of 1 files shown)


brucemiller:
Ah,, I think this actually got fixed by #2580. Could you check? and if so close this & #2326

dginev:
You are correct - The minimal example now works with the latest latexml.

So does Figure 11 from arXiv:1709.07394. 

Great - this PR can be ignored.

PR #2572, opened 2025-06-01
ensure \Big and variants use TeXDelimiter parameters

### LINKED ISSUE #2507: \big. generates an actual dot
In LaTeX, \Big. generates a phantom vertical space which is useful e.g. to stretch delimiters with \left ad \right

In LaTeXML, an actual dot appears.

MWE:

```
\documentclass[11pt]{amsart}
\title{Bug report}
\usepackage{bookml/bookml}
\begin{document}
$\left[ \Big. x+y \right]$
\end{document}
```

dginev:
Fixes #2507. Review most welcome, just my first idea for an approach.

The premise of the PR is that \Big (and its variants) and \left should really take the same parameter types - currently called `TeXDelimiter`.

That requires a slight adjustment to use the `#hint` property check for `.` and a little more care to group reversions only when a T_BEGIN was encountered in the argument.

Seems to work? My test file was derived from the issue and one comment:
```tex
\documentclass[11pt]{amsart}
\begin{document}
$\left[ \Big. x+y \right]$

$\left. \Big. x+y \right.$

$ \Big{((} x-z \Big{))} $
\end{document}
```

The advanced care for `\Big{((}` successfuly reading both parens but only modifying the first hasn't been added here - so both of those parens will continue to have their sizes modified.

### DIFF
--- a/lib/LaTeXML/Engine/Base_ParameterTypes.pool.ltxml
+++ b/lib/LaTeXML/Engine/Base_ParameterTypes.pool.ltxml
@@ -360,19 +360,23 @@ sub readDigested {
   no warnings 'recursion';
   my ($gullet) = @_;
   $gullet->skipSpaces;
-  my $ismath = $STATE->lookupValue('IN_MATH');
-  my @list   = ();
+  my $ismath  = $STATE->lookupValue('IN_MATH');
+  my @list    = ();
+  my $grouped = 0;
   my $token;
   do { $token = $gullet->readXToken(0);
   } while (defined $token && (($token->getCatcode == CC_SPACE) || $token->equals(T_CS('\relax'))));
   if    (!defined $token) { }
   elsif ($token->getCatcode == CC_BEGIN) {
+    $grouped = 1;
     Digest($token);
     push(@list, $STATE->getStomach->digestNextBody()); pop(@list); }    # content w/o the braces
   else {
     push(@list, $STATE->getStomach->invokeToken($token)); }
   @list = grep { ref $_ ne 'LaTeXML::Core::Comment' } @list;
-  return List(@list, mode => ($ismath ? 'math' : 'text')); }
+  my $list = List(@list, mode => ($ismath ? 'math' : 'text'));
+  $list->setProperty('grouped', $grouped) if $grouped;
+  return $list; }
 
 DefParameterType('Digested', \&readDigested,
   undigested => 1,                                           # since _already_ digested.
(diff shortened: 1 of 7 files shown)


lib/LaTeXML/Engine/Base_ParameterTypes.pool.ltxml
```diff
@@ -360,19 +360,23 @@ sub readDigested {
...
+    $grouped = 1;
     Digest($token);
     push(@list, $STATE->getStomach->digestNextBody()); pop(@list); }    # content w/o the braces
   else {
     push(@list, $STATE->getStomach->invokeToken($token)); }
   @list = grep { ref $_ ne 'LaTeXML::Core::Comment' } @list;
-  return List(@list, mode => ($ismath ? 'math' : 'text')); }
+  my $list = List(@list, mode => ($ismath ? 'math' : 'text'));
+  $list->setProperty('grouped', $grouped) if $grouped;
+  return $list; }
```
brucemiller, line 379: Ugh! Adding a `grouped` property seems really klunky, but hard to see the proper alternative...

dginev: Thinking of code precedent... The Whatsit `revert` method already uses certain properties to make decisions, so I could make this technique "more official" by enhancing List's `revert` method?

That would avoid the appearance of a special case, and any List with that property set would then revert braced.

brucemiller: It's still a special case, just more generic :> But maybe `braced` is better, w/o the TeX-group implication? ORR since the braces is really an illusion anyway, maybe the 2nd "(" should just fall outside of the `\left` anyway, so then there's no braces in the reversion in any case? (which would amount to `TeXDelimiter` needing it's own version of `readDigested`?)

lib/LaTeXML/Engine/TeX_Math.pool.ltxml
```diff
@@ -732,20 +732,21 @@ our %DELIMITER_MAP =
   "\x{21D5}" => { lrole => 'OPEN',    rrole => 'CLOSE', name => 'Updownarrow' },    # ??
   );
 
+sub checkDelimiterHints {
+  my ($stomach, $whatsit) = @_;
+  $whatsit->setProperty(hint => 1) if ToString($whatsit->getArg(1)) eq '.';
+  return; }
+
```
brucemiller, line 739: In fact, since we seem to be accumulating layers of helpers, it's almost like we're not getting any benefit from `TeXDelimiter`. Maybe `TeXDelimiter` should already recognize "." and convert it to a whatsit for `\lx@empty@delimiter`, which constructs an `ltx:XMHint`? Then all the constructors are simple `#`, and they don't need `checkDelimiterHints` or a hint property. We'd still have to tweak `addDelimiterRole` and `augmentDelimiterProperties` to recognize the hint. Whadya think?

PR #2573, opened 2025-06-02
Defer digesting frontmatter to \maketitle

### LINKED ISSUE #1929: expansion of \title argument may happen too early

dginev:
Fixes #1929, as well as [ar5iv#174](https://github.com/dginev/ar5iv/issues/174)

A simple add-on for binding `\AND` also adresses the ~30 reports in arXiv resembling [html_feedback#1284](https://github.com/arXiv/html_feedback/issues/1284)

The PR is a spiritual successor to Tom's #1937. In this attempt, I tried to achieve the minimal goals of:
 - defer frontmatter digestion to `\maketitle` (accumulating the tokens until then)
 - mark certain frontmatter as replacable instead of extensible (to avoid duplicated entries, e.g. `\title`).
 - bind `\And`, `\AND` in the common LaTeX kernel to compensate for the limited `\@maketitle` emulation. I also realized emulating fully isn't actually a nice solution here, as they often get bound to `\end{table} ... \hfill ... \begin{table}` which isn't really useful in the latexml context. It is too low-level, we don't rely on `{table}` for the frontmatter XML.
  
What the PR does not attempt is coordinating any internal macros (such as `\@title`, `\@author`, `\@maketitle`, etc) with LaTeXML's frontmatter API. A more thorough approach may be available after the big kernel upgrade, but I wanted to have a simple alternative available, in case we want it.

I was surprise how quickly I got to a working state while only using the existing tricks in the codebase - the `at@begin@maketitle` accumulator is inspired by `at@begin@document`.

### DIFF
--- a/lib/LaTeXML/Engine/Base_Utility.pool.ltxml
+++ b/lib/LaTeXML/Engine/Base_Utility.pool.ltxml
@@ -162,6 +162,14 @@ AssignValue(frontmatter => {}, 'global');
 
 DefConditionalI('\if@in@preamble', undef, sub { LookupValue('inPreamble'); });
 
+# Some frontmatter tags are replaceable by default, only one per document.
+our %ReplaceableFrontmatterTags = (
+  'ltx:title'    => 1,
+  'ltx:toctitle' => 1,
+  'ltx:subtitle' => 1,
+  'ltx:date'     => 1,
+  'ltx:abstract' => 1,
+  'ltx:keywords' => 1);
 # Add a new frontmatter item that will be enclosed in <$tag %attr>...</$tag>
 # The content is the result of digesting $tokens.
 # \@add@frontmatter[keys]{tag}[attributes]{content}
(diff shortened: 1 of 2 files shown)


dginev:
The abstract needs a little more care...

Edit: Or just slightly more careful timing with the use of `\maketitle`. This is in essence a change of behavior in this approach - the `\maketitle` becomes the time of digestion, and generally the time of DOM insertion for the various frontmatter bits that got accumulated. We would need additional "floating" XML directives if we wanted to ensure some strict ordering of elements.

Ready for a round of review.

brucemiller:
Interesting... and Promising! But...

I'm thinking that the `replace` keyword may be better connected to the ultimate elements (`ltx:title,ltx:toctitle...`) rather than in the command, so you wouldn't need to modify all the class files.

As you noticed when modifying test cases, if you've deferred the digestion till `\maketitle`, you get  no title at all without `\maketitle`. And validators complain if you've got no title element. But is having `\title`, but no `\maketitle` a worthwhile usecase?  I don't know if there's a way of cleverly including (digesting!) a fake `\maketitle` when there otherwise isn't one, but I'd rather avoid adding gratuitous `\maketitle` to all the test cases.

dginev:
@brucemiller after the inspirational feedback you offered today, I found both tricks to make the PR minimal (tests unchanged), while landing the improved behavior. Namely:

- Declare certain tags "replaceable" for `\@add@frontmatter` 
- Add fallback hooks at `\section` and latest at the end of the document. I am not sure if we need both hooks, felt safer and more correct somehow. Btw, the `t/structure/titlepage` test requires the end-of-document fallback, as it has no sections.

Ready for another round of review.

brucemiller (approved):
This looks suspiciously simple! (which is a good thing :>)  Thanks!!

PR #2574, opened 2025-06-03
Enable archived JATS output

dginev:
This is possibly a first of several PRs in refreshing the JATS output. The focus here is on three points:

1. When images are post-processed, use the resulting `@imagesrc` to point to the final asset
2. Treat archived JATS the same as we would HTML
    - there is no manifest comparable to ePub
3. JATS does not require the default resources (CSS/JS assets), should be off by default

An additional cosmetic change is lifting the namespace declarations to the `<article>` root, as they otherwise get very verbose with a math-heavy or image-heavy document.

This should enable the conversion run:
```bash
latexmlc minfig.tex --dest=minfig.zip --format=jats
```

over the sample input:
```tex
\documentclass{article}
\usepackage{graphicx}
\begin{document}
\includegraphics{example-image-a}
\end{document}
```

### DIFF
--- a/bin/latexmlpost
+++ b/bin/latexmlpost
@@ -333,7 +333,9 @@ sub validate_args {
     $dographics = 1 unless defined $dographics;
     $picimages  = 1 unless defined $picimages;
     $is_html    = 0;
-    $stylesheet = "LaTeXML-jats.xsl" if !defined $stylesheet; }
+    # JATS is the only format that does not use default resources.
+    $defaultresources = 0                  if !defined $defaultresources;
+    $stylesheet       = "LaTeXML-jats.xsl" if !defined $stylesheet; }
   elsif ($format eq 'tei') {
     $extension = 'xml' unless defined $extension;
     # Add pMML FIRST, unless already there, or was disabled.
--- a/lib/LaTeXML/Common/Config.pm
+++ b/lib/LaTeXML/Common/Config.pm
@@ -546,6 +546,9 @@ sub _prepare_options {
       elsif ($$opts{format} eq "xml")         { delete $$opts{stylesheet}; }
       else                                    { croak("Unrecognized target format: " . $$opts{format}); }
     }
+    if (!defined $$opts{defaultresources}) {
+      # JATS is the only format that does not use default resources.
+      $$opts{defaultresources} = 0 if $$opts{format} eq 'jats'; }
     # Check format and complete math and image options
     if ($$opts{format} eq 'html4') {
       $$opts{svg} = 0 unless defined $$opts{svg};    # No SVG by default in HTML.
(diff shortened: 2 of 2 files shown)


dginev:
Marking a note from our discussion today: The "top-level namespace declarations" would benefit from reusing the parameterized approach we have for MathML in XHTML, and we may want to expose the exact prefix used as a parameter from the command line.

We didn't get to the bottom of what potential issues could be caused by switching to a single declaration at the root node, so for now I will continue to aim for that (it makes the final XML much more readable to declare the namespaces once)

lib/LaTeXML/resources/XSLT/LaTeXML-jats.xsl
```diff
@@ -556,29 +556,29 @@
 
   <xsl:template match="ltx:Math[@mode='inline']">
     <inline-formula>
-      <xsl:apply-templates select="m:math"/>
+      <xsl:apply-templates select="mml:math"/>
```
brucemiller, line 559: I'm wondering that you've changed more `m` to `mml` than is really necessary, and may mislead the reader as to how namespaces are actually handled in xml. It *seems* to me that only the `article` (line 79) and `xsl:element` (line 575) require it (and possibly a redundant `xmlns:mml` line 24 ?) are required.  After all, we're still importing other stylesheets that use the `m:` prefix.

dginev: It's a very unfortunate legacy reason that I did that. Maybe we can safely ignore the old technicalities in 2025 and stick to `m:`? I am just worried if we'll have some mainstream JATS tools failing.

To quote the exact reasoning, it is from:

https://jats.nlm.nih.gov/archiving/tag-library/1.2/element/mml-math.html

> MathML Namespacing: The vast majority of users of this Tag Set use the DTDs as their primary (or only) model, so this Tag Set is maintained to ensure that users of DTDs and DTD-based tools will work well. Because DTDs do not “play well” with namespaces, and most DTD-based tools are not namespace aware, the MathML namespace has been hardcoded to the prefix “mml” in the DTDs. This means that the element name, in the DTDs, really is <mml:math>.

dginev: I took a look, and upon closer inspection, your idea is a bit too ambitious for the XSLT 1.0 allowances we have. The good news, however, is that there is no downside to consistently using `mml:` in `LaTeXML-jats.xsl`.

If I want to use both `m:` and `mml:` in the stylesheet, I need to declare both, i.e.
```xslt
<xsl:stylesheet
    xmlns:m   ="http://www.w3.org/1998/Math/MathML"
    xmlns:mml   ="http://www.w3.org/1998/Math/MathML"
```

However, that leads to a choice between two undesirable alternatives, when globally outputing the namespace on the `<article>` element. Either
 1. Both `m:` and `mml:` are emitted as declared, or
 2. Neither are declared, when only adding `exclude-result-prefixes="m"`

I think the second case is some limitation in xsltproc, (or maybe a subtlety I don't understand).

In any case, the PR is working properly with just `mml:`, so maybe we can live with that?

Issue #2506, opened 2025-02-02
Problems with package tcolorbox


evoludolab:
The minimal LaTeX file `tcolorbox.tex`:
```
\documentclass{article}
\usepackage{tcolorbox}
\newtcolorbox{boxinfo}{}
\begin{document}
\begin{boxinfo}
\textbf{Note:} This is a boxed note.
\end{boxinfo}
\end{document}
```
takes a *looong* time to convert to XML with:
```
latexml --dest=tcolorbox.xml tcolorbox.tex
```
close to 10 mins on my M3 MacBook Pro... but eventually succeeds! Adding the optional argument `[many]` to the `tcolorbox` package takes equally long but then fails with lots of errors. The first few are: 
```
evoludo % latexml --dest=tcolorbox.xml tcolorbox.tex
hauert@lappi listings % latexml --dest=tcolorbox.xml tcolorbox.tex      
latexml (LaTeXML version 0.8.8) processing listings.tex
Package: expl3
Driver file for pgf: pgfsys-latexml.def
Library (tcolorbox): 'tcbraster.code.tex' version '6.2.0'
Library (tcolorbox): 'tcbskins.code.tex' version '6.2.0'
Library (tcolorbox): 'tcbskinsjigsaw.code.tex' version '6.2.0'
Library (tcolorbox): 'tcbbreakable.code.tex' version '6.2.0'
Package pdfcol Info: Interface disabled because of missing PDF mode of pdfTeX.
Package pdfcol Info: pdfTeX's color stacks are not available.
Library (tcolorbox): 'tcbhooks.code.tex' version '6.2.0'
Library (tcolorbox): 'tcbtheorems.code.tex' version '6.2.0'
Error:undefined:\NewCommandCopy The token T_CS[\NewCommandCopy] is not defined. at tcbtheorems.code.tex; line 181 col 15
Error:undefined:\newtcbtheorem The token T_CS[\newtcbtheorem] is not defined. at tcbtheorems.code.tex; line 181 col 29
Error:unexpected:\group_begin: The control sequence \group_begin: should not appear between \csname and \endcsname at tcbtheorems.code.tex; line 187 col 0
Error:unexpected:\tex_escapechar:D The control sequence \tex_escapechar:D should not appear between \csname and \endcsname at tcbtheorems.code.tex; line 187 col 0
[...]

PR #2575, opened 2025-06-07
avoid breakable mode in tcolorbox

dginev:
Fixes the example reported in https://github.com/brucemiller/LaTeXML/issues/2506#issuecomment-2922287738

I am not sure if there is a general+clean approach to veto-ing the load of a tcolorbox module, but I found a reliable low-level approach. That is, to ensure the initialization code for breakable never runs, by mapping it to the usual unbreakable case.

What I think can be reasoned through is that since latexml is not interested in the breaking aspects imposed by pages, avoiding that module entirely is concpetually more correct than loading its boilerplate code fully.

With this in place, I can focus on a couple of sizing questions, now that the conversion reliably produces HTML+SVG again.

### DIFF
--- a/lib/LaTeXML/Package/tcolorbox.sty.ltxml
+++ b/lib/LaTeXML/Package/tcolorbox.sty.ltxml
@@ -17,6 +17,8 @@ use LaTeXML::Package;
 
 # used in tcbbreakable.code.tex assuming it was defined? so:
 DefRegister('\doublecol@number' => Number(0));
+# Ensure only unbreakable mode is possible
+DefMacro('\tcb@init@breakable', '\tcb@init@unbreakable', locked => 1);
 
 RequirePackage('expl3');
 RequirePackage('xparse');


dginev:
With this PR, the remaining problem in https://github.com/dginev/ar5iv/issues/529 was removing the extra stylistic padding in lists. That leads to for example this tcolorbox SVG, made of 3 itemizations, some of which nested



The ar5iv CSS currently adds a `1em` top margin for adjacent list items (`.ltx_item + .ltx_item`), which naturally can only work out if there aren't any fixed-height visual ancestors, such as the colored box frame. Disabling that is very easy with a
```css
foreignObject .ltx_item + .ltx_item {
  margin-top: 0;
}
```

So as a minimal requirement, this PR with the associated CSS tweak get us to visually usable SVGs in that article.

dginev:
Update: @brucemiller I checked this PR patch with your new modes work in #2580 as a base. Sadly the patch is still needed - the "breakable" variants still appear to enter an infinite loop even with the Modes improvement. And similarly, the deactivation patch here continues to avoid the issue and terminates in a timely manner.

But there appears to be a regression with #2580 - the box heights are not getting accumulated anymore, so the resulting SVG appears to mean to contain a single line. Maybe something to investigate there before merging that PR.

brucemiller (approved):
That seems simple enough; Thanks!

Issue #2496, opened 2025-01-23
weird error for ntheorem depending on order of commands


lamBOOO:
I get an error for the `ntheorem` package that seems to weirdly depend on the order of commands:

I redefine the `break` inbuild theorem style and then also define another style

```latex
\documentclass{article}
\usepackage{ntheorem}

\renewtheoremstyle{break}%
{\item[\hskip\labelsep ##1\ HELLO  ##2]\normalfont}%
{\item[\hskip\labelsep ##1\ ##2]{(##3)}\newline\normalfont}

\newtheoremstyle{plain1}%
{\item[\hskip\labelsep ##1\ HELLO  ##2]\normalfont}%
{\item[\hskip\labelsep ##1\ ##2]{(##3)}\newline\normalfont}

\theoremstyle{break}
\newtheorem{theorem}{Theorem}
\theoremstyle{plain1}
\newtheorem{remark}{Remark}

\begin{document}
\begin{theorem}
  This is a theorem.
\end{theorem}
\begin{remark}
  This is a theorem.
\end{remark}
\end{document}
```
and running ` latexmlc --destination=paper.html --format=html5 --path=. test.tex`, I get the following error:

```bash
latexmlc (LaTeXML version 0.8.8)
processing started Thu Jan 23 22:00:29 2025
Error:undefined:\renewtheoremstyle The token T_CS[\renewtheoremstyle] is not defined. at test.tex; line 4 col 18
Error:misdefined:# The token T_PARAM[#] should never reach Stomach! at test.tex; line 6 col 35
Error:misdefined:# The token T_PARAM[#] should never reach Stomach! at test.tex; line 6 col 36
Error:undefined:{theorem} The environment {theorem} is not defined. at test.tex; line 18 col 0
Warning:undefined:theoremstyle Unknown theorem style '\lx@tag', reverting to 'plain'. at test.tex; line 22 col 3
Conversion complete: 1 warning; 4 errors; 2 undefined macros[\renewtheoremstyle, {theorem}] (See /Users/lamberttheisen/Desktop/theoremstyle/test.latexml.log)
Post-processing complete: No obvious problems
Status:conversion:2
```

However, if I just change the order, i.e. first define my own style and then redefine the inbuilt style:

```latex
\documentclass{article}
\usepackage{ntheorem}

\newtheoremstyle{plain1}%
{\item[\hskip\labelsep ##1\ HELLO  ##2]\normalfont}%
[...]

PR #2586, opened 2025-07-04
Add some ntheorem's style commands with the right signature

lamBOOO:
Fixes https://github.com/brucemiller/LaTeXML/issues/2496


t/theorem/ntheoremstyle.tex
```diff
@@ -0,0 +1,24 @@
...
+\newtheorem{theorem}{Theorem}
+\theoremstyle{plain1}
+\newtheorem{remark}{Remark}
+
+\begin{document}
+\begin{theorem}
+  This is a theorem.
+\end{theorem}
+\begin{remark}
+  This is a theorem.
```
dginev, line 22: minor: The `{remark}` text could have said `This is a remark.`

brucemiller:
The perl looks perfectly fine to me. Your (re)definitions of `\newtheoremstyle` don't actually *use* the heading formatters, but (as noted in the comment on #2496), it's probably not really feasible.
So, good job, and thanks for the contribution!

PR #2592, opened 2025-07-17
patch: correct argument list for PassOptions

dginev:
This is a small patch that was discovered while testing.

PassOptions accepts `PassOptions($name, $ext, @options)`, but some code I introduced was wrongly passing in an array reference in 2 cases. A revealing use was the latex:

```tex
\PassOptionsToPackage{numbers, compress}{natbib} 
```


brucemiller (approved):
Aha!

PR #2595, opened 2025-07-23
New orcidlink.sty binding, with test case

brucemiller:
Provides a simple, fast, binding for `orcidlink` providing commands which creates an SVG icon which link to an orcid (https://orcid.org/) in several formats (long, compact, inline).

@duetosymmetry I've taken the liberty of tweaking size & position to correspond more with orcid's apparent guidelines, than the current latex sty; namely the "id" within the icon's circle align with the line height & baseline, rather than the circle itself.  Also: in 2022 texlive, neither hyperref, nor tikz load etoolbox, so no `\ifstrempty`.

@xworld21 Sorry to hijack, but i needed a fun diversion, and graphics is always good for that!

closes: #2395 
fixes: #2386

### DIFF
--- a/lib/LaTeXML/Package/orcidlink.sty.ltxml
+++ b/lib/LaTeXML/Package/orcidlink.sty.ltxml
@@ -17,11 +17,14 @@ use warnings;
 use LaTeXML::Package;
 #======================================================================
 # Note: the original package loads tikz (!) and assumes etoolbox(?)
-RequirePackage('hyperref');
+RequirePackage('hyperref');     # gratuitous
+
+DefConstructor('\lx@orcidlink{}{}',
+  "<ltx:ref title='ORCID #1' class='ltx_orcid' href='https://orcid.org/#1'>#2</ltx:ref>");
 
 # but avoid \ifstrempty (etoolbox)
 DefMacro('\orcidlinkX{}{}{}',
-  '\href{https://orcid.org/#2}{\ifx&#1&\else#1\,\fi\orcidlogo\ifx&#3&\else\,#3\fi}',
+  '\lx@orcidlink{#2}{\ifx&#1&\else#1\,\fi\orcidlogo\ifx&#3&\else\,#3\fi}',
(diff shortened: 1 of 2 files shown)


dginev:
First, a test run:

I grabbed one of the reported arXiv issues to test [here](https://github.com/arXiv/html_feedback/issues/4363), which uses [arXiv:2504.20592](https://arxiv.org/abs/2504.20592).

The resulting SVGs looks somewhat large-ish in the author list? If I am seeing the PDF right they end up at 1em height or so. 
Edit: I now see your comment that you tweaked the sizing. Nice, but we should also make sure existing use of the package looks lovely :-)

The links work well, clicked around successfully.

brucemiller:
Indeed. I was motivated by https://info.orcid.org/documentation/integration-guide/orcid-id-display-guidelines/, which @duetosymmetry had referenced. They don't explicitly mention the sizing, but illustrate it. The pdf version (from the .sty) certainly has smaller icons. Possibly too small, as the "id" within the disk is almost not readable (given the lack of color contrast). So you either end up with "id" matching the current font size, or the whole disk.

If @duetosymmetry made that choice intentionally, I could certainly be persuaded to follow suite.

dginev (approved):

duetosymmetry:
@brucemiller I hacked up the package before I was aware of the ORCID display guidelines. When I became aware of them, I didn't try to emulate getting the baseline of id to line up with the baseline of the surrounding text... that would be a bit above my LaTeX pay grade. So, I don't have any strong feelings about the details of the placement. You did a better job than I did!

dginev:
@brucemiller may I request an `class="orcid_id"` or similar attribute for the SVG elements? 

I think I may slightly tweak some spacing/sizing details.

duetosymmetry:
Handling the sizing/styling in (external) CSS is definitely more flexible than hardcoding `width="1.7em" height="1.7em"` in the `.sty.ltxml` file

PR #2598, opened 2025-08-01
allow intent and arg attributes in schema and pres MathML output

dginev:
This is a minimal changeset that enables bindings to deposit `intent` and `arg` attributes on XMath nodes, making experimentation possible outside of mainline grammar-based XMath generation.

Two pieces to the PR:
 - allow `intent` and `arg` attributes in the schema
 - propagate `intent` in the top-level presentation node and `intent` and `arg` in all inner nodes.

This can certainly be developed further, but it is really helpful as a starting point - otherwise there is no way to persist a deposited `intent` attribute through post-processing.


dginev:
Rebased to the latest master branch, as it had moved forward.

Dec 13, 2025: rebased again.

PR #2601, opened 2025-08-06
Add directory on whatsin to the search paths

dginev:
This was an oversight for a rare kind of input organization such as:
```
mydocument/00README.json
mydocument/main/main.tex
mydocument/sections/section-1.tex
mydocument/sections/section-2.tex
mydocument/figures/figure-1.png
...
```

With the invocation
```
latexmlc --whatsin=directory mydocument/
```

The `00README.json` will specify the path to the `main.tex` as the top-level file, and while the `sourcedirectory` was being correctly updated to `mydocument/` that wasn't actually added to the search paths. So when a conversion starts at `mydocument/main/main.tex`, an `\input{sections/section-1}` will no longer have visibility at the level above.

Simple fix, just add a directory input to the search paths -- in a way it is obviously correct.

### DIFF
--- a/lib/LaTeXML/Common/Config.pm
+++ b/lib/LaTeXML/Common/Config.pm
@@ -220,6 +220,10 @@ sub read {
       $$opts{whatsin} = 'archive'; }
     elsif ($$opts{source} && ($$opts{source} =~ /\/$/)) {
       $$opts{whatsin} = 'directory'; } }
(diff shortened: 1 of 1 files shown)


dginev:
This was uncovered in https://github.com/arXiv/html_feedback/issues/4542
and is testable with `arXiv:2508.00246v1`

PR #2607, opened 2025-08-13
Ar xiv sandbox 2

brucemiller:
Next batch of arXiv sandbox test patches. The main one here is to better manage associating the generated xml elements with the digested boxes responsible for them. In particular, as boxes get accumulated, to avoid keeping them in the cache where they cannot be garbage collected; this was causing massive memory use in certain cases.


dginev (approved):
Very glad we unearthed the memory mystery - will rerun and check if this covers all cases.

PR #2609, opened 2025-08-18
remove non-portable glossaries test

dginev:
I am at a bit of a loss how to remediate this, without introducing "pins" for texlive version in our test suite. What I am observing is that in texlive 2023 (and maybe newer? haven't tested) the glossaries test produces a different PDF, namely the `\GLS` macro suite fails at producing all caps and instead leaves the argument untouched.

One way to proceed is to "exempt" texlive 2023 from the glossaries test. That would require building some infrastructure (such as a `texlive_max` condition for the test harness). 

Another approach, much simpler to accomplish, is what I have done here - comment out the non-portable test line. Somewhat defeatist, but since we did not have consensus on introducing `texlive_max`, I do not have other ideas how to reconcile the two different behaviors in different texlives.

Ideas welcome.

Screenshot from 2025-08-18 12-49-50


dginev:
I suggest merging here to move along the work on CI in #2668

brucemiller:
Odd, I tried a recent texlive that I *think* was 2023, and still get all caps; it seems from the documentation that the intention is still that it should be all caps.

PR #2620, opened 2025-08-26
arXiv sandbox 6

brucemiller:
A few more guards against Fatals. And starting to address some malformed Errors, particularly with algorithm2e.


dginev (approved):
Looks great, thanks!

PR #2621, opened 2025-08-27
Use 'math', not 'inline_math'

brucemiller:
Use `math`, as Knuth does, to name the mode, rather than `inline_math`, avoiding confusion and weirdness. Should have been part of the earlier mode PR, but better late than never.

### DIFF
--- a/lib/LaTeXML/Engine/TeX_Logic.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Logic.pool.ltxml
@@ -120,7 +120,7 @@ DefConditionalI('\ifvmode', undef, sub {
 DefConditionalI('\ifhmode', undef, sub {
   LookupValue('MODE') =~ /horizontal$/; });
 DefConditionalI('\ifinner', undef, sub {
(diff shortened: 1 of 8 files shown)


lib/LaTeXML/Engine/TeX_Logic.pool.ltxml
```diff
@@ -120,7 +120,7 @@ DefConditionalI('\ifvmode', undef, sub {
 DefConditionalI('\ifhmode', undef, sub {
   LookupValue('MODE') =~ /horizontal$/; });
 DefConditionalI('\ifinner', undef, sub {
-  LookupValue('MODE') =~ /^(?:restricted_horizontal|internal_vertical|inline_math)$/; });
+  LookupValue('MODE') =~ /^(?:restricted_horizontal|internal_vertical|math)$/; });
```
dginev, line 123: This regex will now also match `display_math`. Is that intentional?

dginev (approved):
Nice.

I double-checked the `^` and `$` in the regexes, looks good to merge.

PR #2622, opened 2025-08-27
Allow apostrophe in color names

dginev:
This PR is fundamentally a single character change adding apostrophe to the name regular expression for xcolor. This is used in `arXiv:2404.03577` and confused me a bit while looking at the dark mode map. The fallback when color parsing fails is black, and that looks quite broken as it overlaps the foreground color.

The PR also lints and adds a small test showing off the kind of use seen in that arXiv article.


brucemiller (approved):
Weird. But OK. Thanks!

PR #2624, opened 2025-09-03
backquoted character 0

brucemiller:
Make sure backquoted character 0 gets the character code of the actual character (darn Perl)


lib/LaTeXML/Core/Gullet.pm
```diff
@@ -890,7 +890,7 @@ sub readNormalInteger {
     return Number(hex(readDigits($self, '0-9A-F', 1))); }
   elsif ($token->equals(T_OTHER("`"))) {                             # Read Charcode
     my $next = readToken($self);
-    my $s    = ($next && $$next[0]) || '';
+    my $s    = ($next && $$next[0]) // '';
```
dginev, line 893: Oh no, the bars are falling down! Will we see `--` next? (darn Perl)

Issue #2588, opened 2025-07-12
Copy to clipboard for lstlisting


evoludolab:
When translating code listings to html, LaTeXML conveniently creates download links for code snippets. It would be even more user-friendly if this could be changed to a copy-to-clipboard-button.

I started playing around with `LaTeXML-block-xhtml.xsl` and made some progress but it's far from satisfactory... too big a hack to warrant a PR. More specifically, I started by replacing line 153 in `LaTeXML-block-xhtml.xsl` with
```
<xsl:element name="button" namespace="{$html_ns}">
    <xsl:attribute name="onclick">toClipboard(this.parentNode.parentNode)</xsl:attribute>
    <xsl:text>📋</xsl:text>
</xsl:element>
```

However, I would be happy to continue giving this a try if you were willing to give me some directions. More specifically, the issues I ran into were:

- is this the right approach to begin with?
- the clipboard icon is ugly and doesn't scale. An icon like GitHub uses above would be way nicer. Create copy icon and where to place it?
- the custom javascript code for `toClipboard` (or similarly named) method needs to live somewhere. Create javascript code for copying and where to place it?
- the argument `this.parentNode.parentNode` to `toClipboard` works but is ugly and so I expect that I am missing something...
- make this configurable?

Any help and pointers are much appreciated!

dginev:
I certainly prefer copying listings to clipboard on the web, so I like your suggestion.

The difference to keep in mind is that the current download solution works in browsers with JS disabled.

---

Maybe a better direction would be to add a new JS asset (instead of modifying the XSLT) which uses the available data source and instead of download switches the onclick behavior to a clipboard copy?

For customization, if we wanted to support multiple variants in the XSLT we would need to introduce a parameter and use it with `--xsltparameter=name:value`. Btw, If you are making major modifications to one of the stylesheets you probably want to copy it under a new name and load it as an alternative. You could even substitute the entire XSLT pipeline by specifying `--stylesheet=name`.

To add a new conditional JS asset to the final output, we have `--javascript=name`

evoludolab:
Thanks for the suggestions. This makes a lot of sense. However, the scale and complexity of the LaTeXML project is intimidating... 

[...]

PR #2629, opened 2025-09-06
add copy button for lstlistings

evoludolab:
This PR aims to address #2588.

The `lstlisting.js` adds a hook to replace all download links in HTML documents by a download button after the document has been loaded.

Following @dginev's encouragement, I polished my JavaScript and created my own icons to avoid any copyright issues. All SVG icons are in the CSS in order to minimize the number of files, while maintaining customizability.

I marked the PR as a draft because I'm struggling with
- where to make changes or place the files,
- how to best name things,
- how to activate this JS extension during post-processing.

Currently, I pass a JS file to `LaTeXML` with `--javascript=<custom.js>`, which includes the code from `lstlisting.js` and the resulting buttons work as expected.

I'm looking forward to your feedback/guidance on this - or watch what changes you make to the code 😄.


evoludolab:
I had marked this PR as a draft only because I wasn't sure where the CSS and JS code for the copy button of lstlistings should go... Nothing has changed but I will now mark it as ready for review - just bear with me that the above will require some input and guidance from your side.

brucemiller:
Oooh, I like it! I'm gonna want to use it myself!

But I'm a little concerned about naming & packaging. For those that use listing, but not the clipboard, the CSS load is a bit heavy. It would probably be good to separate that out and name both files `ltx-listing-clipboard.(css|js)`; the user would have to add both:
```
--css=ltx-listing-clipboard.css --javascript=ltx-listing-clipboard.js
```
Kinda long.   Or possibly even just `ltx-clipboard.(css|js)` if it seems that it might (eventually) be adaptable to other clipboardable data?

brucemiller:
Oh, and whatever names, they'll need to be added to the MANIFEST file

xworld21:
Very quick comment, because I was looking into this just today as I'd personally like to have both copy & dowload buttons... The clipboard bits can be implemented directly into an onclick handler as
```javascript
url=event.target.nextElementSibling.href;clipboard.writeText(atob(subtring(url.substring(url.indexOf(',')+1))))
```
assuming you add the button right in front of the existing download link (code may be slightly wrong). That's more robust than hiding the line numbers, selecting the text etc.

I also wanted to point out that the download button has its uses, e.g. when the file name is known (because it was inserted by reading a file, or by specifying the `@name` attribute). E.g. try the download button of https://vlmantova.github.io/bookml/get-started.html#get-started-on-overleaf

evoludolab:
Thanks for the feedback! The problem is I'm completely out of my depth with anything perl and at best a novice when it comes to javascript... 

The CSS is mainly a bit heavy because I created and embedded custom svg backgrounds for the buttons and feedback. @xworld21's approach with unicode is of course much more elegant. Albeit some kind of visual feedback seems quite standard (and expected by the user).

With some guidance I'm happy to spend more time on this PR but otherwise I'll just stand by and watch what is happening with #2588.

PR #2630, opened 2025-09-07
allow unbalanced \read and \readline to succeed

### LINKED ISSUE #2429: greek babel broken on texlive 2023

dginev:
Fixes #2429 . But it also merits further review and discussion.

Explanation of the diagnostics and possible choices in the details:

<details>

I first reduced the failing greek babel load to a minimal TeX equivalent:
```tex
\begin{filecontents}{test.ini}
sigma.final.3.0 = { ()"()[σΣ] } }
\end{filecontents}
\documentclass{article}
\begin{document}
\makeatletter

\newread\stream
\openin\stream=test.ini
\loop
    \if T\ifeof\stream F\fi T\relax % Trick, because inside \loop
    \read\stream to \aline
    \message{\meaning\aline}
\repeat
done.
\end{document}
```

Note that the `sigma.final.3.0` line has an actual error, and has been checked in with it in the texlive distribution - there is one more closing brace on that line than needed, making it unbalanced.

Next, I checked what TeX does exactly during that `\read`, and two of the relevant snippets appear to be:
Screenshot From 2025-09-06 21-25-25

This implies that whatever the inner read_toks read does, it gets directly assigned to the CS. And within read_toks we get to the rule:
Screenshot From 2025-09-06 21-52-38

Note the comment: `unmatched ‘}’ aborts the line`

Faced with that, I checked a couple of variants, and indeed TeX drops everything upto the end of the line when an unbalanced `}` is encountered. So the question is how best to implement that behavior - I am not quite sure I have the answer.

</details>

---

As a first attempt, I did one of the more obvious conservative changes:
 - Introduce a new option `allowUnbalanced` which permits an expansion to be unbalanced, by emitting an Error instead of a Fatal (or should this be even an Info?) and trimming down to the balanced portion of the tokens. Note that pdflatex is completely quiet when it does this trim.
 - Implements the option by adding `Tokens::toBalanced` and using it when necessary in `Expansion::new`.

One could also do all of this specifically inside the DefPrimitive code for `\read`, but the flag seems more reusable for `\readline`, which appears rather similar?

I think this is useful enough to share already, but definitely in need of discussion.

### DIFF
--- a/lib/LaTeXML/Engine/TeX_FileIO.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_FileIO.pool.ltxml
@@ -29,10 +29,10 @@ Let('\@currname', '\lx@empty');
 Let('\@currext',  '\lx@empty');
 
 DefMacro('\lx@pushfilename',
-    '\xdef\@currnamestack{{\@currname}{\@currext}{\the\catcode`\@}\@currnamestack}');
+  '\xdef\@currnamestack{{\@currname}{\@currext}{\the\catcode`\@}\@currnamestack}');
 DefMacro('\lx@popfilename', '\expandafter\lx@p@pfilename\@currnamestack\@nil');
 DefMacro('\lx@p@pfilename {}{}{} Until:\@nil',
-   '\gdef\@currname{#1}%
+  '\gdef\@currname{#1}%
     \gdef\@currext{#2}%
     \catcode`\@#3\relax
     \gdef\@currnamestack{#4}');
@@ -80,13 +80,14 @@ DefPrimitive('\read Number SkipKeyword:to SkipSpaces Token', sub {
       AssignValue(PRESERVE_NEWLINES => 2);    # Special EOL/EOF treatment for \read
       AssignValue(INCLUDE_COMMENTS  => 0);
       my @tokens = ();
-      my ($t, $level) = (undef, 0);
+      my ($t, $level, $skip) = (undef, 0, undef);
       while ($t = $mouth->readToken) {
         my $cc = $t->getCatcode;
-        push(@tokens, $t) unless $cc == CC_MARKER;    # End of line marker
-        $level++ if $cc == CC_BEGIN;
-        $level-- if $cc == CC_END;
-        last     if !$level && $mouth->isEOL; }
+        $level++  if $cc == CC_BEGIN;
+        $level--  if $cc == CC_END;
+        $skip = 1 if $level < 0;
+        push(@tokens, $t) unless $skip || ($cc == CC_MARKER);    # End of line marker
+        last if ($skip || !$level) && $mouth->isEOL; }
       $stomach->egroup;
       DefMacroI($token, undef, Tokens(@tokens), nopackParameters => 1); }
     return; });


brucemiller:
Sadly I don't have a 2023 at hand right now. It would be good to understand how this comes up; it shouldn't happen! This shows up rarely but occasionally in arXiv too.
I don't know that adding an (undeclared!) DefMacro option helps, vs simply changing Fatal to Error.

dginev:
I didn't know if there is appetite to entirely remove the Fatal. If there is, I can do so, and avoid the new option. Is that OK?

As to how it comes up - I tried to be extremely detailed in the "Details", feel free to expand them. There is an unbalanced brace checked in babel, which leads to all the troubles.

brucemiller:
Why'd you hide the most important point?!? That's our infallible guide: "What TeX does"!
And interestingly,  *silently* discards the offending rest-of-line, but it does it in `read_toks` which corresponds to the inner loop within `\read` (and for `\readline` hidden away in Mouth's `readRawLine`). So that's where the checking and discarding needs to happen.

The unbalanced tokens never gets into the macro at all (which is `store_new_token`), so Expandable and Tokens need no changes.

dginev:
Thank you for clearly expressing that preference. I have switched to a small skip flag in `\read` which resolves the reported issue.

The `\readline` treatment was a frivolous upgrade that has little to do with the original TeX, so I have removed it. Feel free to merge if you like this formulation, or to make a different variant. It's a simple change.

brucemiller:
Cool! Thanks!! That handles the immediate issue, but can I have a glass of milk to go with it? Namely, your comments sent me on a Whatabout query. Check out this evolution of your nice test case (remember to `rm *.ini` if you've run pdflatex before)
```
\begin{filecontents}{balanced.ini}
sigma.final.3.0 = { { ()"()[σΣ] } }
\end{filecontents}
\begin{filecontents}{longbalanced.ini}
sigma.final.3.0 = { { ()"()[σΣ] }
} more {
}
\end{filecontents}
\begin{filecontents}{extraclose.ini}
sigma.final.3.0 = { ()"()[σΣ] } }
\end{filecontents}
\begin{filecontents}{extraopen.ini}
sigma.final.3.0 = { { ()"()[σΣ] }
\end{filecontents}
\documentclass{article}
\newread\stream
\newcount\nlines
\def\tester#1#2{%
 \message{Test \string#1 on #2:^^J}%
 \nlines=0\relax
 \openin\stream=#2
 \loop
    \if T\ifeof\stream F\fi T\relax % Trick, because inside \loop
    \advance\nlines 1\relax
    \read\stream to \aline
    \message{Line \the\nlines: \meaning\aline^^J}
\repeat
\message{done.^^J}}

\begin{document}
\tester\read{balanced.ini}
\tester\read{longbalanced.ini}
\tester\read{extraclose.ini}
\tester\read{extraopen.ini}

\tester\readline{balanced.ini}
\tester\readline{longbalanced.ini}
\tester\readline{extraclose.ini}
\tester\readline{extraopen.ini}
\end{document}

```
So, as you found, TeX will discard unbalanced close braces, along with the remainder of the line. But, if there are unbalanced open braces on a line, TeX will continue reading complete lines (as does LaTeXML's version) until balanced. However if it reaches EOF in that state, it will signal an Error. If you continue from that error,  it stores the unbalanced tokens in the macro, which may disturb further processing; possibly a Fatal is called for?

It also reveals that `\readline` has exactly the same behavior as `\read`, other than its catcode changes. That kills my naive idea of a modified `Explode`. Probably best would be to pull out the body of `\read` up until the `DefMacroI` into a helper function, which gets also used by `\readline`, but then dumb down the token catcodes appropriately...?

brucemiller:
?? what happened?

PR #2644, opened 2025-09-24
Typos

brucemiller:
This PR fixes a couple of minor issues that turned up: (1) The MathParser could mindlessly try to set the internal `_box` attribute (recording the Box responsible for an XML node) to the id of the box, rather than the actual box.  (2) Boxes record their reversion which should be Tokens, not a string.


dginev (approved):
Thanks, sounds right.

(Windows CI has an independent failure)

brucemiller:
Awesome! I can still get PR reviews!! :> I'll have some more later, but in the meantime, go have a vacation!

PR #2645, opened 2025-09-25
Reading from mouth

brucemiller:
This is a somewhat overdue tweak of the API for `$gullet->readingFromMouth(...` that allows the source to be a string or `Tokens`, rather than only a `LaTeXML::Core::Mouth`. In those new cases, an appropriate `Mouth` is created as needed, simplifying the most common use cases.

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -101,17 +101,30 @@ sub flush {
   $$self{mouthstack} = [];
   return; }
 
-# Do something, while reading stuff from a specific Mouth.
+# Do something, while reading stuff from a specific source Mouth.
+# If $source is a string or Tokens, a Mouth will be created to process it.
 # This reads ONLY from that mouth (or any mouth openned by code in that source),
 # and the mouth should end up empty afterwards, and only be closed here.
 sub readingFromMouth {
-  my ($self, $mouth, $closure) = @_;
+  my ($self, $source, $closure) = @_;
+  my $mouth;
+  my $intokens;
+  my $sourcetype = ref $source;
+  if (! $sourcetype) {          # for a string, create Mouth to read its content
+    $mouth = LaTeXML::Core::Mouth->new($source); }
+  elsif ($sourcetype =~ /^LaTeXML::Core::Tokens?$/) { # Tokens will be unread into Mouth
+    $mouth = LaTeXML::Core::Mouth->new();
+    $intokens = $source; }
+  else {
+    $mouth = $source; }         # Hopefully is a Mouth.
   openMouth($self, $mouth, 1);    # only allow mouth to be explicitly closed here.
+  $self->unread($intokens) if $intokens; # Preload the mouth
   my ($result, @result);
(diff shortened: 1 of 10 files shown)


lib/LaTeXML/Core/Gullet.pm
```diff
@@ -101,17 +101,30 @@ sub flush {
...
+  my $mouth;
+  my $intokens;
+  my $sourcetype = ref $source;
+  if (! $sourcetype) {          # for a string, create Mouth to read its content
+    $mouth = LaTeXML::Core::Mouth->new($source); }
+  elsif ($sourcetype =~ /^LaTeXML::Core::Tokens?$/) { # Tokens will be unread into Mouth
+    $mouth = LaTeXML::Core::Mouth->new();
+    $intokens = $source; }
+  else {
+    $mouth = $source; }         # Hopefully is a Mouth.
```
dginev, line 119: Could we check that it is indeed a Mouth and emit an Error or Fatal otherwise?

lib/LaTeXML/Core/Parameter.pm
```diff
@@ -113,17 +113,8 @@ sub reparse {
...
-        if (@tokens    # Strip outer braces from dimensions & friends
-          && ($$self{type} =~ /^(?:Number|Dimension|Glue|MuDimension|MuGlue)$/)
-          && ($tokens[0]->getCatcode == CC_BEGIN) && ($tokens[-1]->getCatcode == CC_END)) {
-          shift(@tokens); pop(@tokens); }
-        $gulletx->unread(@tokens);    # but put back tokens to be read
-        my $value = $self->read($gulletx);
-        $gulletx->skipSpaces;
-        return $value; }); } }
+    return $gullet->readingFromMouth($tokens, sub {
+      $self->read($gullet); }); } }
```
dginev, line 117: Just checking, is `$self->read` "more correct" than the previously hardcoded reading logic?

lib/LaTeXML/Core/Parameters.pm
```diff
@@ -78,13 +78,9 @@ sub readArgumentsAndDigest {
...
   if (defined $tokens) {
-    return $gullet->readingFromMouth(LaTeXML::Core::Mouth->new(), sub {    # start with empty mouth
+    return $gullet->readingFromMouth($tokens, sub {
         no warnings 'recursion';
-        my ($gulletx) = @_;
-        $gulletx->unread($tokens);                                         # but put back tokens to be read
-        my @values = $self->readArguments($gulletx);
-        $gulletx->skipSpaces;
-        return @values; }); }
+        $self->readArguments($gullet); }); }
```
dginev, line 83: It looks like this snippet no longer does `->skipSpaces` at the end. Is that intentional?

dginev (approved):
Nice cleanup. I like that it is backwards-compatible as well.

lib/LaTeXML/Core/Gullet.pm
brucemiller: You want more than a "Hopefully" comment?  Sure, next commit fixes that :>

PR #2649, opened 2025-09-27
[WIP] Updates for Windows CI

dginev:
Work in progress, investigating Windows CI configuration.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -65,6 +65,7 @@ WriteMakefile(NAME => 'LaTeXML',
     'Image::Size'       => 0,
     'IO::String'        => 0,
     'IO::Handle'        => 0,
+    'IO::Socket'        => 0,
     'JSON::XS'          => 0,
     'LWP'               => 0,
     'MIME::Base64'      => 0,      # Core


dginev:
@brucemiller I investigated today and I think a simple rebuild of our Strawberry Perl cache (for the newer windows-latest image) is all we needed to recover the Windows CI back to ✅ 

Feel free to merge here if you wish.

Makefile.PL
```diff
@@ -65,6 +65,7 @@ WriteMakefile(NAME => 'LaTeXML',
     'Image::Size'       => 0,
     'IO::String'        => 0,
     'IO::Handle'        => 0,
+    'IO::Socket'        => 0,
```
dginev, line 68: Not strictly necessary to add `IO::Socket` today, but might as well future-proof our dependency list, given that it was Socket.c that was tripping up CI.

xworld21:
While you are at it, is this a good time to extend/shift the Perl matrix to 5.42.0.1? There was a long period without Strawberry Perl releases, but they are pretty good now at keeping up with upstream. Chocolatey has added
- 5.42.0.1
- 5.40.2.2
- 5.38.4.1
- 5.36.3.1
- 5.34.3.1

dginev:
Thanks @xworld21 that is indeed already part of our release planning. We are working with Tom Wiesing to update the perls and texlives in CI, I will make separate PRs for that. I am on vacation until October 8, so expect some slowness.

dginev:
Apparently there is also an official GitHub notice about this here:

https://github.blog/changelog/2025-07-31-github-actions-new-apis-and-windows-latest-migration-notice/

I saw the CI failed _again_ with the latest merges, and if I am reading their notice correctly, it may have still used the older runners. I guess I should expire all caches again _tomorrow_ as they said they are completing the transition by October. "Bleeding edge" is right.

brucemiller:
FWIW, it seems that this patch isn't quite enough to get Windows CI working again.

dginev:
See comment above, the windows-latest image is in flux until tomorrow

PR #2650, opened 2025-09-27
extract first rather than last image of multipage images

### LINKED ISSUE #2639: Multi-page .pdf figures in HTML don't default to page 1 as the
Compare figure 2 in https://arxiv.org/html/2509.14185v1 vs https://arxiv.org/pdf/2509.14185

This differs from because figure had multiple pages. The PDF rendering defaults to page 1, but the HTML rendering defaults to a different page (the last one?) instead.

xworld21:
Fix #2639.

(Appending `[N]` to the file name is the documented way of selecting a particular frame, and it starts from 0, see https://imagemagick.org/Magick++/Image++.html.)

### DIFF
--- a/lib/LaTeXML/Util/Image.pm
+++ b/lib/LaTeXML/Util/Image.pm
@@ -448,6 +448,8 @@ sub image_read {
   if (!$source) {
     Error('imageprocessing', 'read', undef, "No image source given"); return; }
   return unless $source;
+  # read first page by default
+  $source .= '[0]';
   my $image = image_object();
   # Just in case this is pdf, set this option; ImageMagick defaults to MediaBox (Wrong!!!)
   image_internalop($image, 'Set',  option => 'pdf:use-cropbox=true') or return;


dginev:
Could it really be that easy? Thanks! I will test with the PR against some of the arXiv reports and write back.

xworld21:
> Could it really be that easy? Thanks! I will test with the PR against some of the arXiv reports and write back.

Yes!

I was going to add that it should also work by patching this line:
https://github.com/brucemiller/LaTeXML/blob/e64d7f5f5139b2a57cb0bc6585400b403c94c95c/lib/LaTeXML/Util/Image.pm#L466

However, reading the correct page from the start is obviously a better solution. If I understand correctly, the current code reads all the pages into memory, *transforms all of them*, then squashes them into one. Not ideal.

I am looking at implementing the `page` option now.

xworld21:
#2677 is better in every way.

PR #2652, opened 2025-09-30
Support calc.sty's sneaking redefinitions

brucemiller:
Here's a tricky one!  LaTeX defines several utility macros `\setcounter,\setlength` (along with `\addtocounter,\addtolength` and others) which basically just assign registers to a Number or Dimension. They are used fairly extensively, and also in several other packages, notably graphicx. Along comes the calc package which supports simple arithmetic calculations of counters and lengths. However, it *redefines* `\setcounter, \setlength` so that automagically, most normal LaTeX commands (eg `\makebox, \includegraphics`) suddenly also accept calc syntax to assign numbers and lengths, not to mention that they also ignore extra braces wrapping the quantities.

These enhancements should only affect quantities that get passed around as macro arguments, eg. `{Dimension}`, but low-level TeX usage like `Dimension` should *not* be affected.  So, the idea here is to define explicitly parameter types `{Number}, {Dimension}`. Previously these would have read a regular arg, then reparsed it as a Number, but now if calc has been loaded, will reparse as calc expressions of the correct type.  Similarly, optional versions `[Number], [Dimension]` are defined. 

[I thought of having calc simply redefine the parameter types, but internal parameter data will already have been incorporated into  Definition's Parameter objects]

### DIFF
--- a/t/complex/figure_dual_caption.xml
+++ b/t/complex/figure_dual_caption.xml
@@ -12,12 +12,12 @@
       <tag role="typerefnum">Figure 2</tag>
     </tags>
     <figure align="center" class="ltx_figure_panel ltx_minipage" vattach="middle" width="172.5pt" xml:id="S0.F2.fig1">
-      <graphics candidates="../graphics/none.png" class="ltx_centering" graphic="../graphics/none.png" options="width=411.93767pt,keepaspectratio=true" xml:id="S0.F2.g1"/>
+      <graphics candidates="../graphics/none.png" class="ltx_centering" graphic="../graphics/none.png" options="width=411.9pt,keepaspectratio=true" xml:id="S0.F2.g1"/>
       <toccaption class="ltx_centering"><tag close=" ">1</tag>Left figure.</toccaption>
       <caption class="ltx_centering"><tag close=": ">Figure 1</tag>Left figure.</caption>
     </figure>
     <figure align="center" class="ltx_figure_panel ltx_minipage" vattach="middle" width="172.5pt" xml:id="S0.F2.fig2">
-      <graphics candidates="../graphics/none.png" class="ltx_centering" graphic="../graphics/none.png" options="width=411.93767pt,keepaspectratio=true" xml:id="S0.F2.g2"/>
+      <graphics candidates="../graphics/none.png" class="ltx_centering" graphic="../graphics/none.png" options="width=411.9pt,keepaspectratio=true" xml:id="S0.F2.g2"/>
       <toccaption class="ltx_centering"><tag close=" ">2</tag>Right figure.</toccaption>
       <caption class="ltx_centering"><tag close=": ">Figure 2</tag>Right figure.</caption>
     </figure>
--- a/t/complex/figure_mixed_content.xml
+++ b/t/complex/figure_mixed_content.xml
@@ -79,7 +79,7 @@
         <tag><text fontsize="90%">(a)</text></tag>
         <tag role="refnum">1(a)</tag>
       </tags>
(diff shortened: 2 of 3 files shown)


lib/LaTeXML/Engine/Base_ParameterTypes.pool.ltxml
```diff
@@ -96,6 +96,44 @@ DefParameterType('XToken', sub { $_[0]->readXToken; });
...
+# been incorporated into definition's Parameters objects; Thus the current hackery:]
+
+# Read a regular arg, but content should be Number,
+# or extended calc syntax if calc.sty is loaded.
+DefParameterType('{Number}', sub {
+    my ($gullet) = @_;
+    my $value = $gullet->readArg();
+    return $gullet->readingFromMouth($value, sub {
+      (LookupValue('calc.sty.ltxml_loaded')
+       ? readCalcExpression($gullet, 'Number')
```
dginev, line 117: This use of `readCalcExpression` will be a namespace headache once it reaches Rust. 

The `readCalcExpression` sub is only available to execute (in the perl sesne) _after_ the calc package has loaded, but it is already used in the Engine, before any bindings load. Ouch. Luckily there is late binding, but it's not easy to port.

Instead, one can have a method defined in Gullet (so always available early) with some generic name describing its behavior, maybe `readArithExpression` or such? And again use it only when calc is loaded. You may want to do this here, or I may just use my comment as a porting guide a bit later.

lib/LaTeXML/Package.pm
```diff
@@ -209,23 +209,33 @@ sub parseParameters {
...
+      # OR read using readOptional, then content (if any) reparsed as Type.
       my ($spec, $inner_spec) = ($1, $2);
-      if ($inner_spec =~ /^Default:(.*)$/) {
-        push(@params, LaTeXML::Core::Parameter->new('Optional', $spec,
-            extra => [TokenizeInternal($1), undef])); }
-      elsif ($inner_spec) {
-        push(@params, LaTeXML::Core::Parameter->new('Optional', $spec,
-            extra => [undef, parseParameters($inner_spec, $for)])); }
+      if(LookupMapping('PARAMETER_TYPES', $spec)) { # If specially defined with []?
+        push(@params, LaTeXML::Core::Parameter->new($spec, $spec, optional => 1)); }
```
dginev, line 229: I am now starting to get confused about having two ways of specifying these. I recall we also had a prefix keyword `Optional`, as with `OptionalNumber`, as well as `[Number]`.

Shouldn't we be normalizing towards one or the other? Is there a risk that a lookup for `OptionalNumber` doesn't find a special mapping for `[Number]` or the other way around?

dginev (commented):
Also tricky to review... Likely good to merge, but I sent some thoughts your way as well.

lib/LaTeXML/Package.pm
brucemiller: Good point, but there's are Lots of the latter

t/graphics/calc.tex
```diff
@@ -88,4 +89,15 @@ \section{Macros?}
 \setcounter{ctra}{3*\maxof{\four+5}{3*\four}}
 Count 3*maxof(\four+5)(3*\four) =\thectra.
 
+\section{Magical Redefinition}
+All but 1st 2 would fail in stock LaTeX:
+\makebox{Foo}
+\makebox[1in]{Foo}
+\makebox[{1in}]{Foo}
+\makebox[{{1in}}]{Foo}
+\makebox[0.5in*2]{Foo}
```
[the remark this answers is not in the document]
brucemiller: The point was to verify that it's using the calc method, not to verify every operator in every command

PR #2653, opened 2025-10-01
[CI] expire cache for new windows-latest github images

dginev:
Follow-up to #2649

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -49,7 +49,9 @@ jobs:
       # ====================
       # Setup Perl
       # ====================
-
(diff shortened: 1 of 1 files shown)


dginev:
Interesting. The fresh run succeeds, but the reuse of the cache consistently hits the Socket.c mismatch. I will take a closer look a little later, there may be an extra path that needs caching now.

dginev:
Well, it appears that the inverse approach is simpler - start with always uninstalling the system-provided strawberry perl, which removes the conflicting dependencies.

Cache reuse works well now. Extra 1 minute penalty for the uninstall step.

PR #2654, opened 2025-10-03
Bump \fmtversion to 2021/06/01 to satisfy memoir class requirements

utkuokur:
This PR updates the hard-coded LaTeX format date in
`lib/LaTeXML/Engine/latex_base.pool.ltxml`:

  - DefMacroI('\fmtversion', undef, '2018/12/01');
  + DefMacroI('\fmtversion', undef, '2021/06/01');

Rationale:
- The `memoir` class (and some other modern packages) require at least LaTeX format `2021/06/01`.
- With the previous value (`2018/12/01`), LaTeXML fails immediately with:

    Class memoir Error: Your LaTeX release is too old.
    The memoir class requires at LaTeX format from at least 2021/06/01 onwards.

- After applying this patch locally, I was able to successfully process documents using `memoir` with LaTeXML.

This should unblock users of `memoir` and other classes that check `\fmtversion`.

**Notes:**
While running make test, I observed 3 failing subtests in t/81_babel.t.
However, the same failures occur when testing against the current master branch, so they appear unrelated to this change.

### DIFF
--- a/lib/LaTeXML/Engine/latex_base.pool.ltxml
+++ b/lib/LaTeXML/Engine/latex_base.pool.ltxml
@@ -253,7 +253,7 @@ EOTeX
 # C.3.1 Making Sentences
 #======================================================================
 DefMacroI('\fmtname',    undef, 'LaTeX2e');
-DefMacroI('\fmtversion', undef, '2018/12/01');
+DefMacroI('\fmtversion', undef, '2021/06/01');
 
 #======================================================================
 # C.3.2 Making Paragraphs


dginev (approved):
That is fine by me, thanks.

But curious if Bruce has some thoughts about the broader impact of that flag, for example if it affects our "make formats" target.

xworld21:
Format 2020 includes stuff like `\NewDocumentCommand` which LaTeXML does not implement directly, so it would break LaTeX3 (e.g. xparse would not load correctly). Rather, does make formats update `\fmtversion`?

utkuokur:
The pull request failed the test. I will retry with a partial solution instead of a global change.

dginev:
@utkuokur your PR was healthy, the failed test was due to CI bitrot (fixed in PR #2653 )

brucemiller:
If I'm reading correctly, with that simple date switch you were able to successfully convert documents reading `memoir.cls` natively? I'm pleasantly surprised (amazed, actually). The memoir class is distinctly non-trivial, and all class files tend to slip in under the semantic/structure radar of LaTeXML (ie. to recognize the significance of front matter and sectioning commands). 

Running `make formats` does set `\fmtversion` to whatever `latex.ltx` file was read, and will probably provide even better support for whatever memoir wants to do.  Getting formats built properly is what we're trying to get standard in future releases, and is the better long-term solution. @xworld21's concern is valid about claiming a version that we dont define even the user-facing commands.

brucemiller:
So, maybe we shouldn't bump `fmtversion` unless we actually support the version. But I'd be quite interested in hearing how `memoir.cls` fares; and moving forward whether a small, surgical binding (within the `make formats` framework) could support it better.
In any case, Thanks for your interest, and the contribution!

PR #2665, opened 2025-10-09
add lang attribute to top-level <html> in ePub

### LINKED ISSUE #2661: EPUB WCAG compliance "Every HTML document needs a lang attribute"
Created an epub file using latexmlc and ran it through Ace by Daisy who indicate this violation of WCAG:

```
Serious: WCAG 2.0 A
html-has-lang
[Learn about: Language](https://github.com/brucemiller/LaTeXML/issues/new#)
The <html> element does not have a lang attribute 
 Ensure every HTML document has a lang attribute 
 <html> element must have a lang attribute
```

Image

I looked at the epub file and the lang is there in the content.opf file but is not included in the xhtml files.  

Somebody with more expertise than I have would need to say if this is actually required by WCAG, or if the Ace by Daisy is reporting a violation that is not really there.

dginev:
Fixes #2661

I believe this was a simple oversight. I carried over the identical rule from `LaTeXML-webpage-xhtml.xsl`

### DIFF
--- a/lib/LaTeXML/resources/XSLT/LaTeXML-epub3.xsl
+++ b/lib/LaTeXML/resources/XSLT/LaTeXML-epub3.xsl
@@ -38,6 +38,14 @@
     <xsl:apply-templates select="." mode="doctype"/>
     <xsl:element name="html" namespace="{$html_ns}">
       <xsl:apply-templates select="." mode="begin"/>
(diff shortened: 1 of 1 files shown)


brucemiller (approved):
Seems like the right thing to do. Thanks!

PR #2668, opened 2025-10-15
updated CI for v0.9 release in 2026

dginev:
Work in progress:
 - [x] CI texlive docker images have been completed with the great aid of @tkw1536 
     - and a tiny build+push of the final runner images from me.
 - [x] Discovery: perl 5.42 has a regression in the test suite, due to `\p{Latin}` regex  changes.
     - Update: a lot more background is now added in tracking issue #2670
 - [x] Discovery: The glossary test is failing under texlive 2024 (I believe @brucemiller had mentioned this in a meeting - please see PR #2609)
 - [x] Discovery: The texlive-enabled linux tests take forever for texlive 2023 and 2024 (55 min).  
     - We probably want to only run those under `make formats`
 - [x] Discovery: The latest choco package for miktex is also failing to install, needs more care.
 - [x] TODO: Add strategic `make formats` targets to the CI matrix, ideally after we fix the dumper in the master branch.

Also, thanks to @xworld21 for suggesting the latest perl versions available via choco for Strawberry Perl.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -274,7 +274,8 @@ InstallTeXStyles
 # Ideally, would depend on /<system>/texmf-dist/.../plain.tex & latex.ltx
 # This is still experimental/optional, otherwise should add formats to pure_all target!
 sub dump_formats {
-  $$MORE_MACROS{INST_FMTDIR} = '$(INST_LIB)/LaTeXML/Engine';
+  $$MORE_MACROS{INST_FMTDIR}  = '$(INST_LIB)/LaTeXML/Engine';
+  $$MORE_MACROS{STRICT_IN_CI} = $ENV{CI} ? "--strict" : "";
   $MORE_MAKERULES .= <<'DumpFormats';
 
 # Create dump file for format plain.tex, latex.ltx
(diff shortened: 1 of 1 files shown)


dginev:
The (first?) perl v5.42 regression is traceable to Font.pm's tests, and in particular what reduces to:

```perl
"\x{02C7}" =~ /\p{Latin}/
```

- true in perl v5.42
- false in perl v5.40 and earlier

I am not sure what is the full range of characters affected yet, but I now know \p{Latin} has been adjusted.

If this only affects modifier letters we may get lucky, as the Font.pm check can use a negative check for `\p{Lm}`. 

Update: The negative Lm filter worked! Regression resolved.
Update 2: Made a tracking issue in #2670 and switched here to `\p{LC}` positive match and `\p{Script=Common}` negative match, to avoid the card suits and letterlike math symbols.

brucemiller:
Interestingly 2023 & 2024 work with `make formats`, but 2025 fails. Probably should fail the build immediately if formats fails (and save the pointless tests)?

dginev:
Done. Let me know if you want to merge this as-is, specifically if I should undo the Font.pm changes. Now that we allow merging with texlive 2025 + perl 5.42 failing, it can also get patched elsewhere.

I will take yet-another look at the MikTeX setup, see if something can be salvaged.

Edit: Checked the choco avenue, not really usable. We may be better off trying to install texlive on windows. But that is outside of the scope of this PR, which just wanted to update the main texlive and perl targets.

brucemiller (approved):
Not perfect, but very close. Thanks for pulling the pieces together!

nschloe:
where v0.9 release in 2025

PR #2677, opened 2025-11-01
implement page option of \includegraphics

### LINKED ISSUE #2508: \includegraphics: page= option not implemented
graphicx supports selecting which page of a PDF should be used, as in:
```latex
\includegraphics[page=3]{figure}
```
This is not implemented in LaTeXML, resulting in the wrong figure.

Cc @leorolla.

xworld21:
Fix #2508, #2639 and supersede #2650. This includes reading the correct page also when determining the size of an image. Specifying multiple page options (should that even happen) behaves like TeX, i.e. the last one wins.

The package Image::Size does not support specifying a particular page, which *shouldn't* matter for those image formats, but I guarded the call to `imgsize` for clarity anyway.

### DIFF
--- a/lib/LaTeXML/Util/Image.pm
+++ b/lib/LaTeXML/Util/Image.pm
@@ -16,7 +16,7 @@ use LaTeXML::Global;
 use LaTeXML::Common::Error;
 use LaTeXML::Common::Dimension;
 use LaTeXML::Util::Pathname;
-use List::Util qw(min max);
+use List::Util qw(first min max pairfirst pairgrep);
 use Image::Size;
 use POSIX;
 use base qw(Exporter);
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Util/Image.pm
```diff
@@ -448,11 +458,14 @@ sub image_read {
   if (!$source) {
     Error('imageprocessing', 'read', undef, "No image source given"); return; }
   return unless $source;
+  my ($page_key, $page) = pairfirst { $a eq 'page' } @args;
+  @args = pairgrep { !($a eq 'page') } @args if defined $page_key;
```
dginev:
```suggestion
  @args = pairgrep { $a ne 'page' } @args if defined $page_key;
```

brucemiller (approved):
Slightly scary, but seems right. Thanks for bravely wading into this!!

PR #2681, opened 2025-11-03
orcidlink.sty.ltxml depends on tikz 

dginev:
Even though tikz is never used in the perl binding, not requiring it leads to surprised users in arXiv. It is a very painful kind of problem to have to wonder why pdflatex succeeded where latexml failed, in an otherwise complex document.

https://github.com/arXiv/html_feedback/issues/5185

Minimal example:

```tex
\documentclass{article}
\usepackage{orcidlink}
\begin{document}
\makeatletter
\@ifundefined{includegraphics}{undefined}{defined}
\end{document}
```

### DIFF
--- a/t/moderncv/orc.tex
+++ b/t/moderncv/orc.tex
@@ -3,6 +3,9 @@
 \usepackage{orcidlink}
 \title{ORCID Test}
 \author{Bruce\orcidlink{0000-0002-2300-0367}}
+\makeatletter%avoid page break portability issues with tikz emulation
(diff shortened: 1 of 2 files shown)


dginev:
The arbitrary test regression in newer texlives due to loading tikz is somewhat depressing. Sigh.

dginev:
Well, I found a creative way to be "portable". Turn the constructor for page breaks into `\relax` during the tiny moderncv test which does not need any breaks.

Ready for review, tests pass everywhere.

PR #2683, opened 2025-11-06
[arXiv] `bibconfig` option for latexml.sty

dginev:
As of yesterday, arXiv submissions now process `.bib` files. To stay backwards compatible, this is done without any additional configuration markup, but via convention. Namely:

- if a `.bbl` is present in the submission, use that file
- if a `.bbl` is not present, and the appropriate `.bib` files are present, run the appropriate bibtex/biber build steps.

In LaTeXML, we had emulated the previous convention (always use `.bbl`, never run bibtex) via adding a `nobibtex` option to latexml.sty.

~~I have extended that a bit, by also adding a `usebbl` option (name suggestions welcome!), which will follow the new arXiv convention.~~

After a round of feedback, I have added a `bblconfig` option which takes a flexible value which can reorder the biblioraphy processors. The usual expectede uses are:
 - `--preload=[bibconfig={bib,bbl}]latexml.sty`
 - `--preload=[bibconfig={bbl,bib}]latexml.sty`
 - `--preload=[bibconfig={bbl}]latexml.sty`
 - `--preload=[bibconfig={bib}]latexml.sty`

Allowing to enable/disable bbl and bib processing, as well as to prefer one over the other.

With this, the arXiv/ar5iv configuration for latexml will migrate from `--preload=[nobibtex]latexml.sty` to `--preload=[bibconfig={bbl,bib}]latexml.sty`.

As latexml already had existing support for the bibtex processing, this PR is *only* a configuration adapter.

### DIFF
--- a/lib/LaTeXML/Engine/latex_constructs.pool.ltxml
+++ b/lib/LaTeXML/Engine/latex_constructs.pool.ltxml
@@ -3957,8 +3957,11 @@ DefMacro('\lx@ifusebbl{}{}{}', sub {
 
     my $bbl_path     = FindFile($jobname, type => 'bbl');
     my $missing_bibs = '';
-    my $bib_config   = LookupValue("BIBTEX_CONFIG") || '';
-    if ($bib_config eq 'BBL_ONLY' or ($bbl_path and $bib_config eq 'BBL_FIRST')) {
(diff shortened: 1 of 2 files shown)


dginev:
I have updated the PR and description to use a new key-value option allowing for flexible ordering. Ready for another round of review.

PR #2687, opened 2025-11-10
add \overunderset to amsmath.sty

dginev:
Tested with [arXiv:2511.05230v1](https://arxiv.org/src/2511.05230v1).

This was reported at https://github.com/arXiv/html_feedback/issues/5230

The PR simply combines the existing definitions of `\overset` and `\underset`. The resulting presentation MathML is a little suspicious however. For the minimal example:

```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}
\[ \overunderset{\text{\rm a.s.}}{\varepsilon\to0}{\longrightarrow} \]
\end{document}
```

The expected behavior is the long right arrow to span both over- and under-accents. In today's Chrome that isn't the case, and the MathML sizes are seemingly rather off. LaTeXML (with this PR) produces:

```xml
  <munderover accent="true" accentunder="true">
    <mo stretchy="false">⟶</mo>
    <mrow>
      <mi mathsize="1.420em">ε</mi>
      <mo mathsize="1.420em" stretchy="false">→</mo>
      <mn mathsize="1.420em">0</mn>
    </mrow>
    <mtext>a.s.</mtext>
  </munderover>
```

So I welcome thoughts on where may be the best place to understand/adjust the sizing details - and stretchy details. Encouragement to do a deeper dive is welcome too.

### DIFF
--- a/t/ams/mathtools.xml
+++ b/t/ams/mathtools.xml
@@ -1170,19 +1170,19 @@ Then a switch of tag forms.</p>
               <XMApp role="ARROW">
                 <XMApp role="UNDERACCENT">
                   <XMTok meaning="times" role="MULOP">⁢</XMTok>
-                  <XMTok font="italic" role="UNKNOWN">u</XMTok>
-                  <XMTok font="italic" role="UNKNOWN">n</XMTok>
-                  <XMTok font="italic" role="UNKNOWN">d</XMTok>
-                  <XMTok font="italic" role="UNKNOWN">e</XMTok>
-                  <XMTok font="italic" role="UNKNOWN">r</XMTok>
+                  <XMTok font="italic" fontsize="70%" role="UNKNOWN">u</XMTok>
+                  <XMTok font="italic" fontsize="70%" role="UNKNOWN">n</XMTok>
+                  <XMTok font="italic" fontsize="70%" role="UNKNOWN">d</XMTok>
+                  <XMTok font="italic" fontsize="70%" role="UNKNOWN">e</XMTok>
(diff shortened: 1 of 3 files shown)


brucemiller:
OK, so you're going to *like* this... (or not).

Firstly, regarding the sizes: It would appear that the over/under accent code that places an argument over/under something like an arrow failed to track the evolution of fontsizing with regular scripts; basically those args should be `InScriptStyle` (or `OptionalInScriptStyle`) to account for the downsizing. But also, the MathML conversion inexplicably downsized *one* of the args when joining an under+over.

As for stretchiness, the way I read it, `\overset` and friends should *not* get stretchy arrows. If you want them to stretch you'd use `\xleftarrow` (and many friends in ams & mathtools). Alas, various oversights led to many of those cases not stretching when they should.

In the process of following all the wtf's, I ended up fixing most/all of the offending cases, and improving the output of `t/math/arrow, t/math/sampler, t/ams/mathtools`.   So, what's going to be the cleanest way to proceed? Should I make a PR with my further changes against your branch? Or just push onto it?

dginev:
I like (and _like_) what you've described - you're welcome to push to this branch directly if you want to reuse this PR.

brucemiller:
OBTW; as I was trying to synthesize the path through the various surprises, I forgot the introductory clause which was to the effect: Your suggested patch/definition of `\overunderset` is almost exactly what is needed; it's just broken by all the other broken things.

brucemiller:
If you're good with my suggested enhancements, we can squash n' merge.

PR #2688, opened 2025-11-11
marvosym binding typo

dginev:
Caught in https://github.com/arXiv/html_feedback/issues/4955


dginev:
This was also subtler than hoped for - the older version of the package appears to have used the `g` name. So we shoudl support both. Updated and ready for review.

PR #2693, opened 2025-11-18
Robust on raw perl value instead of register value

dginev:
This was encountered in `arXiv:2308.06254v1`, where a use of `\fam` was returning a perl `0` scalar value from `readRegisterValue`.

I think we should harden `readRegisterValue` to always return an object, catching these cases and defaulting them to what the coercing would do, or `Number` if nothing else is known.

With the change in the PR the article converts without a Fatal, instead emitting a warning:
```
Warning:malformed:register_value Register \fam did not return a proper value, got 0
```

That can also be made into an error if preferred. Encountered in an old issue when processed with the latest latexml worker, at https://github.com/arXiv/html_feedback/issues/22

Diff without whitespace lints at https://github.com/brucemiller/LaTeXML/pull/2693/files?w=1

### DIFF
--- a/lib/LaTeXML/Core/Gullet.pm
+++ b/lib/LaTeXML/Core/Gullet.pm
@@ -110,26 +110,26 @@ sub readingFromMouth {
   my $mouth;
   my $intokens;
   my $sourcetype = ref $source;
-  if (! $sourcetype) {          # for a string, create Mouth to read its content
+  if (!$sourcetype) {    # for a string, create Mouth to read its content
     $mouth = LaTeXML::Core::Mouth->new($source); }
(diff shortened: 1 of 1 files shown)


brucemiller:
It may be that having a guard on `readRegisterValue` is a good thing, but would also wonder why `\fam` (or `fontfamily`) is returning a simple scalar.

But I can't seem to reproduce your test case; processing that arXiv article gives me the same errors with or without your patch.

dginev:
To reproduce you need the arXiv worker configuration. The two particular pieces that trigger this are the experimental biblatex.sty binding in ar5iv-bindings, as well as the `--preload=[nobibtex]latexml.sty` trigger to use bbl directly.

PR #2694, opened 2025-11-20
protect p{} cell templates with an inner group

dginev:
This came from a discovered regression on `arXiv:2310.06770v3` after raw interpretation was enabled.

I reduced it to a minimal example of a listing showing up inside a cell with a `p{}` template:

```tex
\documentclass{article}
\usepackage{listings}
\begin{document}
\begin{tabular}{p{\linewidth}}
A\begin{lstlisting}
  B
\end{lstlisting}
\end{tabular}
\end{document}
```

The errors produced have to do with TeX mode-matching, in particular the `\hbox` associated with `p{}` terminating the wrong mode at the end.

While fishing for more information I found that making a couple of groups more explicit avoids the error - was it simply a matter of not enough explicit groups? Probably not, as the article continues to encounter a few errors related to mode balancing. Flagging this with a draft PR for now, feedback welcome.

To download the arXiv source for testing:
https://arxiv.org/src/2310.06770v3


dewlove1030-hash (commented):
Sai.net

PR #2708, opened 2025-12-12
Accessibility

### LINKED ISSUE #2638: Regression in \Description

brucemiller:
This PR defines & uses two new metadata elements, `ltx:alternate-text` (for alternate or replacement text for things that may be undisplayable, or not easily understood when they are) and `ltx:long-description` (for detailed descriptions of what an object, especially a figure or image, contains; in contrast to a caption which may focus more on the source or purpose). These are typically useful for accessibility, and may be spoken or displayed by a user agent under certain circumstances.  This metadata is now also allowed within an `ltx:graphics`, which was formerly an empty element.

The XSLT transforms to HTML arrange for these to be stored but not displayed, and their parent (typically the figure or image) points to them via either `aria-labelledby or aria-describedby` (but these are not overwritten if they've already been set).  For JATS, these are converted to `alt-text` or `long-desc`.

Finally, the `acmart` binding for `\Description` is updated & simplified to use this approach, along with the test case output.

Would appreciate feedback from @dginev and @xworld21; I think #2684 can be easily updated to use this approach, and it even easy if there's ever a "long description" keyword added to `\includegraphics`.

Fixes #2638
Closes #2666

### DIFF
--- a/lib/LaTeXML/Package/acmart.cls.ltxml
+++ b/lib/LaTeXML/Package/acmart.cls.ltxml
@@ -73,16 +73,18 @@ DefMacro('\subtitle{}',            '\@add@frontmatter{ltx:subtitle}{#1}');
 DefMacro('\keywords{}',            '\@add@frontmatter{ltx:keywords}{#1}');
 DefMacro('\terms{}',               '\@add@frontmatter{ltx:keywords}{#1}');
 
-# Accessible figure descriptions
-NewCounter('acmlabel');
-DefConstructor('\Description[]{}', '^^<ltx:note xml:id="#id" class="ltx_nodisplay">#1</ltx:note>',
-  properties => sub { ('width' => Dimension(0), 'height' => Dimension(0), RefStepCounter('acmlabel')) },
-  beforeConstruct => sub {
-    my ($document, $whatsit) = @_;
-    # TODO: Is there something useful to do with the short description in our schema?
-    my $figure = $document->getElement;
-    $document->setAttribute($figure, 'aria:labelledby', $whatsit->getProperty('id'));
-    return; });
+# Accessible figures: \Description[alternate text]{long descriptions}
+DefConstructor('\Description[]{}', sub {
+  my ($document, $alt, $desc) = @_;
+  # Could this apply to other than figure?
+  if (my $savenode = $document->floatToElement('ltx:figure')) {
+    $document->insertElement('ltx:alternate-text',   $alt)  if defined $alt;
+    $document->insertElement('ltx:long-description', $desc) if defined $desc;
+    $document->setNode($savenode); }
+  else {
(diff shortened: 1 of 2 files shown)


dginev (approved):
I have no further feedback.

lib/LaTeXML/Package/acmart.cls.ltxml
```diff
@@ -73,16 +73,18 @@ DefMacro('\subtitle{}',            '\@add@frontmatter{ltx:subtitle}{#1}');
...
-    my ($document, $whatsit) = @_;
-    # TODO: Is there something useful to do with the short description in our schema?
-    my $figure = $document->getElement;
-    $document->setAttribute($figure, 'aria:labelledby', $whatsit->getProperty('id'));
-    return; });
+# Accessible figures: \Description[alternate text]{long descriptions}
+DefConstructor('\Description[]{}', sub {
+  my ($document, $alt, $desc) = @_;
+  # Could this apply to other than figure?
+  if (my $savenode = $document->floatToElement('ltx:figure')) {
```
dginev: Ok, I guess I have feedback - this macro can also be used for tables, i.e. `ltx:table`.

lib/LaTeXML/Package/acmart.cls.ltxml
```diff
@@ -73,16 +73,18 @@ DefMacro('\subtitle{}',            '\@add@frontmatter{ltx:subtitle}{#1}');
...
+    $document->insertElement('ltx:alternate-text',   $alt)  if defined $alt;
+    $document->insertElement('ltx:long-description', $desc) if defined $desc;
```
dginev: In practice, often the author provides only one of the two arguments, in which case it is most predictable for us to be setting the primary accessible description only (i.e. ltx:alternate-text). It is also the field requires by accessibility checkers, the long-description is always optional.

So it is unfortunate the optional argument is used for the mandatory field for remediation.

PR #2733, opened 2026-01-06
Allow for comma-separated KeyVal arguments 

dginev:
This PR is a patch follow-up to #2683 . Once I started testing I realized we did not have the right infrastructure in place to recognize and preserve comma-separated values in KeyVals that enter through package options, and in particular via `\usepackage`.

- Adjusts the wrongly written conditional in `\lx@ifusebbl` from my previous PR 
- Enhanced `TrimmedCommaList` to respect inner wrapping braces, as well as to trim outer wrapping braces.
- Moved `TrimmedCommaList` to KeyVal, as it should be in the Core layer and should in theory be useful in value processing.
- I also added tests to ensure I am really seeing the expected behavior this time.

There is likely a better patch out there, and I am happy to pursue it. The current PR allows for a minimal stability path so that we can use the new `bibconfig={bbl,bib}` recipe for arXiv.

### DIFF
--- a/lib/LaTeXML/Core.pm
+++ b/lib/LaTeXML/Core.pm
@@ -168,7 +168,7 @@ sub digestFile {
 sub iniTeX {
   my ($self, $request, $destination, %options) = @_;
   my ($dir, $name, $ext);
-  my $mode = $options{mode} || 'Base'; # normally, w/o TeX (plain) itself
+  my $mode = $options{mode} || 'Base';    # normally, w/o TeX (plain) itself
   if (pathname_is_literaldata($request)) {
     $dir  = undef; $ext = 'tex';
     $name = "Anonymous String"; }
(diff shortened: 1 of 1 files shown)


dginev:
As it turns out, the changes in the PR so far are not yet enough to deal with 

```
--preload=[bibconfig={bbl,bib}]latexml.sty
```

But they suffice for the `\usepackage` variation. I will see if there is an easy extra commit to get preloads working as well.

dginev:
I think I now have `--preload` working as well, again via `TrimmedCommaList` as the single utility sub.

Another detail is that to pass the braces as literal characters in a linux shell, they need to be escaped. So my test example was ultimately:
```
latexmlc test.tex  --preload=[bibconfig=\{bib,bbl\}]latexml.sty
```

brucemiller:
This has a scary number of changes, for what would seem to be ought to be simpler. Keyvals can handle these kinds of values without problems, but I've got to remind myself why we aren't using them; `\usepackage` has `OptionalSemiverbatim` instead, and then manually (wrongly?) massages the argument.

PR #2737, opened 2026-01-12
Added bindings for causets (TikZ extension).

c-minz:
The bindings only contains the standard code for a TikZ extension.

To test it, use the files in the package repository at https://github.com/c-minz/LaTeX-causets (the pdf and html outputs are also included there). My environment:
- `pdflatex ...` to create the corresponding pdf files (on Linux, Kubuntu 24.04)
  pdfTeX 3.141592653-2.6-1.40.28 (TeX Live 2025)
  (including causets v1.5)
- `bin/latexmlc --strict ...` to create corresponding html files

The main test file runs without issues:
[test_causets.tex](https://github.com/c-minz/LaTeX-causets/blob/main/test_causets.tex)

The example files show some warning (coming from other packages), but run also without issues otherwise:
[causets_example1.tex](https://github.com/c-minz/LaTeX-causets/blob/main/causets_example1.tex)
[causets_example2.tex](https://github.com/c-minz/LaTeX-causets/blob/main/causets_example2.tex)

### DIFF
--- a/lib/LaTeXML/Package/causets.sty.ltxml
+++ b/lib/LaTeXML/Package/causets.sty.ltxml
@@ -18,8 +18,6 @@ use warnings;
 use LaTeXML::Package;
 
 #=========================================================================#
-RequirePackage('tikz');
-RequirePackage('pgf');
 InputDefinitions('causets', type => 'sty', noltxml => 1);
 #=========================================================================#
 
--- a/lib/LaTeXML/Package/causets.sty.ltxml
+++ b/lib/LaTeXML/Package/causets.sty.ltxml
@@ -1,0 +1,26 @@
+# -*- mode: Perl -*-
+# /=====================================================================\ #
+# |  causets                                                            | #
+# | Implementation for LaTeXML                                          | #
+# |=====================================================================| #
+# | Part of LaTeXML:                                                    | #
+# |  Public domain software, produced as part of work done by the       | #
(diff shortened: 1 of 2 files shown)


lib/LaTeXML/Package/causets.sty.ltxml
```diff
@@ -0,0 +1,26 @@
...
+# | http://dlmf.nist.gov/LaTeXML/                              (o o)    | #
+# \=========================================================ooo==U==ooo=/ #
+package LaTeXML::Package::Pool;
+use strict;
+use warnings;
+use LaTeXML::Package;
+
+#=========================================================================#
+RequirePackage('tikz');
+RequirePackage('pgf');
```
dginev: You can probably even remove the two `RequirePackage` calls here, as the package itself has `\RequirePackage{tikz}` near the top.

dginev (approved):
The two presented examples look pretty decent in SVG. 

So I am OK with allowing the raw load of causets.sty by default.

brucemiller (approved):
I'm a little concerned that the LaTeXML system overall may not quite be able to fully support causets, but this is still a good first step.  Thanks for the effort!

c-minz:
Thanks. Since the package is TikZ-based, I think most remaining, smaller graphical issues are general TikZ problems. Some of them are reportet in issue [#2738](https://github.com/brucemiller/LaTeXML/issues/2738)

PR #2739, opened 2026-01-14
\href and math mode

remimorvan:
This pull request fixes the behaviour of the `\href` command while in math mode, and of math passed as an argument to `\href`.
- The LaTeX code ```$\href{#}{x = y}$``` did not produce any link. The href attribute was present in the XML file, but not in the HTML. This is fixed by changing the pmml_maybe_resize subroutine in the postprocessor.
- The LaTeX code ```\item foo \href{#}{$x = y$}``` (inside an itemize environment) was producing a link, but a linebreak was present between `foo` and the link. This is fixed by changing the css.

### DIFF
--- a/lib/LaTeXML/resources/CSS/LaTeXML.css
+++ b/lib/LaTeXML/resources/CSS/LaTeXML.css
@@ -241,8 +241,7 @@ table.ltx_eqn_eqnarray tr.ltx_eqn_lefteqn + tr td.ltx_align_right { min-width:2e
 li.ltx_item > .ltx_tag {
     display:inline-block; margin-left:-1.5em; min-width:1.5em;
     text-align:right; }
-.ltx_item .ltx_tag + .ltx_para {
-    display:inline-block; vertical-align:top;}
+.ltx_item .ltx_tag + .ltx_para,
 .ltx_item .ltx_tag + .ltx_para .ltx_p  {
     display:inline; }
 .ltx_item > .ltx_para > .ltx_p:first-child {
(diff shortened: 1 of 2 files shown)


dginev:
Nice initiative, as this synchronizes with another effort. MathML Core is about to introduce a new use for `<a>` anchor elements inside MathML to hold links, in contrast to the MathML 3 use of the `href` attribute on MathML elements themselves. That is tracked in PR https://github.com/w3c/mathml-core/pull/307

I think we should wait just a little for that change to land in MathML Core, then start emitting these `<a>` elements right away in LaTeXML. Bruce likely has his own thoughts on the ideal timeline, logging my own for now.

dginev:
Btw, the CSS change seems unrelated, and is probably discussed better elsewhere.

brucemiller:
Interesting; So, href already works for tokens (handled in `stylizeContent`), but not for mrow expressionsl perhaps patching `pmml_maybe_resize` is the right approach (it's already been abused to do frames & colors, so it's not just resizing after all).

But I don't get what the 2nd point is about.  The XML and HTML do indeed get a newline between the "foo" and the math element, but that shouldn't cause a linebreak in the display unless there's some other line wrapping going on.

lib/LaTeXML/resources/CSS/LaTeXML.css
```diff
@@ -241,8 +241,7 @@ table.ltx_eqn_eqnarray tr.ltx_eqn_lefteqn + tr td.ltx_align_right { min-width:2e
 li.ltx_item > .ltx_tag {
     display:inline-block; margin-left:-1.5em; min-width:1.5em;
     text-align:right; }
-.ltx_item .ltx_tag + .ltx_para {
-    display:inline-block; vertical-align:top;}
+.ltx_item .ltx_tag + .ltx_para,
```
brucemiller, line 244: What is the purpose of this change?

brucemiller (changes requested):

Issue #2316, opened 2024-02-08
Abstract, acknowledgements, references, and (sometimes) appendices do not appear in the post-processing generated table of contents.


chrisjcameron:
We are using the optional post-processing in lateXML to create a TOC and noticed that some elements are excluded. For our use case, it would be ideal if the TOC would provide screenreader users with an easy means of navigation to all major paper sections. 

As examples, see:
https://arxiv.org/html/2401.18052v1
https://arxiv.org/html/2401.18036v1

dginev:
@brucemiller I am refactoring the stopgap code that fills in these omissions in arXiv, and I was wondering if you have thoughts on whether the abstract and references should be in the `.ltx_toclist` produced by `--navigationtoc=context` ?

dginev:
I made a PR with a first concept of this kind of inclusion, although I am sure there are subtleties we need to discuss first. Likely questions around `--split`, which I haven't examined yet.

One bookkeeping comment to the "(sometimes)" description in the issue title. The second article has an example of acknowledgements that do not appear in the ToC even after my changes, which is due to using an intentionally unnumbered section - `\section*{Acknowledgements}` . Currently those are not included in the table of contents, which I think is fine to keep as-is.

brucemiller:
While I haven't yet absorbed the PR, I think the idea should be that the regular TOC (`\tableofcontents`) should mimic what latex includes, but the navigation toc could (should?) be more generous.

dginev:
> For our use case, it would be ideal if the TOC would provide screenreader users with an easy means of navigation to all major paper sections.

Btw @chrisjcameron this is a little less obvious than one may hope. First, the "numbered sections" approach to compiling the table of contents achieves your stated goal to a large degree. Second, and more importantly, I believe screen readers have a dedicated traversal mode based on HTML heading elements themselves (h1...h8), as with the [grouped headings NVDA key + F7](https://srt.csb-cde.ca.gov/nvda/nvda-lesson-2.html) mode in NVDA.

In the "headings-based traversal view", the DOM-based accessible walk is a somewhat parallel/independent concern to the visible Table of Contents. I suspect the visible ToC may even be used more by sighted readers than users in need of screen readers.

The one practical concern I have about this issue is that once we move to exclusively using LaTeXML's generation, the choice becomes "expensive", as changes will require full regeneration of the articles. So it will be good to have a meaningful choice that can last until 2027 at least.

[...]

PR #2746, opened 2026-01-19
Add/ensure ToC includes abstract, acknowledgements, bibliography

dginev:
A working concept for resolving #2316 .

This PR attempts consistency with the current handlers in CrossRef and Scan post-processors, trying to ensure `inlist="toc"` is set on `ltx:abstract`, `ltx:acknowledgements`, `ltx:bibliography`. And that those elements are included and used in the TOC filling algorithms (also requiring a set `xml:id`).
 - The consistency is ensured by subs executed on every element via the `Tag()` technique.

To my surprise this leads to a schema change as well, allowing `Labelled.attributes`.

What is less surprising is that such a change leads to a very large diff in the tests. If nothing else, that change is very uniform.

Feedback most welcome.

### DIFF
--- a/lib/LaTeXML/Engine/latex_constructs.pool.ltxml
+++ b/lib/LaTeXML/Engine/latex_constructs.pool.ltxml
@@ -725,15 +725,15 @@ DefConstructorI('\tableofcontents', undef,
     push(@s, (qw(ltx:appendix ltx:index ltx:bibliography))) if @s;
     (select => join(' | ', @s),
       name => DigestText(T_CS('\contentsname'))); });
-# Ensure we add the TOC-relevant attributes to the relevant elements.
+# Ensure we add an xml:id to navtoc-relevant elements (abstract, acknowledgements).
+# NOTE: we do NOT set inlist="toc" here, as that would disturb \tableofcontents behavior.
+# The navigation TOC (context format) uses $normaltoctypes in CrossRef to include these.
 Tag('ltx:abstract', afterOpen => sub {
     my ($document, $node, $thing) = @_;
-    $document->setAttribute($node, inlist => 'toc');
     GenerateID($document, $node, $thing, 'abstract');
     return; });
 Tag('ltx:acknowledgements', autoClose => 1, afterOpen => sub {
     my ($document, $node, $thing) = @_;
-    $document->setAttribute($node, inlist => 'toc');
     GenerateID($document, $node, $thing, 'acknowledgements');
     return; });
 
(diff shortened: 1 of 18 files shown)


dginev:
Update: I have returned `\tableofcontents` to the mainline behavior, as per feedback.

brucemiller:
Wow, there's a *lot* of diff noise here. Of course, if abstract is going be a valid target, it's going to need an `xml:id`.  But adding `inlist="toc"` to any of these elements is going to affect the normal TOC -- and it does, even though I thought you said you'd backed that off.

However, if we're going to have the navigation TOC include more or different elements than the standard LaTeX TOC, then I guess we're stuck doing one of two things: We could modify `Post::CrossRef` to handle navigation TOC's specially, with hand coded extra candidate elements (which is kinda yucky); or use an extra keyword, like `inlist="navtoc"` to indicate that an element should be included in the navigation toc (or `inlist="toc,navtoc"` for both).  The latter seems the most Correct, but of course, along with new ID's, will also cause lots of diff noise.

dginev:
My memory is:
 - we agreed to back off the changes to `\tableofcontents` so that the original LaTeX behavior is undisturbed
 - we agreed if we are including these new kinds of content in the ToC, we should treat them equivalently to the existing ones, i.e. give them the `inlist="navtoc"` attribute.
 - indeed the PR contains the "kind of yucky" continuing of precedent in Scan.pm, by adding the entries as an allowlist, e.g.

```perl
  $self->registerHandler('ltx:acknowledgements' => \&ack_handler);
```

I don't really have an appetite for a larger refactor right now, and this code has already been deployed on our end, so this can idle until we have a path forward.

brucemiller:
>   * we agreed to back off the changes to |\tableofcontents| so that the
>     original LaTeX behavior is undisturbed

Good. Except that the PR as it stand *is* disturbing the latex behavior.

dginev:
Which LaTeX behavior ?

brucemiller:
Now `\tableofcontents` includes the bibliography, which plain latex omits. (curiously it omits abstract from `\tableofcontents`; though both are included in the navtoc, as you wished)

brucemiller:
This ought to be easy, and almost everything is there to do it, except for a "clean" design hook. We should not be adding `\inlist="toc"` to anything (for this purpose); OTOH, adding `navtoc` to the XML is awkward, as it pollutes all the XML but should be a purely post-processing decision.

dginev:
If you could name the mechanisms you think are capable of achieving this simply, I could try and verify the suggestion.

brucemiller:
It's kinda speculative, ATM, without some experimentation. I'm thinking we should rely more on schema data than explicit lists of elements. Something like all the elements in `FrontMatter.class, document.body.class,BackMatter.class` that allow the attribute `inlist` (which is `Listable.attributes`) would seem to be potential candidates for TOC's (though `ltx:abstract,ltx:acknowlegements` currently don't get that attribute, but seemingly could).  If that list made sense, it might drive `CrossRef`'s selector, maybe only for navtoc, or optionally, or ...

dginev:
I have tried to remove the newly added `inlist="toc"` attributes and use `$normaltoctypes` instead, specifically for `--navtoc=context`.

dginev:
In the call today we agreed this can wait for mainline support and can be approached differently, with a potentially larger refactor that also includes dynamic navtoc element selection based on the schema.

Issue #1734, opened 2021-12-10
si.tex test raises errors under pdflatex?


dginev:
I'm assuming this is some texlive 2021 regression.
```
cd t/complex ; pdflatex si.tex
```

produces a number of "invalid number" errors, with my:
```
(/usr/local/texlive/2021/texmf-dist/tex/latex/siunitx/siunitx.sty
Package: siunitx 2021-11-03 v3.0.35 A comprehensive (SI) units package
```

unsure what the implications are to the latexml binding yet, but thought I'd record the pdflatex problem already.

brucemiller:
I have a vague recollection that I did it intentionally, but...

teepeemm:
I've managed to get si.tex to compile under TeXLive 2023.  The easiest is `\usepackage{siunitx}[=v2]`, but if we want to actually use v3:
* we must use `\complexnum` to parse complex numbers and `\numproduct` or `\qtyproduct` to parse products
* for `\pi` and `\odd` (or `\xi`) to be in a `\num`, we need to declare them as input-digits (see https://tex.stackexchange.com/a/597999/107497)
* prefixes must have units (although you can cheat at this)
* quotients are no longer parsed (to avoid an error, we can add `parse-numbers=false`, but then the rest of it isn't parsed)

Additionally:
* We should change `\si` -> `\unit` and `\SI` -> `\qty`
* option changes: `group-four-digits` -> `digit-group-size=4`, `brackets` -> `bracket`
* column type `s` has been deleted, but can be redefined
* several SIUnits have been deprecated, but the log file tells you how to define them
* Several options have deleted; the log file will print a warning if you use them

This gets si.tex to compile with pdflatex.  Unfortunately, latexml still runs into #2064 (which can be made slightly more minimal with `\usepackage{expl3}` instead of `\usepackage{siunitx}`).

teepeemm:
I would like to come back to this.  There are two ways I can see to approach this:
1. Adjust the existing si.tex so that it compiles under versions 2 and 3 of siunitx
2. Create a new si_3.tex that uses the version 3 syntax

I greatly prefer option 2, since option 1 would involve not testing as many things.

[...]

PR #2751, opened 2026-01-31
expl syntax in siunitx options

teepeemm:
siunitx options are entered in expl syntax.  This matters for a few of the options.  (Note that in the xml, a number of the deletions are non-breaking spaces.)  This is the slight adjustment I mentioned in #1734.

We also cannot just make all of the options inside `\ExplSyntaxOn` and `\ExplSyntaxOff` because there are a few `_` that need to become subscripts.

### DIFF
--- a/lib/LaTeXML/Package/siunitx.sty.ltxml
+++ b/lib/LaTeXML/Package/siunitx.sty.ltxml
@@ -1564,12 +1564,12 @@ RawTeX(<<'EoTeX');
   number-unit-product = \,      ,
   product-units       = repeat,
 
-  list-final-separator = { ~ and ~ } ,
-  list-pair-separator  = { ~ and ~ } ,
-  list-separator       = { , ~ }     ,
+  list-final-separator = { and } ,
+  list-pair-separator  = { and } ,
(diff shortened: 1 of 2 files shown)


brucemiller:
Is the issue that a `~` in expl syntax simply gives a regular space (which are otherwise ignored), rather than a non-breaking one?  If so, your patch makes perfectly good sense.

teepeemm:
It's a bit worse than spaces accidentally becoming non-breaking.  `list-separator` is currently `␣,␣~␣`.  In expl3, this would become the desired `,␣`.  But we're not in expl3, so `\numlist{1;2;3;4}` becomes `1␣,␣~␣2␣,␣~␣3␣~␣and␣~␣4`.  A space between a number and its subsequent comma is the biggest problem, but the `~` become non-breaking spaces, which means that the html leaves the triple spaces in place and doesn't collapse them.  The other options also end up causing triple spaces that don't collapse.

(I'm nearly ready with the siunitx.sty.ltxml for siunitx 3.  But I keep thinking I'm nearly ready, and then finding one last thing that needs fixing.)

Issue #2747, opened 2026-01-20
extra backslashes in format dumps


teepeemm:
The following error comes up if you try to `InputDefinitions('siunitx', type=>'sty', noltxml => 1);`.
``` tex
\documentclass{article}
\begin{document}
\ExplSyntaxOn
\cs_new:Npn \tpmm_good_loop:N #1 {
        Good~
        \quark_if_recursion_tail_stop:N #1
        \tpmm_good_loop:N }
\tpmm_good_loop:N \q_recursion_tail \q_recursion_stop
\par
\cs_new:Npn \tpmm_bad_loop:N #1 {
    Bad?~
    \peek_catcode:NTF \c_group_begin_token { failed } {
        Bad~
        \quark_if_recursion_tail_stop:N #1
        \tpmm_bad_loop:N } }
\tpmm_bad_loop:N \q_recursion_tail \q_recursion_stop
\ExplSyntaxOff
%%1234
\end{document}
```
To parse a key value, siunitx uses something like the bad loop, and the above in LaTeX becomes

> Good
> Bad? Bad

With LaTeXML, the recursion stops in the good loop, but something causes the peek (aka `\futurelet`) to disturb the quark in #1, so that the recursion no longer stops.  We get the errors
```
Error:expected:Until:\\q_recursion_stop Missing argument Until:\\q_recursion_stop for Core::Definition::Expandable[\use_none_delimit_by_q_recursion_stop:w Until:\\q_recursion_stop] at (end of file)
Error:expected:\end{document} Input ended while environment document was open at String; line 1 col 0
```
and the output
``` xml
  <para xml:id="p1">
    <p>Good</p>
  </para>
  <para xml:id="p2">
    <p>Bad? Bad Bad? Bad Bad? Bad Bad? faileddocument</p>
  </para>
```
where the number of "Bad? Bad" is however many tokens there were before the end of the file.

I've been trying (and failing) to get the above to be free of expl3.  But I also noticed the `\\q_recursion_stop` error has an extra backslash compared to the `\use_none...` or `\end{document}`.  Indeed, my latex_dump.pool.ltxml has
``` perl
I(E(C('\\use_none_delimit_by_q_recursion_stop:w'),Ps(P('Until','Until:\\\\q_recursion_stop',extra=>[T(C('\\q_recursion_stop'))])),T(),isLong=>1));
```
whereas in LaTeX `\show\use_none_delimit_by_q_recursion_stop:w` gives
``` tex
[...]

PR #2753, opened 2026-02-09
do not double-escape spec for dumped parameters

dginev:
Addresses the first symptom of #2747 .

It is easy to find the line of code that does the doule-escaping of parameters during the texinit phase. Seemingly removing that does not break any of our tests -- but let's see CI.

Sadly this changed the original issue's example from having the error:
```
Error:expected:Until:\\q_recursion_stop Missing argument Until:\\q_recursion_stop
```

to having the error
```
Error:expected:Until:\q_recursion_stop Missing argument Until:\q_recursion_stop
```

So more troubleshooting remains.


dginev:
@brucemiller I have rebased here on the latest fixes, and CI passes even without the double-escaping. Maybe we should remove it after all -- at the least the debug prints come out clearer, as indicated in #2747

PR #2762, opened 2026-02-25
Lrgroup

brucemiller:
A little thing, but @HowardCohl complained: I found better Unicode for `\lgroup,\rgroup'.


dginev (commented):
I like the initiative for better typography.

I am neutral on the PR, since it appears to have at least one present-day flaw: the new parens aren't stretchy in STIX Two Math.

Example here:
https://codepen.io/dginev/pen/PwGwNNV

brucemiller:
Funny: In firefox, neither the flattened nor regular parens stretch till you wrap with an mrow (which shouldn't be needed, I think); But with the mrow they both do (seemingly indicating the font can do it? Or else Firefox just fakes it?)

In chrome, the flattened don't stretch even with mrow and stretchy="true".

PR #2777, opened 2026-03-13
Fix pstricks raw TeX interpretation for --includestyles mode

dginev:
## Summary

- **Enable raw pstricks loading** alongside the ltxml binding so that internal TeX registers, utility macros, conditionals, and xkeyval key-value handlers are available for downstream raw packages (pst-3d.tex, pst-plot.tex, pstricks-add.tex, etc.) when using `--includestyles`.
- **Fix empty keyset in KeyVals.pm** — the pstricks family list `\pst@famlist` accumulates as `",pstricks"` (leading comma via `\pst@addfams`), so `split(',', ...)` produces `["", "pstricks"]`. The empty keyset caused `keyval_qname("psset","","ArrowInside")` to resolve as `"psset@@ArrowInside"`, colliding with the raw `\def\psset@@ArrowInside#1-#2\@empty#3\@nil{...}` delimited-argument helper macro, triggering 3 "Missing argument" errors per invocation.
- **Fix empty macroprefix handling in KeyVal.pm** — boolean and command keys with an explicitly empty `macroprefix=""` now correctly use the empty prefix instead of falling through to the default qname-based macro name.
- **Fix color key handling in pstricks.sty.ltxml** — tighten the color-matching regex to `[Cc]olor$` and guard against boolean values (`true`/`false`) and undefined color lookups.

## Motivation

Resolves all 101 pstricks-related errors when converting arXiv paper [2405.20126v1](https://arxiv.org/abs/2405.20126) with `--includestyles`. Minimal reproduction case: a `\usepackage{pstricks}` document that triggers `\psset{ArrowInside={}}` (via `\resetArrowOptions` in pstricks-arrows.tex).

## Files changed

| File | Change |
|------|--------|
| `lib/LaTeXML/Package/pstricks.tex.ltxml` | Load raw `pstricks.tex` via `noltxml => 1` instead of redirecting to sty binding |
| `lib/LaTeXML/Package/pstricks.sty.ltxml` | Load raw `pstricks.sty`; add optional `[]` family arg to `\psset`; fix color regex |
| `lib/LaTeXML/Core/KeyVals.pm` | Filter empty strings from keyset lists: `grep { $_ ne '' } split(...)` |
| `lib/LaTeXML/Core/KeyVal.pm` | Use `defined()` checks for empty macroprefix on boolean/command keys |

## Test plan

- [x] All 468 existing tests pass (`make test`) — no regressions
- [x] arXiv 2405.20126v1 converts with 0 pstricks errors (was 101)
- [x] Minimal test `\usepackage{pstricks}` + `\psset{ArrowInside={}}` succeeds
- [x] `latexmllint` passes on all modified files

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brucemiller:
Interesting to see this patch pop up. How does the rendered SVG compare to the pdf (or postscript? can you even produce a pdf?)

I had set a task for myself awhile back to eliminate all the pseudo-svg elements from the schema and directly generate SVG (which we do for tikz, xy).  The `picture` code in core latex would be somewhat easy, but pstricks (and maybe another?) package builds upon it. The catch is that it's hard to come up with the documentation and a good test sample that can be converted to pdf for comparison.  So I've procrastinated. 

Does your arXiv article help to give a good starting point for this?

dginev:
Good point - there is no actual use of pstricks in this article. It only loads the dependencies (likely for a variant before it was revised for PDF). So just enough to Fatal in latexml, but no real benefit to gain after the loads are guarded.

dginev:
If you are looking for examples using it in arXiv, a good strategy has been to find a macro we dont't support, then browse the error page for it. I looked around and found `\psscalebox`, with CorTeX error page [here](https://corpora.mathweb.org/corpus/arxmliv/tex%5Fto%5Fhtml/error/undefined/%5Cpsscalebox?all=false). So that gives 331 articles to experiment with.

From that list `arXiv:2210.16665` seems to be one of the newest ones, [source link here](https://arxiv.org/src/2210.16665). That only builds with `pdflatex` (not latex) and makes a healthy PDF. latexml just prints out the source code for each figure, but otherwise survives the conversion.

Edit: Ah, but of course - they commented out pstricks and instead used a modern replacement:
```tex
\usepackage[off]{auto-pst-pdf}
```

dginev:
[arXiv:1502.04955](https://arxiv.org/abs/1502.04955) is a better example which actually uses pstricks in full. Nicely, with this PR the error in it disappears, but the display is still not quite accurately matching the PDF.

I drew that from the same list of 331 I mentioned above, one just has to keep trying...

brucemiller (commented):
oh, I guess I should have just commented in the first place. Sigh.

brucemiller (approved):
To bad we can't yet generate nice graphics for this, but it is definitely a step in the right direction! Thanks!

PR #2780, opened 2026-03-26
Sample tests for tcolorbox sizing

dginev:
I have not yet completed the "upgrades" PR, but I think I can showcase some tests for tcolorbox sizing.

The one that is particularly useful is `digestion/colorbox_sizes.tex` as it targets lower level constructs. And it's hand-written.

That said, the tcolorbox test directory consists fully of extractions from arXiv articles where users have reported bad sizing for `{tcolorbox}`. As it turns out, the colored backgrounds prove a very potent proving ground for the sizing of the full variety of LaTeX constructs. 

[Aside: I find it reminiscent of the red/green setup for Web Platform Tests]

Naturally, this is a draft PR. And may be superseded - mostly a discussion marker.


dginev:
Tests and a first proposed implementation are now in #2789

PR #2783, opened 2026-04-02
quantikz2 raw interpretation

dginev:
This PR rescues arXiv:2508.15260v1 from a Fatal regression when using raw interpretation.

It ought to be the last PR for this iteration of improving the arXiv conversion regression guards.

### Summary

- fix `\hphantom` display math leak in text/horizontal mode

- tcolorbox: pre-define `\tcb@use@autoparskip` before raw TeX loading, since pgfkeys initialization may not complete and the `\AtBeginDocument` hook at tcolorbox.sty would call it undefined.

- color: define `\current@color`, `\default@color`, and `\reset@color` with safe defaults. These DVI color macros were previously left undefined, causing errors in pgf/tikz color processing.

- latex_constructs: support optional [label] argument in `\AtBeginDocument` and `\AtEndDocument`, matching the modern LaTeX hooks system (since 2020/10/01). Without this, the [label] leaked as literal text into the document output.

### Minimal test

```tex
\documentclass{article}
\usepackage[dvipsnames]{xcolor}
\usepackage{tikz}
\usetikzlibrary{quantikz2}
\definecolor{tintedcolor}{gray}{.80}%

\begin{document}
\begin{quantikz}[background color=tintedcolor]
  \meter[1]{}
\end{quantikz}
Some text with $x=1$.
\end{document}
```

### Disclosure

Diagnostic and patch via Claude Opus 4.6 "heavy effort".

I have manually verified this patch recovers arXiv:2508.15260v1 out of regression status.

PR carried forward from the forked draft at https://github.com/dginev/LaTeXML/pull/12

### DIFF
--- a/lib/LaTeXML/Engine/math_common.pool.ltxml
+++ b/lib/LaTeXML/Engine/math_common.pool.ltxml
@@ -356,7 +356,7 @@ DefRewrite(select => ["descendant-or-self::ltx:XMTok[text()='\x{FF0F}' and \@mea
       if (my $id = $not->getAttribute('xml:id')) {
         foreach my $n ($doc->findnodes("descendant-or-self::ltx:XMRef[\@idref='$id']")) {
           $doc->removeNode($n); } }    # ? Hopefully this is safe.
-} });
+    } });
 
 #----------------------------------------------------------------------
 # \joinrel
@@ -400,7 +400,7 @@ DefConstructor('\@@joinrel{}{}', sub {
         my $role = (scalar(keys %roles) == 1 ? [keys %roles]->[0] : ($roles{ARROW} ? 'ARROW' : 'RELOP'));
         map { $node->removeChild($_) } @rels;
         $document->insertElement('ltx:XMTok', [map { $_->textContent } @rels], role => $role);
-  } } },
+      } } },
   reversion => '#1\joinrel #2');
 
 #----------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


lib/LaTeXML/Engine/math_common.pool.ltxml
```diff
@@ -655,9 +655,24 @@ DefConstructor('\phantom{}',
...
+      $stomach->beginMode('restricted_horizontal');
+      AssignValue('_hphantom_mode_override' => 1); }
+    else {
+      AssignValue('_hphantom_mode_override' => 0); }
+    return; },
   afterDigest => sub {
-    my $whatsit = $_[1];
+    my ($stomach, $whatsit) = @_;
+    if (LookupValue('_hphantom_mode_override')) {
+      $stomach->endMode('restricted_horizontal'); }
```
brucemiller: Might it not be simpler to rename it as an internal `\lx@hphantom@`, and then define `\hphantom` as `\lx@phantom@{\hbox{#1}}` ?

brucemiller:
I'm curious: Why quantikz version 2, specifically? It's not yet available in my conservative distributions, and less frequently used in arXiv (though that will change, of course)

lib/LaTeXML/Engine/math_common.pool.ltxml
brucemiller: Or better yet
```
DefMacro('\hphantom','\protect\ifmmode\lx@math@hphantom\else\lx@text@hphantom\fi');
```
with `\lx@(math|text)@hphantom` defined as fairly simple constructors, with mode, alias, as appropriate.

brucemiller (changes requested):

lib/LaTeXML/Engine/math_common.pool.ltxml
dginev: Please take another look, I updated.

brucemiller:
Cool! Thanks!!

PR #2784, opened 2026-04-03
Siunitx version 3

### LINKED ISSUE #1734: si.tex test raises errors under pdflatex?

teepeemm:
I've gone through the documentation for siunitx version 3, and tested out as many options as I could.  As I mentioned, I tried to allow v2 or v3 syntax without throwing any warnings (in contrast to the tex version).  There are a few instances where something is valid in both versions but gives different results; these have gone into separate (and short) test files.  There is also a test file for v3 options that are only allowed to change in the preamble (or even only while loading the package).  Note that implementing the v2 tests does mean that we now have a texlive_max option that we'll want to use sparingly.

There's many commits because I was trying to keep diffs to a minimum, except where necessary.  I think the commit messages should indicate where that happened and why.

This fixes #1734


teepeemm:
I may need some help debugging why the tests are failing.  For TexLive 2021, I'm not sure why `class="ltx_figure_panel"` is getting added.  For TexLive 2022, siV3 having `10.0` instead of `10` indicates that the binding is operating in `$VERSION_TWO` mode, which means that `\fmtversion` reported a year prior to 2021.
@dginev , do you have any insight?

dginev:
@teepeemm first thought before looking at any of your changes -- try syncing with the upstream master branch before continuing:

```
$ cd your-fork-directory
$ git remote add upstream https://github.com/brucemiller/LaTeXML
$ git fetch --all
$ git checkout siunitx-tpmm
$ git rebase upstream/master
$ git push --force
```

We have CI passing now, so getting to the latest state will make this easier to review by everyone.

teepeemm:
Catching up to upstream/master did occur to me, and I think I did that (somewhere around the [6afbc9f](https://github.com/brucemiller/LaTeXML/pull/2784/commits/6afbc9f4b805a53b5697d9d1a594efea1563413b) commit).  `git diff --name-status upstream/master` reports these 16 files, and nothing else.

It seems like I was failing because of the class `ltx_figure_panel`, which comes from a width calculation in a table or longtable.  I'm not sure why that wasn't happening on my end, but I've taken out the tables (but left in the tabulars).  

I've completely disabled the siV3 test, so we'll need to look into why that wasn't working.  I'm not seeing `\fmtversion` in any of the test files.  Is it possible it expands to something different in GitHub actions?  (I don't suppose they share the `make formats`?)

I'm not sure why 2021 is still failing.  I'm not seeing what would cause `sectional-block` instead of `logical-block`, and why 2021 would be special in that regard.

dginev:
Congrats on passing CI! I will take some time to review next week, apologies for being a little short on time.

teepeemm:
I figured out why siV3 was having trouble, and left a comment in `t/80_complex.t`: `lib/LaTeXML/Engine/latex_base.pool.ltxml` sets `\fmtversion` to 2018/12/01.  If `make formats` is called, `blib/lib/LaTeXML/Engine/latex_dump.pool.ltxml` overrides `\fmtversion` to the actual value.  The CI calls `make formats` only for 2023 onward, so I restricted siV3 to just that.  This also means `make test CI=true` will fail without `make formats` (unless we have expl3 always `make formats` if it's not already present?).

I'm still not sure why `t/complex/si` was failing 2021, and that last commit just removed that test.  So passing CI is a bit of a cheat.  I'll try to look into that.

teepeemm:
I'm unable to get the `sectional-block` tag that the failing test (t/complex/si with TexLive2021) is getting.  I've installed TL2021, run `make clean`, `make`, and used `LATEXML_KPSEWHICH=/usr/local/texlive/2021/bin/universal-darwin/kpsewhich`, but I can only get output identical to LT2022--25 with `logical-block` tags.  (I've also tried TL2020.)

teepeemm:
@dginev  Were you able to get around to reviewing this, or were you waiting for me to figure out the failing test?  I'm not quite sure how to debug why a GH action is differing than my local result.  Is there an easier way than editing the action to focus on that test?

dginev:
@teepeemm I have disappeared in a maze of corners to examine since March 2026, so no, no updates from me on your PR.

I will get back to the open PRs from a very intriguing perspective come August (post-release) that may be a bit controversial. Basically, the Rust rewrite of LaTeXML is completed (core-complete since April), and is likely to be released in summer (and made public). At which point I plan to transpile all PR suggestions over as Rust suggestions.

The controversial bit is that this effort succeeded in very large part due to AI assistance, in particular the Claude Opus series of models from Anthropic. And I plan on continuing to use them until we map out all of the required coverage for arXiv. So, with all of those caveats, I will be happy to look at your PR in a couple of months!

In the meantime Bruce may beat me to it, which would also be great for the perl side.
 I am giving you the long story since you tagged me. More "soon"!

PR #2767, opened 2026-03-05
Frontmatter refactor

brucemiller:
Revised infrastructure for managing frontmatter, especially navigating the maze of strategies used in different classes to associate a document creator (author, editor, ...) with various contact information (affiliation, address, email, ...).
Classic LaTeX uses only `\author` with multiple authors separated by `\and`, and affiliations (presumably) separated from each author by `\\`.  Others use `\author` for single authors, followed by commands for `\affiliation`, etc.  Another scheme uses `\inst{marks}` within the author, and the marks refer to marks in the affiliation, or the sequence, Or, or, or.

Will need to flesh out the description here. But while it's too early for a full review, it's a good time to start scoping out the scheme; There's undoubtedly lots of not-quite-right about.  Probably easiest approach is to pick a class you're fond of and see how the binding has changed, whether it's readable, sensible, even intuitive?

PR #2785, opened 2026-04-03
Fix \label inside longtable not producing xml:id

vanatteveldt:
## Summary

`\label{...}` inside a `longtable` environment does not produce a matching `xml:id` on the generated `<ltx:table>` element, causing cross-references (`\ref{...}`) to that label to become dangling.

**Root cause:** `\lx@longtable@label` (line 133) stores the label in `LONGTABLE_LABEL`, and `longtableBindings` (line 79) clears it on entry, but the `afterDigest` callback in `\@@longtable` never reads `LONGTABLE_LABEL` to set the `label` property on the whatsit. Since the constructor template uses `labels='#label'`, the property must be set for the `xml:id` to appear.

**Fix:** Add 3 lines to `afterDigest` to read `LONGTABLE_LABEL` and set it as the `label` property, matching how `LONGTABLE_PROPERTIES`, `LONGTABLE_CAPTIONS`, etc. are already handled.

## Minimal reproduction

```latex
\begin{longtable}{ll}
\caption{Example}\label{tab:example}\\
\hline
A & B \\
\end{longtable}

See Table~\ref{tab:example}.
```

Without the fix, the `\ref` produces `labelref="LABEL:tab:example"` in the intermediate XML but no element has a matching `xml:id`, so `latexmlpost` emits "Missing Target for Label: LABEL:tab:example".

## Test plan

- [x] Verified the fix resolves the dangling reference in our test document
- [x] Verified that regular `table` floats with `\label` are unaffected

### DIFF
--- a/lib/LaTeXML/Package/longtable.sty.ltxml
+++ b/lib/LaTeXML/Package/longtable.sty.ltxml
@@ -46,6 +46,9 @@ DefConstructor('\@@longtable [] Undigested DigestedBody',
   afterDigest  => sub {
     my ($stomach, $whatsit) = @_;
     $whatsit->setProperties(%{ LookupValue('LONGTABLE_PROPERTIES') || {} });
+    # Apply the stored \label to the table element.
+    if (my $label = LookupValue('LONGTABLE_LABEL')) {
+      $whatsit->setProperty(label => $label); }
     # Insert caption and toccaption, if any were encountered.
     if (my $captions = LookupValue('LONGTABLE_HEAD_CAPTIONS') || LookupValue('LONGTABLE_CAPTIONS')) {
       my ($toccaption, $caption) = @$captions;


vanatteveldt:
Note that I used Claude Code to help prepare this PR because we encountered the problem when using latexml to create JATS XML versions of manuscripts for our journal (computationalcommunication.org).

Specifically, an author had a 5-page longtable  in the appendix which they referred to using a standard \ref / \label pair, but the longtable did not create the xref in the xml. This generated a JATS validation error as the reference didn't resolve.

Happy to provide more details if useful!

dginev (approved):
Seems deceptively correct. Or just correct. Thanks!

vanatteveldt:
Thanks! I'll have a look at this and PR2786 later this week and add a test case for this as well if I manage.

brucemiller:
Nice catch, but you confused me. It turns out I've already fixed that in the frontmatter refactoring PR #2767... but then forgot I had! Hopefully that'll be mereable soon...

brucemiller:
OK, PR #2767 is merged, and fixes this problem. Thanks for the report, and even moreso: Thanks for the patch, even if it ultimately wasn't needed.

PR #2786, opened 2026-04-03
booktabs: mark rows between \toprule and \midrule as table headers

vanatteveldt:
## Summary

Tables using booktabs (`\toprule`, `\midrule`, `\bottomrule`) don't produce `<thead>`/`<tbody>` grouping in the output. All rows end up in `<tbody>`, so downstream converters (JATS, HTML) cannot distinguish header rows from data rows.

**Root cause:** The current `\toprule` and `\midrule` macros simply expand to `\hline` without setting the `in_tabular_head` flag on the Alignment.

**Fix:** Change `\toprule` and `\midrule` from simple macro expansions to DefMacro subs that set `$$alignment{in_tabular_head}` before returning the `\hline` tokens. `\toprule` enables the flag (marking subsequent rows as headers), `\midrule` disables it (ending the header region). This matches the universal booktabs convention where header rows appear between `\toprule` and `\midrule`.

## Minimal reproduction

```latex
\begin{tabular}{ll}
\toprule
Name & Value \\
\midrule
A & 1 \\
B & 2 \\
\bottomrule
\end{tabular}
```

**Before:** All rows in `<tbody>`, no `<thead>`.
**After:** "Name & Value" row in `<thead>`, data rows in `<tbody>`.

## Test plan

- [x] Verified correct `<thead>`/`<tbody>` output for booktabs tables in our JATS conversion pipeline
- [x] Verified that tables without booktabs are unaffected
- [x] Verified that `\bottomrule` behavior is unchanged (no header flag manipulation needed)

### DIFF
--- a/lib/LaTeXML/Package/booktabs.sty.ltxml
+++ b/lib/LaTeXML/Package/booktabs.sty.ltxml
@@ -19,10 +19,17 @@ use LaTeXML::Package;
 # Except for the encodings of hline as t (normal) or T (thin)
 # there's currently no support for variable thickness.
 
-# \toprule[thickness]  doubled
-DefMacro('\toprule[Dimension]', '\hline\hline');
-# \midrule[thickness]
-DefMacro('\midrule[Dimension]', '\hline');
+# \toprule[thickness]  doubled; marks subsequent rows as header rows
+DefMacro('\toprule[Dimension]', sub {
+    if (my $alignment = LookupValue('Alignment')) {
+      $$alignment{in_tabular_head} = 1; }
+    return (T_CS('\hline'), T_CS('\hline')); });
+
+# \midrule[thickness]  ends the header region
+DefMacro('\midrule[Dimension]', sub {
+    if (my $alignment = LookupValue('Alignment')) {
+      $$alignment{in_tabular_head} = 0; }
+    return (T_CS('\hline')); });
 # \bottomrule[thickness] doubled
 DefMacro('\bottomrule[Dimension]', '\hline\hline');
 


vanatteveldt:
Note that I used Claude Code to help prepare this PR because we encountered the problem in the wild when using latexml to create JATS XML versions of manuscripts for our journal (computationalcommunication.org). We recommend booktabs to our authors, and this didn't create valid JATS XML as the thead/tbody structure was not generated properly . 

Happy to provide more details if useful!

dginev (approved):
Bruce would know better than me, but to me this reads well.

It is certainly a reasonable heuristic guess. Maybe we can dig up some tables which use `\toprule` without having a true header, but the vast majority of examples I have seen would indeed follow the PR's guess.

dginev:
@vanatteveldt you will also need to update the tests, as you will change certain cases - also a good opportunity to review if those cells are meaningfully improved or damaged.

To regenerate, you can use a call to
```
perl tools/maketests t/.../name-of-changed-test.tex --xml
```

vanatteveldt:
Thanks, I'll have a look at it later this week!

brucemiller:
This is pretty cool... But!  I'm concerned that it only works if people use the strict pattern "top header mid data bottom". I can imagine someone using multiple midrules (or cmidrule). If they have no mid, the whole table will be header; There's some pattern sniffing in the Alignment module (which is very opaque); maybe that can be adapted?

brucemiller:
While I was contemplating how I might recognize the corner cases (eg both top & bottom rules, but no mid) in order to unmark the headers, I tested out your example above. The current development code seems to be converting this table, and guessing the headers correctly. The jats output has the first row within a `thead` and the remainder in `tbody`.  Are you using an older LaTeXML? Or perhaps your real use cases are confusing LaTeXML?

PR #2787, opened 2026-04-06
acl.sty binding with new frontmatter framework

dginev:
This is an experiment (though not a draft - it works) in writing bindings with AI support.

It _partially_ uses the new frontmatter framework, but had to deviate due to math-mode superscript marks not having an obvious existing API. Maybe this is motivation to extend with something on the lines in the PR here?

The binding itself is an often reported problem in arXiv's html_feedback repository, widely used by ACL conference submissions.


dginev:
To jot down two notes from the private discussion:

 - the "scripted marks" infrastructure belongs in the main frontmatter API, this initial PR state only showcases the need for it. 
 
 - there is a need for a proper manual readthrough of acl.sty docs as well as a good handful more examples from arXiv. We can not rely on the AI survey to be exhaustive.

PR #2789, opened 2026-04-09
[WIP] Incrementally better box sizes

dginev:
[Deyan]: I have been collecting, studying and testing with the oversized `{tcolorbox}` examples reported to us by arXiv users. There is a very diverse range of constructs that are affected. In a way, the colored boxes become a similar testing harness to the red/green background Web Platform Tests -- we create an SVG which must set a predetermined width/height size, so whenever the size computation does not completely accurately reflect the way the browser renders the contents of the SVG, we see obvious breakage in a browser window.

This is both very useful for debugging, and very challenging to correct - since we need everything from TeX registers to Box modes for enumerations to be set accurately for the full sizing math to be accurate.

This PR uses Opus 4.6 to do an exploratory pass allowing  eleven tcolorbox sizing configurations to be near-accurate. Especially the changes to Font.pm, and I suspect others, require very high level judgement calls, so this is definitely a draft PR aiming to start a discussion.

The `{tcolorbox}` height mismatches were the most reported issue on arXiv's HTML pages in 2025.   

### Summary (generated)

This branch improves LaTeXML's estimation of TeX box dimensions (height, width, depth) used for SVG foreignObject sizing in the HTML output. The primary motivation is accurate tcolorbox rendering on arXiv, but the changes benefit all content that relies on box measurement —
minipages, parboxes, captions, list environments, and xy-pic diagrams.

### Core changes

- **Box sizing algorithm overhaul** (`Font.pm`): Rewrites `computeBoxesSize` to group loose character boxes into pseudo-paragraphs, track per-line `\baselineskip`, handle CJK line-height (1.1x multiplier), insert display math skip glue, and flatten inline formatting wrappers
(`\emph`, `\textbf`, etc.) for accurate line-wrapping estimation — with guards to protect pre-sized Whatsits like xy-pic arrowheads.

- **Baselineskip synchronization**: `MergeFont` now updates `\baselineskip` when font size changes, mirroring TeX's `\selectfont` chain. `\@setfontsize` is rewritten as a primitive. `\setstretch` stores a global factor applied to baselineskip. The active `\baselineskip` is
captured at paragraph-end in `repackHorizontal` and used for faithful vertical stacking.

- **foreignObject sizing pipeline** (`TeX_Box.pool.ltxml`): Sets `font-size` in px on foreignObjects so em-based CSS widths resolve correctly in SVG coordinates. Computes line-height from baseline/font-size ratio. Converts descendant widths from pt to em. Clamps `\hbox` width
near `\hsize` to prevent minor tabular overflows.

- **TeX register defaults**: Sets `\hsize`, `\spacefactor`, `\hangafter`, and list spacing registers (`\topsep`, `\partopsep`, `\itemsep`, `\parsep`) to real LaTeX 10pt article values.

### Paragraph and list environment fixes

- Brace groups propagate horizontal mode outward (matching TeX's nest/save stack separation)
- `\addvspace` implemented as `\vskip` (was a no-op)
- `\preitem@par` gets `afterDigest` + height properties for proper paragraph closing
- List environments migrated from `beforeDigestEnd` to `afterDigestBody` across 6 packages
- `\textcolor` adds `\leavevmode` to prevent vertical-mode height inflation

### Package-specific fixes

- **tcolorbox**: Equal-height group support, `\tcbline` override, `noparskip`, `\tcb@use@autoparskip` pre-definition
- **setspace/doublespace**: Spacing environments use `internal_vertical` mode
- **fancyvrb**: Pre-define `\FancyVerbSpace` register
- **CJK**: New `CJK.sty.ltxml` with proper environment definitions
- **soul**: Sizer and mode for `\hl` inline highlight

### Tests

- **New tcolorbox test suite** (`t/68_tcolorbox.tt`): 11 CI-only test cases covering plain text, enumerations, equal heights, CJK, verbatim, resizebox, separators, tabular captions, titled equations, and tool-calling content
- **New colorbox_sizes test** for vbox dimension precision
- **27 regenerated test XMLs** across alignment, babel, complex, fonts, graphics, structure, and theorem suites


lib/LaTeXML/Common/Font.pm
```diff
@@ -652,20 +663,82 @@ sub computeBoxesSize {
...
+        @hrun = ();
+        push(@grouped, $box); } }
+    if (@hrun >= 10) {
+      my $l = LaTeXML::Core::List->new(@hrun);
+      $l->setProperty(mode                => 'horizontal');
+      $l->setProperty('_pseudo_paragraph' => 1);
+      push(@grouped, $l); }
+    elsif (@hrun) {
+      push(@grouped, @hrun); }
+    foreach my $box (@grouped) {
```
brucemiller, line 689: This whole block should be removed; it's trying to patch up lists that haven't been properly finished, but it's kinda too late. The main problem is that most (but not all) cases where `internal_vertical` mode is begun,  should be preceded by `leaveHorizontal` (or leaveHorizontal_internal`) in order to pack the preceding paragraph.
The exceptions to that are effectively `inline_internal_vertical`, where the content is in vertical mode, but they don't disrupt the current flow. This would include `\footnote,{picture}`, and even `\vbox`.  And this suggests a minimally disruptive patch to fix it, which works, but we should probably discuss.  And it even seems to tighten the height of some of your color boxes.

brucemiller:
This should all be covered by #2798, right? (other than adding complex tests). Can we close?

PR #2790, opened 2026-04-09
[CI] update actions/checkout to v6

dginev:
We were getting CI warnings that `actions/checkout@v2` is about to be phased out in September 2026, so updating to the current version.


brucemiller (approved):
Ah, ok, sounds good; Thanks!

PR #2794, opened 2026-04-16
Simplify perlcritic configuration (to work with stock ubuntu install)

brucemiller:
As it says... It still seems to check the basics/important stuff.


dginev (approved):
Simpler. Thanks!

PR #2713, opened 2025-12-17
update list environments to internal_vertical

dginev:
This is a first step towards addressing the mixed-mode lists that are encountered when dealing with latex's itemization environments.

- The main contribution is declaring them all as `mode => "internal_vertical"`
- `\preitem@par` seemed to have an inaccuracy (I think) in checking the current element name is not an itemization element.
- interestingly, the final digested `\par` doesn't really seem needed for our tests to pass. The autoOpen/autoClose schema guarantees may be sufficient to deal with the markup side.

What is still missing here is ensuring the horizontal material of the last item gets repacked at an itemization's end.

PR #2798, opened 2026-04-23
Leavehorizontal

brucemiller:
This PR corrects the mode of many commands and environments to `internal_vertical`. When entering a vertical mode, we generally must "finish the paragraph being built", which means that `$stomach->beginMode('internal_vertical');` (or `display_math`) should call `leaveHorizontal` in order to pack up any previously digested boxes into a horizontal (paragraph) list.  However, a number of commands, notably `\vbox`, are actually vertical *inside*, but stay in the flow; that is they are inline-blocks. For these we introduce a variant mode `inline_internal_vertical`, which does *not* call `leaveHorizontal`.

In the process of testing, it was discovered that hooks, which often cause material to be processed by `beforeDigest` daemons defer inserting them into the `@LaTeXML::LIST`, so that they fail to be collected by such repacking.  So, `beforeDigest,` at the very least, should *not* be deferred but immediately pushed onto `@LIST`. It is probably the case that many other such deferred pushes should be immediate, but this will take some care.

This PR is a preliminary to addressing the issues raised in #2789; it avoids the need for `$font->computeBoxesSize` to attempt to guess whether the boxes failed to be repacked into a paragraph. (This PR does not yet recover the sizing improvements of #2789, however; There are many other useful bits there!).


dginev (approved):

brucemiller:
Phew! OK, I think I've got this done finally.  The handling of modes is much more TeX-like and consistent. I've refactored the sizing algorithms to be more TeX-like, as well, and they include the effects of more TeX parameters (`\baselineskip`, in particular).  It seems to do a tolerable job, if not perfect, with most of the test cases that were proposed in #2789.  In the process, I think it also incorporated the changes from #2713 (please check).

Also, since I had to deal with frames and padding, perhaps this also solves the issue raised by @xworld21 in #2820 ?

*AND* it looks like we've got a Windows problem in CI again.

dginev:
Windows CI issue was transient, but also a good moment to add a bit of robustness to survive Chocolatey flakes. (transient HTTP 503 failures to fetch). I pushed here, but if that is a problem please revert and I will make a new PR.

As to #2713 , here is a brief outline of what remains outside of this new PR, produced by the "co-author" of PR 2713:

---
  This branch supersedes the core of PR 2713 nicely — {itemize}/{enumerate}/{description} are internal_vertical with the beforeDigestEnd \par dropped, the real \topsep/\itemsep/\parsep/… defaults are in, and the reworked \preitem@par (emitting \vskip\itemsep/\vskip\parsep). In addition, leaveHorizontal finally closes PR 2713's open "repack the last item" problem. IEEEtran/elsarticle/enumitem are carried over too.

  A few of PR 2713's edits didn't make it in, though, so it's not a full pickup:

  1. paralist.sty.ltxml — not ported at all. PR 2713 set mode on all 11 environments (inparaenum, compactenum, asparaenum, inparaitem, compactitem, asparaitem, the redefined enumerate/itemize, inparadesc, compactdesc, asparadesc); here they still carry no mode, so they digest in horizontal mode and miss both
  the repack and the inter-item spacing the new \preitem@par now assumes.
    - Note: with inline_internal_vertical now available, the right split is probably inline lists (inparaenum/inparaitem/inparadesc → ltx:inline-*) on inline_internal_vertical and the block compact/aspara lists on internal_vertical, rather than PR 2713's blanket internal_vertical.
  2. sv_support.sty.ltxml {description} — not ported. Still beginItemize('description'); PR 2713 added the '@desc' label key, mode => 'internal_vertical', and beforeDigest => Let('\makelabel','\descriptionlabel') (a real label fix, not just mode).
  3. parskip.sty.ltxml — trivial header/TODO comment, harmless to skip.

---

Is this helpful?

brucemiller:
Very perceptive comments; I did indeed overlook changes to `paralist,sv_support`.  I'll deal with those on Monday.

Thanks for the CI fix!

xworld21:
> Also, since I had to deal with frames and padding, perhaps this also solves the issue raised by @xworld21 in #2820 ?

My list of issues in #2820 is not exhaustive. The problem is the blanket `.ltx_framed { padding: 3pt; }` rule which gets applied indiscriminately to anything framed, including e.g. \hrule and \vrule, not just underlined text. The padding should really be only for framebox type things (and many of those need to read `\fboxsep`). So this PR improves by moving under/overlining away from framed, but it doesn't address the underlying problem I was trying to fix. I think my PR is mostly orthogonal to this one.

In fact, I don't see any changes to the rule constructors here? I should send another PR then! Some very small changes to improve the accuracy of \hrule, \vrule, and getting rid of the non-breaking whitespace which causes me so much grief.

I'll check properly during the weekend though, I might be looking at the code wrong. It's a lot of changes!

brucemiller:
I've incorporated the missing bits @dginev noticed.   I think it'll be best to hold off any further framing tweaks to a followup PR. So, if this otherwise meets with satisfaction, would be good to call it done & merge.

lib/LaTeXML/Package/paralist.sty.ltxml
```diff
@@ -113,34 +117,40 @@ DefMacro('\setdefaultitem{}{}{}{}', sub {
...
     "<ltx:itemize xml:id='#id'>#body</ltx:itemize>",
-    properties => sub { beginItemize('itemize', '@item'); },
+    properties       => sub { beginItemize('itemize', '@item'); },
     afterDigestBegin => sub { setItemizationStyle($_[1]->getArg(1)); }); }
 
 #======================================================================
 # Descriptions
 DefEnvironment('{inparadesc}',
-  "<ltx:inline-description xml:id='#id'>#body</ltx:inline-description>",
+  "<lttx:inline-description xml:id='#id'>#body</ltx:inline-description>",
```
dginev: `lttx` typo

lib/LaTeXML/Package/paralist.sty.ltxml
```diff
@@ -113,34 +117,40 @@ DefMacro('\setdefaultitem{}{}{}{}', sub {
...
-  properties => sub { beginItemize('itemize', '@item'); },
+  mode             => 'internal_vertical',
+  properties       => sub { beginItemize('itemize', '@item'); },
   afterDigestBegin => sub { setItemizationStyle($_[1]->getArg(1)); });
 
 if (IfCondition(T_CS('\if@plnewitem'))) {
   DefEnvironment('{itemize} OptionalUndigested',
     "<ltx:itemize xml:id='#id'>#body</ltx:itemize>",
-    properties => sub { beginItemize('itemize', '@item'); },
+    properties       => sub { beginItemize('itemize', '@item'); },
```
dginev, line 138: missing `mode => 'internal_vertical'` in this redefined `{itemize}` ?

dginev (commented):
Two small changes, then good to merge by me 👍🏻

dginev (approved):

PR #2806, opened 2026-05-18
Add dirtytalk binding

AMBRA7592:
Summary

Adds a LaTeXML binding for the `dirtytalk` package, covering the common `\say{...}` command and nested quotation behavior.

- Defines `dirtytalk.sty.ltxml` with default outer/inner quotation marks.
- Handles package options for custom `left`, `right`, `leftsub`, and `rightsub` quote symbols.
- Adds a focused structure test for plain and nested `\say{...}` usage.
- Updates `MANIFEST` for the new binding and fixtures.

Validation
- `git diff --check`
- `CI=true perl t/97_manifest.t`
- Syntax/load check of `dirtytalk.sty.ltxml`

Full `t/50_structure.t` was not run locally because this environment is missing LaTeXML's Perl dependency chain, starting with `Text::Unidecode`.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -486,6 +486,7 @@ lib/LaTeXML/Package/dcolumn.sty.ltxml
 lib/LaTeXML/Package/deluxetable.sty.ltxml
 lib/LaTeXML/Package/diagbox.sty.ltxml
 lib/LaTeXML/Package/ding.fontmap.ltxml
+lib/LaTeXML/Package/dirtytalk.sty.ltxml
 lib/LaTeXML/Package/doi.sty.ltxml
 lib/LaTeXML/Package/doublespace.sty.ltxml
 lib/LaTeXML/Package/dsfont.sty.ltxml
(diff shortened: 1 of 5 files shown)


dginev:
> Full t/50_structure.t was not run locally because this environment is missing LaTeXML's Perl dependency chain, starting with Text::Unidecode.

I'll run these in CI, but you should really install the full dependency chain to preempt any failures.

brucemiller:
Actually, this looks quite reasonable, other than some minor bookkeeping: missing `dirtytalk.pdf`, which is useful for quick comparison. `tools/maketest t/structure/dirtytalk` will create it for you, but you'll want to add another line in `MANIFEST`. Thanks in advance!

AMBRA7592:
I just added `t/structure/dirtytalk.pdf` and the MANIFEST entry in 7c24ef6. Thanks!

brucemiller (approved):
Cool! Thanks!!

Issue #2253, opened 2023-10-30
llncs: `\tocauthor` rendered in page


nschloe:
MWE:
```latex
\documentclass{llncs}

\begin{document}

\toctitle{Lecture Notes in Computer Science}
\tocauthor{Authors' Instructions}

Lorem.
\end{document}
```

LaTeXML output:
```xml
  <para xml:id="p1">
    <p>Authors’ Instructions</p>
  </para>
  <para xml:id="p2">
    <p>Lorem.</p>
  </para>
```

LaTeX output:

![screenshot_2023-10-30-150739](https://github.com/brucemiller/LaTeXML/assets/181628/25da9afe-02d6-4c8c-88a6-080433f0934e)

PR #2815, opened 2026-06-09
llncs: don't leak \tocauthor & friends into the body (#2253)

jackwillis:
\tocauthor, \authorrunning, \titlerunning and \toctitle were defined as token registers. A bare register doesn't consume a following {...}; it relied on TeX's optional-'=' assignment to swallow the argument, which is fragile and position-dependent, so e.g. \tocauthor{Authors' Instructions} leaked its argument into the body text.

These commands take an argument that only feeds the running heads / table of contents (which LaTeXML doesn't render), so define them as argument-gobbling macros instead. The internal \authrun/\titrun token registers are left unchanged.

Adds t/85_llncs.t and t/llncs/toc.{tex,xml,pdf}.


lib/LaTeXML/Package/llncs.cls.ltxml
```diff
@@ -75,12 +75,17 @@ DefRegister('\instindent' => Dimension(0));
 
 # DefMacro('\lastand',Tokens());
 
-DefRegister('\authrun'       => Tokens());    # Box!
-DefRegister('\authorrunning' => Tokens());
-DefRegister('\tocauthor'     => Tokens());    # Box!
-DefRegister('\titrun'        => Tokens());    # Box!
-DefRegister('\titlerunning'  => Tokens());
-DefRegister('\toctitle{}'    => Tokens());
```
brucemiller, line 83: Actually, llncs defines these as token registers, and changing them to macros would cause `\authorrunning={something}` to fail.  In fact, I'd originally had them as macros, but changed them to registers to be more "correct". Unfortunately, I made a typo, leaving the `{}` in the definition of `\toctitle`! That's actually the cause of the problem you saw; it get's the processing out of step.

brucemiller (changes requested):

brucemiller:
This got fixed by #2847

PR #2818, opened 2026-06-13
listings: do not look up ltxml files when reading raw files

### LINKED ISSUE #2816: \lstinputlisting should not read .ltxml files
Funny bug: the following example from the listings package
```
\lstinputlisting[lastline=4,language=TeX]{listings.sty}
```
will read `listings.sty.ltxml` istead of `listings.sty`.

The solution is probably as easy as adding `noltxml => 1` here:
https://github.com/brucemiller/LaTeXML/blob/51fea96a31444f4210e0f3056a7ee83135d9cc1e/lib/LaTeXML/Package/listings.sty.ltxml#L323

xworld21:
Fix #2816.


brucemiller (commented):
Ah, of course!! (and explicit `ltxml` extensions still work!) Thanks!!

PR #2819, opened 2026-06-13
listings: create group around identifiers

### LINKED ISSUE #2817: \underbar behaves incorrectly in listings example
Bug that surfaces in multiple ways depending on the LaTeXML version (or maybe BookML patches?). At a minimum, `\underbar` applies to the first character only, rather than the entire keyword. The further problem is that it creates a frame around the character instead of just underlining it. 
```latex
\documentclass{article}
\usepackage{xcolor}
\usepackage{listings}
\begin{document}
\lstset{language=Pascal,numbers=left,frame=tblr}
{\lstset{% general command to set parameter(s)
       basicstyle=\small,          % print whole listing small
       keywordstyle=\color{black}\bfseries\underbar,
                                   % underlined bold black keywords
       identifierstyle=,           % nothing happens
       commentstyle=\color{red}, % white comments
       stringstyle=\ttfamily,      % typewriter type for strings
       showstringspaces=false}     % no special string spaces
\begin{lstlisting}
for i:=maxint to 0 do
begin
    { do nothing }
end;

Write('Case insensitive ');
WritE('Pascal keywords.');
\end{lstlisting}}
\end{document}
```
which I have taken from the listings documentation renders as

Image

using "LaTeXML 0.8.8 (51fea96a)".

Running with LaTeXML 0.8.8 (and some BookML patches), I get this

Image

which creates the correct border, still only on the first character (the undesired padding is a different issue, probably already solved).

xworld21:
Easier than I thought it would be. Fix #2817.


xworld21:
I have read my PR more carefully, and noticed that one has to go just a little deeper. So much for writing a quick patch! This is now a three lines change instead of one, including a tiny bugfix (the index code used to push the 'begin' tokens, but omit the 'end' tokens, which becomes a problem when you open and close an additional group).

I have added one use of \underbar to the test.

brucemiller (approved):
Nice! Basically, the extra `{}` copes with the style being a command taking an argument, rather than just changing the current style.  Thanks!

PR #2824, opened 2026-06-21
do not add frame and background to inline listings

### LINKED ISSUE #2823: \lstinline should not have frame and background
Per listings documentation, 'code snippets' (i.e. anything `\lstinline`) do not support frames and background colors. LaTeXML however adds both.

xworld21:
Fix #2823.

### DIFF
--- a/lib/LaTeXML/Package/listings.sty.ltxml
+++ b/lib/LaTeXML/Package/listings.sty.ltxml
@@ -58,6 +58,7 @@ DefMacro('\lx@lstinline OptionalKeyVals:LST', sub {
     lstActivate($keyvals);         # But do BEFORE reading arg, since some options screw things up.
                                    # get opening delim from gullet, before cattable swap,
                                    # in case token was already peeked for the optional args!
+    AssignValue('LISTINGS_INLINE' => 1);
     my $init = $gullet->readToken;
     my $body = listingsReadRawString($gullet, (Equals($init, T_BEGIN) ? T_END : $init));
     return (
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/listings.sty.ltxml
```diff
@@ -954,7 +957,7 @@ DefPrimitive('\lst@@backgroundcolor Until:\end', sub {
     lstPushValueLocally(LISTINGS_PREAMBLE_BEFORE => T_CS('\lst@@@set@background')); });
 
 DefPrimitive('\lst@@@set@background', sub {
-    MergeFont(background => LookupValue('LISTINGS_BACKGROUND')); });
+    LookupValue('LISTINGS_INLINE') ? undef : MergeFont(background => LookupValue('LISTINGS_BACKGROUND')); });
```
brucemiller: Might be slightly cleaner to not return an explicit `undef`. How about:
```
DefPrimitive('\lst@@@set@background', sub {
    MergeFont(background => LookupValue('LISTINGS_BACKGROUND'))
        unless LookupValue('LISTINGS_INLINE');
  return;  });
```

brucemiller (changes requested):

lib/LaTeXML/Package/listings.sty.ltxml
xworld21: Makes sense.

brucemiller:
Beautiful! Thanks!!

Issue #2822, opened 2026-06-18
binary operators can get wrong spacing when formula is split by &


xworld21:
Another MathML operator issue. When a formula is broken into cells, operators appearing next to `&` get the wrong treatment in MathML. For example,
```latex
\begin{align*}
a & +b
\end{align*}
```
will output `<td><math><mrow><mo>+</mo>...`. MathML and MathJax classify that `+` as a prefix operator, and so render it with no spacing to the right.

TeX is smarter and uses the binary operator spacing even when `&` breaks the formula. LaTeXML should (somehow) know to output `<mo form="infix">+</mo>`, or do like MathJax and inject an empty `<mi></mi>`. Either of those will produce the correct spacing.

xworld21:
Interesting: this behaviour is from amsmath, not TeX. Aligned equations in amsmath have an extra `{}` at the beginning of even columns. Indeed, you are always supposed write `&=`, and not `=&`! On the bright side, this should be easy to implement, no guessing or anything, just systematically adding an empty token in the even column to the environments that use this trick.

PR #2825, opened 2026-06-21
add empty identifier in left-aligned columns of AMS aligned equations

xworld21:
First attempt at fixing #2822: add a `<XMTok meaning="absent"/>` in the same place AMS would add `{}`, and filter it out when building the non-aligned part of MathFork.

This produces the same XML in most cases (equations starting with a binary relation, like `=` or `<`), as the math parser was already adding its own `<XMTok meaning="absent"/>`. The difference is when the line starts with something that could be a prefix unary operator, like `+`: the tests show that you now get the correct markup.

This produces the right markup in most[^1] cases, while being visually accurate in all cases. The remaining problem is when the left-aligned column starts with a non-binary operator, as the math parser creates `absent *`, so an extra `<mo>`, which can interfere with screen readers. This should be fixable in construction by checking if 'absent' is followed by identifier or similar... but I have run out of time today, so I am putting this out as a starting point, and I'll probably come back to it in a few days.

[^1]: Claim not supported by data!


lib/LaTeXML/Package/amsmath.sty.ltxml
```diff
@@ -446,11 +446,14 @@ DefConstructor('\@@amsalign SkipSpaces DigestedBody',
   afterConstruct => sub { rearrangeAMSAlign($_[0], $_[0]->getNode->lastChild); });
 DefPrimitive('\end@amsalign', sub { $_[0]->egroup; });
 
+# Placeholder token to make operators infix when at the start of a left-aligned cell (see MathParser.pm for the 'absent' meaning)
+DefConstructorI('\@ams@absent', undef, '<ltx:XMTok meaning="absent"/>', reversion => '', alias => '');
```
xworld21, line 450: I am currently playing with this crude hack. However, it would be just better if the parser could recognise the sequence 'absent identifier' followed by 'autoinserted invisible times' and get rid of it (or better yet, not generate it in the first place).
```suggestion
DefConstructorI('\@ams@absent', undef, '<ltx:XMTok meaning="absent"/>', reversion => '{}', alias => '');
DefConstructorI('\@ams@absent@cleanup', undef, '', reversion => '', alias => '', beforeConstruct => sub {
    my ($document) = @_;
    my $cell       = $document->getElement;
    my $absent     = $document->getFirstChildElement($cell);
    my $node       = $document->findnode('node()[2]', $cell);
    if ($node) {
      my $role = $node->getAttribute('role');
      if (!defined $role) {
        my $tag = $document->getModel->getQName($node);
        # TODO resolve XMRef (is it possible at this stage?)
        if ($tag eq 'ltx:XMDual') {
          my ($content, $presentation) = element_nodes($node);
          $role = $content->getAttribute('role') || $presentation->getAttribute('role') || ''; } }
      # remove the empty identifier if it is not followed by a possible binary operator
      if (!grep { $role eq $_ } ('RELOP', 'ARROW', 'METARELOP', 'MODIFIEROP', 'ADDOP', 'MULOP', 'BINOP', 'APPLYOP', 'COMPOSEOP')) {
        $document->replaceNode($absent, ()); } }
});
```

xworld21:
This PR is good as proof of concept, but it's bad at everything else. I am currently shipping the following version
https://github.com/vlmantova/bookml/blob/6b4764a2f8b93a3b6464d2c260dbc52c0bf61b88/bookml.sty.ltxml#L641-L726
which at least is robust, but still problematic. The issue is ultimately about the grammar and it should not be solved in the binding. What's missing is a hint that tells the grammar that the following operator should not be prefix.

PR #2829, opened 2026-06-24
Framing

brucemiller:
This PR defines a handy helper function `framedProperties(%options)`, which takes options for `margin,rule,color,backgroundcolor`. The margin and rule can be Dimensions, or Register commands, or just strings containing them. This function provides more consistent attributes, and also adds properties for padding which helps size calculations. Since the margin is more consistently handled, the class `.ltx_framed` no longer adds padding by default.

I've used this on several framing commands and environments, except the ones addressed in #2820. @xworld21 Please take a look & comment; if you approve, apply them to your PR


lib/LaTeXML/Package/framed.sty.ltxml
```diff
@@ -95,24 +83,25 @@ DefEnvironment('{snugshade*}', sub {
 DefEnvironment('{leftbar}', sub {
     my ($document, %props) = @_;
     $_[0]->maybeCloseElement('ltx:p');    # this starts a new vertical block
-    insertBlock($document, $props{body}, framed => 'left', framecolor => $props{framecolor},
-      cssstyle => 'border-width:3pt;padding-left:10pt'); },
-  properties => sub { (framecolor => Black); });
+    insertBlock($document, $props{body}, %props); },
+  properties => sub { (framedProperties(
+        frame => 'left', color => Black, margin => '10pt', rule => '3pt')); });
```
xworld21: This is where the abstraction is too weak. Sizing will pick up further 13pt from all sides, will it not?

lib/LaTeXML/Engine/latex_constructs.pool.ltxml
```diff
@@ -4653,7 +4653,7 @@ DefConstructor('\mbox {}',
   sizer   => '#1',
   robust  => 1);
 
-our %makebox_alignment = (l => 'left', r => 'right', s => 'justified');
+our %makebox_alignment = (l => 'left', r => 'right', c => 'center', s => 'justified');
```
xworld21: We might as well do
```suggestion
our %makebox_alignment = (l => 'left', r => 'right', c => 'center', s => 'stretched');
```
The class `ltx_align_justified` does not exist anyway.

Unfortunately, emulating `stretched` in CSS seems pretty hard!

lib/LaTeXML/Engine/latex_constructs.pool.ltxml
```diff
@@ -4695,15 +4695,12 @@ DefConstructor('\@framebox[Dimension][]{}',
...
-      ($sep ne '3.0pt' ? (cssstyle => 'padding:' . $sep_pts) : ()),
-      ($_[1]           ? (width    => $_[1])                 : ()),
-      padtop => $pad, padbottom => $pad, padleft => $pad, padright => $pad,
-    ); },
+    my $width = $_[1];
+    my $align = ToString($_[2]);
+    $align = 'c' if !$align && $width;
+    (framedProperties(margin => '\fboxsep', rule => '\fboxrule'),
+      ($width ? (width => $width)                     : ()),
+      ($align ? (align => $makebox_alignment{$align}) : ())); },
```
xworld21, line 4706:
```suggestion
      ($align ? (align => $makebox_alignment{$align} // 'center') : ())); },
```
to respect the default value.

lib/LaTeXML/Engine/latex_constructs.pool.ltxml
xworld21, line 4690: Reminder (but should be a separate discussion) that this loses with, color and alignment. But really, to be discussed somewhere else. The correct MathML is tricky.

xworld21 (commented):
The approach is sound, I have suggested a couple of tweaks, but nothing major.

Mikropenis (commented):
Pm

lib/LaTeXML/Package/framed.sty.ltxml
brucemiller: Yeah, I knew I was overgeneralizing. Good to get called out :>

lib/LaTeXML/Engine/latex_constructs.pool.ltxml
brucemiller: Sure, that seems legit.

lib/LaTeXML/Engine/latex_constructs.pool.ltxml
brucemiller: That's what line 4700 should be doing, but no need to set the alignment if there isn't a width.

lib/LaTeXML/Engine/latex_constructs.pool.ltxml
brucemiller: I patched that it doesn't (over)simplify when there's a width, so that should be good. Color maybe lost?

xworld21: `<menclose>` uses the current `@mathcolor`, not the CSS border color, so we should be fine?

lib/LaTeXML/Engine/latex_constructs.pool.ltxml
xworld21: Right, I missed that.

xworld21 (commented):
LGTM

brucemiller:
Thanks all!

Issue #2725, opened 2025-12-30
keep class to distinguish multline, gathered, etc


xworld21:
It can be useful to know which environment has produced a particular aligned equation. I am thinking of multline, which looks good only after applying some suitable CSS (roughly speaking, setting `width:100%` on the table, with details depending on how the HTML is generated).

However, LaTeXML does not seem to add a suitable class like `ltx_multline` so it's hard to guess when the `width:100%` is appropriate. In XML, the only trace of the environment is in
```xml
<XMArray name="multline" vattach="middle">
```
and that info is completely lost in HTML.

PR #2835, opened 2026-06-27
use correct alignment rules for AMS multline and implement \shoveleft…

### LINKED ISSUE #2838: amsmath multline uses wrong default alignment, does not implement \shoveleft, \shoveright
To help you assign milestones and such: amsmath `\begin{multline}` should align the middle rows to the center, or to the left if `fleqn`, and `\shoveleft`, `\shoveright` should be available in amsmath as well, not just mathtools. Fixed in #2835.

xworld21:
…, \shoveright

I am being lazy and not reporting a separate issue. The PR fixes the following issues
- LaTeXML aligns multline rows to the right by default, instead of centering them [edit:] or aligning to the left under fleqn
- amsmath does not implement \shoveleft, \shoveright

I haven't tried replicating what multline does when given a single line, which I hope is very rare.

I copied \shoveleft, \shoveright from mathtools, but left the original in mathtools untouched, because that one is supposed to do more, see the TODO comment.

Fixes #2838.


xworld21:
> because that one is supposed to do more, see the TODO comment.

...and just by looking at the PR here on GitHub, I realised it's trivial to implement. I'll push something in a moment.

lib/LaTeXML/Package/mathtools.sty.ltxml
```diff
@@ -586,17 +586,17 @@ DefMacroI('\multlined', '[][]', '\@multlined@tmp{'
...
-    my $rowalignment = LookupValue('MULTIROW_ALIGNMENT_RULE');
-
-    $$rowalignment{$cur_row_num} = ToString($shove_dir);
-    return; });
-
-DefMacro('\shoveright[]{}', '\@MT@shove{right}#2');
-DefMacro('\shoveleft[]{}',  '\@MT@shove{left}#2');
+DefMacro('\shoveleft[]{}', sub {
+    my ($gullet, $dim, $eqn) = @_;
+    $dim = readExpression($gullet, 'dimen', $dim) if defined $dim;
```
xworld21, line 591: Maybe better to introduce new parameter types in calc.sty, like `CalcDimension`, `CalcNumber` instead of calling `readExpression` manually?

lib/LaTeXML/Package/amsmath.sty.ltxml
```diff
@@ -242,9 +245,11 @@ DefConstructor('\@@multline DigestedBody',
     . "</ltx:equation>",
   mode         => 'display_math',
   properties   => sub { RefStepCounter('equation') },
-  beforeDigest => sub { $_[0]->bgroup; },
-  afterDigest  => sub {                              # Nasty trick required to make "body" be the arg.
-    $_[1]->setProperty('MULTIROW_ALIGNMENT_RULE' => { '0' => 'left', '-1' => 'right' });
+  beforeDigest => sub {
+    AssignValue('MULTIROW_ALIGNMENT_RULE' => { 'default' => IfCondition('\if@fleqn') ? 'left' : 'center', '0' => 'left', '-1' => 'right' });
+    $_[0]->bgroup; },
```
brucemiller, line 250: Sorry you've got to be patching such convoluted code! So, you've adopted the bit from `mathtools` where we've got to set a State variable in order for `\shoveleft` to see it, but then it needs to be copied to the Whatsit in order to be preserved for construction time (sigh!). 

But what puzzles me is why I had it *after* the `bgroup`; maybe there's already a grouping, but it looks like the assignment is global. Could you try moving the bgroup in front of the assignment? (it just seems safer/more correct). If it works, we should do that to *all* these `MULTIROW_ALIGNMENT_RULE` environments (amsmath, mathtools)

Also, `\@@gathered` should get this same two-step treatment as well (shoving sorta works there, too, right?); likewise mathtools `\@@lgathered,\@@rgathered`.

lib/LaTeXML/Package/amsmath.sty.ltxml
```diff
@@ -1315,8 +1324,21 @@ DefRegister('\multlinetaggap' => Glue('10pt'));
...
+sub amsShove {
+  my ($shove_dir)  = @_;
+  my $cur_row_num  = LookupValue('Alignment')->currentRowNumber - 1;
+  my $rowalignment = LookupValue('MULTIROW_ALIGNMENT_RULE');
+
+  $$rowalignment{$cur_row_num} = $shove_dir if $rowalignment;
+
+  return;
+}
+
```
brucemiller, line 1341: I like that you've turned these into constructors (or primitives); the previous timing of macro seems dubious.

However, `\shoveleft` could get misused outside of an alignment, so we should probably be more careful, like wrapping the code with:
```
if(my $alignment = LookupValue('Alignment')) {
  my $cur_row_num = $alignment->currentRowNumber - 1;
 ... }
return;

File: lib/LaTeXML/Package/mathtools.sty.ltxml [hunk 4]
```diff
@@ -586,17 +586,17 @@ DefMacroI('\multlined', '[][]', '\@multlined@tmp{'
...
+    return (T_CS('\lx@ams@shoveleft'), T_BEGIN, ($dim ? (T_CS('\kern'), Tokenize($dim->toString)) : ()), $eqn, T_END);
+});
+
+DefMacro('\shoveright[]{}', sub {
+    my ($gullet, $dim, $eqn) = @_;
+    $dim = readExpression($gullet, 'dimen', $dim) if defined $dim;
+    return (T_CS('\lx@ams@shoveright'), T_BEGIN, $eqn, ($dim ? (T_CS('\kern'), Tokenize($dim->toString)) : ()), T_END);
+});
```
brucemiller, line 599: [These as macros should be OK, however]

brucemiller (changes requested):
Looks pretty good, particularly given the awkwardness of the original code.

lib/LaTeXML/Package/amsmath.sty.ltxml
xworld21: > But what puzzles me is why I had it _after_ the `bgroup`; maybe there's already a grouping, but it looks like the assignment is global.

...right. The bgroup/egroup problem comes from the `\lx@end@alignment` token, I think? The group is closed at the end of the body, so before `afterDigest` runs. So I need to set the variable before starting the group. A little odd indeed.  Shall I add an extra `bgroup/egroup`?

> Also, `\@@gathered` should get this same two-step treatment as well (shoving sorta works there, too, right?); likewise mathtools `\@@lgathered,\@@rgathered`.

Not according to documentation, `\shove` is only meant for multline, but I can check if it has an effect or not and emulate it accordingly.

lib/LaTeXML/Package/amsmath.sty.ltxml
xworld21: An error is probably in order as well.

xworld21: > I like that you've turned these into constructors (or primitives); the previous timing of macro seems dubious.

Yes, the previous timing was odd. I have used a constructor in order to preserve `\shove` in reversion. Except for a little problem: the reversion is not wrapped in `\begin{multline}`! This is a variant of #2725... the environment gets lost somewhere in XMath.

I guess for the moment I should simplify the reversion, so that mathimages don't break.

xworld21: I have hidden the reversion, because `\shove` is not valid outside of `multline`, and mathtex/mathimages would consequently break, because the reversion is missing the `\begin{multline*}` around it (separate bug!).

lib/LaTeXML/Package/amsmath.sty.ltxml
xworld21: It turns out that `\shoveleft` is almost a no-op outside of multline (it creates a bit of whitespace which does not look intentional, but it does nothing to the alignment). I have added a guard to check for 'multline' specifically, but no error message, in line with amsmath. Also an extra grouping to keep the assignments local.

t/ams/mathtools.xml
```diff
@@ -4600,7 +4600,7 @@ Then a switch of tag forms.</p>
                           <XMTok meaning="times" role="MULOP">⁢</XMTok>
                           <XMRef idref="S9.p1.m1.6"/>
                           <XMApp>
-                            <XMTok role="SUPERSCRIPTOP" scriptpos="post5"/>
+                            <XMTok role="SUPERSCRIPTOP" scriptpos="post6"/>
```
xworld21: What? Is this a consequence of the extra grouping? I doesn't make that much sense, though.

brucemiller: Yeah, it's weird, but the grouping level is relevant to figure out which scripts pair-up. So consequently, adding extra grouping isn't always free.  It's best not to introduce extra when not necessary, since it causes churn & confusion.

I think the grouping was fine as it was, just a little odd looking. (sorry for the wild goose chase)

lib/LaTeXML/Package/mathtools.sty.ltxml
brucemiller: I'd thought about a "Calc" version of those types in #2652, but *so* many dimensions potentially get sneakily passed through calc, that it's impractical to figure out which are `CalcDimension` and which are plain `Dimension`.  I decided to err on the side of *all* being calc, if calc is loaded. Unfortunately, we haven't yet settled on the clean, efficient way to detect that, but...

So you don't actually need to parse the dimension here, at all; `\kern` will do it for you, so you don't need either the `readExpression` nor the `Tokenize($dim->toString)`.

For the record, with Primitives or Constructors, it's handy to use `[Dimension]` for the parameter. It's less appealing for Macros (which deal *only* in Tokens), since it ends up causing extra cycles of parse/revert.

xworld21: > So you don't actually need to parse the dimension here, at all; `\kern` will do it for you, so you don't need either the `readExpression` nor the `Tokenize($dim->toString)`.

I don't get it, it doesn't work when I try that. The mathtools example is of the form `\shoveleft[\widthof{$A+B$}]`, and if I pass all of that to `\kern`, it complains of 'Missing number' and the following tokens `\widthof ...` make their way into the output. As far as I can tell, the current implementation of `Dimension` does not use calc, nor calc.sty.ltxml redefines `Dimension`.

t/ams/mathtools.xml
xworld21: > I think the grouping was fine as it was, just a little odd looking. (sorry for the wild goose chase)

I reverted the extra grouping.

lib/LaTeXML/Package/mathtools.sty.ltxml
brucemiller: Sorry (again) about my mangled response; that's exactly one of those cases where a LaTeX op is secretly enhanced if the calc package is loaded.  What I should have said was that *if* #2652 were merged, you wouldn't need to deal with the dimension. But, for now, you do; so your code is fine.

xworld21: > _if_ #2652 were merged

I should have inferred!

PR #2837, opened 2026-06-29
use correct number of columns in \hdotsfor

### LINKED ISSUE #2839: \hdotsfor is '\hdotsfor[]{}` and it should span multiple columns
The argument of `\hdotsfor` is not how many dots to produce, but how many cells to occupy. Plus, the argument can be wrapped in braces. This leads to further content on the same line being put in the wrong cell, so matrices may end up being semantically wrong.

xworld21:
`\hdotsfor{N}` creates dots spanning `N` columns within a matrix. LaTeXML does not.

This is a simple fix that creates all the columns required, with some dots in each. It also fixes the prototype (it was wrong!).

The ideal implementation would create a multicolumn cell with an inline-block with dotted border... or something of that sort. That's beyond the scope of this PR though. I just want the matrices to be semantically correct.

Fixes #2839.

### DIFF
--- a/lib/LaTeXML/Package/amsmath.sty.ltxml
+++ b/lib/LaTeXML/Package/amsmath.sty.ltxml
@@ -841,8 +841,11 @@ DefMacroI('\DOTSI', undef, Tokens());
 DefMacroI('\DOTSX', undef, Tokens());
 Let('\hdots', '\lx@ldots');
 
-DefMacro('\hdotsfor Number', sub {
-    (map { T_CS('\hdots') } 1 .. $_[1]->valueOf); });
+# The optional argument determines the visual space between the dots, currently ignored
+DefMacro('\hdotsfor[]{Number}', sub {
+    my $cols = $_[2]->valueOf;
+    $cols = 1 if $cols < 1;
+    (map { T_CS('\hdots'), T_ALIGN } 1 .. $cols - 1), T_CS('\hdots'); });
 
 # The basic idea is simple enough (in TeX world);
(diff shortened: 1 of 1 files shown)


lib/LaTeXML/Package/amsmath.sty.ltxml
```diff
@@ -841,8 +841,11 @@ DefMacroI('\DOTSI', undef, Tokens());
 DefMacroI('\DOTSX', undef, Tokens());
 Let('\hdots', '\lx@ldots');
 
-DefMacro('\hdotsfor Number', sub {
-    (map { T_CS('\hdots') } 1 .. $_[1]->valueOf); });
+# The optional argument determines the visual space between the dots, currently ignored
+DefMacro('\hdotsfor[Number]{Number}', sub {
+    my $cols = $_[2]->valueOf;
+    $cols = 1 if $cols < 1;
+    (map { T_CS('\hdots'), T_ALIGN } 1 .. $cols - 1), T_CS('\hdots'); });
```
brucemiller, line 848: Cool....but apparently the optional arg is a "scaling" that can be non-integer. Since we aren't using it anyway, it's probably best to code it as `[]`.

brucemiller (changes requested):
Cool! Thanks!!

lib/LaTeXML/Package/amsmath.sty.ltxml
xworld21: Oops.

PR #2843, opened 2026-07-05
tabular vertical align

### LINKED ISSUE #2732: incorrect vertical-align settings with tabular
The following
```latex
\documentclass{article}
\usepackage{array}
\begin{document}
    \begin{tabular}{c|p{2cm}|c}
         a&  a very long line, enough to wrap (p) & b
    \end{tabular}

    \begin{tabular}{c|m{2cm}|c}
         a&  a very long line, enough to wrap (m) & b
    \end{tabular}

    \begin{tabular}{c|b{2cm}|c}
         a&  a very long line, enough to wrap (b) & b
    \end{tabular}
\end{document}
```
should look like

Image

but it comes out as

Image

Even after temporarily setting `.ltx_td { vertical-align: baseline !important; }`, you get

Image

There are three bugs here:
1. The width of `p{2cm}` gets lost.
2. All cells should be `vertical-align: baseline`, no matter what.
3. The content of the column `m` should have `ltx_align_middle` instead of `ltx_align_top`.

Changing all three manually gives you

Image

xworld21:
Fix #2732 plus *more side effects* (hopefully all good).

The three notable changes are:
- Columns `p`, `m`, `b` don't get `vattach` anymore. This is a no-brainer: the `vattach` is already on the `inline-block` for those column types (in fact, you had already fixed the core issue, the only remaining problem was the extra `@vattach`).
- The top/bottom vertical alignments now opt into using `baseline-source`, which behaves like TeX; this also produces better alignment between tags and text in lists. This should be low risk. I am only worried by MathML, because some CSS stops working there.
- **All** table cells default to `vertical-align:baseline` instead of `middle`. This is dangerous! I had to revert the change in *equation* cells, because tags are pretty much hard-coded to `ltx_align_middle`. I believe that all other `<table>`s are generated from `tabular`, `\halign`, and other TeX constructs for which the baseline alignment is correct, so *in theory* this is all fine.

I checked that the result works in all three main browsers (Chrome, Safari, Firefox). Firefox does not implement `baseline-source`, so it falls back to the less precise `vertical-align:top`.

Note: while the CSS spec says authors should prefer the shorthand `vertical-align: first baseline`, Chrome only supports the longhand `baseline-source`. That's why I used `@supports`.

### DIFF
--- a/lib/LaTeXML/Engine/TeX_Tables.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Tables.pool.ltxml
@@ -68,9 +68,8 @@ DefColumnType('r', sub {
 # BUT @{} might add other contents to the cell!  So, we need an inline-block
 DefColumnType('p{Dimension}', sub {
     $LaTeXML::BUILD_TEMPLATE->addColumn(
-      before  => Tokens(T_CS('\lx@tabular@p'), T_LETTER('t'), T_BEGIN, $_[1]->revert, T_END, T_BEGIN),
-      after   => Tokens(T_END),
-      vattach => 'top',
+      before => Tokens(T_CS('\lx@tabular@p'), T_LETTER('t'), T_BEGIN, $_[1]->revert, T_END, T_BEGIN),
+      after  => Tokens(T_END)
     ); return; });
 # Make sure the content sees the desired width as \hsize
 DefMacro('\lx@tabular@p{}{}', '\hsize=#2\relax\lx@tabular@p@{#1}{#2}');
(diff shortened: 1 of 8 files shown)


brucemiller:
This is clearly doing something right! The 1st sample in `t/alignment/array` comes out very nice in Chrome, but really bad in firefox: Although the 1st cell starts below the other cells, the height of the row hasn't been affected, so it lies on top of the next row.

dginev:
Could you comment on "works in Safari" - is the MDN page outdated? They claim it isn't included in the browser Baseline effort due to Safari not having a plan to add support, if I am reading it right:

image

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/baseline-source#browser_compatibility

xworld21:
> the height of the row hasn't been affected

Today I learn that Firefox's `vertical-align:baseline` is *broken* for table cells. Content that increases the line height does not contribute to the cell height! Try adding text after an inline block with `vertical-align:top` (or `baseline-source`... the end result is the same). The cell will increase in height by the number of lines you add, but the content will keep overflowing the cell. Clearly the code that computes the cell height simply counts the number of lines without considering that lines can grow.

Bad, bad Gecko.

https://bugzilla.mozilla.org/show_bug.cgi?id=1936392, ~probably~. Edit: definitely, the test case at the bottom exhibit exactly the issue we have with t/alignment/array. Possibly a regression introduced in 2024.

> Could you comment on "works in Safari" - is the MDN page outdated?

I meant to say, the way I added `@supports` ensures that Safari gets less accurate but acceptable rendering based on `top/bottom` instead of `baseline-source`.

PR #2846, opened 2026-07-07
Leave preamble at right place

### LINKED ISSUE #2754: empty lines in \AtBeginDocument{} do not split paragraphs
```latex
\documentclass{book}
\AtBeginDocument{
  Paragraph 1.

  Paragraph 2.
}
\begin{document}
  Text.
\end{document}
```
produces

Image

The PDF would show two separate lines, with 'Text' on the same line as 'Paragraph 2'.

brucemiller:
Assert we're leaving the preamble after `@document@preamble@atend` (eg. etoolbox's `\AtEndPreamble`, but before `@at@begin@document` (eg. `\AtDocumentBegin`).

Fixes #2754

### DIFF
--- a/lib/LaTeXML/Engine/latex_constructs.pool.ltxml
+++ b/lib/LaTeXML/Engine/latex_constructs.pool.ltxml
@@ -325,9 +325,9 @@ DefConstructorI(T_CS('\begin{document}'), undef, sub {
 
     if (my $ops = LookupValue('@document@preamble@atend')) {
       push(@boxes, $stomach->digest(Tokens(@$ops))); }
+    AssignValue(inPreamble => 0);    # We're now leaving the preamble (!?)
     if (my $ops = LookupValue('@at@begin@document')) {
       push(@boxes, $stomach->digest(Tokens(@$ops))); }
-    AssignValue(inPreamble => 0);    # atbegin is still (sorta) preamble
     if (my $ops = LookupValue('@document@preamble@afterend')) {
       push(@boxes, $stomach->digest(Tokens(@$ops))); }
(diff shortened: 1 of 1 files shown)


dginev (approved):
Thanks

dginev:
Hm, this may have been premature - the latexml-oxide translation had a spike of errors seemingly due to this change. See for example a technique used by the raw inconsolata.sty, minimally:

```tex
\documentclass{article}
\AtBeginDocument{\RequirePackage{xcolor}}
\begin{document}
And {\color{red} red}.
\end{document}on my part 
```

which works in pdflatex.

Do you have a minimal example of what you were trying to fix?

brucemiller:
well, yeah; see the fixes #2754

dginev:
Right, thanks for the reminder.

It looks like a better approach would be to improve `\par` behavior to be mode-based instead of preamble-based and make it a no-op in vertical mode - a comment based on a note you had left in TeX_Paragraph.sty.ltxml .

I'll attempt a PR and get back to you.

brucemiller:
I'd thought that this was the point of etoolbox's `\AtEndPreamble`, but I guess it works with plain `\AtBeginDocument`, too.  A superficial look at latex.ltx suggests there's no `inPreamble` flag (a LaTeXML expedient?), but they simply let all the preamble only commands to an error somewhere within `\document`.    That suggests taking a look at all uses of `inPreamble` to see if the logic can be done a better way...

dginev:
After some head-scratching and AI co-brainstorming I arrived at a pragmatic suggestion: https://github.com/brucemiller/LaTeXML/pull/2848

Certainly one of many, many, variants for improvement.

PR #2848, opened 2026-07-08
Make \par context-aware instead of retiming inPreamble (#2754, #2846)

### LINKED ISSUE #2754: empty lines in \AtBeginDocument{} do not split paragraphs

dginev:
Supersedes the earlier `inBeginDocumentHook` approach on this branch (force-pushed).

Fixes the `\RequirePackage`-in-`\AtBeginDocument` regression from #2846 while keeping #2754 fixed — **without** a second flag.

### The problem

#2846 moved *leaving the preamble* (`inPreamble => 0`) to **before** `@at@begin@document`, so that a blank line inside `\AtBeginDocument` breaks a paragraph (`\par` is a NOOP while `inPreamble`; #2754). But `inPreamble` **also** gates the `\@onlypreamble` guard, so this disabled the preamble-only commands too early: a `\RequirePackage`/`\usepackage` deferred to `\AtBeginDocument` now wrongly errors *"can only appear in the preamble"*.

```latex
\documentclass{article}
\AtBeginDocument{\RequirePackage{xcolor}}
\begin{document}
And {\color{red} red}.
\end{document}
```

Real-world witness: `inconsolata.sty` does `\AtBeginDocument{...\usepackage{upquote}}` → `upquote.sty`'s `\RequirePackage{textcomp}`.

### The fix

Stop routing `\par` through `inPreamble` at all:

1. **Revert #2846** — clear `inPreamble` **after** the begindocument hooks again (`latex.ltx` disables `\@preamblecmds` only after the hook), so a deferred `\RequirePackage`/`\usepackage` stays legal.
2. **Make `\lx@normal@par` context-aware** — a NOOP **only in the raw preamble**: `inPreamble` is set *and* `document` is not yet on the `current_environment` stack. `\begin{document}` pushes `document` at its start, so it is present throughout the begindocument hooks and the body.

So a blank line inside `\AtBeginDocument` still splits paragraphs (#2754), a deferred `\RequirePackage` stays legal, and a body-level `\RequirePackage` still errors.

### Why context, not "NOOP in vertical mode"?

This is the pragmatic reading of the existing note in `TeX_Paragraph.pool.ltxml` (*"`\par` should be a NOOP in vertical mode ... if we can be sure we're tracking modes correctly"*). We can't rely on mode: LaTeXML stays vertical after a display equation (a mode test would drop the blank line between `$$…$$` groups, and inside `\AtBeginDocument{\[x\]…}`), and raw-preamble text is horizontal — so mode can't tell it from a hook `\par`. **Context** — are we inside the `document` environment — is the stable signal. The check walks the `current_environment` *stack* (so a hook that opens a nested environment still counts as "in document") and only runs while `inPreamble` is set, so the hot body path short-circuits.

Fixes #2754. Follow-up to #2846.

### DIFF
--- a/lib/LaTeXML/Engine/TeX_Paragraph.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Paragraph.pool.ltxml
@@ -91,12 +91,18 @@ sub alignLine {
 # Remember; \par _closes_, not opens, paragraphs!
 # Generally, we want to close both an open ltx:p and ltx:para (if either are open).
 # But when executed "internally" (by $stomach->leaveHorizontal), only close ltx:p.
-# NOTE Also that the whole inPreamble bit is, I think, overused.
-# For example, \par should be a NOOP in vertical mode, and that would generally make it
-# ignored in the preamble. (if we can be sure we're tracking modes correctly).
+# \par is a NOOP only in the RAW preamble (before \begin{document}); everywhere else it
+# closes the paragraph being built.  This is the pragmatic reading of the old note here
+# ("\par should be a NOOP in vertical mode ... if we can be sure we're tracking modes
+# correctly"): we can NOT rely on mode (we stay vertical after a display equation, so a
+# mode test would drop the blank line between $$..$$ groups; and raw-preamble text is
+# horizontal, so mode can't tell it from an \AtBeginDocument \par), so we key on CONTEXT
+# — whether \begin{document} has opened the 'document' environment — instead (afterDigest).
+# This lets a blank line inside \AtBeginDocument split paragraphs while inPreamble is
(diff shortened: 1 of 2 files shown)


dginev:
@brucemiller refined the fix to a slightly more elegant one that does not add any new machinery, just slightly tweaks the `\par` check. I'll leave this PR alone at this point. 

The "neutralized vertical mode" idea is promising, nothing new on that front

brucemiller (approved):
Hmm, clever. Might work as intended! :>  Thanks!!

Issue #2466, opened 2024-12-31
improve pt/px ratio accuracy


xworld21:
The current internal DPI for LaTeXML is 100, but that seems arbitrary. I suggest to change it to 96 like CSS, or even better to 96.36, because:
- the CSS dots-per-inch is always 96, where in fact dot = 1px
- in CSS, 72pt = 1inch (= 96px)
- in TeX, 72.27pt = 1inch, and LaTeXML pretends that CSS points are actually TeX points

Thus, to make pixels and points match their TeX meaning even when exported to HTML, the internal dpi value should be 72.27*96/72 = 96.36.

Using the correct DPI value will improve the accuracy of alignment in TikZ and of image sizing, mostly.

At the end of the day, the glitches I see come from LaTeXML using both pxValue and ptValue in the output. I am not sure if there is a general rule for choosing between the two, or if it just happened organically. Everything would work naturally if all logical sizes were consistently exported as (TeX) points, with postprocessing taking care of translating to pixel when required (e.g. for images width/height attributes).

brucemiller:
FWIW, the choice of pt vs px, was definitely inorganic, and somewhat toxic. Originally, the preference for lengths in LaTeXML  were either em or pt (except for `img` lengths).  But basically, I found myself stuck in a corner where only by disabling svg's `viewBox`, and storing all svg lengths in px could I get sorta decent svg. With your persistence, we've finally got enough inconsistencies & bugs repaired where we could re-enable `viewBox` (Thanks!). Hopefully, we'll get enough of the rest to remove the awkward need to explicitly request px.
At that point, this PR will be Icing! Right now, I'm not sure if it would make it easier or harder to detect the current code inconsistencies.

Issue #2377, opened 2024-07-10
support font sizes provided by the class


xworld21:
LaTeXML pretends that the base font size is always 10pt, for simplicity, except for the a0poster class. This is quite inconvenient with TikZ and tikz-cd, because the figures are generated based on 10pt, even though they might have been written with 12pt in mind.

I am running some experiments with BookML and using the font sizes prescribed by the class options results in visible improvements.

Without adjusting for 12pt class option (top LaTeXML, bottom LaTeX+dvisvgm):
![image](https://github.com/brucemiller/LaTeXML/assets/1962985/3936d127-0fc8-4674-90e8-83496bfd129e)
After adjusting for 12pt:
![image](https://github.com/brucemiller/LaTeXML/assets/1962985/51b43e49-380d-462f-9151-fede6ecd8f0c)

(For both examples, the images have bene rescaled so that the characters have the same size as the surrounding HTML.)

xworld21:
@brucemiller will the current refactors and improvements include code related to this? I have a crude implementation of the `XXpt` options in BookML (with corresponding CSS that rescales the SVGs in the output) and I was considering implementing `\@setfontsize` and similar properly, so that files like `size10.clo` can be read raw. That might conflict with your plans, though.

xworld21:
Reviving this issue: I figured that reading .clo files is already feasible with small changes. The chain is `\normalsize` -> `\@setfontsize` -> `\fontsize` -> `\set@fontsize`, while the actual font change is in `\selectfont`. To read size10.clo raw,  one needs to:
- copy the raw definitions of `\@setfontsize`, `\fontsize`, `\set@fontsize` into latex_base, remove the existing ones from latex_constructs (or reimplement them in Perl, but following the content of latex.ltx, i.e. set `\f@size`, adjust the line spread, etc)
- add `\f@size` among the parameters handled in the `\selectfont` primitive
- make the class options 10pt, 11pt, etc read the relevant sizeNN.clo, bkNN.clo, and in very few cases copy the definitions of `\normalsize` and others from the .cls

The advantages are more faithful box sizing (e.g. updating `\strutbox`) and preserving the proportion between font sizes and absolute lengths, in particular the correct pt -> em conversion.

PR #2849, opened 2026-07-09
Foreign object tuning

### LINKED ISSUE #2695: foreignObject width & font size mismatch
The variable `--fo_width` uses em units, based on the font size of its content. However, changes in font size are applied at different times in HTML. In this example:
```latex
\documentclass{article}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
  \draw (0,1) node {Hello.};
  \draw (0,0) node {\scriptsize Hello.};
\end{tikzpicture}
\end{document}
```
both foreign objects get approximately the same size in em units, because LaTeXML is applying `\scriptsize` first.

However, the HTML looks like
```html
<foreignObject style="--fo_width :2.54em;--fo_height:0.61em;--fo_depth :0em;" height="6.73" overflow="visible" transform="matrix(1 0 0 -1 0 6.73)" width="28">
  <span class="ltx_foreignobject_container">
    <span class="ltx_foreignobject_content">
      <span class="ltx_text" style="font-size:70%;">Hello.</span>
    </span>
  </span>
</foreignObject>
```
The width is applied to `ltx_foreignobject_container`, where the font size is still `100%`. Somehow the `font-size:70%` should float up for this to work correctly.

brucemiller:
This PR improves svg foreignObject sizing & positioning.

Since text & graphical items within an svg must be scaled together to preserve the relationship, but we don't know what font size will be used in the browser, we must set the svg size in font-relative units. This was done in a previous PR. But since this introduces an unknown scaling seen by the foreignObject (and in case the svg itself has a stray scaling), we must reset the font size on the foreignObject. (Contrary to what some on the web assert, the svg scaling *does* affect the font size). However, we  should reset it to the font-size used on the parent svg, not the font within the object (% sizes will accumulate!).

I also improved sizing of fractions, and recognized a tricky way that tikz does centering in minipage.

I've removed the use of `--ltx-fo-*` variables from `LaTeXML.css`, since they were used for heuristic tweaks that caused as many problems as they solved, and in fact made it harder to solve the problems correctly. Really, CSS should be used for styling, not for fixing bugs.  I've put the variables back to pts instead of ems, but I'd really prefer to remove them.

fixes #2695

### DIFF
--- a/lib/LaTeXML/Engine/TeX_Box.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Box.pool.ltxml
@@ -407,12 +407,14 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
     # Reminder: TeX/LaTeXML think origin is left & baseline
     # svg foreignObject content will draw down from the top
     if ($whatsit) {
+      my $svg = $document->findnode('ancestor::svg:svg', $node);
       my ($w, $h, $d) = $whatsit->getSize;
       my $H = $h->add($d);
       my $W = $w;
       my $x = Dimension(0);
-      my $y = $h;
-      my $f = $whatsit->getFont;
+      my $f = ($svg && $document->getNodeFont($svg)) || $whatsit->getFont;
+      # Heuristic: Browsers seem set on creating at least a lineheight box, so adjust
+      my $y = $h->larger(Dimension($f->getSize . 'pt'));
       Debug("SVG:foreignObject " . $w->pxValue . ' x ' . $h->pxValue . ' x ' . $d->pxValue . "; "
           . " => H=" . $H->pxValue . " @ y=" . $y->pxValue
           . " for Whatsit=" . ToString($whatsit)) if $LaTeXML::DEBUG{svg_verbose};
@@ -422,12 +424,14 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
       # Set top of FO height above current, as if baseline will line up w/outside
       $node->setAttribute(transform => 'matrix(1 0 0 -1 ' . $x->pxValue . ' ' . $y->pxValue . ')');
       $node->setAttribute(overflow  => 'visible');
+      # Reset the font size, so not affected by any svg scaling.
+      my $style = '';
       # Store expected size for experimentation Temporary?
-      $document->setAttribute($node,
-        style => '--ltx-fo-width:' . $w->emValue(undef, $f) . 'em;'
-          . '--ltx-fo-height:' . $h->emValue(undef, $f) . 'em;'
-          . '--ltx-fo-depth:' . $d->emValue(undef, $f) . 'em;'
-          . 'font-size:' . $f->getSize . 'pt;');
+      $style .= '--ltx-fo-width:' . $w->toAttribute . ';'
+        . '--ltx-fo-height:' . $h->toAttribute . ';'
+        . '--ltx-fo-depth:' . $d->toAttribute . ';';
+      $style .= 'font-size:' . $f->getSize . 'pt;';
+      $document->setAttribute($node, style => $style) if $style;
 } });
 
 sub isVAttached {
--- a/lib/LaTeXML/Engine/TeX_Box.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Box.pool.ltxml
@@ -413,8 +413,10 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
       my $W = $w;
       my $x = Dimension(0);
       my $f = ($svg && $document->getNodeFont($svg)) || $whatsit->getFont;
+      my $y = $h;
       # Heuristic: Browsers seem set on creating at least a lineheight box, so adjust
-      my $y = $h->larger(Dimension($f->getSize . 'pt'));
+      my $vhack = $h->subtract(Dimension($f->getSize . 'pt'));
+      $vhack = Dimension(0) if $vhack->valueOf > 0;
       Debug("SVG:foreignObject " . $w->pxValue . ' x ' . $h->pxValue . ' x ' . $d->pxValue . "; "
           . " => H=" . $H->pxValue . " @ y=" . $y->pxValue
           . " for Whatsit=" . ToString($whatsit)) if $LaTeXML::DEBUG{svg_verbose};
@@ -430,7 +432,9 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
       $style .= '--ltx-fo-width:' . $w->toAttribute . ';'
         . '--ltx-fo-height:' . $h->toAttribute . ';'
         . '--ltx-fo-depth:' . $d->toAttribute . ';';
-      $style .= 'font-size:' . $f->getSize . 'pt;';
+      # Reset font-size to parent svg's, so not affected by svg scaling.
+      $style .= 'font-size:' . $f->getSize . 'pt;'
+        . '--ltx-fo-vhack:' . $vhack->emValue(2, $f) . 'em;';
       $document->setAttribute($node, style => $style) if $style;
 } });
 
--- a/lib/LaTeXML/Engine/TeX_Math.pool.ltxml
+++ b/lib/LaTeXML/Engine/TeX_Math.pool.ltxml
@@ -303,7 +303,7 @@ sub cleanup_XMText {
             $document->replaceNode($m, map { $_->childNodes } $m->childNodes); }
           else {                                           # Otherwise, wrap whatever it is in an XMText
             $document->wrapNodes('ltx:XMText', $m); }
-        } } }
+    } } }
     # And now we don't need the XMText any more.
     foreach my $attr ($textnode->attributes) {    # Copy the child's attributes (should Merge!!)
       $table->setAttribute($attr->nodeName => $attr->getValue); }
@@ -426,7 +426,7 @@ sub scriptHandler {
     my $c = (($op eq 'SUPERSCRIPT') ? '^' : '_');
     Error('unexpected', $c, $stomach, "Script $c can only appear in math mode");
     return Box($c, undef, undef, (($op eq 'SUPERSCRIPT') ? T_SUPER : T_SUB));
-  } }
+} }
 
 DefPrimitiveI(T_SUPER, undef, sub { scriptHandler($_[0], 'SUPERSCRIPT'); });
 DefPrimitiveI(T_SUB,   undef, sub { scriptHandler($_[0], 'SUBSCRIPT'); });
@@ -547,7 +547,7 @@ DefRewrite(xpath => 'descendant::ltx:Math[child::ltx:XMath[child::ltx:XMApp[' .
             elsif ($role eq 'FLOATSUBSCRIPT') {
               $document->insertElement('ltx:sub', $text);
               return; }
-          } } } }
+    } } } }
     # should never happen, but just in case:
     Info("rewrite", "footnotemark", "Failed to find floating node in: " . $math->toString(1));
     $document->getNode->appendChild($math);
@@ -822,7 +822,7 @@ sub augmentDelimiterProperties {
       if (exists $$entry{char}) {                    # replace the char content!
         if (my $char = $$entry{char}) {
           $delim->firstChild->setData($char); } }
-    } }
+  } }
   return; }
 
 # Useful for afterConstruct of delimiter sizing macros (eg. \bigl)
@@ -1058,10 +1058,22 @@ sub adjustMathStyle_internal {
 
 sub fracSizer {
   my ($numerator, $denominator) = @_;
-  my $w = $numerator->getWidth->larger($denominator->getWidth);
-  my $d = $denominator->getTotalHeight->multiply(0.5);
-  my $h = $numerator->getTotalHeight->add($d);
-  return ($w, $h, $d); }
+  my $nfont     = $numerator   && $numerator->getFont;
+  my $dfont     = $denominator && $denominator->getFont;
+  my $style     = $nfont       && $nfont->getMathstyle || 'text';
+  my $isdisplay = ($style eq 'display');
+  my $syfont    = 'cmsy' . ($nfont && $nfont->getSize || 10);
+  my ($wnum, $hnum, $dnum) = map { $_->valueOf } $numerator->getSize;
+  my ($wden, $hden, $dden) = map { $_->valueOf } $denominator->getSize;
+  my ($wnom, $hnom, $dnom) = map { $_->valueOf } $nfont->getNominalSize;
+  my $th      = Dimension('0.4pt')->valueOf;
+  my $xheight = getFontDimen($syfont, 5);
+  my $maxis   = $xheight / 2 + $th;
+  my $pad     = ($isdisplay ? 3.5 : 1.5) * $th;
+  my $h       = $pad + max($dnum, $dnom) + $hnum + $maxis;
+  my $d       = $pad + max($hden, $hnom) + $dden - $maxis;
+  my $w       = max($wnum, $wden) + Dimension('2.4pt')->valueOf;    # should be what?
+  return (Dimension($w), Dimension($h), Dimension($d)); }
 
 # \lx@generalized@over{reversion}{keyvals}{top}{bottom}
 # keyvals: role,meaning, left,right, thickness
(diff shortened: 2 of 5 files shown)


lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -407,12 +407,14 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
...
+      my $svg = $document->findnode('ancestor::svg:svg', $node);
       my ($w, $h, $d) = $whatsit->getSize;
       my $H = $h->add($d);
       my $W = $w;
       my $x = Dimension(0);
-      my $y = $h;
-      my $f = $whatsit->getFont;
+      my $f = ($svg && $document->getNodeFont($svg)) || $whatsit->getFont;
+      # Heuristic: Browsers seem set on creating at least a lineheight box, so adjust
+      my $y = $h->larger(Dimension($f->getSize . 'pt'));
```
xworld21: I... don't like this. You are now lying again about the position of the foreignObject as intended by PGF, which makes enhancing the output using `--ltx-fo-*` quite a bit harder (I'll basically have to parse the `transform` attribute to undo the adjustment). You should let the foreignObject be where PGF wants it and solve the problems on the HTML/CSS side.

This is also reintroducing the problems that were caused by `\baselineskip`, on a smaller scale, because the font size is a bit smaller. For instance: the line height in the browser is likely determined by the *content* font size, so in my example you need 7pt, not 10pt. And what if the content uses multiple font sizes, superscripts and subscripts, or MathJax is manipulating the MathML... ? None of that will work correctly.

I still think that flex and optionally text-box-trim are the only sane ways to address this issue.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
```diff
@@ -422,12 +424,14 @@ Tag('svg:foreignObject', autoOpen => 1, autoClose => 1,
...
       # Store expected size for experimentation Temporary?
-      $document->setAttribute($node,
-        style => '--ltx-fo-width:' . $w->emValue(undef, $f) . 'em;'
-          . '--ltx-fo-height:' . $h->emValue(undef, $f) . 'em;'
-          . '--ltx-fo-depth:' . $d->emValue(undef, $f) . 'em;'
-          . 'font-size:' . $f->getSize . 'pt;');
+      $style .= '--ltx-fo-width:' . $w->toAttribute . ';'
+        . '--ltx-fo-height:' . $h->toAttribute . ';'
+        . '--ltx-fo-depth:' . $d->toAttribute . ';';
+      $style .= 'font-size:' . $f->getSize . 'pt;';
```
xworld21: Since we are using variables, why not set `--ltx-svg-font-size` on `svg:svg` and have a generic `foreignObject { font-size: var(--ltx-svg-font-size); }` rule? Repeating the font size on each FO seems wasteful.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
brucemiller: I think I/we need to understand exactly why we're sometimes getting this vertical position wrong. It seems to be that the browser is doing it, but it could still be something overlooked by LaTeXML. In the latter case, we haven't put it where pgf wanted it. In the first case, you're probably right that it would be better solved in CSS, if we knew what/when to apply it. But the previous flex approach simply vertically centered everything, which is wrong.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
brucemiller: Yeah, that could work, too.

lib/LaTeXML/Engine/TeX_Box.pool.ltxml
xworld21: > why we're sometimes getting this vertical position wrong

Yeah, what I am trying to say is that we should keep the concerns separated. In my mind, the issues either happen in the stomach (the box is not generated correctly) or in the browser (inevitably, the browser box is different). So I'd say that the constructor must trust the stomach because it's not where the problem is.

> But the previous flex approach simply vertically centered everything, which is wrong.

We need to decide what is 'right', though. I suppose we agree that we'd like to deposit the text on the baseline set by PGF? I got the best results, by far, doing that. Even got acceptable figures with the wrong font size. Centering was the desperate fallback for browsers not implementing text-box-trim (Firefox), although at some point I did something crazy with `1lh - 1cap`... I'd have to look at it again.

brucemiller: Was just brainstorming with @dginev and we pretty much concluded that this extra space is the leading, although it isn't clear why it doesn't appear consistently. And, as you suggest, `text-box-trim`, or possibly something with `::first-line { line-height:0; }`, should help. But I'm having trouble getting either of those to work consistently in Chrome in non-svg cases, and not at all in our svg, or in firefox.

xworld21: Have you tried
https://github.com/brucemiller/LaTeXML/pull/2488/changes#diff-c25f8a2b2cc3e2b234e08745990985488db6ffc118e40c185c6634df7f4e8896?
that is
```css
/* move center of TeX box (height + depth tall) to center of CSS line (ascent + descent tall)
   0.306em is the harcoded value of (ascent - descent)/2 for Computer Modern */
foreignObject { translate: 0 calc(0.306em + (var(--lx-depth) - var(--lx-height)) / 2); }
```
The 0.306em would have to be rescaled appropriately based on the font size of the content. Still not 100% reliable. Nothing really is, apart from `text-box-trim` (in that same PR), and I admit I don't remember testing multiline content very much. It worked sufficiently well for purely horizontal boxes, and very well in Chrome.

xworld21: > it isn't clear why it doesn't appear consistently

By the way, do you have a list of reference examples that you are using?

brucemiller: I'm mostly working with a long list of TeXamples codes and the TColorBox samples that @dginev posted with a draft PR; some of them are in `LaTeXML/t/(pgf|tikz)`. Many of them have multiline foreignObject, and some have a short entry at the top of a longer outer box, so vertical centering is not really desirable.

Assuming I can figure out the amount to shift & when, I suppose it might be a bit "Cleaner" to store it as a variable and do the actual translation in CSS, similar to what you suggested.  All with the hope/expectation of replacing it with `text-box-trim` someday.

brucemiller: I just pushed a commit that treats this Hack adjustment as I described above.  Hopefully we can remove it once the appropriate CSS (eg. `text-box-trim`) is available.

dginev:
I can also contribute two recent regression tests ("witnesses") with plenty of foreignObject, which would ideally render *beautifully* in HTML. The examples are in the appendixes for both:

**`arXiv:2605.02240`** (tcolorbox) and **`arXiv:2605.00468`** (typewriter/verbatim)

Also, for 2605.00468, we can observe a typewriter-specific bug of 5% size mismatch, due to "quad" (em-size) not matching the design size ("drawn to be used at this size").

A handy command to verify these:

```bash
$ tftopl "$(kpsewhich cmtt10.tfm)" | grep -iE 'DESIGNSIZE R|QUAD R'
(DESIGNSIZE R 10.0)
   (QUAD R 1.049991)
```

LaTeXML uses both `getEMWidth` (quad) and `getSize` in Font.pm (design size) and we may have sized with the former but tell the browser the latter. I am just beginning to understand these details, so no thoughts on the best way to sync everything, but I can appreciate that we "merely" have to become completely consistent with our sizing.

xworld21:
> LaTeXML uses both `getEMWidth` (quad) and `getSize` in Font.pm (design size) and we may have sized with the former but tell the browser the latter.

I'd classify this as 'font metric' mismatch, which is going to happen no matter what, unless you do silly stuff like https://vlmantova.github.io/bookml/index.plain.html (Latin Modern!), but even with the same font, browsers use different typesetting so you get different results unless you are rendering single characters or words. (For your tcolorbox example: it does seem like LaTeXML is overestimating the height, presumably some `\baselineskip` shenanigans rather than font metrics? It's also weird that the width of the foreignObject is *less* than the width of the minipage. Something's off.)

The question here (for me) is orthogonal: given that the HTML box *will* be different, what is the least ugly answer?

From what I understand from the examples: for small horizontal boxes, e.g. in commutative diagrams, my baseline + line centering fallback works *great*. I keep pestering you about foreignObject because I know that tikz-cd can look basically perfect, but it's very easy to break. On the other hand, tcolorbox seems obviously better with straight centering – your arXiv example looks way more convincing with `align-items: center` instead of `align-items: end`.

Maybe the right answer is to use different strategies depending on single vs multiple lines? Say use centering when foreignObject contains an `inline-block`, and aim for the baseline otherwise. It can even be done entirely in CSS via some variant of `:has(>.ltx_inline-block:only-child)`.

(I admit I am thinking and talking about this way too much.)

brucemiller:
I think it's perfectly natural to focus, or obsess, on the foreignObjects, as the textual bits tends to draw the eye, or at least *my* eye! And of course, that's the area where the browser & TeX are each switching gears in quite different ways.

By default, TeX's boxes, and it's measurements, are intended to convey where the *ink* is (though the font designer can lie for various good reasons); of course those measurements can, and are, then adjusted with wild abandon when composing. OTOH, the browser apparently constructs "lineboxes" with *half* leading above & below (TeX puts them between lines), and quite possibly has a different set of "lies" about where the top (or bottom) of the ink may be (?). If we were able to predict or guess the distance between the top of the browser's linebox and it's electrons/ink,  *or* if `text-box-trim` eventually does it for us, we've solved one big problem.

Using CSS to vertically center goes too far. Think of `\parbox[t][10cm]{10cm}{Hello!}` with tall given height. And with a lot of layout going on, you can get an arbitrary number of "container" elements before you get content; selector could be tricky.  Do you have a commutative diagram sample where the current code is noticably off?

Although TColorBox is affected by this offset issue, it *also* suffers from insufficiencies in LaTeXML's sizing code. But I still need to dig into the source code to see exactly what they're computing the size *of*.

xworld21:
Small foreignobject stress test:
[foreignobject-stress-test.zip](https://github.com/user-attachments/files/29923762/foreignobject-stress-test.zip)

The result is not *terrible* but all text content is too high, and the `\large B` is quite off (whereas text-box-trim + Latin Modern gets basically pixel perfect results).

The test includes a DPI adjustment to 96.36 (as per #2466), otherwise the `--ltx-fo-*` variables are systematically some 4% off the `@width`, `@height` attributes even now that they are in pt units, plus there's a bit of CSS to set the font to Latin Modern Math 10pt, to have matching font metrics.

> Think of `\parbox[t][10cm]{10cm}{Hello!}` with tall given height.

LaTeXML omits the height of the `\parbox` in the XML output, so CSS will definitely have a hard time aligning it properly. But if it were to have its own height, there wouldn't be any issue, the foreignObject would already have the same size as the parbox.

Maybe the inline-block heuristic is too weak, but I still don't understand why centering is wrong in the abstract. For nodes in a TikZ figure, centering is definitely more pleasant than having all nodes shifted in the same direction.

But putting that aside, the VO hack is something I really struggle to understand: the number you want is ascent + half lead, assuming the ascent doesn't get pushed up (e.g. by a superscript). You can't predict that number, it depends on the font, the line-height property, and the mood of the browser. The fact that it looks vaguely reasonable in some example is just a coincidence: in my browser, for LM 10pt, the ascent should be 7pt, plus half-lead probably 1pt, so 8pt. That happens to be close to 10pt, and it would explain why all nodes are a few pixel too high, and even higher when you use the wrong DPI (ignoring `\large B` which pushes the ascent further up, hence it ends up in the completely wrong place).

The summary of my understanding is: without text-box-trim, the only anchors available in CSS are the corners of the padding box and the center. Trying to find another anchor like the baseline is a fool's errand.

xworld21:
> it depends on the font, the line-height property, and the mood of the browser

I forget I had all this stuff figured out last year. I should have kept notes. You cannot anticipate the HTML line box, but of course you can compute the equivalent TeX 'line' box, i.e. decide where the 'descent + half line skip' would be below the baseline, and likewise 'ascent + half line skip' for top edge, when computed from TeX. Then you can decide that one of those is your anchor point for the HTML box.

But LaTeXML can't do that yet, because its implementation of `\normalsize` et alii is too naive: it only sets the font size, but it doesn't update `\baselineskip` nor it accounts for `\baselinestretch` (a plausible reason for the exceedingly wrong vertical sizes in the arXiv examples). I mentioned `\@setfontsize` in #2377 and surely in a couple of other comments since then. The recent Opus-generated PR #2789 included improvements about it (not too dissimilar from my crude https://github.com/vlmantova/bookml/blob/4c3c01d808b1f969b865bfeb2c29ee75ca8d06d0/bookml-init.sty.ltxml#L470-L482).

I suppose that reading `latex.ltx` raw changes things, because you can rely on the native `\fontsize` implementation and target the lowest level font setting primitives, so you can in principle get the line skip adjustments for free. I don't think you are doing that, though.

brucemiller:
Without having a clear understanding of the boxes that the browser is constructing, I can't really simulate it on the TeX side. And chasing the "center" anchor is just as much a fools errand as any other. 

For the most part, your stress test looks fairly decent in Chrome (with or without your font & dpi changes), and less so in Firefox, especially the last 2 line case.
Can you also post the html+css which you find is significantly better?

xworld21:
Warning: long text ahead! Line boxes are just complicated.

> Without having a clear understanding of the boxes that the browser is constructing, I can't really simulate it on the TeX side. And chasing the "center" anchor is just as much a fools errand as any other.

I haven't been clear about it. The way browsers construct lines is explained in the CSS inline module, and it is something like (barring memory failures and misunderstandings): ascent metric + half leading above baseline, descent metric + half leading below baseline, plus extra space if there are inline blocks which are higher or deeper then ascent and descent. Without text-box-trim, the browser can only align content based on over-edge, under-edge, and center of a line. With text-box-trim, you can also target the baseline (and a few other lines). Also note that ascent and descent are properties of the font, independent of the actual glyph metrics (so as far as I know, an oversized glyph in text does NOT alter the height of the line in the browser). The leading is either suggested by the font, or determined by `line-height`.

Since from the TeX side we do have the font metrics, we *can* determine where the browser under-edge would be if the browser were using the same font. Your VO hack is a crude version of this: you are trying to pin the over-edge as per TeX font metrics to the over-edge that will come out in the browser. I say crude, because you use 1em, instead of using the ascent metric (0.694em for Computer Modern) and half of the line skip. I suppose max(1em,ascender) is also not quite right, but it should be in the vast majority of cases (the exceptions should be oversized glyphs in text, which are rare).

I would personally target the under-edge, rather than the over-edge, because I am fixated with baselines. The centre of the line is not necessarily better, but it has a huge advantage in that it depends on ascent and descent only, so it's implementable right now without modifying `\normalsize` and it's oblivious to `line-height` changes.[^1]

Targeting the baseline, instead, has none of those issues: with text-box-trim, you just deposit the baseline exactly where TeX wanted it, and stop thinking about it.

For reference, https://www.w3.org/TR/css-inline-3/#inline-height

> Can you also post the html+css which you find is significantly better?

This is the text-box-trim version:
```css
foreignObject {
  translate: 0 var(--ltx-fo-depth);
}

.ltx_foreignobject_container {
  display: flex;
  height: 100%;
  width: 100%;
  align-items: end;
}

.ltx_foreignobject_content {
  translate: initial;
  text-box: trim-end cap alphabetic;
}
```
Note that it relies on the foreignObject being the TeX box (width x totalheight) with an extra variable `--ltx-fo-depth`. I find this the obviously correct answer and the thing we should be approximating as much as possible.

Under the same assumptions, my 'improved center line' hack (really, a different take on the VO hack) looks like
```css
foreignObject { translate: 0 calc(0.306em + (min(var(--ltx-fo-depth),0.306em) - min(var(--ltx-fo-height),0.694em)) / 2); }
.ltx_foreignobject_container { /* ...flex... */ align-items: center; }
```
Which shows issues when 1em is not the actual font size, e.g. if the content is wrapped in `<span style="font-size:70%">`. I just improvised the `min()` addition to match multiple lines correctly, and indeed it's pretty much a simplified version of the VO hack.

If you want to match the over-edge instead of the center, that should also be reasonable. I would probably try the under-edge because it might result in closer baselines on average, but it also might not, who knows. My only complaint issue is the VO hack does not match over-edge to over-edge – it just feels like it does it by accident.

**Note:** never, ever translate the *content* of a foreignObject (the fO itself is fine) nor use `position:relative`. It triggers a decades old WebKit bug which completely destroys the figure. Nobody seems interested in fixing it.

[^1]: Except for Firefox, as usual. I distinctly remember Firefox doing something different when `line-height` is smaller than 1.

brucemiller:
Thanks for this; it looks helpful! I've got some things to do, but will work through this and experiment shortly.

PR #2850, opened 2026-07-11
make listings line numbers not selectable

xworld21:
Alleviate an issue related to #2629: currently, if you select code in a listing across multiple lines, you also grab some of the line numbers. Making numbers not selectable seems more sensible.


dginev (approved):

brucemiller (approved):
Oh, cool! I didn't know there was such a property (even if it isn't yet widespread). Thanks!

PR #2852, opened 2026-07-20
standalone.sty: a subfile's \documentclass options are not packages

dginev:
`\@standalone@documentclass` `RequirePackage`s the comma-split **optional** argument of an included subfile's `\documentclass`, but that argument holds class *options*. An ordinary subfile preamble therefore reports a missing file:

```latex
% main.tex
\documentclass{article}
\usepackage{standalone}
\begin{document}\input{child}\end{document}

% child.tex
\documentclass[12pt]{article}
\begin{document}Child text.\end{document}
```

```
Warning:missing_file:12pt Can't find binding for package 12pt at child.tex; line 1 col 1
Conversion complete: 1 warning; 1 missing file[12pt.sty]
```

`\documentclass[border=2pt]{standalone}` — a very common subfile preamble — misses the same way. Content is unaffected; the damage is a false `missing_file` in the log and the missing-file tally.

**The fix follows `standalone.sty` itself.** `\sa@documentclass` compares `#2` against `standalone` before honoring `#1`, and only under `obeyclassoptions`, which `\newif` defaults to false — so the package consults a subfile's class options only when that subfile's class is literally `standalone`, and then hands them to a keyval family rather than `\RequirePackage`. This gates the loop the same way, and within a `standalone` subfile requires only the options `standalone.cls` turns into a same-named package load (`tikz`, `pstricks`, `preview`, `varwidth`, `multido`) — which is what makes `\documentclass[tikz]{standalone}` from #1432 work, and is why the loop is kept rather than dropped.

Verified: the MWE above goes `1 warning; 1 missing file` → `No obvious problems`, and #1432's `\documentclass[tikz,border=2pt]{standalone}` subfile still loads tikz and renders its picture.

Also adds a `t/structure` case for subfile inclusion, which had no coverage. In fairness it does **not** guard this fix: the bug is a spurious warning that leaves the XML unchanged, so `latexml_ok` cannot observe it — I could not find a harness that asserts on warnings. It does guard subfile inclusion generally.

Found while porting this binding to [latexml-oxide](https://github.com/dginev/latexml-oxide) (dginev/latexml-oxide#309, reported by a user); the equivalent Rust fix is dginev/latexml-oxide#310.


brucemiller (approved):
Ah, nice. Thanks!

PR #2855, opened 2026-07-28
Preserve existing logs for archive output

### LINKED ISSUE #2836: latexmlc --dest=.epub deletes log file
Starting with a minimal `file.tex`:
``` latex
\documentclass{article}
\begin{document}
content
\end{document}
```
Then:
``` shell
> latexmlc --dest=file.html file
latexmlc (LaTeXML version 0.8.8; revision cb2f810a)
processing started Sat Jun 27 12:41:19 2026
Conversion complete: No obvious problems
Post-processing complete: No obvious problems
Status:conversion:0
> ls
file.html		file.latexml.log	file.tex		LaTeXML.css		ltx-article.css
> latexmlc --dest=file.epub file
latexmlc (LaTeXML version 0.8.8; revision cb2f810a)
processing started Sat Jun 27 12:41:38 2026
Conversion complete: No obvious problems
Post-processing complete: No obvious problems
Status:conversion:0
> ls
file.epub	file.html	file.tex	LaTeXML.css	ltx-article.css
```
Not only has `latexmlc` not created a log file, it actually deleted the existing one.

yuchenmiao04:
Fixes #2836.

## Summary

- use a compact `$keep_log_file` flag to identify archive output before initializing the client log
- preserve pre-existing external log files for archive outputs
- keep default archive log names relative so a fresh conversion log is packaged inside EPUB/ZIP output
- keep EPUB regression coverage in `t/931_epub.t` and move the ZIP variant to `t/932_zip.t`

## Root cause

`latexmlc` unconditionally removed the selected log file before the output type had been normalized. Archive conversions buffer their log for packaging instead of writing that external file, so this could delete a pre-existing log without replacing it. The previous absolute default log path also prevented the fresh log from being added to the archive.

The preflight flag tracks the raw output-shape, format, and destination-extension signals that produce archive output. This keeps the live conversion configuration unchanged while preserving non-archive log handling.

## User impact

Running an EPUB, MOBI, or ZIP conversion no longer deletes or overwrites an existing external log with the same name. The archive still contains a fresh conversion log.

## Validation

- `prove -lv t/931_epub.t t/932_zip.t` (25 tests)
- related client, format, profile, runtime, configuration, fatal-path, and manifest tests (81 tests)
- `CI=true prove -lv t/97_manifest.t` (2 tests)
- `tools/latexmllint --precommit bin/latexmlc t/931_epub.t t/932_zip.t`
- `git diff --check`

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -860,6 +860,7 @@ t/91_latexmlc_api.t
 t/92_profiles.t
 t/93_formats.t
 t/931_epub.t
+t/932_zip.t
 t/94_runtimes.t
 t/95_complex_config.t
 t/96_fatal.t
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -182,8 +182,7 @@ my $destination = $opts->get('destination');
 # Always write log to a file, either specified via --log or 'latexml.log'
 # with the exception of *archive* outputs, where the log is in the ZIP
 my $whatsout   = $opts->get('whatsout');
-my $is_archive = $keep_log_file
-  || ($whatsout && ($whatsout =~ /^archive/));
+my $is_archive = $whatsout && ($whatsout =~ /^archive/);
 
 if ($log && !$is_archive) {
   chomp($log);
--- a/bin/latexmlc
+++ b/bin/latexmlc
@@ -20,6 +20,7 @@ my $RealBin_safe;
 use FindBin;
 use File::Spec::Functions;
 use File::Which;
+use Storable qw(dclone);
 
 BEGIN {
   if ($FindBin::RealBin =~ /^([^\0]+)\z/) {    # Valid Unix path TODO: Windows, revisit regexp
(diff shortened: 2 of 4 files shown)


dginev (changes requested):
This is headed in the right direction, but I believe there is a more compact way to get the effect.

Can you avoid using `dclone` and prepare a flag variable, maybe `$keep_log_file`, responsible for tracking all possible reasons for the log to be kept?

Similarly, the ZIP portion of the test work could move to a new `t/932_zip.t` file to keep the variants more readable.

yuchenmiao04:
@dginev Thanks for the review. I updated the implementation in `3afb43d0`:

- removed `dclone` and replaced the preflight normalization with a single `$keep_log_file` flag covering `whatsout`, format, and destination-derived ZIP output
- moved the ZIP regression coverage into `t/932_zip.t` and added it to `MANIFEST`

The focused EPUB/ZIP tests pass 25/25, the related client and format suite passes 81 tests, and the CI manifest check passes 2/2. Would you mind taking another look?

dginev:
Verified against master and against the real `latexmls`. The fix works — I reproduced #2836 and confirmed it's fixed, and the log now actually lands inside the archive (`OPS/file.latexml.log`), which master never managed. Worth calling that out in the description; it's the more valuable half.

One more change to request, this is not really helpful:

```perl
my $is_archive = $keep_log_file
  || ($whatsout && ($whatsout =~ /^archive/));
```

This is currently dead weight. Locally it's a no-op — `prepare_session` has already run `Config::check`, so `$whatsout` is `'archive'` whenever `$keep_log_file` is set (reverting just this line keeps `t/931_epub.t` and `t/932_zip.t` green, so nothing covers it).

Using the old variant as-is should be fine:
```perl
my $is_archive = $whatsout && ($whatsout =~ /^archive/);
```

Keep the `binmode` switch to `$is_archive` regardless; it fixes a real `Use of uninitialized value $whatsout ... line 189` that master emits on that path.

*(Reviewed by Claude Opus 5, driven by @dginev.)*

yuchenmiao04:
@dginev Thanks for catching this. I updated is_archive to remain derived solely from whatsout, leaving keep_log_file responsible only for the unlink decision. I reran t/931_epub.t and t/932_zip.t (25 tests), the related regression suite (81 tests), and the CI manifest checks; all pass.

brucemiller (approved):
Looks reasonable, and if @dginev is happy, I am! Thanks!!

PR #2875, opened 2026-09-16
mtable @align should get baseline1 or baseline-1 for top or bottom

### LINKED ISSUE #2729: MathML mtable/@align looks better with baseline 1, -1 instead of top, bottom
...at least when running the examples in t/ams/mathtools.tex. For instance the current 'bottom1' renders as

Image

whereas 'baseline 1' looks like

Image

Compare with the PDF

Image

I get more accurate alignment also with 'baseline -1' in place of 'bottom'.

I tested the above with both Firefox 146 and MathJax 4. Chrome and Safari do not support the attribute altogether (and I bet it's not in MathML Core).

The line that would need to change is
https://github.com/brucemiller/LaTeXML/blob/d81e955babef9508b4908af4ed08583f5c5e7b59/lib/LaTeXML/Post/MathML.pm#L439

brucemiller:
This PR implements @xworld21 suggestion in #2729.

Note that `mtable@align` is not in MathML Core, but we were already using it. This PR improves the appearance in Firefox; We'll have to come up with a better CSS based solution once CSS is better supported by MathML.

Fixes #2729.


xworld21:
> This PR improves the appearance in Firefox

And in MathJax too (I mentioned elsewhere that MathJax is still the best MathML renderer in the browser).

dginev:
> Note that mtable@align is not in MathML Core, but we were already using it. 

What is the long-term plan for moving this to MathML Core? There is no expectation for the extended set to make it to all browsers.

dginev (approved):

brucemiller:
> What is the long-term plan for moving this to MathML Core? 

Presumably, it'll have to be handled with CSS; I'm open to a relatively simple, cross-browser CSS solution (if & when).

